PR #54, opened 2019-10-16
docs: Create StorybookDemo component

sebpearce:


sebpearce:
This is just the component that will hold our Storybook iframe embeds when we hook them up.

Looks like this:

image

PR #56, opened 2019-10-17
docs: Order Components nav items with Overview at top

sebpearce:
We're already sorting the GraphQL data for the nav items on the Guidelines page; this PR does the same for Components (slightly different as they don't need to be in separate sections in the sidebar).

### DIFF
--- a/site/src/components/MainNav.tsx
+++ b/site/src/components/MainNav.tsx
@@ -2,7 +2,7 @@ import {
   Link as NavLink,
   NavigationBar,
 } from "@cultureamp/kaizen-component-library"
-import { graphql, useStaticQuery, withPrefix } from "gatsby"
+import { withPrefix } from "gatsby"
 import * as React from "react"
 
 type MainNavProps = {
@@ -10,21 +10,6 @@ type MainNavProps = {
 }
 
 const MainNav: React.SFC<MainNavProps> = ({ currentPath = "" }) => {
-  const data = useStaticQuery(graphql`
-    query MainNavComponentsQuery {
-      allMdx(filter: { fields: { slug: { regex: "^/components/" } } }) {
-        edges {
-          node {
-            fields {
-              slug
-            }
-          }
-        }
-      }
-    }
-  `)
-  const firstComponentPath = data.allMdx!.edges[0]!.node!.fields!.slug
-
   const currentPathStartsWith = (path: string) =>
     currentPath.startsWith(withPrefix(path))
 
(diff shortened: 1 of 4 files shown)


site/src/templates/componentPage.tsx
```diff
@@ -13,12 +13,35 @@ import {
   SidebarSection,
   SidebarTab,
 } from "../components/SidebarAndContent"
+import { sortSidebarTabs, stripTrailingSlash } from "./util"
+
+const renderSidebarTabs = (pages, currentPath) => {
+  return pages.map((node, i) => (
+    <SidebarTab
+      href={node!.node!.fields!.slug}
```
mbylstra: I realise this wasn't introduce in this PR, but could this `node!.node!.fields!.slug` shenaningens be moved to a variable, so it at least gets a name explain what it is?

site/src/components/MainNav.tsx
```diff
@@ -2,29 +2,14 @@ import {
...
-          node {
-            fields {
-              slug
-            }
-          }
-        }
-      }
-    }
-  `)
-  const firstComponentPath = data.allMdx!.edges[0]!.node!.fields!.slug
```
mbylstra: I have no idea what this was, but I'm glad it's gone!

site/src/templates/componentPage.tsx
mbylstra: Actually, it's the `node!.node!` bit that's the hardest to understand, and it seems really susceptible to changes of structure? Perhaps a function like like `getFooNode` (I don't actually know what Foo is to give it a name)

sebpearce: > it seems really susceptible to changes of structure?

Yeah 👎 Not sure what you mean by using a function as a way to remedy that though. Wouldn't it be just as susceptible?

site/src/components/MainNav.tsx
sebpearce: Before, there was no components overview, so we had to query to find out what the first component was and make a link for that. Now we just hard-code `components/overview`.

site/src/templates/componentPage.tsx
sebpearce: The `node.node` bit is part of the structure of Gatsby GraphQL queries, not very intuitive.

sebpearce: Actually I'll just merge this and raise an issue to clean this up as tech debt. So far we've been pretty good on striking off issues so there's a good chance I'll get to it.

sebpearce: Done. See https://github.com/cultureamp/kaizen-design-system/issues/60

PR #72, opened 2019-10-23
docs: Create dos and don'ts components

sebpearce:
This PR allows docs contributors to add dos + don'ts components like this:

image

Note that MDX doesn't seem to allow named imports like `import { Do, Dont, DosAndDonts }`.

### DIFF
--- a/site/src/docs-components/DosAndDonts.scss
+++ b/site/src/docs-components/DosAndDonts.scss
@@ -1,0 +1,48 @@
+@import "~@cultureamp/kaizen-component-library/styles/color";
+
+.container {
+  display: flex;
+  justify-content: space-between;
+}
+
+.card {
+  margin-top: var(--ca-grid);
+  width: 48%;
+  padding: var(--ca-grid);
+  box-sizing: border-box;
+  border: 1px solid $ca-border-color;
+  box-shadow: 0px 1px 1px rgba(0, 0, 0, 0.19);
+  border-radius: 3px;
+  border-top-width: 7px;
+  border-top-style: solid;
+
+  :global(.md-ul.md-ul) {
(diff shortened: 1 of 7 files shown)


site/src/docs-components/DosAndDonts.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  :global(.md-ul.md-ul) {
+    padding-left: 1.25rem;
+  }
+
+  :global(.md-li.md-li) {
+    font-size: 0.875rem;
+    line-height: 1.5;
+    margin-top: 0.75rem;
+  }
```
sebpearce: Just an explanation on this weirdness:

We had an issue when porting the old site where we wanted to use some existing site components like the colour palette showcase, but those old components had things like `<h2>`s in them that were getting our global Markdown styles (`h2 {}`) and messing up the look of the component.

In order to ship this quickly and not go into that rabbithole I decided to instead generate an `.md-` class for anything that our Markdown parser spits out, in order to distinguish Markdown content from things like `<h2>`s inside site components.

With the _new_ site components, however, we _do_ want to inherit our Markdown styles inside the components, since the component styles are often a slight variation on the Markdown styles – like in this case, where the `li` style involves tweaking the size, line-height and margin while keeping everything else the same.

Unfortunately this meant a bit of a specificity battle, so I reached for this fairly-harmless-but-slightly-gross hack which involves repeating the classname to bump up its specificity.

Ideally one day I'd like to just update the ported components to not use `<h2>`s and have their own scoped classes, so that we can just use regular `<h2>`s in our Markdown and site components. But it'll be a bit more work to get there. Thinking of raising an issue for this as tech debt. Open to alternatives though!

PR #73, opened 2019-10-23
feat: Create SplitButton component

jeremy8883:
This PR mostly just copy/pastes the `SplitBox` component from murmur.

image

The component was needed for the perform frontend, as per CORE-169 (ie. the cycle tables found under "admin -> self reflections" and "admin -> employee evaluations").

There are some minor differences with this kaizen implementation:

* I noticed there was some screenreader text for the dropdown icon. I made that text customisable as part of a `dropdownAltText` prop. I figured that different systems would be handling translations differently, so I didn't think kaizen should be the one to handle it. But lmk if there's a better way.
* Instead of passing down a `Locale` object to determine if the dropdown is LTR or RTL, it's now just part of the `dir` property. Again, I'm assuming different systems will probably have different locale objects.
* As to keep things simple, I didn't include the `Tooltip` component that murmur uses. I think its purpose is when the user hovers over the down chevron, it shows a "More" tooltip. The component required the `react-bootstrap` dependency, so I was also being cautious of the bundle size. Currently in murmur, it's only used under `SurveyListContainer`, and I couldn't get it working in my local environment, so maybe no-one will even notice.
* Before the button was an `a`nchor element. I updated the logic so it will only be an `a`nchor, if a `href` is supplied - otherwise it will be a button. This allows for better ux and a11y.
* Flow and js files have been converted to typescript


jeremy8883:
I snuck in a new update:

> Allow SplitButton menu width to be greater than the SplitButton, if required

See updated PR description for a snapshot of the change.

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -0,0 +1,171 @@
...
+    // https://github.com/WICG/focus-visible#backwards-compatibility
+    :global(.js-focus-visible) & {
+      // hide native focus ring when :focus but not :focus-visible
+      &:focus {
+        outline: none;
+      }
+
+      // show custom focus ring when :focus-visible
+      &:global(.focus-visible) {
+        border: $focus-border-width solid $ca-palette-ocean;
```
jeremy8883: Most of the code here was pasted from murmur. There are a couple of minor things that I noticed, that I didn't spend time on, but let me know if I should clean it up.

One of the things was this line. When the user tabs onto the SplitButton, the added 2px border will change the size of the component. Ideally I'd say the size should remain the same.

jeremy8883:
I got back from the designer in regards to getting sign-off for this component. There were some minor centering issues that I will check out before merging this one (the down chevron doesn't align with the menu icons). The only problem is that issue is in regards to the `MenuItem` component, which was already an existing component. I'll investigate and see what I can do, but things might get awkward given that the MenuItem has no knowledge of the SplitButton and vice versa.

ActuallyACat:
> I got back from the designer in regards to getting sign-off for this component. There were some minor centering issues that I will check out before merging this one (the down chevron doesn't align with the menu icons). The only problem is that issue is in regards to the `MenuItem` component, which was already an existing component. I'll investigate and see what I can do, but things might get awkward given that the MenuItem has no knowledge of the SplitButton and vice versa.
> 
> Edit: Or alternatively, lmk if this is still good to merge, given that this would be a tricky change to make.

If this is moved to be a draft component we can get this through sooner (if this component is blocking work for you). 

I'd like to get @didoesdigital opinion on whether the split button dropdown should use the `MenuItem` as a child, or if this is a special enough case that something like `SplitButtonMenuItem` could (or should) exist. My hunch is no - in which case both will need to be refactored slightly so that they work together.

ActuallyACat:
My mistake - it looks like this is a draft. LGTM to merge as is.

PR #74, opened 2019-10-23
feat: Minor enhancements to the Table component

jeremy8883:
* Remove `overflow: hidden` from `TableCard` component. The reason was because in the perform product, we wanted to include a `SplitButton`, but it was cutting off the dropdown like so:

image

I think this `overflow: hidden` existed because of the expandable table rows, but I couldn't seem to replicate the feature. When I try it out in storybook, I get the following:

expanded example

So maybe this feature isn't finished yet? I'll do some further investigation before getting this PR merged.

* Table cell widths to allow for flex values. We currently only allow setting a cell width via a `number` ratio value from 0 to 1. I wanted a bit of flexibility with setting the row widths. In perform, I wanted some columns to be fixed width, and one of the columns to fill the remaining width.

* Add ability for table rows to link to routes. It turns out we no longer need this functionality in perform, but I kept it in anyway since I figured that it may be useful.

* Add property to force the hover state, even through there is no `onClick` or `href` property passed down. The reason being, in perform, I wanted the ability to make every cell clickable, except for the last column with the `SplitButton`. So I set the `href` properties at the cell level instead of row level. However, I still wanted the hover state to show, hence this new `forceHoverState` prop.

image

LMK of any other preferred alternatives for solving any of these issues.


sebpearce:
Sorry for the wait – will play around with this tomorrow (Wednesday).

packages/component-library/draft/Kaizen/Table/Table.tsx
```diff
@@ -109,10 +129,17 @@ export const TableRow: TableRow = ({ children }) => (
 )
 
 type TableRowCell = React.FunctionComponent<{
```
ActuallyACat: @sebpearce and I were discussing and think that this could be confusing for consumers having width as a percentage and allowing a flex string. WDYT of annotating so it's clear when people use this component?

```suggestion
/**
 * @param width value between 1 and 0, calculated as a percentage
 * @param flex CSS flex shorthand as a string
 */
type TableRowCell = React.FunctionComponent<{
```

ActuallyACat: Like this :D 

image

packages/component-library/draft/Kaizen/Table/Table.tsx
```diff
@@ -22,10 +22,14 @@ export const TableHeaderRow: TableHeaderRow = ({ children }) => (
...
 
+const ratioToPercent = (width?: number) =>
+  width != null ? `${width * 100}%` : width
+
 type TableHeaderRowCell = React.FunctionComponent<{
   labelText: string
   onClick?: (e: React.MouseEvent<HTMLButtonElement>) => any
-  width: number
+  width?: number
+  flex?: string
```
ActuallyACat: Ditto for this one

PR #75, opened 2019-10-23
feat: Create a package for tokens

ActuallyACat:
These are a copy pasta job from `kaizen-component-library`. No token variables or folder structures have been changed to preserve the imports in sass. 

Next step is to do some investigation into using this in Murmur / Perform. If all goes well, references can change from `@cultureamp/kaizen-component-library/styles/thing` to `@cultureamp/kaizen-tokens/styles/thing`. 

If that doesn't go to plan, this can be deleted with no implications / side effects and we'll look into another approach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,18 +8,18 @@
     "node": "10.16.3"
   },
   "scripts": {
-    "storybook": "start-storybook -c storybook",
-    "storybook:build": "build-storybook -c storybook -o storybook/public",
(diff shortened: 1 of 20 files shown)


mbylstra:
[Extract tokens from kaizen-component-library into new npm package](https://trello.com/c/o9m2stPK/572-extract-tokens-from-kaizen-component-library-into-new-npm-package)

PR #84, opened 2019-10-25
docs: Add Storybook demos for components

sebpearce:
## Iframe path issue

A limitation we’ll have is that I can only point the story iframe to https://cultureamp.design/storybook-static/iframe.html, which means it won’t show any changes on a branch when you look at the branch preview URL. If I point it to the relative path, /storybook-static/iframe.html, as a way of getting around this, it of course breaks because it needs to have the branch name at the start of the path. We might just have to settle with changes not showing up here on branch previews.

## Visual issues

Here are some of the nastier results that we need to make a decision on before merging this.

They're variously related to:
- the **height of the iframe** (we can't use JS to dynamically size the height from our page, because Gatsby is server-side-rendered and doesn't have access to `window`)
- the fact that the **width of the iframe is narrower than a desktop viewport** (affects desktop-only components like Global Notification)
- the fact that whether or not a component is **centred** is determined by each story
- the **width of the content inside the iframe** (determined by each story)
- any **padding above the component** makes the iframe look wrong (determined by each story)

image
image
image
image
image
image


sebpearce:
This is the only thing stopping this from getting through now:

image

I'll try to find some answers on how we can allow CORS here.

site/src/templates/componentPage.tsx
```diff
@@ -42,6 +43,14 @@ export default ({ data, pageContext, location }) => {
     allPages.filter(el => el.node.frontmatter.navTitle !== "Overview")
   )
 
+  const renderStorybookIFrame = () => {
+    return md.frontmatter.demoStoryId ? (
```
ActuallyACat: Do you think it would be worthwhile logging an error to the console here? Just thinking about what might be useful for a non-developer trying to update the docs and not seeing their example. 🤔

sebpearce: Superb idea. Will do 👍

sebpearce:
@ActuallyACat Added this:
image

PR #85, opened 2019-10-25
feat: Add a React version of the Select component

lijuenc:
### Storybook preview

https://dev.cultureamp.design/lijuenc/react-select/storybook-static/?path=/story/select--single

### Context

This PR adds a React version of the Select component. Until now we've only had an Elm version in Kaizen and if you needed the same thing in React, you had to use [react-select](https://github.com/JedWatson/react-select) directly and manually apply styles to make the react-select control look like a Culture Amp control.

### Implementation

As a company we're generally familiar with react-select because it's used in murmur, but it's a very old version. This Select component uses the latest version of react-select internally and is basically a skin.

- Instead of using global CSS to target and override react-select's built-in styles, we use the components API to **replace** parts of its internals with Culture Amp looking components (without replacing the behaviour!).
- All props on this component get passed through to react-select, so the react-select documentation is an accurate representation of the API: https://react-select.com.
- Because it's nothing more than a skin, it should be easier to keep up to date with newer react-select versions and it should be less likely that we end up stuck on an old version as we have in murmur.

### Something to be aware of

React version:
Screen Shot 2019-10-25 at 4 08 46 pm

Elm version:
Screen Shot 2019-10-25 at 4 08 57 pm

Notice the slight differences in the multi select between the React and Elm versions.

The Elm Select uses the Elm Tag component which is, feature-wise, way ahead of the React Tag component. It has things which the React Tag doesn't (like dismiss buttons).

Because the React Tag is out of date, the React Select component skips using it and instead just uses react-select's tag.

There should be a separate piece of work to bring the React Tag up to parity (I'll do this next). That will allow us to start using it in the React Select component.


sebpearce:
Nice work!! Will be great to have a React equivalent finally.

> There should be a separate piece of work to bring the React Tag up to parity (I'll do this next). That will allow us to start using it in the React Select component.

Will this work include merging together the React and Elm stylesheets somehow, like breaking out any common styles for them to both inherit from? The only thing that worries me in this PR is the splitting of these two and the possibility of them drifting apart as time goes on. Or is this inevitable given that we're modifying the internals of react-select?

lijuenc:
> Will this work include merging together the React and Elm stylesheets somehow, like breaking out any common styles for them to both inherit from? The only thing that worries me in this PR is the splitting of these two and the possibility of them drifting apart as time goes on. Or is this inevitable given that we're modifying the internals of react-select?

I think some of that is unavoidable given that the the Elm component was built from scratch while the React component is a skin on top of react-select.

For the React component, I've tried to inherit as much as possible from more generic kaizen styles (like borders, shadows, and form input styles) so its styles already fairly minimal.

Perhaps we can do once over for the Elm component's CSS and make sure that it's using those generic kaizen styles rather than defining everything from scratch. I've tried to do some of that in this PR, replacing manual styles with references to variable values.

sebpearce:
Roger.

> Perhaps we can do once over for the Elm component's CSS and make sure that it's using those generic kaizen styles rather than defining everything from scratch. I've tried to do some of that in this PR, replacing manual styles with references to variable values.

Yeah, this is great. If we just get it to the point where it's duplicating as little as possible then that's the best we can hope for I think.

lijuenc:
I've resolved the merge conflict, could you take a look again please?

PR #86, opened 2019-10-25
fix: select content alignment in ie11 off

ellisjas:


ellisjas:
[IE11 bugs roundup](https://trello.com/c/OyHPNWeN/1667-ie11-bugs-roundup)

PR #90, opened 2019-10-29
feat: Add scss variable and function

kemblesong:
Thoughts? kebab-case? 🤔

### DIFF
--- a/packages/binary-assets/README.md
+++ b/packages/binary-assets/README.md
@@ -9,10 +9,12 @@ Binary asset management for Culture Amp's Kaizen Design System.
 Returns the full URL of the asset at `path` managed by the
 `kaizen-design-system-assets` service.
 
+**TypeScript/JavaScript**
 ```ts
 assetUrl("some/blob.png") // -> "https://<origin>/some/blob.png"
 ```
 
+**SCSS**
 ```scss
 asset-url("some/blob.png") // -> "https://<origin>/some/blob.png"
 ```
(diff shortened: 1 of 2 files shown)


packages/binary-assets/lib/index.scss
```diff
@@ -0,0 +1,8 @@
+$kaizen-asset-origin-base-url: "https://kaizen-assets.s3-us-west-2.amazonaws.com";
+
+// Returns the full URL of the asset at `path` managed by the
+// `kaizen-design-system-assets` service.
+// @return string The full URL of the aset at $path
```
sebpearce:
```suggestion
// @return string The full URL of the asset at $path
```

packages/binary-assets/README.md
```diff
@@ -13,6 +13,10 @@ Returns the full URL of the asset at `path` managed by the
 assetUrl("some/blob.png") // -> "https://<origin>/some/blob.png"
 ```
 
+```scss
+asset-url("some/blob.png") // -> "https://<origin>/some/blob.png"
+```
+
```
sebpearce: Now that there are 2 examples, I think it'd be helpful for people who stumble on this for the first time to see labels above each example saying "JavaScript" and "SCSS". Otherwise the kebab case is the only thing to tell them it's SCSS (on GitHub's Markdown preview).

sentience:
Looks good to me. I think kebab-case is our established standard for Sass functions/mixins.

packages/binary-assets/lib/index.scss
kemblesong: i liek aset

sebpearce: me to

PR #92, opened 2019-10-29
chore: update codeowners to ping the DST for any new packages

ActuallyACat:
Props to @zioroboco for the suggestion 👍 

This essentially adds us as an owner for any new packages in this repo. It will need to be updated once draft components are split out of the Design System.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,7 +6,7 @@
 /packages/ @cultureamp/design-systems
 
 # Exclude modifications to existing packages by assigning it to no one
-/packages/component-library/ @ghost
+/packages/component-library @ghost
 /packages/binary-assets @ghost
 
 # Pipelines are maintained by Delivery Engineering:
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,8 +1,13 @@
 # CODEOWNERS
 # https://help.github.com/articles/about-code-owners/
 
-# Require Design Systems Team approval to modify this file:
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,8 +1,13 @@
...
 # https://help.github.com/articles/about-code-owners/
 
-# Require Design Systems Team approval to modify this file:
+# Require Design Systems Team approval for any new packages or this file
 /CODEOWNERS @cultureamp/design-systems
+/packages/ @cultureamp/design-systems
+
+# Exclude modifications to existing packages by assigning it to no one
+/packages/component-library/ @ghost
+/packages/binary-assets @ghost
```
sebpearce: Does it matter whether there's a trailing slash? (Lines 9 & 10)

ActuallyACat: Good catch! I've just found this (codeowners uses gitignore regex) 

https://git-scm.com/docs/gitignore
> If there is a separator at the end of the pattern then the pattern will only match directories, otherwise, the pattern can match both files and directories.

It might be more effective to exclude _any_ changes within component-library and binary-assets. Else we'll get pinged if someone makes a new file within the top level of these dirs, for example, someone adds or modifies a README or other .md files here. 

WDYT?

sebpearce: Yep I think you're right. We don't want to block people when they're doing boring stuff, just becomes another headache.

PR #93, opened 2019-10-30
chore: Create token package

ActuallyACat:
Similiar to my other (closed) PR, this one goes a step further and removes anything that's not an option token. 

Still a working draft.


packages/tokens/styles/layout.scss
```diff
@@ -0,0 +1,5 @@
+@import "type";
+
+$ca-layout-max-width: 1416px;
+$ca-layout-side-padding: 24px;
+$ca-layout-content-width: 1080px;
```
ActuallyACat: Not 100% sure how I feel about these ones. They're very layout focused.

packages/tokens/styles/type.scss
```diff
@@ -0,0 +1,29 @@
...
+$ca-default-font-base-size: 0.875rem; /* 14px */
+$ca-default-font-descender-height: 0.115;
+
+// Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
+$ca-ideal-sans-font-family: "Ideal Sans A", "Ideal Sans B",
+  $ca-default-font-family;
+$ca-ideal-sans-font-base-size: 1rem; /* 16px */
+$ca-ideal-sans-font-descender-height: 0.14;
+
+// Locale-specific fonts
```
ActuallyACat: Locale-specific fonts can probably go too, WDYT?

packages/tokens/styles/color.scss
```diff
@@ -0,0 +1,32 @@
...
+// Secondary colours
+$ca-palette-seedling: #45ad8f;
+$ca-palette-ocean: #1b7688;
+$ca-palette-wisteria: #727193;
+$ca-palette-lapis: #253c64;
+$ca-palette-peach: #f3786d;
+$ca-palette-yuzu: #ffce1e;
+
+// Tertiary colours
+$ca-palette-stone: add-tint($ca-palette-ink, 97%);
```
mbylstra: `add-tint` is a mixin that comes from kaizen-component-library/styles/color.scss, not a scss built in. So, in it's current state this file won't compile.

There might be a cyclic dependency issue if kaizen-component-library depends on kaizen-design-tokens which depends on kaizen-component-library (for the mixins). Scss helpers like this might need to be moved into another package to avoid the dependency, or they could be moved here (but then the nice the token vs mixin separation you have going here is lost). A more controversional way to do it is just hardcode hex values for the ones here that use the mixins 🤔 . Perhaps that's not such a terrible thing. We are moving towards having hex values for all tokens anyway.

packages/tokens/styles/layout.scss
mbylstra: I guess layouts and content widths are design decisions, so it makes sense for them to be tokenised, even if they can be considered legacy tokens (we may have a completely different non backwards compatible approach tokens when it comes time to doing layout for the design vision)

packages/tokens/styles/color.scss
mbylstra: It'd make compilation a few nanoseconds faster :)

ActuallyACat: Really good catch! This is one of the reasons I think it's good to think about where these tools and helpers might belong. Some might best fit in the component library, but others would be better suited to their own special package. 

In this case i'll change this to a hex/rgba value.

packages/tokens/styles/layout.scss
sebpearce: Also, we have a `Layout` component. There's an ongoing problem with the question of where to put these values – if we put them with the Layout component, people in other repos then use those styles directly and it becomes brittle, plus a few other Kaizen components will end up doing the same :( I'm pretty sure the reason these are here now is that they used in be in the Layout component and got moved out once other components like `TitleBlock` and `EmptyState` started needing them.

"Layout tokens" are something we should make into a wider conversation I reckon.

See this convo @dougmacknz and I having btw: https://github.com/cultureamp/kaizen-design-system/pull/103#issuecomment-548649459

packages/tokens/styles/type.scss
sebpearce: Hmmm, we do want these to apply in every repo though. I see them as the same as the standard type styles but for other languages. They need more fleshing out though.

packages/tokens/styles/color.scss
```diff
@@ -0,0 +1,32 @@
+$white: #fff;
+$black: #000;
+
+// Primary colours
+$ca-palette-ink: #3e4543;
+$ca-palette-paper: #f2edde;
+$ca-palette-coral: #f04c5d;
+
+// Secondary colours
```
didoesdigital: @ActuallyACat nit: we're moving away from these groupings of primary, secondary, tertiary for our colors so you could leave out the comments here

didoesdigital:
@ActuallyACat are we able to change the package to `design-tokens`? This would align with the direction of [W3C](https://www.w3.org/community/design-tokens/) and package naming in other design systems.

PR #97, opened 2019-10-31
feat: Add restore icon

jhuangnz:
Add new restore icon made for Perform camp (Future of Work team)
Screen Shot 2019-10-31 at 12 01 10 pm


sebpearce:
You just need a conventional commit message. Usually this is just done in the PR title _except_ when there's only one commit (thanks GitHub) in which case the commit itself needs to have the conventional commit message format.

Try `feat: Add restore icon`.

jhuangnz:
Thanks @sebpearce, have amended. TIL about conventional commits :)

PR #103, opened 2019-10-31
feat: Auto generate anchors for headings

### LINKED ISSUE #8: Markdown headings don't have IDs so can't be linked to
We need to decide whether to:
1) generate dynamic IDs in `snake-case` based on the heading text, or
2) use fixed IDs that are unrelated to the text

dougmacknz:
Closes #8 

* Auto generates an ID for all headings generated from MDX content so that we can link to specific sections on a page (eg. https://cultureamp.design/guidelines/badge-tag/#when-to-use)
* Adds a link that's viewable on hover of a header (or via keyboard navigation) that you can use to go to that anchor and grab the URL easily.
image

---

I originally went down a slightly different path with this where I wrote the ID generation logic into `markdownComponents.tsx`, but writing it there had a flaw where we couldn't detect duplicate heading IDs for a single page (only site-wide).

I then tried implementing the [`gatsby-remark-autolink-headers`](https://www.gatsbyjs.org/packages/gatsby-remark-autolink-headers/) plugin, but found that the `offsetY` param (to offset our fixed header) didn't seem to work at all, and the logic to generate the IDs was a bit flawed.

So eventually I decided to create a local version based off that plugin, which I noticed `reactjs.org` have done as well, and modified/stripped it to suit us.

---

The main thing here is we'll want to agree on the rules for the ID generator, ideally it doesn't change from the point this is merged. I've currently got it accepting ampersands and forward slashes (something Gtithub doesn't do), because I found that it makes some IDs make a bit more sense, but I'm open to hearing what others think.

Note: Gatsby seems to have a bug when using the develop build that it won't jump to the anchor on page load, but it works fine with `gatsby build/serve`:
https://github.com/gatsbyjs/gatsby/issues/6309


site/src/plugins/autolink-headers/index.js
```diff
@@ -0,0 +1,94 @@
...
+        // Register this as a heading used
+        headingOccurrences[id] =
+          headingOccurrences[id] !== undefined ? headingOccurrences[id] + 1 : 1
+      }
+    }
+
+    // No custom ID found? Create one from the heading content
+    if (!id) {
+      const slug = toString(node)
+        .replace(/[^A-Za-z0-9_ -&/]/g, "") // remove anything that isn't alphanumeric, with a few exceptions
```
damncabbage: My first reaction to `&` being left alone was confusion, but I don't *think* it'd be a problem.

sebpearce:
Looks good (and intense!).

Re: @damncabbage's comment on using `$ca-navigation-bar__height` – my first impression was that I would prefer not to import component SCSS into the `markdown.scss` if we can avoid it.

There are two alternatives I can think of: duplicating the vars from the nav bar and icon components, or setting them as CSS vars in `global.scss` or similar. I've been trying to make the site as uncoupled to individual components as possible.

Totally open to different arguments though if people feel strongly about this. What do you reckon?

site/src/plugins/autolink-headers/index.js
dougmacknz: Interestingly, Wikipedia/mediawiki appear to leave all the punctuation in their anchor IDs. I tried it out, but I found that it broke on one of our headings that has an ellipsis in it, so I resorted back to a whitelist.

site/src/plugins/autolink-headers/index.js
```diff
@@ -0,0 +1,96 @@
...
+module.exports = ({ markdownAST }) => {
+  const headingOccurrences = {
+    "need-to-know": 1, // There's a 'Need to know' heading on most pages - register one occurrence of it by default
+  }
+
+  visit(markdownAST, "heading", node => {
+    let id
+
+    // Support custom IDs, eg. `### My Great Heading {#custom-id}`
+    if (node.children.length > 0) {
```
dougmacknz: Worth noting that this whole block was copy pasted from the official Gatsby plugin, I didn't change anything.

dougmacknz: Except for adding the `headingOccurrences` stuff at the bottom, oops.

dougmacknz:
> There are two alternatives I can think of: duplicating the vars from the nav bar and icon components, or setting them as CSS vars in global.scss or similar. I've been trying to make the site as uncoupled to individual components as possible.

I don't have too strong of an opinion here but duplicating it seems like a bit of a bummer. What about if the navigation height got moved out into `@cultureamp/kaizen-component-library/styles/layout`?

sebpearce:
> I don't have too strong of an opinion here but duplicating it seems like a bit of a bummer. What about if the navigation height got moved out into @cultureamp/kaizen-component-library/styles/layout?

I definitely agree with the idea here, although I think there's already at least one case (in Murmur, but possibly also elsewhere) where non-Kaizen repos are reaching into the nav bar component's styles to get its height, so this could be messy.

My main concern is that the component's API in React or Elm should be its only interface. If other repos are reaching into the styles, we introduce the possibility of someone breaking some random repo when they make a simple CSS change to a component. Maybe it's unavoidable in some cases (and maybe the navbar is one of those cases). The thing about Murmur I mentioned above is one example of why this is a pain point.

dougmacknz:
> I think there's already at least one case (in Murmur, but possibly also elsewhere) where non-Kaizen repos are reaching into the nav bar component's styles to get its height, so this could be messy.

We could have a stop gap where we set the local `$ca-navigation-bar__height` in `NavigationBar/styles.scss` to the newly created token, which should allow those cases to continue working and they can switch over to using the proper token in their own time.

> My main concern is that the component's API in React or Elm should be its only interface. If other repos are reaching into a component's styles, we introduce the possibility of someone breaking some random repo when they make a simple CSS change to the component.

Yeah I totally agree now that you've pointed that out.

I'm happy to defer to whatever you and team tokens @ActuallyACat & @mbylstra want to do here. But it does seem like people need a way of getting the nav bar height, I've needed at least one other time on top of this, to get sticky headers to sit right below the nav bar. I was also somewhat surprised to see that no one has needed to use the icon width yet.

ActuallyACat:
The content of this PR LGTM 👍 

As I understand it there are two pain points with navigation:
1. Jumping/scrolling to anchors on the page
2. Sticky headers (not sure if this is needed right now)

I would prefer _not_ to statically expose the height of the navbar within the component library if we can avoid it. My concerns would be requiring the height in both javascript and sass, which would result in some duplication. There's also the (potential) use case in the future where the navbar is dynamic; such as different sizes for mobile/desktop, or fixing more content to the bottom of the navbar which would break this. 

I would prefer the navbar to expose its height dynamically. If we're cautious about throttling events I feel as though this is a future-proof solution. I can POC something together in the coming days about what the API and usage might look like.

ActuallyACat:
@dougmacknz This is my preferred approach:  
https://github.com/cultureamp/kaizen-design-system/compare/ally/navbar-poc
There's a little bit of setup required for the NavbarContext for the consumer, but well worthwhile as things are less likely to unexpectedly break. Let me know if you want to chat about it :) 

Branch build: https://dev.cultureamp.design/ally/navbar-poc/storybook/

dougmacknz:
Thanks @ActuallyACat. that's a cool approach. However, correct me if I'm wrong, as this PR currently stands we wouldn't to be able to use it here, given this is a Gatsby plugin in plain JS that doesn't have access to React & Context Providers.

I believe we'd have to move this code from a plugin into the React code, which is what I originally had but wasn't able to create unique heading IDs *per page* like I was able to with the plugin. It might be possible though. I probably won't have a chance to look at this for at least a couple weeks though.

dougmacknz:
Sorry @sebpearce, two cases of CSS reachy going in.

PR #111, opened 2019-11-07
chore: token package for colours

ActuallyACat:


lijuenc:
In SASS it's pretty common for us to use fractions or multipliers of `$ca-grid`, for example:

```
margin: $ca-grid * 2;
```

Will we have a similar mechanism for this in JS?

ActuallyACat:
> In SASS it's pretty common for us to use fractions or multipliers of `$ca-grid`, for example:
> 
> ```
> margin: $ca-grid * 2;
> ```
> 
> Will we have a similar mechanism for this in JS?

Yup, You'd be able to import the JSON tokens for spacing if you need to do any inline overrides
```
@import token from @cultureamp/kaizen-tokens/tokens/spacing; 

export const MyComponent = () => (
   <div style={{padding: `${token.spacing["grid-unit"]} * 2`}}
);
```

ActuallyACat:
Now I'm looking at it in action I'm not sure how I feel about allowing hyphens in the tokens. It makes accessing the object in javascript a bit icky. WDYT?

lijuenc:
> Yup, You'd be able to import the JSON tokens for spacing if you need to do any inline overrides

Does that work? `1.5rem * 2` is not a valid CSS unit, unless you meant `calc(1.5rem * 2)`. If the expectation is for the consumer to use `calc`, then that's fine, but just wanted to clarify :)

> Now I'm looking at it in action I'm not sure how I feel about allowing hyphens in the tokens. It makes accessing the object in javascript a bit icky. WDYT?

Yeah, camelcasing would make it easier to use.

packages/design-tokens/tokens/color.json
```diff
@@ -0,0 +1,60 @@
...
+      "500": "#FA7558"
+    },
+    "ash": "#ECECEF",
+    "white": "#FFFFFF",
+    "stone": "#F6F6F6"
+  },
+  "DEPRECATED": {
+    "color": {
+      "lapis": {},
+      "ocean": {}
```
ActuallyACat: Ally fix this - ink, lapis, ocean

sebpearce:
> Now I'm looking at it in action I'm not sure how I feel about allowing hyphens in the tokens. It makes accessing the object in javascript a bit icky. WDYT?

> Yeah, camelcasing would make it easier to use.

What you reckon, are we acting on this? The way we've got them now does kind of bias them towards Sass usage.

packages/design-tokens/tokens/color.json
```diff
@@ -0,0 +1,64 @@
+{
+  "kz": {
+    "color": {
+      "wisteria": {
+        "100": "#F0F1F4",
+        "200": "#E1E2EA",
+        "300": "#C4C5D4",
+        "400": "#A7A9C1",
+        "500": "#898BA9",
+        "600": "#5B5E7E",
```
sebpearce: Looking at the Zen UI Kit colours page in Figma, it has `#6B6E94` for Wisteria-600. Worth double checking maybe

packages/design-tokens/tokens/color.json
```diff
@@ -0,0 +1,64 @@
...
+      "seedling": {
+        "100": "#E9F6F3",
+        "200": "#BCE3DC",
+        "300": "#90D1C5",
+        "400": "#64BFAE",
+        "500": "#21A38B",
+        "600": "#177261"
+      },
+      "yuzu": {
+        "100": "#FFFDF4",
```
sebpearce: Figma has `#FFFAE8` for Yuzu-100.

packages/design-tokens/README.md
```diff
@@ -0,0 +1,36 @@
+# Kaizen Tokens
+Design tokens for all platforms.
+
+## About
+Design Tokens are the heart of every Design System. The tokens represented here are platform-agnostic (JSON), as this will help us contribute to & facilitate the maintenance of living style guides. 
+
+In its current state, this package supports Sass and generates three different outputs: color, typography, helpers.
```
mbylstra: This needs to be updated to just color. Also, we should mention that we only include option tokens and not decision tokens (in case people see Warning/Instructional in Figma and wonder why they aren't in this package)

packages/design-tokens/tokens/color.json
ActuallyACat: Fixed, thanks :D

packages/design-tokens/README.md
```diff
@@ -0,0 +1,36 @@
+# Kaizen Tokens
+Design tokens for all platforms.
+
+## About
+Design Tokens are the heart of every Design System. The tokens represented here are platform-agnostic (JSON), as this will help us contribute to & facilitate the maintenance of living style guides. This package defines all the option tokens in Kaizen.
```
sebpearce:
```suggestion
Design Tokens are the heart of every Design System. The tokens represented here are platform-agnostic (JSON), as this will help us contribute to & facilitate the maintenance of living style guides. This package defines all the option tokens in Kaizen.

(An "option token" is a token like `$kz-color-wisteria-800: #35374A` that is one of the options in a palette and contains no information about _how_ it should be used, as opposed to a "decision token" like `$default-text-color: $kz-color-wisteria-800` which does information about how the token should be used.)
```

sebpearce: Optional suggestion here for an explanation of what an option token is – I figured it couldn't hurt to spread awareness.

mbylstra:
What Trello card does this PR related to? It should be linked from this PR (and PR linked from the Trello card)

packages/design-tokens/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Usage
+### Sass
+```
+## Note helper functions are provided
+@import "~@cultureamp/kaizen-design-tokens/lib/sass/[color/depth/layout/spacing/typography/helpers]";
+```
+
+### Javascript 
+```
+@import * as tokens from @cultureamp/kaizen-tokens/tokens/[color/depth/layout/spacing/typography]
```
mbylstra: Should be `@cultureamp/kaizen-design-tokens/...`?

sebpearce: Just had another idea on this – wonder if there's a way we could do what we did with the component library and use `index.js` barrel files to make an API that isn't coupled to a particular dir structure.

So people can just do
```scss
@import "~@cultureamp/kaizen-design-tokens/[color/depth/layout/spacing/typography/helpers].scss"
// no lib/sass/
```
```js
@import * as tokens from "@cultureamp/kaizen-design-tokens/[color/depth/layout/spacing/typography]"
// no tokens/
```

ActuallyACat: Good catch @mbylstra, I'll update the docs. 

I wasn't sure what the future of tokens would look like/if we needed to support Less tokens in the future, so that's why I moved it to its own directory. I can flatten this to `"~@cultureamp/kaizen-design-tokens/sass/[...]`? Thoughts?

ActuallyACat: Actually @didoesdigital had a comment here that I previously missed: 
https://github.com/cultureamp/kaizen-design-system/pull/93#discussion_r342850300

Thoughts on changing to `@cultureamp/design-tokens`, or even `@kaizen/design-tokens`?

sebpearce: > I wasn't sure what the future of tokens would look like/if we needed to support Less tokens in the future, so that's why I moved it to its own directory. I can flatten this to "~@cultureamp/kaizen-design-tokens/sass/[...]? Thoughts?

I don't mean we would have to get rid of any of our internal dirs like `/lib/sass` or anything :)

I mean like what we've done with the component library, where we've got `index.ts` files that export everything and allow you to import directly from the top level dir without having to know anything about the structure beyond that:

image

sebpearce: Re: Thoughts on changing to `@cultureamp/design-tokens`, or even `@kaizen/design-tokens`?

I'd prefer `@kaizen/design-tokens` as long as time doesn't become an issue.

ActuallyACat: > Because of that I think I prefer design-tokens without the kaizen- bit.
Agreed, I'll change this. 

Oh, I see what you mean. This works for javascript because you can import something specific (or everything) from a list of exports... but it doesn't work so well with Sass which makes everything global as soon as you import the file.

sebpearce: Ah yeah, bummer. Maybe just having `/sass` is a good compromise then? Just to make it easy to remember if writing the import manually.

mbylstra: If we leave `@cultureamp/kaizen-component-library`  the same, then removing `kaizen` from `@cultureamp/kaizen-design-tokens` seems inconsistent. Perhaps we should wait until if we decide to use `@kaizen` and then do a big refactor for both: eg (`@kaizen/component-library` and `@kaizen/design-tokens`)

mbylstra: I just read the orginal comment DI left, but I think that was more about the lack of `design` in the name (in the original PR the package name was `kaizen-tokens`), rather than the specifically the addition of `kaizen` int the name.

mbylstra: I haven't delved in depth for ramifications, but other ideas:
- less and scss could end up in the same directory to avoid needing `sass` directory
- could use symlinks to shortcut the deep folder heirachy for a nicer API. That might mess with tooling though.

It be good if we could make the imports less verbose (and more like what people are currently used to)

packages/design-tokens/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Installation
+```
+yarn add @cultureamp/kaizen-tokens
+```
+
+## Usage
+### Sass
+```
+## Note helper functions are provided
+@import "~@cultureamp/kaizen-design-tokens/lib/sass/[color/depth/layout/spacing/typography/helpers]";
```
sebpearce:
```suggestion
@import "~@cultureamp/kaizen-design-tokens/sass/[color/depth/layout/spacing/typography/helpers]";
```

packages/design-tokens/README.md
didoesdigital: That's right @mbylstra, I was referring to using `design-tokens` to match industry naming instead of `tokens`, and have no strong feelings on how we include `kaizen`.

PR #120, opened 2019-11-12
feat: Add Primary prop into SplitButton

vicchirino:
## Objective
Add the design needed into the `SplitButton` component, recently added [here](https://github.com/cultureamp/kaizen-design-system/pull/73), to make it usable for this [JIRA](https://zugata.atlassian.net/browse/FOW-447).

Here is the [Figma link](https://www.figma.com/file/sCtQWGVxrsrNtnTw2LtMAi/Team-Goals-V2?node-id=172%3A4584)

### Changes
Added a `variant` prop into the SplitButton component that is optional and set a default as `"default"`.
If `variant = "primary"` the new color is displayed.
Updated the _scss_ to have different style when that prop is true.
Reused code for updating the states style.
This prop is passed into the `dropdown` used by the `SplitButton` to apply the style there. 

Added stories pointing to the new component style


ActuallyACat:
The border colour will cause issues on any non-white background. I'd consider changing the border colour to seedling (to keep the height consistent with the default button) and have the borders that bump against each other to be white. 

Screen Shot 2019-11-13 at 9 39 07 am
Screen Shot 2019-11-13 at 9 39 26 am

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
```diff
@@ -24,6 +26,7 @@ export default class Dropdown extends React.Component<Props, State> {
   static displayName = "Dropdown"
   static defaultProps = {
     dir: "ltr",
+    primary: false,
```
jeremy8883: I'm wondering if we should change this to a string? So something like `color: "default" | "primary"`. That way, we get more flexibility if we want to have a third theme.

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -70,9 +71,57 @@ $button-width: 124px;
...
 }
 
-.button {
+.buttonPrimary {
+  &:hover {
+    text-decoration: none;
+    color: inherit;
+  }
+
+  &:not(.disabled):not(:disabled) {
```
jeremy8883: With the css tree being a duplicate from above, developers will need to make sure that both `.buttonDefault` and `.buttonPrimary` are kept in sync. I'm thinking a mixin could be used here to solve this issue.

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
ActuallyACat: I really love this idea because a primary button and secondary button are exclusive, yet currently can accept both within the component `<Button primary={true} secondary={true}/>`. The intention is not clear. 

I do see this a Kaizen-wide change though. Button and Link would also need to change, and it would be breaking.

vicchirino: Thanks for the the comments! Noticed that the `<Tag />` component has a `variant` prop that is more like @jeremy8883 is saying. I took the approach of the `<Button /`> that has the `primary` and `secondary` prop, so the third theme will be the default i guess.

packages/component-library/draft/Kaizen/SplitButton/styles.scss
vicchirino: created a `mixin` for duplicate code. Nor sure if this what you are pointing out or if there is a better way to do it

vicchirino:
@jeremy8883 @ActuallyACat this is ready for another round of review. Updated the border color to be seedling keeping the divider white. 

Added a `mixn` for duplicate css styles.

Related to the `width` of the dropdown menu when opening i'm not sure how to do it. 

IMG_FFECFDA946AF-1

At the moment we have a fixed `min-width` to the _primary button_ and then we calculate the _dropdown menu_ width using that `min-width` and the size of the _dropdown button_.
`width: calc(#{$button-width} + 100%);`

But if the text of the _primary button_ is larger than the `min_width` we will have a _dropdown menu_ shorter:

image

Since the _dropdown menu_ is child from the _dropdown button_ not sure how to set up the `width` to be 100% of the `root`. Should i give the possibility to set the `width` inside the component? like a prop or something like that? Will be very helpful some comments 🙏

packages/component-library/draft/Kaizen/SplitButton/styles.scss
jeremy8883: Cool! I was also thinking of including parameters in the mixin. So something like:

```scss
.buttonPrimary {
  @include button-mixin(
    $background-color: add-tint($ca-palette-seedling, 10%),
    $border-color: $ca-palette-seedling,
    $active-background-color: $ca-palette-seedling,
    $disabled-background-color: add-tint($ca-palette-ink, 90%),
    $disabled-border-color: add-tint($ca-palette-ink, 90%)
 )
}
```

That way we don't need to synchronise the css tree structure between variants.

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
jeremy8883: Cool cool! I suppose my vote would be to use a `variant` prop, despite the inconsistency with `Button`, but up to you guys.

jeremy8883:
Interesting problem. I remember having issues with the dropdown width myself. I can pull the branch down to take a look. The downside of fixed widths is that they may not work well with different languages, such as German.

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
ActuallyACat: Good call out about Tag, I didn't realise there was already component API discrepancy. 🤦‍♀ In that case, my preference is to move this to a `variant` prop and the DST can work out a plan/pitch for tidying up the API for other components.

jeremy8883:
> Since the dropdown menu is child from the dropdown button not sure how to set up the width to be 100% of the root

Not sure what your thoughts may be @ActuallyACat, but the way I'm thinking is to surface the `DropdownMenu` to the parent element. That way your css will simply be `width: 100%`, rather than `width: calc(#{$button-width} + 100%)`. No fixed widths required, the width will automatically expand as necessary.

I did a quick test in the chrome dev tools, and it seems to work with a couple css tweaks.

Screen Shot 2019-11-14 at 5 47 32 pm

Right now, the `DropDownMenu` is coupled with the corresponding dropdown button, so a small refactor may be required.

This all assumes that we want the dropdown menu to always equal the width of the parent SplitButton.

packages/component-library/draft/Kaizen/SplitButton/styles.scss
vicchirino: Nice, to be honest didn't know about that. Just updated the PR to include colors! Thanks @jeremy8883

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
vicchirino: Remove the `primary` prop and include the `variant` prop. 2 vs 1 🤓 .  Thanks for the comments

vicchirino:
@jeremy8883 yeah! i thought about that too, to move the dropdown to the parent component. Talked with Jack and we need at least for the case we are using it, that the dropdown has the size of the button. Maybe we can set a max-width to avoid the button being to big? 
Can you push what you did so can continue from there?

jeremy8883:
Cool! What I had done was only in the chrome dev tools, I didn't make any updates to the source code. Here's a snapshot, but it's a bit convoluted

Screen Shot 2019-11-14 at 5 45 17 pm

The SplitButton-styles__menuContainer has been moved, its width has been changed to `100%`, the parent has been changed to `position: relative`, and the `top` of the SplitButton-styles__button has been set to `0` (I'm not quite sure why the `top` and `bottom` was set to `-6.24px`, maybe for vertical alignment?).

vicchirino:
@jeremy8883 @ActuallyACat hey folks had some time to comeback to this PR and update it. 
So i moved out the `DropdownMenu` component to `SplitButton` _(parent element)_.
Now, the dropdown content get the size of the SplitButton make it adjustable to the title label.

Also removed logic to figure out how to center the dropdownMenu since wasn't needed anymore.

**Before:** 
📹 http://g.recordit.co/dQ8HdEK9gq.gif

**Now**
📹 http://g.recordit.co/J6C7r6vqES.gif

Probably this will be a breaking change for performance if merge.

packages/component-library/draft/Kaizen/SplitButton/Dropdown.tsx
```diff
@@ -1,89 +1,57 @@
...
     super(props)
-
-    this.state = {
-      isMenuVisible: Boolean(props.menuVisible),
-    }
   }
 
   toggleDropdownMenu = (e: SyntheticEvent<HTMLButtonElement>) => {
     e.stopPropagation()
-    const currentState = this.state.isMenuVisible
```
vicchirino: moved this into `SplitButton`.

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
```diff
@@ -5,12 +5,6 @@ const styles = require("./styles.scss")
 type Props = {
   children: React.ReactNode
   hideDropdownMenu: () => void
-  position: {
```
vicchirino: This wasn't needed anymore

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
```diff
@@ -24,36 +18,13 @@ export default class DropdownMenu extends React.Component<Props> {
...
     window.addEventListener("resize", this.handleDocumentResize, false)
-    this.positionMenu()
   }
 
   componentWillUnmount() {
     document.removeEventListener("click", this.handleDocumentClick, false)
     window.removeEventListener("resize", this.handleDocumentResize, false)
   }
 
-  positionMenu() {
```
vicchirino: neither this

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -166,7 +240,6 @@ $button-width: 124px;
 }
 
 .menuContainer {
-  position: absolute;
-  width: calc(#{$button-width} + 100%);
```
vicchirino: Now we don't need this width calculation for the dropdown content making it to be the size of his parent.

jeremy8883: Sick! 👌

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -158,6 +220,22 @@ $button-width: 124px;
...
+
+.dropdownButtonPrimary {
+  composes: buttonReset buttonPrimary;
+
+  // Making the left | right border white to separate the button with dropdown button
+  border-left-color: $white !important;
+
+  [dir="rtl"] & {
+    border-left-color: $ca-palette-seedling !important;
+    border-right-color: $white !important;
```
jeremy8883: Using `!important` is generally considered an anti-pattern. Just confirming if these are needed?

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
```diff
@@ -24,36 +18,13 @@ export default class DropdownMenu extends React.Component<Props> {
...
-    const menu = this.menu
-    if (!this.props.position || !menu) {
-      return
-    }
-    const pos = this.props.position
-    const { innerHeight } = window
```
jeremy8883: I see there's some logic here that determines if the popover should be above or below, depending on what best fits the browser viewport window. Just confirming that this logic still exists?

packages/component-library/draft/Kaizen/SplitButton/styles.scss
vicchirino: i see. I couldn't find another way to override this. The thing with the primary color is that the border color is the same color (`seedling`) instead of white (https://github.com/cultureamp/kaizen-design-system/pull/120#issuecomment-553151667) but i put the white color for separating the dropwdown with the main button. If we don't like that we can just make the separator same color as the border `seedling`.
will look like this:

image

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
vicchirino: no, i missed this. Shouldn't being deleted. Added it again but pass the reference until the parent. Thanks for noticing this 🙏

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -159,6 +218,23 @@ $button-width: 124px;
...
+}
+
+.dropdownButtonPrimary {
+  composes: buttonReset buttonPrimary;
+
+  // Making the left | right border white to separate the button with dropdown button
+  border-left-color: $kz-color-white !important;
+
+  [dir="rtl"] & {
+    // equal to add-tint($ca-palette-seedling, 10%). Temporary until we upgrade to Zen (Wisteria)
```
vicchirino: added this comment that is the same that Michael added in [this](https://github.com/cultureamp/kaizen-design-system/commit/1bc36311c7ba674f8525fbe7b15e2c705caeca8b) commit  in [this](https://github.com/cultureamp/kaizen-design-system/commit/1bc36311c7ba674f8525fbe7b15e2c705caeca8b#diff-5e77baee7dee0b91c4ef99b35bbc011b) file

packages/component-library/draft/Kaizen/SplitButton/styles.scss
```diff
@@ -65,15 +67,58 @@ $button-width: 124px;
...
+    $hover-border-color: $kz-color-wisteria-400,
+    $active-background: $kz-color-ash,
+    $active-border-color: $kz-color-wisteria-500,
+    $disabled-background: $kz-color-white,
+    $disabled-border-color: $kz-color-wisteria-200,
+    $disabled-color: $kz-color-wisteria-300);
+}
+
+.buttonPrimary {
+  // equal to add-tint($ca-palette-seedling, 10%). Temporary until we upgrade to Zen (Wisteria)
```
vicchirino: added this comment that is the same that Michael added in [this](https://github.com/cultureamp/kaizen-design-system/commit/1bc36311c7ba674f8525fbe7b15e2c705caeca8b) commit  in [this](https://github.com/cultureamp/kaizen-design-system/commit/1bc36311c7ba674f8525fbe7b15e2c705caeca8b#diff-5e77baee7dee0b91c4ef99b35bbc011b) file

jeremy8883:
I pulled down the branch to test if it worked alright with admin self reflections / employee evaluations, but I was seeing the following:

Screen Shot 2020-01-21 at 5 54 22 pm

My code is just adding `variant="primary"`. Is there anything else that I need to add?

Also, a nitpick, but I noticed a very slight pixel difference between the before and after:

Before:
Screen Shot 2020-01-21 at 5 49 21 pm

After:
image

Notice that in the before snapshot, you couldn't see the bottom border radius of the SplitButton.

vicchirino:
@jeremy8883 hmm interesting.. It seems that the cycle row is taking the size of the splitbutton when expanded? How did you test it that locally? I tried with `yarn link` but without success! 
Also if the intention of the button is to be white, should be `variant = "default"`.

Regarding to the pixel will take a look. 👀

vicchirino:
@jeremy8883 made a couple of fixed. 
1. Set a fixed height into the root container to avoid expansion when the dropdown menu is visible.
2. Update the reference logic to decide were display the dropdown (up or down) this was not working correctly with my changes. Tested in the performance-ui project. What i did, is move the components manually to perf-ui and tested there. 

image

image

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
```diff
@@ -40,11 +40,13 @@ export default class DropdownMenu extends React.Component<Props> {
...
     const rect = menu.getBoundingClientRect()
+
     if (pos.bottom > innerHeight - rect.height) {
-      menu.style.bottom = "40px"
+      const bottom = rect.height + 40
+      menu.style.bottom = `${bottom}px`
       menu.style.top = "auto"
     } else {
-      menu.style.top = "40px"
+      menu.style.top = "-2px" // This is to hide the border of the buttonsContainer class
```
vicchirino: Here i'm not sure other way to do it. The intention of set the top `-2px` is to move up the dropdown menu to fix this --> https://github.com/cultureamp/kaizen-design-system/pull/120#issuecomment-576544686

jeremy8883:
> How did you test it that locally? I tried with yarn link but without success!

My method was by going into the package.json, and changing the `"@cultureamp/kaizen-component-library": "^17.8.0",` line to `"@cultureamp/kaizen-component-library": "../kaizen-design-system/packages/component-library/",`, then doing `yarn install`, then manually removing some `_` characters in some numbers in the kaizen source code. It's not the best way of doing it though, because you need to call `yarn install` every time a change is made. Getting `yarn link` to work properly would be better.

packages/component-library/draft/Kaizen/SplitButton/DropdownMenu.tsx
jeremy8883: I think that's how it was done before, so it should be fine. Thanks!

keirah:
> > How did you test it that locally? I tried with yarn link but without success!
> 
> My method was by going into the package.json, and changing the `"@cultureamp/kaizen-component-library": "^17.8.0",` line to `"@cultureamp/kaizen-component-library": "../kaizen-design-system/packages/component-library/",`, then doing `yarn install`, then manually removing some `_` characters in some numbers in the kaizen source code. It's not the best way of doing it though, because you need to call `yarn install` every time a change is made. Getting `yarn link` to work properly would be better.

I feel like this hack should be pinned to the Camp Perform FE Guild channel 😁

ActuallyACat:
Nothing out of the ordinary here, LGTM.

jeremy8883:
Ah, I noticed another positioning issue with this PR. If it's not affecting performance, then it should be fine to merge this. I've fixed it in [this one](https://github.com/cultureamp/kaizen-design-system/pull/255)

PR #129, opened 2019-11-19
chore: remove sidebar background color

mbylstra:
This is unnecessary, as everywhere the sidebar is used should already have a background.
If we wanted this sidebar to be used on a different colour background like white, it wouldn't work
in it's current implementation anyway (there's no padding, so there'd be a harsh transition from
the side of the sidebar to the white background)


mbylstra:
Not ready for release yet. I haven't been able to test this locally yet (I couldn't get yarn linking to work).

PR #135, opened 2019-11-21
feat: Add 'cautionary' confirmation modal

jhuangnz:
`<ConfirmationModal type="cautionary"`

Screen Shot 2019-11-21 at 7 31 26 pm


packages/component-library/draft/Kaizen/Modal/presets/ConfirmationModal.tsx
```diff
@@ -53,17 +66,15 @@ const ConfirmationModal = ({
...
           <div className={styles.iconContainer}>
             <svg className={styles.iconBackground}>
               <circle cx="75" cy="75" r="75" />
             </svg>
             <div className={styles.icon}>
-              <Icon
-                icon={type === "positive" ? success : information}
-                role="presentation"
-              />
+              <Icon icon={getIcon(type)} />
```
ellisjas: Should this icon still have a role of presentation?

sebpearce:
Adding this here for visibility:

image

packages/component-library/draft/Kaizen/Modal/presets/ConfirmationModal.tsx
sebpearce: Yes, all icons need to have a `role`. I think the build will even break if these are removed.

There's an explanation of what a role is and why it's important in the source code for the Icon component: `packages/component-library/components/Icon/Icon.tsx`.

jhuangnz: Good to know! FYI I noticed the builds did pass without it (I had accidentally deleted it before but added it back in a later commit)

packages/component-library/draft/Kaizen/Modal/presets/ConfirmationModal.tsx
```diff
@@ -31,6 +33,17 @@ interface Props {
 
 type ConfirmationModal = React.FunctionComponent<Props>
 
+const getIcon = type => {
+  switch (type) {
+    case "affirmative":
```
lijuenc: where is `affirmative` coming from?

sebpearce: @jhuangnz Looks like it's called `positive` in line 25.

lijuenc: We could probably type the `type` parameter of this function with the same type as the `type` prop, and TypeScript would catch this sort of thing for us.

lijuenc: Or if `getIcon` were made a regular function instead of an anonymous arrow function, I think TypeScript might be smart enough to automatically infer the type of the `type` argument. This works in Flow but not sure about TypeScript :)

jhuangnz: 😅 good question! Wires crossed today... thanks!

jhuangnz: @lijuenc do you mean something like:
```
const getIcon = (type: Props["type"]) => {
  ...
}
```

lijuenc: Close, but that looks like invalid syntax. Something like:

```
type ModalType = "positive" | "informative" | "negative" | "cautionary"

interface Props {
  ...
  type: ModalType
  ...
}

const getIcon = (type: ModalType) => {
  ...
}
```

jhuangnz: Thanks @lijuenc. Ha! My editor actually made the suggestion! But I'm not sure if it would have worked. I have amended with the syntax you have suggested.

packages/component-library/draft/Kaizen/Modal/presets/ConfirmationModal.tsx
```diff
@@ -20,7 +22,7 @@ const styles = require("./ConfirmationModal.scss")
 
 interface Props {
   readonly isOpen: boolean
-  readonly type: "positive" | "informative" | "negative"
+  readonly type: ModalType
```
sebpearce: This is a nitpick, but calling this `Mood` would make the language consistent with how the designers talk about modals.

sebpearce: (I see you didn't introduce the word `type`, but it could be a good improvement)

packages/component-library/draft/Kaizen/Modal/presets/ConfirmationModal.scss
```diff
@@ -2,11 +2,6 @@
 @import "~@cultureamp/kaizen-component-library/styles/layout";
 @import "~@cultureamp/kaizen-component-library/styles/responsive";
 
-// override Murmur global styles :(
-.header h1 {
-  color: $white;
-}
-
```
jhuangnz: I removed this as it didn't seem to do anything. It was also interfering with my ability to change the color of the title specifically for the cautionary modal. Does anyone see a problem with removing this styling here?

jhuangnz: @sebpearce @lijuenc @ellisjas

lijuenc: The comment says that it was an override for murmur. I think murmur has styles which apply globally to all `h1` elements and that interfered with this Kaizen component's styles.

sebpearce: I would assume it does do something when you see it in Murmur rather than Storybook... @lijuenc is right, there are likely global `h1` styles in Murmur that will make it some heinous colour instead of white if you remove it.

Instead of removing it, you could try putting a class on the `h1` when it's cautionary and targeting that with `.header h1.h1NameGoesHere` maybe?

jhuangnz: Thanks @sebpearce! Just paired with @lijuenc to get it right. 🙏

PR #138, opened 2019-11-22
feat: make mobile version of navigation bar menu optional

ellisjas:
The freemium version of the navbar menu in murmur is already mobile friendly and hence does not need to change shape at lower resolutions.


ellisjas:
[Freemium responsive view breaks header](https://trello.com/c/aWeWOLAL/1696-freemium-responsive-view-breaks-header)

PR #142, opened 2019-11-28
feat: Component Library > Tag > Classes prop to add/override styles

dmisdm:
I'd like to be able to add some extra style (in my case `white-space: nowrap`) to an instance of a Tag component.
I thought id take a page out of the material-ui framework and add a `classes` prop to add classes to the components within.


dmisdm:
closed because the problem in my instance was fixed by modifying the flex behaviour of the container to grow and not shrink (`flex: 1 0`)

PR #146, opened 2019-12-03
feat: Add DatePicker component

frimoldi:


packages/component-library/package.json
```diff
@@ -32,12 +33,15 @@
...
     "@types/lodash": "^4.14.132",
+    "@types/moment": "^2.13.0",
     "@types/react-select": "^3.0.5",
     "@types/uuid": "^3.4.4",
     "classnames": "^2.2.6",
     "lodash": "^4.17.11",
+    "moment": "^2.24.0",
     "motion-ui": "^2.0.3",
     "react-animate-height": "^2.0.15",
+    "react-calendar": "^2.19.2",
```
ActuallyACat: Looks like you're not using `react-calendar` anywhere in here, this could be removed.

packages/component-library/draft/Kaizen/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const dateIcon = require("@cultureamp/kaizen-component-library/icons/date-start.icon.svg")
+  .default
+const dateRangeIcon = require("@cultureamp/kaizen-component-library/icons/date-range.icon.svg")
+  .default
+
+const styles = require("./styles.scss")
+
+export type DatePickerProps = {
+  id: string
```
ActuallyACat: Does this need an `id` to function, and if it does, could this be an optional prop?

ActuallyACat: Ah, my mistake. Looks like this is required for TextField. Disregard!

packages/component-library/draft/Kaizen/DatePicker/primitives/Calendar.tsx
```diff
@@ -0,0 +1,118 @@
+import { Button, Text } from "@cultureamp/kaizen-component-library"
```
ActuallyACat: You're not using the Kaizen button in here, this can be removed.

ActuallyACat:
/nit There's nothing stopping the user from selecting a second date that is before the first date. For example:
Screen Shot 2019-12-16 at 9 58 53 am

ActuallyACat:
/nit /feature You may want to allow the developer to set an initial date/date rate, for when a customer returns to a prefilled form. Happy to move this to be a issue in Github if this component is required sooner.

frimoldi:
Thank you for the feedback, @ActuallyACat !
I'll come back to this PR with updates once we meet our deadline for Unilever in January

ActuallyACat:
No problemo 👌

ActuallyACat:
Deciding to close this but keep the branch as DST continues to research an accessible and generic date picker that will work for all teams.

PR #149, opened 2019-12-06
feat: Added types to colours 

ActuallyACat:
Added types to colours and a helper function so that consumers don't need to worry about the structure of the colour token. 

**The good**
Nice type hints for the available colours
Screen Shot 2019-12-06 at 2 55 55 pm

**The bad**
The colour scale is different depending on what colour you are trying to access. The hints do not reflect that
Screen Shot 2019-12-06 at 2 56 12 pm


sebpearce:
@ActuallyACat PR looks great.

How would you feel about an alternative way of doing the types for the colour scales? I'm wondering if it'd be more "semantic" to define an interface (or just a type?) for each number e.g. `200`, and then do something like
```
type Wisteria = x100 | x200 | x300 | x400 | x500 | x600 | x700 | x800
type Cluny = x100 | x200 | x300 | x400 | x500
```
This is a nitpick though, and only because it felt a bit weird seeing the darker colours referred to as an extended colour scale.

ActuallyACat:
> @ActuallyACat PR looks great.
> 
> How would you feel about an alternative way of doing the types for the colour scales? I'm wondering if it'd be more "semantic" to define an interface (or just a type?) for each number e.g. `x200`, and then do something like
> 
> ```
> type Wisteria = x100 | x200 | x300 | x400 | x500 | x600 | x700 | x800
> type Cluny = x100 | x200 | x300 | x400 | x500
> ```
> 
> This is a nitpick though, and only because it felt a bit weird seeing the darker colours referred to as an extended colour scale.

Yeah, I'm not really seeing the benefit of this approach anymore. I wonder if going one step simpler would be better? 

```
type Wisteria = "100" | "200" | "300" | "400" | "500" | "600" | "700" | "800"
```

ActuallyACat:
Not perfect, but better.
```
type Cluny = Pick<ColorScale, "100" | "200" | "300" | "400" | "500">
```

zioroboco:
```ts
import { kz as tokens } from "./color.json"

export type ColorTokens = typeof tokens.color
export type ColorNames = keyof ColorTokens

export const color: <N extends ColorNames>(
  name: N,
  variant: keyof ColorTokens[N]
) => ColorTokens[N] = (n, v) => tokens.color[n][String(v)]
```
See https://cultureamp.slack.com/archives/CETLTFG9G/p1579001266057400

PR #150, opened 2019-12-06
feat: Update color docs with Zen palette

sebpearce:
- Creates a new set of palette showcase components for Zen.
- Uses contrast checking to determine whether the text on a colour item should be white or dark.
- Moves legacy palette stuff into a `legacy/` dir.
- Creates individual components for 3 of our palette colour groups (Cluny, Wisteria, Neutrals). The intention is to add more once design documentation has been made for them.

Note that we need to use individual components for each group because we have chosen to avoid using props in our Markdown docs, so we prefer
```js
<ZenCluny />
```
to
```js
<ZenColorGroup color="cluny" />
```
This makes them easier to work with for non-tech contributors, and also simplifies the issue of the Neutrals color group which is unlike the others.

## Important

Types have not yet been added, and should be. I chose to wait because we have an [open PR](https://github.com/cultureamp/kaizen-design-system/pull/149) that will add types for our colours from within the `kaizen-design-tokens` package. I would prefer to use those types rather than duplicate them here. Whether we want to merge this without types or wait until the other PR has merged and update this before merging is up for discussion.

image


sebpearce:
[Update Colors page on Kaizen Site](https://trello.com/c/e0tsL8qT/588-update-colors-page-on-kaizen-site)

site/src/docs-components/color/ZenColor.tsx
```diff
@@ -0,0 +1,56 @@
...
+const ZenColor = ({ name, value }) => {
+  const borderStyles = value === colorTokens.kz.color.white ? withBorder : {}
+  const textStyles = shouldUseWhiteText(value)
+    ? { ...defaultTextStyles, ...withWhiteText }
+    : defaultTextStyles
+  return (
+    <div
+      style={{ ...containerStyles, ...borderStyles, backgroundColor: value }}
+    >
+      <span style={{ ...textStyles }}>{name}</span>
```
ActuallyACat: nit/ this could also be `<span style={textStyles}>{name}</span>`, similar to line 51

site/src/docs-components/color/ZenColorGroup.tsx
```diff
@@ -0,0 +1,21 @@
...
+  gridColumnGap: "2%",
+  justifyContent: "space-between",
+}
+
+const ZenColorGroup = ({ colors }) => (
+  <div style={colorGroupStyles}>
+    {Object.keys(colors)
+      .reverse()
+      .map(c => (
+        <ZenColor name={c} value={colors[c]} />
```
ActuallyACat: Is this complaining about a missing `key` prop?

site/src/docs-components/color/ZenColor.tsx
sebpearce: Good catch, that was a case of "expecting to add something else here later but didn't end up doing it".

site/src/docs-components/color/ZenColorGroup.tsx
sebpearce: Oh, weird... it's not, no error in the console on the branch preview nor in the editor. 🤔 Maybe the `name` prop is weird because of the HTML attribute `name` and that makes it unique or something?

sebpearce: Anyway I'll add it!

PR #152, opened 2019-12-06
fix: Fix typo in README for tokens import in JS

sebpearce:
This removes an `@` from the start of the import statement in JS, which was likely there after being copied from the Sass import and modified.


sebpearce:
[Fix typo in README for JS import of KDT](https://trello.com/c/QywKHK4x/660-fix-typo-in-readme-for-js-import-of-kdt)

sebpearce:
Just realised that should've been `docs:` and not `fix:` 🤦‍♂ Ah well.

PR #161, opened 2019-12-13
feat: add noBottomMargin prop to inlineNotification

dougmacknz:
I'm working on something where I could use this component but need to remove the margin-bottom.

image

This PR just adds a prop to the React version of `InlineNotification` to allow us to remove the bottom margin.

### DIFF
--- a/packages/component-library/components/Notification/InlineNotification.tsx
+++ b/packages/component-library/components/Notification/InlineNotification.tsx
@@ -13,6 +13,7 @@ type Props = {
   hideCloseIcon: boolean
   onHide?: () => void
   automationId?: string
+  noBottomMargin?: boolean
 }
 
 const InlineNotification = ({
--- a/packages/component-library/components/Notification/_styles.scss
+++ b/packages/component-library/components/Notification/_styles.scss
@@ -122,6 +122,12 @@ $ca-notification-slide-right: transform $ca-duration-slow ease-out;
       opacity: 1;
     }
   }
+
+  // Modifiers
+
+  &%ca-notification---noBottomMargin {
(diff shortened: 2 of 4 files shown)


sebpearce:
@ActuallyACat This is a thing that has come up a lot before. Our running policy has been that things have a default bottom margin and a prop (`noBottomMargin`) to turn it off.

It's part of a bigger conversation around whether components should be in control of their exterior layout/positioning. I can see a conflict here, because I recently explained to someone that needing to wrap a button component for layout purposes was intentional, i.e. buttons don't control how they're positioned on the page at all, which runs counter to the "components should have a default bottom margin" policy. :/

For now in this case, I think adding an optional prop to turn it off is the safest way to go since it'll leave the current ones unaffected.

But we should surface this topic again for chats!

ActuallyACat:
> @ActuallyACat This is a thing that has come up a lot before. Our running policy has been that things have a default bottom margin and a prop (`noBottomMargin`) to turn it off.
> 
> It's part of a bigger conversation around whether components should be in control of their exterior layout/positioning. I can see a conflict here, because I recently explained to someone that needing to wrap a button component for layout purposes was intentional, i.e. buttons don't control how they're positioned on the page at all, which runs counter to the "components should have a default bottom margin" policy. :/
> 
> For now in this case, I think adding an optional prop to turn it off is the safest way to go since it'll leave the current ones unaffected.
> 
> But we should surface this topic again for chats!

Makes sense in this context, for now. Let's come back to this one in the future.
LGTM, feel free to merge @dougmacknz

dougmacknz:
^ thank you

PR #164, opened 2019-12-16
docs: Make all md links underlined except on hover

sebpearce:


sebpearce:
[Make links underlined when in a paragraph on Kaizen Site](https://trello.com/c/zfjPLFQn/670-make-links-underlined-when-in-a-paragraph-on-kaizen-site)

PR #168, opened 2019-12-16
feat: Add confirmation variant to elm modal

Confidenceman02:
## Background
The elm modal has only had the GenericModal variant available. To bring it closer to its React counterpart the first confirmation variant has been added. Having the new variant created makes it easier for anyone to go in and add the other missing confirmation variants due to the code patterns being basically identical.

Along with this the elm modal was using its own css files which lived along side the react stylesheets. This PR also combines many of the styles to better unify the styling between both elm and react modals. Whilst the respective views are built very similarly, there are some slight differences in how the styling is applied.

## Work completed
Informative confirmation variant of the elm modal

## Improvements
Accessibility labels were not added to the elm modal which are found in the react version.
Sensible focus sites on open have not been built out.
Building out all other modal variants.


packages/component-library/draft/Kaizen/Modal/Modal.elm
```diff
@@ -131,33 +143,56 @@ view (Config config) =
...
-                GenericModal.view
-                    [ modalBox content size ]
-                    (genericModal |> GenericModal.events genericModalEvents)
+                GenericModal.view (GenericModal.Custom size)
+                    content
+                    (genericModalConfig |> GenericModal.events genericModalEvents)
+
+            Confirmation confirmationType configs ->
+                let
+                    withOnDismiss confirmationConfig =
```
juliohr: what do you think about extracting this to external functions to be piped like `confirmLabel`, `dismissLabel`, etc?

packages/component-library/draft/Kaizen/Modal/Modal.elm
```diff
@@ -243,14 +278,20 @@ mapDuration duration =
 
 
 type Variant msg
-    = Generic (List (Html msg)) Size
+    = Generic (List (Html msg)) ( Float, Float )
+    | Confirmation ConfirmationType (ConfirmationConfig msg)
```
juliohr: what are the benefits of having confirmation type outside confirmation config?
would putting inside simplify things?

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.scss
```diff
@@ -13,6 +13,11 @@
   max-width: 600px;
 }
 
+.elmModal {
```
juliohr: why did you call `elmModal`? is there any other modal overlapping?

packages/component-library/draft/Kaizen/Modal/Modal.elm
Confidenceman02: `ConfirmationModal.elm` does have those functions already and Modal.elm does in fact call them. This config is more so a contract that `Modal.elm` needs for the variant. So when a consumer asks for a Confirmation variant from `Modal.elm` this config is the required data Modal needs in order to use `ConfirmationModal.elm`. You could call `ConfirmationModal.elm` directly if you wanted to build some sort of hybrid modal and use those functions you are talking about. We want to discourage people from doing that though.

packages/component-library/draft/Kaizen/Modal/Modal.elm
Confidenceman02: No real benefits, it could potentially live inside the config!

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.scss
Confidenceman02: Yes there is a `.modal` style and it does differ in style implementation a bit. `.elmModal` does compose `.modal` but it adds a few different styles required by the elm version of Modal. This is mainly due to elm Modal being built around the same time as React version. Styling wise they are identical but there are a few minor differences.

PR #169, opened 2019-12-17
Ally/security fix

ActuallyACat:
Update :allthethings: 

There was a security issue in `serialize-javascript`, brought in by:
webpack -> terser-webpack-plugin -> serialize-javascript

Read more here: https://github.com/webpack-contrib/terser-webpack-plugin/issues/188


ActuallyACat:
Stale + irrelevant now

PR #174, opened 2019-12-18
fix: apply fastclick "opt-out" classname to enable React Select to work on mobile

okeeffed:
React Select does not work unless the option has `needsclick` attached.

Here I have manually hardcoded it because I am sure there are bountiful Select components that need this applied across the main CA repos, but am happy for feedback to change this to a prop and manually add the prop to each instance of `Select`.


packages/component-library/draft/Kaizen/Select/Select.tsx
```diff
@@ -64,7 +64,7 @@ const Menu: typeof components.Menu = props => (
 const Option: typeof components.Option = props => (
   <components.Option
     {...props}
-    className={classNames(styles.option, {
+    className={classNames("needsclick", styles.option, {
```
ActuallyACat: Can you drop a TODO in here? Something like:

`TODO - needsclick class disables fastclick on this element. Remove when fastclick is removed from consuming repos`

PR #178, opened 2019-12-18
fix: Disable fastclick so mobile devices don't require double click

okeeffed:
Hello! Ran into two more components this morning that fall victim to `fastclick` - not the worst outcome, but they require double tapping to select.

If this is the desired effect, please let me know! Gif attached. Might be hard to see, but you can see the first tap does not select anything (but some styles update) and requires a double click to activate. 

Not sure how the double tap impacts when it comes to those using features like `VoiceOver` but also worth considering. 

When I am off the simulator and using my device for QA swarm this afternoon, I will investigate the VoiceOver usage.

double-clicking

### DIFF
--- a/packages/component-library/draft/Kaizen/Form/Primitives/Checkbox/Checkbox.tsx
+++ b/packages/component-library/draft/Kaizen/Form/Primitives/Checkbox/Checkbox.tsx
@@ -4,6 +4,7 @@ const checkIcon = require("@cultureamp/kaizen-component-library/icons/check.icon
 const minusIcon = require("@cultureamp/kaizen-component-library/icons/minus.icon.svg")
   .default
 import * as React from "react"
+import classnames from "classnames"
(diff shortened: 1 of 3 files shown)


packages/component-library/draft/Kaizen/Form/Primitives/Checkbox/Checkbox.tsx
```diff
@@ -56,7 +57,8 @@ const Input: Input = ({
       data-automation-id={automationId}
       // This si only used as a handle for unit testing
       data-indeterminate={checkedStatus === "mixed"}
-      className={styles.checkbox}
+      // TODO - needsclick class disables fastclick on this element to prevent double tap on mobile. Remove when fastclick is removed from consuming repos
```
ActuallyACat: :D Thanks for this

PR #181, opened 2019-12-19
feat: Add `boxOffset` property to the `Popover` component

jeremy8883:
Add `boxOffset` property to the `Popover` component. It will shift the popover box a specified amount, without moving the popover arrow.

image

*(the snapshot above is slightly off centre, but that has now been fixed)*

This task links to a requirement of ticket CORE-348 in perform. We needed to show a popover, have the popover arrow centered to the sibling button, but the box to not go past the main body margin. This new property will give us the flexibility to achieve this.

Side updates:
* Removal of some dead scss code in the `arrow` mixin
* Removal of the `position` `left` and `right` options, since they aren't yet implemented.
* Separate the popover arrow to its own element, as to give more flexibility on how it is positioned.

Notes:
One of the ugly things that I've introduced is some fragmentation of styling. Because we now allow for manually updating the position via javascript, we are forced to have some styling in javascript, and some in scss. I'm not sure the best way to get around this, other than going all in on javascript styles (which I would be very happy to do 😛).

I've gone through all of the storybook stories, to make sure there are no visual regressions, and it all looks good on my end.

### DIFF
--- a/packages/component-library/draft/Kaizen/Popover/Popover.tsx
+++ b/packages/component-library/draft/Kaizen/Popover/Popover.tsx
@@ -66,7 +66,7 @@ const Popover: Popover = forwardRef<HTMLDivElement, Props>(({
     <div className={mapVariantToBoxClass(variant)}>
       <div className={styles.header}>
         {variant !== "default" && (
-          <span className={styles.icon}>
+          <span className={classNames(styles.icon, mapVariantToIconClass(variant))}>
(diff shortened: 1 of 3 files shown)


ActuallyACat:
Looks like there's some missing styled on the icons in the popover header

Screen Shot 2019-12-19 at 3 50 01 pm
Screen Shot 2019-12-19 at 3 49 59 pm

jeremy8883:
Oops! Good pickup, all fixed.

PR #182, opened 2019-12-19
fix: Increase z-index of close icon to prevent it being hidden

ellisjas:


ellisjas:
[Increase z-index of close icon in modal](https://trello.com/c/MtIKdFyj/1820-increase-z-index-of-close-icon-in-modal)

PR #183, opened 2019-12-19
feat: Create loading Collapsible components

dougmacknz:
Creates a loading version of the `Collapsible` & `CollapsibleGroup` components using the existing `LoadingPlaceholder` component and the existing styles for the `Collapsible` component.

This is a new pattern for Kaizen (creating loading versions of existing components). I've chatted about this with @didoesdigital, and briefly with @ActuallyACat & @sebpearce and we seem to be happy with the introduction of this. We're not totally sure how many other components this might apply to as well but `Table` seems like a good guess at least.

Di & I will write up some documentation for this once we're happy with this and it passes through.

Storybook examples:
https://dev.cultureamp.design/dug/loading-collapsibles/storybook-static/?path=/story/collapsible--loading-collapsible-group

Example usage in Murmur (see loaders at the bottom):
loading-collapsibles-2


ActuallyACat:
Have you considered a loading prop on the component? I looked through the components and there's no clear stance on preference for loading on props or having a separate "loading" component. So this is open for discussion :D 

I think this will mostly come down to preference, and I'm interested to hear if anyone has anyone in the practice has strong opinions on this. I slightly prefer a loading prop - It's a little more succinct when using it. Thoughts?

`
const Example = ({ loading, content }) => (
    <Collapsible loading={loading}>
      {content}
    </Collapsible>
)
`
vs 
`
const Example = ({ loading, content }) => (
  <div>
    {loading && <LoadingCollapsible />}
    {!loading && (
      <Collapsible>
        {content}
      </Collapsible>
    )}
  </div>
)
`

sebpearce:
☝️ This is a good point. It would make it more of a self-contained unit. Hmmmmmmm.

packages/component-library/draft/Kaizen/Collapsible/Collapsible.tsx
```diff
@@ -11,125 +12,163 @@ const chevronUp = require("@cultureamp/kaizen-component-library/icons/chevron-up
...
+  | {
+      loading: true
+      separated?: boolean
+      group?: boolean
 
-  /* Will avoid rendering the content until required (especially important when you have queries inside sections). Removes animation. */
-  lazyLoad?: boolean
+      /* This won't actually do anything in this context, but it's required to keep
+         the internalOpen state hook at the top of the function below */
+      open?: boolean
```
dougmacknz: The other option here is to remove this pattern and allow all props along with `loading`. Neither is ideal, I decided to go with this to stay consistent with `CollapsibleGroup`, which has more reason for this pattern because of the `collapsibles` prop that only applies on `loading`. I could go either way on this.

dougmacknz:
> Have you considered a loading prop on the component?

Ok so having explored both options now I feel quite strongly that we should stick with separate components. I'll do my best to explain my reasoning:

* Putting it into the same component makes prop types & usage more confusing and difficult then it needs to be.
Eg. On the `CollapsibleGroup` component I want a `collapsibles` prop for the loading version only. This means that you need to adjust the prop types based on whether `loading` is true or false. You're basically dealing with trying to squeeze two different components into one component at that point.
image

* I'm having a hard time imagining a case where you don't need or want to use a conditional to show a loader. In real life situations: you're not going to know how many items, or any details of items in the loading state. So even with a `loading` prop on the same component, I would still imagine you need some of conditional:
```
{loading && <Collapsible loading={true} />}
{!loading && <Collapsible id="1" title="some data" />
```

* With the way we have things structured - I'm not sure if it makes any difference but it's possible there's a small performance improvement by not having them contained in the one component. Eg. if you need to use a `Collapsible` but not a `LoadingCollapsible` - there's a bit of code that doesn't need to be loaded there.

All in all, I suppose I just discovered that a loading placeholder for a component and the component are two separate entities with their own concerns, trying to make them into a single entity makes life difficult.
I did push through hoping that even though it makes the component code more complicated, maybe it would make for easier usage. I'm not sure that that's the case though because now you have props that only apply in certain situations.

> If we scaled this, there would be 2 components for every component – the main one, and the loading one. Could get a bit cluttered.

Looking through the current components in Kaizen, I don't think there's too many others that are actually going to need `loading` versions. I think `Table` and `Tabs` were the only other ones that stand out to me as probables for this.

dougmacknz:
I had a chat to @didoesdigital about the two possible patterns. She raised a concern about possible duplicate data when implementing this pattern elsewhere.

This is an example of a bad implementation on a loading `Table` component:

```
{loading && <LoadingTable firstColHeading="Resource name" secondColHeading="Supplementary Information" />}
{!loading && <Table firstColHeading="Resource name" secondColHeading="Supplementary Information" />
```

The way I see this working instead would be something like this, where only the required component is turned into a `Loading` version
```
<Table>
  <TableHeader>
    ...
  </TableHeader>
  {loading && <LoadingTableBody cols="3" rows="3" />}
  {!loading && 
    <TableBody>
      ..
    </TableBody>
  }
</Table>
```

dougmacknz:
Closing this after a chat with @ActuallyACat, until further discussions have happened around the standard we want for loading components. For now I'm going to make loaders for my case alongside the usage in Murmur (I will the link the PR here for reference shortly)

PR #185, opened 2019-12-23
docs: update contributing docs the steps to add elm stories

juliohr:
I recently added elm stories, but couldn't see them showing on the storybook. I spent a while troubleshooting, digging docs, without success. I was only able to fix the issue after talking to an old contributor.
This PR documents an extra step you have to do after adding an elm story.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,6 +101,22 @@ You can also clean up generated files:
 
 `yarn clean`
 
+### Elm components stories
+Besides creating your Elm story on an Elm file, there is an extra step in order to make it appear on the storybook. 
+You have to refer it on the component's JS story.
+
+```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -65,6 +65,22 @@ You can also clean up generated files:
...
+
+```
+-- MyComponent.tsx
+import { loadElmStories } from "@cultureamp/elm-storybook"
+
+// JS stories
+
+loadElmStories("MyComponent (Elm)", module, require("./MyComponent.elm"), [
+  "Default",
+  "Default /w icon",
```
didoesdigital: To give more visibility here to the engineer's current task of adding a story, can we rename one of these to something more prominent? E.g. "Your New Story Name"?

PR #194, opened 2019-12-30
feat: component library zen colors uplift

mbylstra:
### Overview

The main motivation for this PR is to upgrade Ocean to Cluny for buttons and links in the Kaizen Component Library. This is causing jarring clashes in consuming websites that have already been upgraded to Zen colors.

 [Color tokens in Kaizen Component Library: Ocean -> Cluny buttons](https://trello.com/c/9REVJgDM/680-color-tokens-in-kaizen-component-library-ocean-cluny-buttons)

### Solution

*before*: https://cultureamp.design/storybook/
*after*: https://dev.cultureamp.design/dst/zen-colors-uplift/storybook/

### Implementation overview

The colors have been automatically upgraded using our Original to Zen color mapping. So, although the main requirement is to upgrade Ocean to Cluny, all of the colours have been upgraded anyway via the automation. What we haven't done is manually matched the new Zen UI Kit perfectly. We have however fixed contrast issues where they existed (eg background too dark on). We are not aware of any remaining AA constrast issues.

#### Button component
This has been given to most effort to match the new Zen UI Kit colors (such as hover, disabled and focus states). However:
- we kept the the primary button Seedling (it's Wisteria in Zen UI Kit), to ease to the transition to the new colors, as buttons are very widespread. Upgrading to Wisteria would be a follow up card.
- we kept the reversed buttons transparent background on coloured text. In Zen UI Kit, the reversed buttons will be coloured background with transparent text. This was too big a change considering the scope of this PR.
- Icon Buttons remain dark wisteria, rather than upgrading to Cluny like in Zen UI Kit (which also has circular background hover and focus states). This is quite a big change, and the space needed for the background circle when hovering may be bigger than current implementations allow for (so this out of scope for now)

#### Tag component
Should match the colors in  Zen UI Kit

#### styles/color.scss
- The file styles/color.scss is now legacy, and contains the legacy colors. They have not been removed as some web sites may still be using them. Let's make removing them a follow up task.
- The legacy colours and the `add-tint` and `add-shade` mixins are no longer used _within_ Kaizen Component Library.
- It also contains existing "decision tokens" such as `$ca-link-hover-text-color`. As `kaizen-design-tokens` does not yet contain any decision tokens, they will remain. _some_ of the decision tokens have been upgraded to zen:

```
$ca-base-text-color
$ca-link-text-color
$ca-link-hover-text-color
```
And the rest have not, because either it's unclear if they will exist in Zen, or they are too risky to change:

```
$ca-palette-positive-text
$ca-palette-negative-text
$ca-border-color
$ca-border-between-rows-color
```

For these, let's try to fix any thing that doesn't pass design review.

#### The rest of the components
Have had some minor background color updates  - see code changes.


mbylstra:
[Color tokens in Kaizen Component Library: Ocean -> Cluny buttons](https://trello.com/c/9REVJgDM/680-color-tokens-in-kaizen-component-library-ocean-cluny-buttons)

packages/component-library/package.json
```diff
@@ -29,6 +29,7 @@
     "react": "^16.9.0"
   },
   "dependencies": {
+    "@cultureamp/kaizen-design-tokens": "^0.1.2",
```
ActuallyACat: The tokens package needs to be a peer dependency, as we want to tell all consumers to also add tokens as a  dependency whenever they use Kaizen. This is a really good read: https://nodejs.org/es/blog/npm/peer-dependencies/

packages/component-library/components/Button/styles.scss
```diff
@@ -51,19 +52,18 @@ $caButton-verticalPaddingForm: calc(
...
+    border-color: $kz-color-wisteria-500;
+    color: $kz-color-wisteria-800;
 
-    &:hover,
-    &:active {
-      background: $ca-palette-stone;
-      border-color: add-tint($ca-palette-ink, 60%);
+    &:hover {
+      background: $kz-color-wisteria-100;
+      border-color: $kz-color-wisteria-700;
```
ActuallyACat: Beautiful attention to detail 😍 I notice that the border colours and their states (hover/active/etc) align to the new Zen UI kit.

ActuallyACat:
Mega nitpicking
I notice there's no hover state on the primary button anymore. `<Button label="Label" primary={true} />` 
Same for The primary reversed button, which also looks like it's disabled in this state: `<Button label="Label" primary={true} disabled={false} reversed={true} />`

I only bring these up because they may be a usability issues

mbylstra:
> Mega nitpicking
> I notice there's no hover state on the primary button anymore. `<Button label="Label" primary={true} />`



Oh, good pickup. I remember why I did this now (and forgot to document it): the Original button was lighter than Seedling-500, which got converted to Seedling-400 which is even lighter, and in my opinion made the button look like it was disabled (and also did not meet AA color contrast). I didn't think of using Seedling-600 for the hover state. Now I've added a commit that uses Seedling-500 for inactive state, Seedling-600 for hover and Seedling-600 for active. You can distinguish the colors between hover and active, but the button also dips down on the active state @didoesdigital : I'll leave it to you for the design review, to see if you think this is a good compromise, or if you have any other ideas.

> Same for The primary reversed button, which also looks like it's disabled in this state: `<Button label="Label" primary={true} disabled={false} reversed={true} />`

The background should change from white to a light grey background on hover. This should be unchanged from Original implementation (although a slightly different shade of grey). Do you mean it looks like it's disabled when you aren't hovering, or when you are? (It doesnt' look disabled to me when not hovering, but does look similar to the disabled state when hovering). It's hard to compare against the original Storybook, because there was a bug in the original story, where the story said "Primary Reversed", but the parameter had `disabled={true}`. I intended to leave this alone as much as possible, as the Zen reversed primary button is very different from Original (out of scope).

packages/component-library/package.json
mbylstra: I made a commit with the peer dependency as `0.x` (forgetting that you'd pasted some code), thinking I should restrict it to the 0 major version in case we made some breaking changes to the tokens. What do you think?

didoesdigital:
@mbylstra 

"Secondary" button and "Secondary w/ Icon" active states background should be Cluny 100, not like this:
image

didoesdigital:
After discussing the trade-offs with @mbylstra of the Primary button hover state, we've decided to use the original hex/rgb values for the default/inactive and hover/focus and active states without `add-tint` or `add-shade` instead of the token value as an interim measure until we can switch them to Wisteria.

didoesdigital:
Icon in destructive button should be Coral 600:
image

didoesdigital:
"Secondary Reversed Disabled w/ Icon" hover background should be Coral 100:
image

didoesdigital:
@mbylstra this destructive action hover state in Menu List is similar to what exists at the moment. If it's a fast change, it could be updated to match the secondary destructive button styles above (hover style is Coral 600 Icon/Text and Coral 100 background):

image

packages/component-library/package.json
ActuallyACat: That's a good call, definitely the right choice.

didoesdigital:
@mbylstra can we drop this bullet/disc down to Wisteria 700?

image

didoesdigital:
@mbylstra the "closed" tag is the only one of the Status tags that does not have a matching border and background color. The background color and border color should be Wisteria 200:

image

PR #200, opened 2020-01-03
fix: Specify ie11 for @babel/preset-env

JakePitman:
This PR adds makes storybook work in ie11 by adding @babel/preset-env.

In a previous PR (https://github.com/cultureamp/kaizen-design-system/pull/198), I added the name of the preset only, and this led to a break when running `yarn gatsby develop` (Which uses setBabelPreset to add an options object with a `plugins` field to each preset - which @babel/preset-env doesn't accept)

To address this break, this PR also adds a specific options object to @babel/preset-env (specifying ie11), and some logic in gatsby-node.js to check if a preset already has it's own options (and use that instead if it exists)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,14 @@
   },
   "babel": {
     "presets": [
-      "@babel/preset-env",
+      [
+        "@babel/preset-env",
+        {
+          "targets": {
+            "ie": "11"
+          }
+        }
+      ],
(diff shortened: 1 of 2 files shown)


ActuallyACat:
LGTM - though I don't have a VM with IE11 so I can't verify it properly.

PR #209, opened 2020-01-08
feat: Chromatic QA integration (with Github Action)

mbylstra:
This integrates with Chromatic for visual regression testing of the Kaizen Storybook stories.
It's currently using a trial account (30 days) using my personal github account. This PR using a Github Action for the CI server to generate the visual regression snapshots that are sent to Chromatic (as opposed to Buildkite). This is basically ready to go, but would require approval from security to use Github Actions as opposed to Buildkite (what we normally use).

As the Chromatic app has not been given Oauth access to the Culture Amp organisation, there is no interactive Github integration. You can look into the chromatic build, and click a url inside the log to get to the Chromatic build, and then approve/reject changes, but the github UI does not update when you approve changes. All it shows is that there weren't any errors with the Chromatic build (not whether it was approved or not)
image

To see what it looks like with the Oauth intergration, have a look at https://github.com/mbylstra/kaizen-design-system-chromatica-trial/pull/1, which is a fork of the kaizen-design-system under my personal github account (with the buildkite config removed).
image

Notes:

- I needed to use a very specific node js version ('10.16.3') in the GIthub action, or the build would fail. It might be hard to remember to change this if we update Node (although the build would break, so it should be pretty clear from the Chromatic error message)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -5,15 +5,15 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Use Node.js
-      uses: actions/setup-node@v1
-      with:
-        node-version: '10.16.3'
-    - name: Install dependencies
-      run: yarn setup
-    - uses: chromaui/action@v1
-      with:
-        appCode: azwytw4amgi
-        token: ${{ secrets.GITHUB_TOKEN }}
-        buildScriptName: storybook:build
+      - uses: actions/checkout@v1
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10.16.3"
+      - name: Install dependencies
+        run: yarn setup
+      - uses: chromaui/action@v1
+        with:
+          appCode: azwytw4amgi
+          token: ${{ secrets.GITHUB_TOKEN }}
+          buildScriptName: storybook:build
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -14,6 +14,6 @@ jobs:
         run: yarn setup
       - uses: chromaui/action@v1
         with:
-          appCode: azwytw4amgi
+          appCode: ${{ secrets.CHROMATIC_APP_CODE }}
           token: ${{ secrets.GITHUB_TOKEN }}
           buildScriptName: storybook:build
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,19 @@
+name: "Chromatic"
+on: push
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.16.3'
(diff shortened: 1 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,19 @@
...
+    - uses: actions/checkout@v1
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.16.3'
+    - name: Install dependencies
+      run: yarn setup
+    - uses: chromaui/action@v1
+      with:
+        appCode: azwytw4amgi
```
mbylstra: I intended to use `${{ secrets.CHROMATIC_APP_CODE }}` and put the value in the Github settings for repo, but I don't seem to have the right level of access to do that. The code is for a free trial account. We'd creata new account and/or change the key before going to production, and also delete this account/"chromatic project".

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: actions/checkout@v1
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10.16.3"
+      - name: Install dependencies
+        run: yarn setup
+      - uses: chromaui/action@v1
+        with:
+          appCode: ${{ secrets.CHROMATIC_APP_CODE }}
```
mbylstra: Note: this secret is manually added in the github settings for the repo (requires the right level of access). There is no way to do automatic key rotation with Chromatic (you need to email and get them to manually change the API key if you want to rotate it)

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: "10.16.3"
```
mbylstra: Note: a very specific version of node is required, or the build won't build.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Install dependencies
+        run: yarn setup
```
mbylstra: Note: this crucial step (yarn or npm) is not mentioned in the Chromatic github action readme (https://github.com/chromaui/action). I got most stuck on this, as the error message is very confusing if you don't do this! I will get them to update their docs.

.github/workflows/chromatic.yml
zioroboco: @didoesdigital or @ActuallyACat would be able to set up the secret.

.github/workflows/chromatic.yml
zioroboco: This restriction is set in the root `package.json`:
https://github.com/cultureamp/kaizen-design-system/blob/master/package.json#L7-L9

mbylstra:
Not actually ready for review, I just wanted to get a screeshot of what it would look like when it's ready for review!

mbylstra:
Closed in favor of https://github.com/cultureamp/kaizen-design-system/pull/224.

We'd prefer to stick to Buildkite, even though the Github Action was easier to implement

PR #210, opened 2020-01-09
half failed attempted at automatically upgraded storybook stories

mbylstra:
it failed half way, but at least some of them were upgraded


mbylstra:
Managed to do this completely automaticlaly in the end in https://github.com/cultureamp/kaizen-design-system/pull/253 🎉

PR #219, opened 2020-01-14
feat: Add ability to specify if elm popover has variable height and apply relevant styles

cset00:
## Context
https://cultureamp.slack.com/archives/CETLTFG9G/p1578874817025300

TLDR: 
Popover’s styles prevent it from being positioned above an element when the contents of the popover have an unpredictable height.

We've decided to put it behind a prop for now because doing a major release means we have to update all existing implementations of Popover which will take a bit of time.

Example screenshots:
Screen Shot 2020-01-13 at 10 23 32 am
Screen Shot 2020-01-13 at 10 23 26 am

## Changes in this PR
I've added the ability for us to set a `withVariableHeight` as `True` and get the css attribute we need. This defaults to `False` and shouldn't affect any existing usages of Popover.elm.

### DIFF
--- a/packages/component-library/draft/Kaizen/Popover/Popover.elm
+++ b/packages/component-library/draft/Kaizen/Popover/Popover.elm
@@ -14,10 +14,12 @@ module Kaizen.Popover.Popover exposing
     , withContent
     , withHeading
     , withTipPosition
+    , withVariableHeight
     )
 
 import CssModules exposing (css)
 import Html exposing (Html, button, div, span, text)
+import Html.Attributes
 import Html.Events exposing (onClick)
 import Icon.Icon as Icon
 import Icon.SvgAsset exposing (svgAsset)
@@ -40,6 +42,7 @@ type alias ConfigValue msg =
     , heading : Maybe String
     , variant : Variant
     , tipPosition : ( Side, Position )
+    , forVariableHeight : Bool
     }
 
 
@@ -53,6 +56,7 @@ defaults =
     , heading = Nothing
     , variant = Default
     , tipPosition = ( Bottom, Center )
+    , forVariableHeight = False
     }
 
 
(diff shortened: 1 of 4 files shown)


sebpearce:
@cset00 

> I've added the ability for us to set a custom attribute to the Popover Elm component.
> This will allow me to set a class where I'm using Popover to have bottom: 100%

Do you expect a range of different values being passed in, or just `bottom: 100%`? If it's the latter maybe it'd be better to just have a boolean prop that turns on that particular rule. 🤔

cset00:
Ah yep! That was what I was a bit concerned about (allowing people to just add any custom attributes). Having a bool sounds good - i'll update the PR. Just to confirm, both your suggestions are the same right? @mbylstra @sebpearce 

I think i get it, but i'll ping you again once i've updated it. Thanks for taking a look!

sebpearce:
>  both your suggestions are the same right

Looks that way to me, yep.

cset00:
Also note that I'm getting this error in screenshot when i try to run storybook.. Am i not doing the storybook code properly? 😬 

Screen Shot 2020-01-14 at 10 24 58 pm

cset00:
I'm closing this PR because we've decided to adjust the design instead so this will no longer block us. We're putting the popover at the bottom of the button so we don't have to worry about aligning the popover at the bottom. Thanks for the help!

sebpearce:
@cset00 Roger. By the way, try running `yarn clean` in `kaizen-design-system` before running Storybook – as far as I know that error you posted is usually a result of cached compiled stuff hanging around, not a result of your code.

mbylstra:
Hi. Sorry, I've been out sick. Yeah, I think we meant the same thing (I didn't see Seb's comment when I was drafting mine - Seb's was much shorter and to the point!)

BTW, that was the exact same solution (putting the popover underneath instead of on top) we came up with in my former team was blocked by this issue :) 

I think there's consensus that we'll be replacing the internals (and possibly changing the API) so this won't be an issue some time in the future (for React or Elm), so I think what you've done is a totally fine temporary solution.

PR #220, opened 2020-01-14
WIP: Migrate packages to `@kaizen` scope

zioroboco:
Re-releases all packages under their own `@kaizen` module scope, freeing up the `@cultureamp` scope for internal use. Will reset all package versions to `1.0.0`.

### todo:
- [x] Add @cultureamp-ci service account to [`kaizen`](https://www.npmjs.com/org/kaizen) npm org
- [x] Write a codemod to update consuming projects
- [ ] Update branch and reapply changes
- [ ] Release with a `BREAKING CHANGE` annotation


packages/component-library/package.json
```diff
@@ -21,7 +21,7 @@
     "!tsconfig.dist.json"
   ],
   "sideEffects": false,
-  "version": "17.6.2",
+  "version": "0.0.0",
```
ActuallyACat: Why is this 0.0.0 not 1.0.0? Will the build process automagically take care of the bumping here? If so, should this PR instead be `BREAKING CHANGE:`?

zioroboco: The last commit to the branch includes a `BREAKING CHANGE` annotation, so this will get released at `1.0.0`. It's a minor quirk of the release pipeline that there isn't another way to release packages at `1.0.0` — manually setting the major version will always lead to `1.0.1` or `1.1.0` for the initial release. :joy:

storybook/webpack.config.ts
```diff
@@ -83,7 +83,6 @@ const svgIcons: Rule = {
 
 const elm: Rule = {
   test: /\.elm$/,
-  exclude: [/elm-stuff/, /node_modules/],
   use: [
```
ActuallyACat: Why was this rule removed?

zioroboco: This is being overwritten by the condition to exclude external modules, so was only causing confusion. Must have gotten caught up in a copy-and-paste of the elm loader rule.

zioroboco:
@ActuallyACat We're looking for places where the scope update pattern doesn't make sense. Cases that have come up so far are `someone@cultureamp.com` email addresses, and `@cultureamp/team` names in `CODEOWNERS`.

zioroboco:
Closing (see #242 )

PR #224, opened 2020-01-16
ci: Chromatic integration via Buildkite

mbylstra:
This PR enables Chromatic integration ( chromaticqa.com)

## Overview:

- Buildkite is used to create storybook build, and then the chromatic CLI tool is used to send the build to Chromatic.
- Currently we are not permitted to do the proper github integration that would show the status inside github, so the way to use Chromatic is to click on the Buildkite build and follow the link to Chromatic where you can do the review
- Currently the Chromatic build is always green, regardless of whether there are any visual changes or not. This is becuase of the `--exit-zero-on-changes` flag. The idea is that we remove that flag once we're confident there aren't any flaky tests and that we have good instructions for using Chromatic before we publicise it to the rest of the practice. For now it's just mean to be for team DST to trial
- Chromatic API key is stored in AWS secrets manager. At the time of writing the key is for my personal account for a project named `Testing`. It's a badly named project, but it's the first project I created and I can't change the name without creating a new project and new key. It's not much work changing the key in AWS, but would prefer to get this rolling first - we can do that later.

The PR https://github.com/cultureamp/kaizen-design-system/pull/416 was required to get this work, as we had in the meantime since creating this PR created a mix of `CSF` and `storiesOf` stories within the one file. Chromatic is happy as long as they are in separate files.


mbylstra:
Still waiting on approval/purchase of Chromatic.

mbylstra:
This is just a random unrelated comment to see if Github Slack integration is working

PR #226, opened 2020-01-16
fix: Fix elm styling regression for the Popover component

jeremy8883:
* Fix elm styling regression for the Popover component. Updates were made to the react component, therefore changing the styles, but the elm component wasn't updated accordingly.

[Related PR](https://github.com/cultureamp/kaizen-design-system/pull/181)

Before:
image

After:
image


packages/component-library/draft/Kaizen/Popover/Popover.elm
```diff
@@ -170,54 +175,136 @@ view (Config config) =
...
-        , div [ styles.class .container ] [ config.content ]
+        [ div
+            [ styles.classList
+                [ ( mapVariantToBoxClass config.variant, True ) ]
+            ]
+            [ case config.heading of
+                Just heading ->
+                    div [ styles.class .header ]
+                        [ if config.variant == Default then
+                            span [] []
```
jeremy8883: I wasn't sure how to return no element here

packages/component-library/stories/PopoverStories.elm
```diff
@@ -68,16 +68,16 @@ main =
             div [ style "margin-top" "1.5rem" ]
                 [ Popover.view
                     (Popover.default
-                        |> Popover.withHeading "Default"
+                        |> Popover.withHeading "Arrow start"
```
jeremy8883: Minor changes made to the elm stories, as to exactly match the react stories

packages/component-library/draft/Kaizen/Popover/Popover.elm
sebpearce: If it's returning an `Html Msg`, you can do `text ""` to return a blank text node if you want to avoid random empty stuff in the DOM.

sebpearce:
```suggestion
                            text ""
```

packages/component-library/stories/PopoverStories.elm
sebpearce: Nice.

PR #231, opened 2020-01-17
chore: add Knobs addon stories to some stories

mbylstra:
https://trello.com/c/QZC8nZEW/708-add-knobs-to-storybook

## Background

Adding “knobs” to storybook ~would replace the need for so many stories*~ and will also make it clear what is a component and what is a prop.

*In the end we decided to keep the the old stories as they will be needed for future visual regression tests.

## Preview
https://dev.cultureamp.design/dst/storybook-knobs/storybook/
(React Stories for Button though to IconButton have an extra story named "with Knobs" that has knobs)

## Summary of changes:
- upgrade storybook and all storybook addons to `5.3.5`
- add and register `@storybook/addon-knobs` dependency
- add a "with Knobs" story to React stories for Button, CheckboxField, CheckboxGroup, EmptyState, GobalNotification, HeroCard, Icon, and IconButton (roughly A though to I)

## Comments:

- It's only currently possible to add knobs to React stories. It seems possible to do so for Elm stories, but the Elm story package (integration between Elm and React Storybook) would need new features and it seemed like too much work for this PR. I still feel the React stories are useful for Elm developers as they give a good overview of the features of the component even if it doesn't completely accurately represent the Elm version.
- Although there is now inconsistency as only some of the stories have knobs, I feel this is releasable as it provides good examples for others to contribute knobs stories (without any, but just the npm dependencies, it would be harder to describe how to do it)
- Knobs stories are included as well as static stories, because the static stories will be required for Chromatic visual regression testing.. I searched for Chromatic integration with the knobs addon but I couldn't find anything.  Also it can be sometimes easier to tell a story of how to use a component with more manual static stories (knobs aren't always perfect as you are limited by the constraints of the Knobs feature)
- I didn't add knobs for every parameter either because it was too fiddly to do (time constraints) or it didn't seem possible (I don't think it's possible to add a React Node as a knob for an optional field such as content fields). I still feel that a subset of all the parameters is useful (as many of the parameters now in knobs are not in any of the static stories)
- This still uses the old `storiesOf` format as opposed to the Component Storybook Format that you see in the latest docs for knobs: https://www.npmjs.com/package/@storybook/addon-knobs. I think the examples provided here should make it easy enough for others to contribute even though they don't reflect the official documentation. We plan to upgrade to CSF in a later release.
- I added `//@ts-ignore` to some of the stories contain radio widgets because we couldn't figure out how to make Typescript happy (didn't spend too much time on it), stories should be easy to write, we want people to contribute we don't expect people to be type wizards to contribute.
- I discovered https://github.com/storybookjs/addon-smart-knobs after manually writing stories, which would mean completely automatic generation of stories. Pros: get the stories for free and much less maintenance burden, stories can't get out of date with production code. Cons: less ability to customise (labels, radio vs select control widget). But, I didn't quite get it working as I ran out of time. It might not be that much extra work. The problem is that if we encourage people to write the rest of the stories, and then later decide to switch this would mean wasted manual work. **Do you think we should release and encourage people to write more stories, or you think we should freeze this PR until we've investigated smart knobs further?**

There's definitely tradeoffs with this release (it giveth but taketh a little), so have a think about whether you think it's a net win.


mbylstra:
[Add Knobs to Storybook](https://trello.com/c/QZC8nZEW/708-add-knobs-to-storybook)

packages/component-library/stories/CheckboxGroup.stories.tsx
```diff
@@ -42,6 +43,47 @@ export default class CheckboxGroupExample extends React.Component<Props> {
...
+  .addDecorator(withKnobs)
+  .add("with Knobs", () => {
+    return (
+      <CheckboxGroup labelText={text("labelText", "Checkbox Group Label")}>
+        <CheckboxGroupExample
+          render={({ checkedStatus, onCheckHandler }) => (
+            <CheckboxField
+              onCheck={onCheckHandler}
+              id="checkbox-1"
+              checkedStatus={checkedStatus as any}
```
ActuallyACat: nitpicking, but could this be cast to a string instead? That will bring it inline with the typescript definition on `CheckboxField`

ActuallyACat:
Screen Shot 2020-01-24 at 2 07 03 pm

First win for knobs - The ability to quickly toggle the props in components made this bug apparent (Hero component in full width mode, illustration scales out of container) 😬 
Screen Shot 2020-01-24 at 2 07 03 pm

yarn.lock
```diff
@@ -1338,6 +1338,16 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+"@chromaui/localtunnel@1.10.1":
```
ActuallyACat: This dependency must have carried over from what you were doing with the Chromatic stuff.

yarn.lock
```diff
@@ -17892,6 +17887,35 @@ store2@^2.7.1:
   resolved "https://registry.yarnpkg.com/store2/-/store2-2.10.0.tgz#46b82bb91878daf1b0d56dec2f1d41e54d5103cf"
   integrity sha512-tWEpK0snS2RPUq1i3R6OahfJNjWCQYNxq0+by1amCSuw0mXtymJpzmZIeYpA1UAa+7B0grCpNYIbDcd7AgTbFg==
 
+storybook-chromatic@^2.2.2:
```
ActuallyACat: This one too 😬 Might be a good idea to check out the lock file from master and run `yarn install` to regenerate this with your dependencies

packages/component-library/stories/CheckboxGroup.stories.tsx
mbylstra: I have to admit I copied and pasted the earlier example, but yeah, that could be fixed.

yarn.lock
mbylstra: Ooooops! Good pickup.

mbylstra:
I had another quick timeboxed go at trying to use Smart Knobs. This PR is as far as I got: https://github.com/cultureamp/kaizen-design-system/pull/245.

Although I didn't get it working, I'm still very hesistant to release this PR as I think automatically generated knobs will be far more maintainable than manually handcrafted ones. I don't want to encourage people to manually create them, and then have to delete them.

Part of smart knobs is https://github.com/strothj/react-docgen-typescript-loader which I think is extremely useful. So, if we get that working, then we can see if smart knobs actually works. So, I think we should park this PR until we get a chance to work on `react-docgen-typescript-loader`

Issue #188, opened 2019-12-24
There's no date picker guidance


didoesdigital:
A few people have asked about date picker design elements and component needs. Some considerations that need to be addressed on Kaizen Site include:

- Overall size including vertical height of button padding
- Confirmation and clear buttons
- Validation and other states
- Default date (e.g. not today for date of birth?)
- Date ranges vs single date pickers
- Date and time pickers
- Accessibility
- Elm and React implementations
- Localisation: datepicker text that isn’t a number will need to be localised
- Can the user type in the date manually without selecting from calendar? If so, how do we validate format?
- What do we do on tiny screens? What do we do on touch devices (small or large)? Should we use native Android/iOS/etc. widgets?

didoesdigital:
Resources:

https://polaris.shopify.com/componen

- ts/forms/date-picker
- https://www.atlassian.design/guidelines/product/components/date-picker
- https://www.lightningdesignsystem.com/components/datepickers/#site-main-content
- https://www.lightningdesignsystem.com/components/datetime-picker/#site-main-content
- https://material.io/components/pickers/
- https://material.io/components/text-fields/#anatomy
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date
- https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html
- https://github.com/airbnb/react-dates

PR #240, opened 2020-01-22
docs: add date picker documentation

didoesdigital:
Branch preview:

- https://dev.cultureamp.design/di/add-date-picker-documentation/components/date-picker/
- https://dev.cultureamp.design/di/add-date-picker-documentation/components/text-field/

image

Fixes: #188


reneetan:
Looks really good. I like that the calendar closes when you select a second date in the range.

Only minor thing is I think we could have a little more space between the month heading and the dates, it's looking a little squished.

image

didoesdigital:
Early feedback from engineers suggests that we won't have the flexibility to increase the spacing on the date picker UI between the calendar visual and the text month/year and buttons, but if that constraint opens up, let's increase the spacing.

PR #245, opened 2020-01-27
Dst/storybook smart knobs

mbylstra:
This is a an attempt at using https://github.com/storybookjs/addon-smart-knobs to automatically generate Knobs stories at runtime rather than manually create them. I have no idea if Smart Knobs actually works or works well (I couldn't find any examples on line of it in action). This PR could be picked up again in a future cooldown, particularly the `react-docgen-typescript-loader` part.

Worth noting that the https://github.com/strothj/react-docgen-typescript-loader required for Smart Knobs is also very useful for automatically generating documentation. So getting that to work (I don't think it's working in this PR), would also enable that.  However, https://storybook.js.org/docs/configurations/typescript-config/ recommends using `awesome-typescript-loader` instead of `ts-loader` (used here). That's something else to try.

I timeboxed this but continued to run into compilations and never got it to work. There are some comments below.


storybook/webpack.config.ts
```diff
@@ -151,10 +151,25 @@ export default ({ config }) => {
...
+      require.resolve("react-docgen-typescript-loader"),
+    ],
+  })
+
   config.module.rules.push(
     ...[babel, styles, svgs, svgIcons, elm].map(excludeExternalModules)
   )
 
+  config.module.rules[0].use[0].options.plugins = [
+    require.resolve("babel-plugin-react-docgen"),
```
mbylstra: It wasn't clear if you needed to add both `react-docgen` _and_ `typescript-docgen`. I didn't try it without `react-docgen`

storybook/webpack.config.ts
```diff
@@ -151,10 +151,25 @@ export default ({ config }) => {
   // Required for the storysource storybook addon
   config.module.rules.push(storybookSource)
 
+  config.module.rules.push({
+    test: /^((?!\.stories).)*\.tsx?$/, // match all .tsx files other than .stories.tsx files
```
mbylstra: This was because I was getting a lot of typescript errors on the *.stories.tsx files. We don't need documentation on the stories themselves (that would be very meta), so I exclude them.

PR #246, opened 2020-01-28
feat: Make right/left arrows look like forward/backward

sebpearce:
_This PR will be a "breaking design change" (the look of the arrows is changing) but not a breaking change._

We currently have two pairs of left and right arrows in the platform that look different:
image
image
Designers would like to keep the bottom pair and get rid of the top pair.

However, due to the fact that "right" and "left" are more RTL-friendly than "backward" and "forward", we want to _rename_ the bottom pair to "Arrow Right" and "Arrow Left".

**This PR preserves the filenames, but makes the top pair look like the bottom pair.**

It does not delete the now-redundant bottom pair yet, because my plan is to first remove references to the bottom pair in consuming repos, and then do another PR in `kaizen-design-system` to remove the icons themselves.

### DIFF
--- a/packages/component-library/icons/arrow-left.icon.svg
+++ b/packages/component-library/icons/arrow-left.icon.svg
@@ -4,7 +4,7 @@
     <title>arrow-left</title>
     <desc>Created with Sketch.</desc>
     <defs>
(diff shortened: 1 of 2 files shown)


sebpearce:
[Remove/substitute two conflicting directional arrow icons.](https://trello.com/c/J8cMSvGC/674-remove-substitute-two-conflicting-directional-arrow-icons)

PR #250, opened 2020-01-30
feat: Add elm modal positive, negative variants

Confidenceman02:
This adds the positive and negative modal variants.

## Positive
Screen Shot 2020-01-30 at 3 01 13 pm

## Negative
Screen Shot 2020-01-30 at 3 01 48 pm

### DIFF
--- a/packages/component-library/draft/Kaizen/Modal/Modal.elm
+++ b/packages/component-library/draft/Kaizen/Modal/Modal.elm
@@ -66,6 +66,7 @@ type Duration
 type ConfirmationType
     = Informative
     | Positive
+    | Negative
 
 
 type alias ConfirmationConfig msg =
@@ -193,6 +194,20 @@ view (Config config) =
                         GenericModal.view GenericModal.Default
                             [ ConfirmationModal.view
                                 (ConfirmationModal.positive
+                                    |> withOnDismiss
+                                    |> withOnConfirm
+                                    |> withBodySubtext
+                                    |> ConfirmationModal.confirmLabel configs.confirmLabel
+                                    |> ConfirmationModal.dismissLabel configs.dismissLabel
+                                    |> ConfirmationModal.title configs.title
+                                )
+                            ]
+                            (genericModalConfig |> GenericModal.events genericModalEvents)
+
+                    Negative ->
+                        GenericModal.view GenericModal.Default
+                            [ ConfirmationModal.view
+                                (ConfirmationModal.negative
                                     |> withOnDismiss
                                     |> withOnConfirm
                                     |> withBodySubtext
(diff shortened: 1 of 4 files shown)


didoesdigital:
Thanks for the PR, @Confidenceman02! Our modals are really valuable components for reuse so this will have a great impact on our Elm development.

Some thoughts:

- The "!" icon needs to be an "i" icon for an informative modal.
- The "!" icon needs to be a tick icon for the positive modal.
- The "i" icon needs to be an "!" icon for the React negative modal (probably out of scope here).
- Can you have the modal shown by default in Storybook? This will help people see what a modal looks like more quickly when skipping through stories than clicking the button to open it (and would match the React modal)
- I giggled at the placeholder button text, but history has shown it can help designers to see sensible defaults, such as the consistent use of "Cancel" for the cancel button label. Would you mind updating it to match the React modals?
- I believe (but could be wrong here) the React modal focuses the confirmation button by default to move the focus from the triggering button to the modal to ensure people using screen readers don't get lost. I noticed that the background button stays focused in the Elm modal. @lijuenc might have more info here.
- Once the modal has focus, I _believe_ the accessible thing to do is trap keyboard focus within the modal (meaning you can't accidentally tab out of it). Check out [W3C ARIA practices](https://w3c.github.io/aria-practices/#dialog_modal), which I grabbed from the [Kaizen Modal guidelines](https://cultureamp.design/components/modal/).
- Nit: capitalisation is different where React stories show `Confirmation (informative)` vs Elm stories show `Confirmation (Informative)`
- @Confidenceman02 do you have scope to stress test a long content story like [the React Test - scrolling modal and scrolling content story](https://dev.cultureamp.design/confidenceman02/elm-modal-variants/storybook-static/?path=/story/modal-react--test-scrolling-modal-and-scrolling-content)

Confidenceman02:
Some thoughts:

The "!" icon needs to be an "i" icon for an informative modal.
Done
The "!" icon needs to be a tick icon for the positive modal.
Done
The "i" icon needs to be an "!" icon for the React negative modal (probably out of scope here).
Done
Can you have the modal shown by default in Storybook? This will help people see what a modal looks like more quickly when skipping through stories than clicking the button to open it (and would match the React modal)
I giggled at the placeholder button text, but history has shown it can help designers to see sensible defaults, such as the consistent use of "Cancel" for the cancel button label. Would you mind updating it to match the React modals?
Done
I believe (but could be wrong here) the React modal focuses the confirmation button by default to move the focus from the triggering button to the modal to ensure people using screen readers don't get lost. I noticed that the background button stays focused in the Elm modal. @lijuenc might have more info here.
Once the modal has focus, I believe the accessible thing to do is trap keyboard focus within the modal (meaning you can't accidentally tab out of it). Check out W3C ARIA practices, which I grabbed from the Kaizen Modal guidelines.
These are known missing accessibility things. There is a github issue tracking the focus issue you are talking about. Yet to be built!
Nit: capitalisation is different where React stories show Confirmation (informative) vs Elm stories show Confirmation (Informative)
Done
@Confidenceman02 do you have scope to stress test a long content story like the React Test - scrolling modal and scrolling content story
This can be a future story.

PR #252, opened 2020-01-30
fix: tag in Titleblock misaligned in IE11

ellisjas:


ellisjas:
[Title Block - Some parts of the app are using old version](https://trello.com/c/TMJaUkLc/1835-title-block-some-parts-of-the-app-are-using-old-version)

PR #256, opened 2020-02-02
feat: close elm modal with escape

Confidenceman02:
The elm modal currently does not support closing by pressing the escape key.

By subscribing to onKeyDown and decoding for the escape key we can ensure the modal runs the closing animation.

Care was taken to ensure a subscription to onKeyDown is only applicable when the modal is not closed. This means that pressing escape whilst the modal is opening will disrupt the animation and close it again.


sebpearce:
LGTM – once this passes design review feel free to ping me for a ✅ .

didoesdigital:
@Confidenceman02, assuming focus management is still out of scope for this (e.g. escape key to return focus to trigger button), it looks good to me.

PR #258, opened 2020-02-03
fix: added tslint:latest to linting rules

ActuallyACat:
* Fixed the lowest hanging linting issues
* Added severity: none to rules we will eventually clean up

### DIFF
--- a/packages/component-library/components/Button/components/GenericButton.tsx
+++ b/packages/component-library/components/Button/components/GenericButton.tsx
@@ -31,7 +31,7 @@ type LabelProps = {
 
 type Analytics = {
   eventName: string
-  properties: Object
+  properties: object
 }
 
 export type IconButtonProps = GenericProps
--- a/tslint.json
+++ b/tslint.json
@@ -1,5 +1,5 @@
 {
-  "extends": ["tslint-config-prettier"],
+  "extends": ["tslint-config-prettier", "tslint:latest"],
   "linterOptions": {
     "exclude": ["**/*.js?(x)"]
   },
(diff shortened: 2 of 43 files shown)


tslint.json
```diff
@@ -9,6 +9,35 @@
...
       ["@kaizen", "@storybook", "@testing-library", "react", "react-dom"]
     ],
-    "ordered-imports": true
+    "ordered-imports": true,
+    "semicolon": false,
+    "no-submodule-imports": false,
+    "no-var-requires": false,
+    "interface-name": false,
+
+    "interface-over-type-literal": {
```
ActuallyACat: The idea here is to _eventually_ switch a rule to a warning -> fix the warnings across the repo -> switch said rule to an error. Rinse and repeat for all the `"severity": "none"` rules. I've disabled a couple of the rules that came with tslint:latest for these reasons:

```
    "semicolon" // this seems to be a standard we roughly follow. Not my taste, but I won't argue it. 
    "no-submodule-imports" // enabling this throws errors for importing Kaizen components. We're safe to use submodule imports for our stories, so i'm disabling this. 
    "no-var-requires" // We `require` our CSS modules, so this needs to stay disabled
    "interface-name": false, // There doesn't seem to be a standard here, but I personally dislike the prefixing my interfaces with "I" and would prefer to use a different convention (like "ButtonProps") 
```

I didn't want to spend too much time on this knowing Delivery Engineering might be creating a shared linting file to keep all repos consistent. This will help us keep things somewhat reasonable until then.

mbylstra:
@ActuallyACat The build is failing on a type check: https://buildkite.com/culture-amp/kaizen-design-system/builds/1215#a3f9bb26-6f7a-4f8d-8116-4cc84d839b06/165-193

ActuallyACat:
> @ActuallyACat The build is failing on a type check: https://buildkite.com/culture-amp/kaizen-design-system/builds/1215#a3f9bb26-6f7a-4f8d-8116-4cc84d839b06/165-193

Thanks @mbylstra, I missed this 😓 Should be good now.

packages/component-library/components/Button/components/GenericButton.tsx
```diff
@@ -31,7 +31,7 @@ type LabelProps = {
 
 type Analytics = {
   eventName: string
-  properties: Object
+  properties: object
```
sebpearce: Love it.

PR #259, opened 2020-02-03
feat: Add header images to docs pages

sebpearce:
## Changes

This PR adds header image functionality to guidelines and component docs pages.

Header images will be SVGs and should be no larger than 200px. They live in `site/src/header-images/`.

### Example

If `modal-informative.svg` lives in the `header-images/` dir, and you want to use it, you would go to the corresponding `.mdx` file for the page you want to edit (either a `README.mdx` in the Component Library, or one of the pages in `docs/components/` or `docs/guidelines/`) and add the following frontmatter:

`headerImage: modal-informative`

### Why is modal-informative.svg part of this PR?

If we don't add at least one image, and at least one use of the `headerImage` frontmatter in an `.mdx` file, Gatsby will not compile.

### Preview

image


sebpearce:
[Custom images per Kaizen Site page e.g. component SVG next to README](https://trello.com/c/WVDTLwZq/721-custom-images-per-kaizen-site-page-eg-component-svg-next-to-readme)

didoesdigital:
@sebpearce is there an empty `alt=""` tag somewhere for these header images?

sebpearce:
@didoesdigital Because the `header-images/` dir is empty, the build breaks. I could commit our test image, but then it'd be in the git history forever, so I'll wait until you've got at least one image ready that we can put in here.

image

PR #268, opened 2020-02-11
feat: Add typography tokens 

ActuallyACat:
Let's get this package to v1. 

I have changed a few things that would constitute a breaking change: 
* I've moved the type definitions to their own file, rather than rely on tokens/index.ts. I don't think the latter will scale, especially if we start adding helper functions to access some of the more deeply nested tokens. So now to import the typed variables, you would `import * as typography from '@kaizen/design-tokens/tokens/color'
* I've changed the colour helper from `color` to `get`, so that file names are consistent with the main export (typography.ts exports typography as the full object, colour.ts exports color, etc)

I think these are the final things that needed fixing before this package is considered "stable". From here, new tokens should be relatively straightforward


packages/design-tokens/less/typography.less
```diff
@@ -0,0 +1,31 @@
+/** THIS IS AN AUTOGENERATED FILE **/
+
+@kz-typography-family-mono: "IBM Plex Mono";
+@kz-typography-family-display: "Greycliff CF";
+@kz-typography-family-text: "Inter";
+@kz-typography-weight-bold: 600;
+@kz-typography-weight-regular: 400;
+@kz-typography-display-fontSize-1: 4.5rem;
+@kz-typography-display-fontSize-2: 2.25rem;
+@kz-typography-display-fontSize-3: 1.875rem;
```
ActuallyACat: It slightly irks me that the transform library doesn't convert tokens to different cases. A potential improvement? 🤔

sebpearce:
Here's what a  "purely options" set of tokens would look like if we're treating things like font size as an option in themselves:
```
{
  "kz": {
    "typography": {
      "family": {
        "plex": "\"IBM Plex Mono\"",
        "greycliff": "\"Greycliff CF\"",
        "inter": "\"Inter\""
      },
      "weight": {
        "600": "600",
        "400": "400"
      },
      "fontSize": {
          "4.5rem": "4.5rem",
          "2.25rem": "2.25rem",
          "1.875rem": "1.875rem",
          "1.375rem": "1.375rem",
          "1.25rem": "1.25rem",
          "1.125rem": "1.125rem",
          "1rem": "1rem",
          "0.875rem": "0.875rem",
          "0.75rem": "0.75rem"
      },
      "lineHeight": {
          "4.5rem": "4.5rem",
          "2.25rem": "2.25rem",
          "1.875rem": "1.875rem",
          "1.5rem": "1.5rem",
          "1.125rem": "1.125rem"
      }
    }
  }
}
```

sebpearce:
When I read things like this it makes me less confident about doing it the way that we've defined options :(
image

sebpearce:
If I were to make an uncharitable diagram of the comparison between the approach we're leaning towards and the one in that article, it'd be:
image

mbylstra:
This is what it might look like if you had "decisions" (which I've named `styles`) in addition to atomic options. The first part is a copy n paste of Seb's example above. The pretty ugly `"{"$ref": "#/kz/typography/family/greycliff"}` is from the `json-reference` standard (https://tools.ietf.org/html/draft-pbryan-zyp-ref-03) but there are simpler ways you could do that (just make something up such as use a string like `*family/greycliff", if you're willing to give up the tooling that has been made for json-reference.

```
  {
    "kz": {
      "typography": {
        "family": {
          "plex": "\"IBM Plex Mono\"",
          "greycliff": "\"Greycliff CF\"",
          "inter": "\"Inter\""
        },
        "weight": {
          "600": "600",
          "400": "400"
        },
        "fontSize": {
            "4.5rem": "4.5rem",
            "2.25rem": "2.25rem",
            "1.875rem": "1.875rem",
            "1.375rem": "1.375rem",
            "1.25rem": "1.25rem",
            "1.125rem": "1.125rem",
            "1rem": "1rem",
            "0.875rem": "0.875rem",
            "0.75rem": "0.75rem"
        },
        "lineHeight": {
            "4.5rem": "4.5rem",
            "2.25rem": "2.25rem",
            "1.875rem": "1.875rem",
            "1.5rem": "1.5rem",
            "1.125rem": "1.125rem"
        }
      },
      "styles": {
          "display0": {
              family: "{"$ref": "#/kz/typography/family/greycliff"},
              weight: "{"$ref": "#/kz/typography/weight/600"},
              fontSize: "{"$ref": "#/kz/typography/fontSize/4.5rem"},
              lineHeight: "{"$ref": "#/kz/typography/lineHeight/4.5rem"}
          },
          "heading1": {
              family: "{"$ref": "#/kz/typography/family/greycliff"},
              weight: "{"$ref": "#/kz/typography/weight/600"},
              fontSize: "{"$ref": "#/kz/typography/fontSize/2.25rem"},
              lineHeight: "{"$ref": "#/kz/typography/lineHeight/2.25rem"}
          },
          ...
        }
      }
   }
}
```

You might be asking why do `weight: "{"$ref": "#/kz/typography/weight/600"},` when you could just do `600`. The reason is that if you removed 600 as an option, the json-reference (or custom tooling) would complain - it'd force you to use a weight that is an "option".

mbylstra:
Another one FWIW:

```
  {
    "kz": {
      "typography": {
        "family": {
          "plex": "\"IBM Plex Mono\"",
          "greycliff": "\"Greycliff CF\"",
          "inter": "\"Inter\""
        },
        "weight": {
          "600": "600",
          "400": "400"
        },
        "fontSize": {
            "4.5rem": "4.5rem",
            "2.25rem": "2.25rem",
            "1.875rem": "1.875rem",
            "1.375rem": "1.375rem",
            "1.25rem": "1.25rem",
            "1.125rem": "1.125rem",
            "1rem": "1rem",
            "0.875rem": "0.875rem",
            "0.75rem": "0.75rem"
        },
        "lineHeight": {
            "4.5rem": "4.5rem",
            "2.25rem": "2.25rem",
            "1.875rem": "1.875rem",
            "1.5rem": "1.5rem",
            "1.125rem": "1.125rem"
        }
      },
      "textStyles": {
          "display0": {
              family: \"Greycliff CF\",
              weight: "600",
              fontSize: "4.5rem",
              lineHeight: "4.5rem"
          },
          "heading1": {
              family: "\"Greycliff CF\"",
              weight: "600",
              fontSize: "2.25rem",
              lineHeight: "2.25rem"
          },
          ...
        }
      }
   }
}
```
This does away with the json-reference complexity, but there's less guarantees. You'd have to manually make sure you weren't using values that weren't options.
```

**Comment by mbylstra** [comment]
If you wanted even more guarantees, you could use json-schema to enforce that a text style always had a family, weight, fontSize, lineHeight (this is really heading into overengineering territory), so you could guarantee that if you swapped out `display0` with `heading1` it would still compile.

**Comment by mbylstra** [comment]
YAML has anchors and references which are a similar thing to json-reference: https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html

**Comment by ActuallyACat** [comment]
Whoever called out that coupling font-size and line-height to a font family (mono/display/text) really made me ponder. It did start to feel a little "decision token-y". Here's a version using the animal scale, which will accommodate new sizes should we need them. I'm not married to this version though.

```
{
  "kz": {
    "typography": {
      "family": {
        "mono": "\"IBM Plex Mono\"",
        "display": "\"Greycliff CF\"",
        "text": "\"Inter\""
      },
      "weight": {
        "bold": "600",
        "regular": "400"
      },
      "fontSize": {
        "whale": "4.5rem",
        "elephant": "2.25rem",
        "hippo": "1.875rem",
        "horse": "1.375rem",
        "goat": "1.125rem",
        "dog": "1rem",
        "quokka": "1.25rem",
        "mouse": "0.875rem"
        "flea": "0.75rem"
      },
      "lineHeight": {
        "whale": "4.5rem",
        "elephant": "2.25rem",
        "hippo": "2.25rem",
        "horse": "1.875rem",
        "goat": "1.5rem",
        "dog": "1.5rem",
        "quokka": "1.5rem",
        "mouse": "1.5rem"
        "flea": "1.125rem",
      }
    }
  }
}
```

The more I think about it, the more I think the approach Theo uses isn't overengineering. Something like what @mbylstra is suggesting might be a good middle ground 🤔

**Comment by sebpearce** [comment]
This approach, which is a modification of @mbylstra's one above, means we can accurately model designers' perception of type styles as options (assuming that's the case generally) but also keep the families and weights as their own thing:

image

**Comment by sebpearce** [comment]
@ActuallyACat @mbylstra 

I'm trying to draft a list of questions that seem to keep coming up in my mind as we think about structure. Feel free to suggest more and I'll edit.

- Could a designer look at this structure and agree that it accurately represents their model of the design system?
- Could an engineer look at this structure and reasonably understand what the designers intended?
- Does this structure provide a single source of truth for design decisions? If not all, what % of decisions does it capture and are we happy with this?
- Does this structure allow for easy modification later?
- What incentives does this structure create for engineers?

**Comment by ActuallyACat** [comment]
> @ActuallyACat @mbylstra
> 
> I'm trying to draft a list of questions that seem to keep coming up in my mind as we think about structure. Feel free to suggest more and I'll edit.
> 
>     * Could a designer look at this structure and agree that it accurately represents their model of the design system?
> 
>     * Could an engineer look at this structure and reasonably understand what the designers intended?
> 
>     * Does this structure provide a single source of truth for design decisions? If not all, what % of decisions does it capture and are we happy with this?
> 
>     * Does this structure allow for easy modification later?
> 
>     * Can this structure be consumed by future unknown tech stacks without too much hassle?
> 
>     * What incentives does this structure create for engineers?

Additionally, things we all (seem to) agree about:
* We want teams to stop using Sass from Kaizen. Instead they should be using the Design Tokens.

**Comment by ActuallyACat** [comment]
To consolidate a few ideas: 

**Bundle of rules**
```
{
  "kz": {
    "typography": {
      "heading0": {
        family: "greycliff",
        weight: "600",
        fontSize: "4.5rem",
        lineHeight: "4.5rem"
      },
      "heading1": {
        family: "greycliff",
        weight: "600",
        fontSize: "3rem",
        lineHeight: "3rem"
      }
      ...
    },
  }
}
```

**Compiles to**
```
$kz-typography-display-heading0-family: "Greycliff"
$kz-typography-display-heading0-weight: 600
$kz-typography-display-heading0-size "4.5rem"
```

*pros*: 
- Aligns with the design intention in Figma
- Easy for non-DST developers to contribute to Kaizen

*cons*: 
- We need to be SUPER CLEAR to designers that any modifications in the UI kit will have pretty catastrophic consequences for the component library

*misc*: mixing the styles from different objects is allowable and makes Ally sad (e.g. heading0-size and heading3-lineHeight)

*Animal farm*
```
{
  "kz": {
    "typography": {
      "family": {
        "mono": "\"IBM Plex Mono\"",
        "display": "\"Greycliff CF\"",
        "text": "\"Inter\""
      },
      "weight": {
        "bold": "600",
        "regular": "400"
      },
      "fontSize": {
        "xxl": "4.5rem",
        "xl": "2.25rem",
        ...
      },
      "lineHeight": {
        "xxl": "4.5rem",
        "xl": "2.25rem",
        ...
      }
    }
  }
}
```

**Compiles to**
```
$kz-typography-family: "Greycliff"
$kz-typography-weight-bold: 600
$kz-typography-display-fontSize: "4.5rem"
```

*pro* 
- decoupled from intention / unopinionated 
- locks down the "allowable" list of properties within the system to properties in this file. No snowflakes.
- "encourages" people to use components instead of reaching into Sass vars

*con* 
- Less clear for non-DST developers how to build UI. Especially painful when comparing
- requires an additional layer of "decisions" in components (maybe a pro?)


**Bundle of animals**
```
{
 "kz": {
    "typography": {
      "family": {
        "mono": "\"IBM Plex Mono\"",
        "display": "\"Greycliff CF\"",
        "text": "\"Inter\""
      },
      "weight": {
        "bold": "600",
        "regular": "400"
      },
      ...
      "display0": {
              family: "{"$ref": "#/kz/typography/family/greycliff"},
              weight: "{"$ref": "#/kz/typography/weight/600"},
              fontSize: "{"$ref": "#/kz/typography/fontSize/4.5rem"},
              lineHeight: "{"$ref": "#/kz/typography/lineHeight/4.5rem"}
          },
          "heading1": {
              family: "{"$ref": "#/kz/typography/family/greycliff"},
              weight: "{"$ref": "#/kz/typography/weight/600"},
              fontSize: "{"$ref": "#/kz/typography/fontSize/2.25rem"},
              lineHeight: "{"$ref": "#/kz/typography/lineHeight/2.25rem"}
          },
      ...
    },
  }
}
```

**Compiles to**
```
$kz-typography-family: "Greycliff"
$kz-typography-weight-bold: 600
$kz-typography-display-fontSize: "4.5rem"
```
**Not quite sure what the "decisions" will (or should) compile to**


*pro* 
- Literally all the pros of the first two options

*con* 
- Unclear about how we should compile the decisions

**Comment by sebpearce** [comment]
I went ham.

image

**Comment by ActuallyACat** [comment]
Thinking more about this premise: 
>  represents their model of the design system
We need to build a combined understanding of the system. We're the experts in implementation, and if their mental model doesn't nicely translate to our model then we're going to have problems. As an example, the idea that a `heading-0` as a token when it is defined as a combination of CSS properties is not compatible with what we have historically considered being a token (one single variable). How we're currently compiling our tokens into Sass/Less variables **doesn't support this concept**. I believe this is where we keep hitting a wall when we discuss tokens. 

For "bundle of rules" I would disagree with "does this allow for easy modification later" point, _especially_ if we're keeping the rigid naming convention that's been defined in Figma. We have the same issue as introducing levels in individual properties (line-height-sm, line-height-md, etc). 

This is why I think it's so important for us to define primitives for each line-height/font-size, font-family. We wouldn't let someone use a snowflake colour, so why would we allow them to use a snowflake font-size? Or snowflake font-family?

**Comment by sebpearce** [comment]
> As an example, the idea that a heading-0 as a token when it is defined as a combination of CSS properties is not compatible with what we have historically considered being a token (one single variable).

`One token == one variable` isn't something I've put much stock in, since the thing we're trying to represent is design decisions and not CSS rules. The concept of a design decision exists on an abstract layer above individual CSS rules, so whether a design decision can be captured by a single CSS variable is basically down to chance, i.e. whether CSS has a rule for that. For example, with colours we were lucky, because a single hex value can capture a colour. But what if CSS didn't support that? What if CSS forced you to use 3 different rules for R, G, and B? Now something that a designer thinks of as one thing would require 3 things to capture it, and would not qualify as a token under this line of reasoning.

> For "bundle of rules" I would disagree with "does this allow for easy modification later" point, especially if we're keeping the rigid naming convention that's been defined in Figma. We have the same issue as introducing levels in individual properties (line-height-sm, line-height-md, etc).

I don't get this yet, might need an example or two before I can grok what you mean. To clarify, what are some examples of later modifications that could be difficult under the "bundle of rules" approach?

> This is why I think it's so important for us to define primitives for each line-height/font-size, font-family. We wouldn't let someone use a snowflake colour, so why would we allow them to use a snowflake font-size? Or snowflake font-family?

We wouldn't let someone use a snowflake colour because a colour exists as an atomic unit in a designer's mind. Font sizes and line heights are subatomic to designers (like saturation for a colour) when it comes to the type styles. The designers are not saying, "we only want people to use these font sizes and line heights, but you can combine them in any way you like." They're saying, "we only want people to use these specific combinations of font sizes and line heights." If a designer decides that Combination X needs a different font size that's 1px larger, it's irrelevant to designers whether or not that new font size is part of an approved list, because they are not thinking in terms of font sizes, but in terms of type styles. To summarise, a colour can't be compared to a font size or font family because they exist at different levels in the designer's mental model (colours are at the surface, and font sizes are one layer down below type styles).

**Comment by ActuallyACat** [comment]
> One token == one variable isn't something I've put much stock in, since the thing we're trying to represent is design decisions and not CSS rules. 

Exactly! The only thing holding these properties together is the naming convention, and I don't think this is enough. We can't just throw out the concept of decision/option tokens since it is not compatible with the way we write code. If this is the model we're following, Sass variables in their current state _are not the right tool_. I think there's a lot of validity in the concept, so I would prefer to create primitives and come up with a better way to group what designers consider to be option and decision tokens. 

Screen Shot 2020-02-13 at 11 06 54 am

> I don't get this yet, might need an example or two before I can grok what you mean. To clarify, what are some examples of later modifications that could be difficult under the "bundle of rules" approach?

We have heading0, heading1, but what about something between those? The naming convention adopted in Figma doesn't allow for that. That's one of the drawbacks we identified with using t-shirt sizes pretty early on, but it also applies to this scenario. 

> Font sizes and line heights are subatomic to designers (like saturation for a colour) when it comes to the type styles
Not quite. Marc chose line heights and font sizes that are fractions of our grid unit. Why wouldn't we consider that in the system?

> If a designer decides that Combination X needs a different font size that's 1px larger, it's irrelevant to designers whether or not that new font size is part of an approved list, because they are not thinking in terms of font sizes, but in terms of typestyles.

**Comment by sebpearce** [comment]
Just leaving this here for reference as I used it in my chat with @ActuallyACat. Trying to leave some sort of paper trail!
sebs model

**Comment by ActuallyACat** [comment]
Since this is something that's blocking both streams of work (typography tokens and uplift) @sebpearce and I made a call to proceed with Bundle of Rules. _Should we_ want to pivot to one of the other options, we can do that after this cycle with little repercussions. 

Further, we both agreed we've made assumptions about what designers want without fully understanding their intentions. We'd like to get clarification on: 
1. Are the definitions of option and decision tokens used when building and explaining new UI? 
1a. What would you call this combination of styles:
Screen Shot 2020-02-13 at 5 37 55 pm

(I think there was another, @sebpearce will remember)

File: packages/design-tokens/less/typography.less [hunk 2]
```diff
@@ -0,0 +1,43 @@
+/** THIS IS AN AUTOGENERATED FILE **/
+
+@kz-typography-heading0-fontFamily: "Greycliff CF";
```
ActuallyACat: It still irks me that the case here is consistent

packages/design-tokens/tokens/typography.json
```diff
@@ -0,0 +1,69 @@
+{
+  "kz": {
+    "typography": {
+      "heading0": {
```
sebpearce: Rename to `display0`

sebpearce:
@ActuallyACat Did we decide to stick to `px` or `rem` in the end for font sizes? I vaguely remember talk about using `px` in the tokens (to help keep them in sync with Figma etc. since it's easy to notice when they're different) and then having an extra layer to convert them to `rem`, but I can't remember what we decided on.

ActuallyACat:
@sebpearce This is a little tricky. We'd probably need to annotate each property since there's quite a few that we might want to switch to px in certain circumstances. As an example, we would want our borders and box shadows to be px, but out typography to remain rem.

sebpearce:
Good point. Let's rem it up

mbylstra:
I have concerns of the readability of the rem units because of the math involved (it'd be easier if the root font size was 1px!), but seeing as we're not exporting and you don't use them directly, it's not so bad (it's only people that are working on the Typography tokens and components that would need to do the conversion).

packages/design-tokens/tokens/typography.json
```diff
@@ -0,0 +1,83 @@
+{
+  "kz": {
+    "typography": {
+      "display0": {
+        "fontFamily": "\"Greycliff CF\"",
+        "fontWeight": "400",
```
sebpearce: I just realised these `400` weights for Greycliff are wrong. Fixing now

mbylstra:
Just following up on this from a few comments back:

> Not quite. Marc chose line heights and font sizes that are fractions of our grid unit. Why wouldn't we consider that in the system?

Clarification: font-sizes are not in grid units (I'm pretty sure), but line-heights definitely are (so things lined up on the baseline grid). 

I think we should consider using grid units for line height, but I also think we can refactor this later (we need some way of referencing JSON tokens from within JSON tokens which is a whole other kettle of fish). The good thing is if we replaced `lineHeight: 0.567rem` with something like `lineHeight: { "$ref": "#/kz/spacing/unit-1-pt-5}` it'd be an internal only refactor (wouldn't make any difference to the consumers). I imagine at some point we will add spacing tokens as they are a big part of the design system, but that seems out of scope for the current cycle.

sebpearce:
We haven't included the type styles for these yet, but they will be needed by people like Kemble who are working on buttons.

`Buttons Primary` is like `Heading 4`, except that it's a weight of 700 instead of 600. I would like to add these styles so we don't have a situation where the designers think "We're using the `Buttons Primary` type style for button labels" and the engineers think "We're using the `Heading 4` style with a font weight override for button labels", which could lead to problems down the road if things change.

image

mbylstra:
I think we just leave the current scope to heading tokens as we can easily make follow up PRs to add more tokens (even today if we have time).

packages/design-tokens/less/typography.less
sebpearce: We could change the script so that it compiles to kebab case for sass/less?

sebpearce:
@ActuallyACat @mbylstra I've also just realised we may want to have the line height expressed as a unitless value instead of in `rem`s, in case somebody overrides the line height in a component and uses `px` instead.

mbylstra:
I'm not following: `rem` is relative to the body font size, not the element's font size, so the line height won't change if someone overrides and changes the font size, regardless of what unit they use (which we can't control)

sebpearce:
> I'm not following: rem is relative to the body font size, not the element's font size, so the line height won't change if someone overrides and changes the element's font size, regardless of what unit they use (which we can't control)

Exactly, and this is the problem. If they override the font size, the line height will stay the same, which will potentially look wrong. If we have unitless sizes, it will shift with the font size.

mbylstra:
It needs prettier to be run.

mbylstra:
" If they override the font size, the line height will stay the same," This is a design question. In the base grid land, it's exactly what we wanted. Who knows what we want now, but we can ask. There is currently no automatic relationship between font size and line height, and we'd be creating one.

sebpearce:
@mbylstra Alright you've convinced me. Prettier done also ✅

mbylstra:
I think we can potentially refactor and revisit when we have more clarity on whether we'll be using the baseline grid.

sebpearce:
Can't figure out how to make Moai satisfied with the PR title 🤔

ActuallyACat:
> This is a breaking change because of color.ts, right?

That's right.

PR #271, opened 2020-02-13
quick and rough spike on uplifting headings to greycliff (not for merging)

mbylstra:
This is an experiment to update headings to Greycliff and see how they look.

### DIFF
--- a/packages/component-library/styles/type.scss
+++ b/packages/component-library/styles/type.scss
@@ -2,6 +2,10 @@
 @import "color";
 @import "layers";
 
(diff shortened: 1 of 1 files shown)


mbylstra:
[Spike on replacing Ideal Sans heading with Greycliff in Murmur and/or Perform to get a sense of how much things will change (keep baseline shift)](https://trello.com/c/cEjJIxQ8/767-spike-on-replacing-ideal-sans-heading-with-greycliff-in-murmur-and-or-perform-to-get-a-sense-of-how-much-things-will-change-keep)

PR #273, opened 2020-02-14
feat: @kaizen/deprecated-kaizen-helpers package for typography scss mixins

mbylstra:
This PR provides typography mixins in new package, that can be used to replace the existing mixins for the OG styles. They should map cleanly to the existing mixins so all we need to do is find and replace strings.

Some decisions I made that are definitely up for debate:

- The name `@kaizen/deprecated-component-library-helpers`. I wanted to keep it pretty generic so we didn't have to keep creating new packages for mixins, vars, etc. This is where all the stuff we'd want to deprecate would go.

- The name of the story. It sticks out like a sore thumb, but it it's not a component, so it breaks the mold (but making a story for this was extremely useful, it revealed numerous bugs in my blindy written code, so I wouldn't want to remove the story just because it's not a component). It's ugly, but hopefully we can remove it when we eventually completely deprecate it.

- The fact that this not only contains the new mixins, but a fork of everything in `component-library/styles/type.scss` . The reason is that I think everything in `@kaizen/component-library/styles/type.scss` should be moved here, but we can't do that until we've renamed imports in mumur/BFR/perform.

I've added some comments below that explain the existence of the unit conversion functions: `to-unitless-pixels` and `to-grid-units`


packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,360 @@
...
+    $descender-height: $ca-greycliff-font-descender-height,
+    $line-height-in-grid-units: $line-height-in-grid-units,
+    $letter-spacing: $letter-spacing,
+    $weight: $weight,
+    $args...
+  );
+}
+
+@mixin kz-typography-display-0(
+  $size: to-unitless-pixels($kz-typography-display-0-font-size),
```
mbylstra: The reason for this is that I wanted to make the mixins backwards compatible with the old mixins, so that we can just to a simple string match and replace based on the mapping. Eg: find `ca-type-ideal-display` to `kz-typography-heading-3` . Sometimes arguments are passed into the existing mixings such as `ca-type-ideal-display(32)` which overrides the font size. The units are in "Unitless pixels" (there is no `px` at the end), so this function converts from the token in rems to "unitless pixels".

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,360 @@
...
+  $args...
+) {
+  @include kz-typography-greycliff(
+    $size: $size,
+    $weight: $kz-typography-display-0-font-weight,
+    $letter-spacing: $kz-typography-display-0-letter-spacing,
+    $line-height-in-grid-units:
+      to-grid-units($kz-typography-display-0-line-height),
```
mbylstra: The existing baseline shift system wants line heights in grid units, so this function converts from rems to grid units. Eg: `1.5rem` (24 pixles) becomes `1`

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,359 @@
...
+$ca-default-font-descender-height: 0.115;
+
+// Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
+$ca-ideal-sans-font-family: "Ideal Sans A", "Ideal Sans B",
+  $ca-default-font-family;
+$ca-ideal-sans-font-base-size: 1rem; /* 16px */
+$ca-ideal-sans-font-descender-height: 0.14;
+
+$ca-greycliff-font-base-size: 1rem; /* 16px */
+$ca-greycliff-font-descender-height: 0.098;
```
mbylstra: This is the value I created by hand. I'm sure there's a way to get this mathematically perfect, but this seems close enough.

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,359 @@
...
+
+@function to-unitless-pixels($value-in-rems) {
+  @return strip-unit($value-in-rems) * 16;
+}
+
+@function to-grid-units($value-in-rems) {
+  @return $value-in-rems / $ca-grid;
+}
+
+@mixin kz-typography-greycliff(
```
mbylstra: I like the `type` to `typography` naming we used in the tokens. "type" is such an overloaded word in programming and in general! It always confused the hell out of me.

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,359 @@
...
+    $size: $size,
+    $weight: $kz-typography-display-0-font-weight,
+    $letter-spacing: $kz-typography-display-0-letter-spacing,
+    $line-height-in-grid-units:
+      to-grid-units($kz-typography-display-0-line-height),
+    $args...
+  );
+}
+
+@mixin kz-typography-heading-1(
```
mbylstra: Also note: there's no need to put "greycliff" in the name like we did for "ideal", because the `kz` prefix means it can't clash with mixins like `ca-type-heading`

sebpearce:
> The name @kaizen/deprecated-component-library-helpers. I wanted to keep it pretty generic so we didn't have to keep creating new packages for mixins, vars, etc. This is where all the stuff we'd want to deprecate would go.

On this, I love the idea of making it clear that it's deprecated – the thing is this means we'll have a situation for a few weeks where the new stuff has the word "deprecated" in it, and the old stuff we're moving away from doesn't (`styles/type.scss`). I wonder how we could mitigate that.

packages/deprecated-component-library-helpers/styles/type.scss
sebpearce: image
I don't think I understand how the formula works 🤔 I thought it would be the `descender height in units`/`units per em`, which for Greycliff would be `389/2048 = 0.1899`. That looks quite different to what you got by hand, which makes me think that these metrics I mentioned here don't represent what I thought they represented. Hmmm

ActuallyACat:
Screen Shot 2020-02-18 at 9 20 35 am

This is an interesting concept. Following on from this idea... Should we chat about reformatting our stories?

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,359 @@
...
+
+@mixin kz-typography-greycliff(
+  $size,
+  $weight,
+  $line-height-in-grid-units: 1,
+  $letter-spacing: $kz-typography-heading-1-letter-spacing,
+  $args...
+) {
+  @include ca-type(
+    $size-ratio: $size / 16,
```
ActuallyACat: If I'm understanding this correctly, the division by 16 is to convert the size to `rem`?

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -0,0 +1,360 @@
...
+  @include kz-typography-greycliff(
+    $size: $size,
+    $weight: $kz-typography-heading-3-font-weight,
+    $letter-spacing: $kz-typography-heading-3-letter-spacing,
+    $line-height-in-grid-units:
+      to-grid-units($kz-typography-heading-3-line-height),
+    $args...
+  );
+}
+
```
mbylstra: These shouldn't be deprecated, so they should be removed

sebpearce:
[Uplift type mixins (type.scss) in Kaizen](https://trello.com/c/tmp9YSHv/751-uplift-type-mixins-typescss-in-kaizen)

mbylstra:
@ActuallyACat  by the way, we ended up deciding to just delete the stories. They were really useful for development, but they don't need to remain. They really didn't fit in.

PR #275, opened 2020-02-14
feat: Add font-face declarations for Zen fonts

sebpearce:


ActuallyACat:
I agree with @mbylstra, a more generic directory would work better here.

I've also added this to our dev backlog: https://trello.com/c/A5cX7wsQ/428-add-woff2-support-for-greycliff

sebpearce:
Roger that, I'll put it in `styles/`.

PR #276, opened 2020-02-14
feat: Add Generics to Select/AsyncSelect options types

frimoldi:
Hello DST 👋🏼!

## Context

Back in Perform Camp, we found ourselves in the need of using `AsyncSelect` to display a `Select` with a set of options. But, instead of just displaying a label for each option, we wanted to render a custom component, like Avatars, multiple labels, etc).

Good news is that `react-select` allows you to do that, by using the `formatOptionLabel` function, which takes an `option` parameter and returns a `React.Node`. So far so cool.

## Problem

The problem is that the default type for that `option` parameter one with the shape of:
```
{ label: string, value: string }
```

While this is enough for displaying a simple label for each option, it didn't work for us, since we also need a more complex object as option, like a `User` object, for instance.

Also good news, `react-select` types support that, through Generic Types! 🎉

## Solution

The issue was that our very own `Select` and `AsyncSelect` were not allowing developers to pass the type along to `react-select`. So my change was to make `Select` and `AsyncSelect` to accept a type argument, to be used as the type for the option objects by passing it to the `react-select` components.

Note that this is not a breaking change at all, both `Select` and `AsyncSelect` will continue working even if you don't specify this type at all, defaulting to the `react-select` default -> `{label: string, value: string}`

## Usage

```
type CustomOptionType = { name: string, avatar: string }

<Select<CustomOptionType>
  options={[
    { name: "Fran", avatar: "foo/bar.png" },
    { name: "Rick Sanchez", avatar: "foo/picklerick.png" }
  ]}
  formatOptionLabel={option => (
    <div>
      <Text>{option.name}</Text>
      <Avatar imageURL={option.avatar} />
    </div>
  )}
/>
```


ActuallyACat:
Closing as this PR is now stale. Please reopen if I've closed this in error.

PR #280, opened 2020-02-17
feat: Add Guidance Block draft component

juliohr:
## Overview
The call to action component is a pattern that emerged organically. There is no guidelines yet, but the common theme is an illustration on the left side, title and description in the middle, and a button on the right that triggers an action.

It is optional to have a close icon on the top right corner to dismiss it. 
When dismiss action is triggered, there is an animation to fade out and collapse the banner, then on transition end, it is removed from the DOM.

## Usage
This draft component is tailored for the `streamlined onboarding survey call-to-action` on murmur's survey index page. 
It requires some investigation in order to replace the other call to action components. The other implementations of the call to action are:
- murmur/app/client/modules/ca-onboarding/SurveyTrainingBanner.js
- murmur/app/client/modules/ca-action-framework/components/GuidanceBanner/GuidanceBanner.js

### streamlined onboarding survey 
Survey_Index

### survey trainer
Survey_Index

### Action framework
Engagement__One-click_Onboarding

## Limitations of the current draft
- It does not account for different types of button style
- dismiss is mandatory. (some CTAs are not dismissible)

## Murmur usage
The plan is to first replace the streamlined onboarding survey cta, which can be done with little effort. Over time we can start expanding this component to fit the needs of other areas.


sebpearce:
Thanks for the detailed summary, this really helps. 👌

sebpearce:
@juliohr Code is looking good at first glance. The thing I would mainly be concerned about is the design.

There's a WIP component in the upcoming Zen UI Kit that matches this and it's called a Guidance Block – see below.

The two things it'd be good to clear up are:
- Let's make sure the design used for this component is exactly what designers want. I noticed that all 3 of the screenshots you posted of the existing ones are different – different size image, different font weights and sizes, different spacing between heading and text. It'd be great if we could get 100% clear on which of these is the "correct" one moving forward and then make this component look like that.
- Since the component in the new world is called a Guidance Block, it would be a good idea to sync with designers on the right name for this component. Calling it the same name as the upcoming Zen equivalent would make it easier to do the transition later. Or maybe designers may decide that the one in the Zen UI Kit should be called a Call To Action instead.


image

juliohr:
This is a screenshot from the storybook.
Is there any design changes in regards to colours, fonts, margins, padding, etc?

GuidanceBlock___React___-_Default_⋅_Storybook

sebpearce:
Yes, there are design changes needed. You'd have to check with a designer to be 100% sure (both of the designers in DST are not working today), but I think the third example from the Action Framework is the most correct. The key things I noticed were:
- the text colour, which should be Wisteria 800 instead of pure black,
- the space between the heading and body text, which should look like the screenshot below,
- the name of the component, which designers have said should be "Guidance Block" (https://cultureamp.slack.com/archives/CETLTFG9G/p1582164184216300?thread_ts=1581901325.191700&cid=CETLTFG9G).

image

packages/component-library/draft/Kaizen/GuidanceBlock/GuidanceBlock.scss
```diff
@@ -36,7 +37,7 @@ $ca-banner-collapse-height-delayed: margin-top $ca-duration-fast
   padding: $ca-grid / 2 $ca-grid / 2 0 $ca-grid / 2;
 
   h3 {
-    margin-bottom: 0;
+    margin-bottom: $ca-grid / 2;
```
sebpearce: image

The way the Text component is intended to be used is that you pick ` style` prop for your style, and the `tag` prop can be whatever is semantically correct for your use case on the page (i.e. if it's a heading then `h3` is good).

So in your case,
image
would look like:
```
<Text style="heading" tag="h3">{this.props.text.title}</Text>
<Text style="notification" tag="p">{this.props.text.description}</Text>
// (I think it's notification, could be paragraph though, need to check visually)
```
I just checked our Storybook and realised the stories don't match this, which sucks and I'll aim to change them so they do demonstrate this approach – sorry for any confusion.

For the `margin-bottom`, rather than reaching into the component (which makes the component brittle), we encourage people to wrap the Text component in a div and put padding/margin on that. The upcoming text components will make this much easier by the way!

mbylstra: > You can strip out any in-built margins from the Text component by using the inline prop.
Does this also change it do `display: inline` -  I don't know that that's what you'd want?

mbylstra: (looking at the component implementation, it adds `inline-flex`... still, I don't know if that's want you want if you just want to remove margins)

sebpearce: > Does this also change it do display: inline - I don't know that that's what you'd want?

If you're wrapping it in a `div` to for the margin, and a `div` is `display: block`, the `inline` becomes irrelevant.

mbylstra: That makes sense - it's just not super intuitive. All good.

sebpearce: Yeah agreed, we're scrapping it in the next batch of text components.

juliohr: what do I need to change in the end? 🤔

juliohr:
I made some designs changes to make it conform to our updated banner version. Now the button is of type default and has an arrow forward arrow. 

Also, removed custom style on Text component, and simplified the text area (mid section) styles.

sebpearce:
image
This gap between the heading and body text still seems quite large – I noticed there is still no div with a custom margin-bottom wrapping the Text component used for the heading. Isn't that what you intended to do to fix this?

Disclaimer: I'm not a designer so can't make calls here, but I'm expecting this to come up when you do talk to someone so I'm mentioning it early.

juliohr:
Design reviewed with Mindy.

Storybook

Some responsiveness: 
Storybook

@katsambarcus Mindy asked you to give a final look?

katsambarcus:
Thanks for this @juliohr – there are a couple of minor things that I can see which will need to be tweaked.

1. The illustration icon is looking a bit smaller than the example in [Zen UI Kit](https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A39077). The size of the image should be 155 x 155px.

--

2. Padding on the left of the icon illustration is different (36px) to the padding on the right of the button (24px). These would need to be the same for balance. I would suggest giving the padding on the right more space (increase padding by 12px to 36px).

--

3. The spacing between the H3 and Paragraph Body text is too big. Looking at the [Guidance Block](https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A39077) in Figma, the spacing between H3 and Body is 6px (1/4 Grid Unit).

image

--

4. The width of div container that holds the text appears to be too short. The max width of a paragraph of text should be 650px. If this width runs into the button, then there should also be 36px margin between text and button.

image

--

5. To comply with our copy and language guidelines, please remove the exclamation mark (!) from the button.

sebpearce:
I'm putting this here for the record – there is not yet any mobile responsiveness. I assume this is because this was mainly copied from an existing use case which doesn't have it.

This is OK for now if the existing versions don't do it (that is, nothing will break when people replace those with an instance of this component), but this will need to be fixed when this component makes the transition to Zen.

packages/component-library/draft/Kaizen/GuidanceBlock/GuidanceBlock.scss
```diff
@@ -0,0 +1,58 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/component-library/styles/layout";
+@import "~@kaizen/component-library/styles/color";
```
sebpearce:
```suggestion
@import '~@kaizen/deprecated-component-library-helpers/styles/color';
```

packages/component-library/draft/Kaizen/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <img
+            src={this.props.img.src}
+            alt={this.props.img.alt}
+            height="155px"
+            width="155px"
+          />
+        </div>
+        <div className={styles.descriptionContainer}>
+          <div className={styles.headingWrapper}>
+            <Text inline={true} tag="h3">
```
sebpearce:
```suggestion
            <Text inline={true} tag="h3" style="display">
```

sebpearce: We're about to merge a PR that makes sure that all Text components have a `style` prop.

(This is what I was trying to explain in [this comment](https://github.com/cultureamp/kaizen-design-system/pull/280#discussion_r384292864).)

packages/component-library/draft/Kaizen/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -0,0 +1,128 @@
...
+              {this.props.text.title}
+            </Text>
+          </div>
+          <Text tag="p">{this.props.text.description}</Text>
```
sebpearce:
```suggestion
          <Text tag="p" style="body">{this.props.text.description}</Text>
```
Same as above. I went and checked the styles for the Text component to see what style the `p` tag falls back to, and it's `body`.

packages/component-library/stories/GuidanceBlock.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default {
+  title: "GuidanceBlock (React)",
+}
+
+export const Default = () => (
+  <div style={{ display: "flex", justifyContent: "center", margin: "100px" }}>
+    <GuidanceBlock
+      img={{ src: bannerImg, alt: "Guidance block" }}
+      text={{
+        title: "This is the Guidance title",
```
sebpearce:
```suggestion
        title: "This is the Guidance Block title",
```

PR #283, opened 2020-02-17
refactor: Make elm modal completely dependant on subscriptions

Confidenceman02:
Instead of returning Cmd ModalMsgs when triggering the modal it makes
more sense to simply ask the modal to run. This is a simple field in the
modal data that will be picked up by subscriptions.

This will then trigger the modal update in which case the modal can do
all the normal things it was doing previously. This moves all the modal
running logic to just one flow.

# Breaking Changes
Old trigger
```
triggerModal : (ModalState msg, Cmd (ModalMsg), ModalStatus)
triggerModal =
    trigger modalState
```
New trigger
```
triggerModal: ModalState msg
triggerModal =
    trigger modalState
```

### DIFF
--- a/packages/component-library/draft/Kaizen/Modal/Modal.elm
+++ b/packages/component-library/draft/Kaizen/Modal/Modal.elm
@@ -123,8 +123,8 @@ type alias ModalData =
     , firstFocusableId : FirstFocusableId
     , lastFocusableId : LastFocusableId
     , defaultFocusableId : DefaultFocusableId
-    , runModal : ModalStep
-    , previousTime : Maybe Time.Posix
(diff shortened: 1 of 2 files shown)


Confidenceman02:
@mbylstra I completely agree about the story with the button. They were originally setup to be triggered by a button but to make it the same as the react stories I removed them. I will add one back though because it will show the full flow of how to ideally hook it up.

I know what you mean about it being non standard. I am am trying to experiment with different patterns and approaches and its yet to be seen if they will pay off.

The move to subscriptions was definitely worth it as queries the modal will do for focus-lock can be done in animation frames. A trick I learned about from the react-beautiful-dnd.

PR #287, opened 2020-02-18
feat: Added heading and box

ActuallyACat:
~Opening this up as a draft to seek some feedback from you all about the approach.~
The initial feedback indicated that there doesn't seem to be anything spicy in here so I'm moving this to a Pull Request.

### Allowing classNames on components
This can be quite a controversial approach. As much as possible, it's important for our team to try to anticipate the needs of those we support. In practice, and especially at Culture Amp, a Design System is reactive. Restricting components to **not allow** classNames will force people into hacks or approaches that don't use the system. I have added `classNameAndIHaveSpokenToDST` to `Heading` and `Box`, but would like to hear some opinions on this. 

### Component Safelists
~See here for all the props I think we should allow on components: https://github.com/cultureamp/kaizen-design-system/pull/287/files#diff-aa90a6cd282df351d6c0f030c45a4ecaR3-R12
`automationId` is a prop I see throughout the system that comes to mind. IMO, since teams are writing the automated tests, it should be up to them to add whatever data attributes they need on the components. I've also added `Item*` props to the safelist.~

Removed this feature until we figure out the best way to support users wanting to target specific elements within the component (for example aria- should appear on the button element, whereas other attributes might be okay on the wrapper

### Weirdly atomic-but-not-really classes
I'm referring to the spacing classes here. In the future, it might be neato to allow a subset of these classes in other components. As an example: 
```
// rough scaffold of what Well.tsx might look like
export type ResponsiveSpacingForWell =
  | {
      [key in keyof Breakpoint]?: "1" | "4" // this fictional component only allows 1 unit and 4 units of margin
    }
  | GridFractions

export const Well = (
  m: ResponsiveSpacingForWell
) => {
  const classes: string[] = classnames(
    styles.defaultWellStyles,
    ...responsiveMarginClasses({ m }),
  )

  return (
    <div className={classes}>{children}</div>
  )
}

// Using the component
<Well m={{ mobile: "1", desktop: "4" }}>
  {/* More stuff */}
</Well> 
```

I think Sass also generates the same hash when a class is reused, so there's an ever so slight benefit of not bloating our stylesheets. *CITATION NEEDED*. 

### Spacing units not consistent with guidelines
~The guidelines I'm referring to are these ones: https://cultureamp.design/guidelines/spacing/
In practice, I've seen people use `$ca-grid/8`. I've seen some units hardly used at all (3.5, 2.5 both come to mind). This might be a good time to sanity check what we actually need/use.~ 
After discussing with @katsambarcus, we think these units will cover most consumer use cases:
```
0 / 0px
0.25 / 6px
0.5 /  12px
0.75 / 18px
1 / 24px
1.25 / 30px
1.5 / 36px
1.75 / 42px
2 / 48px
2.5 / 60px
3 / 72px
3.5 / 84px
4 / 96px
```

### Why no Elm?
It's coming! I am hoping for feedback on this approach before I reimplement this in Elm :D 

👈😀👈


packages/component-library/components/Box/Box.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export interface BoxProps {
+  children: React.ReactNode
+  classNameAndIHaveSpokenToDST?: string
+}
+
+export const Box = ({
+  children,
+  classNameAndIHaveSpokenToDST,
+  p,
```
ActuallyACat: I've borrowed the naming conventions from `styled-system`: https://styled-system.com/api#space

A couple of thoughts:
* Should we allow long names too? (m and margin, pl and paddingLeft, etc?)
* Should our grid accept negative numbers?

packages/component-library/components/Spacing/padding.ts
```diff
@@ -0,0 +1,77 @@
...
+      classes.push(styles[`p-${convertFractionToString(unit[key])}--${key}`])
+    })
+  }
+
+  return classes
+}
+
+// @TODO - i18n. Use padding before / padding after
+// Seb can help with this
+export const responsivePaddingClasses = ({
```
ActuallyACat: Hey @sebpearce could you help with this 👼? Do we switch paddings and margins for LTR languages?

packages/component-library/components/Spacing/Base.module.scss
```diff
@@ -0,0 +1,15 @@
+@import "~@kaizen/component-library/styles/fonts";
+@import "~@kaizen/component-library/styles/type"; // just for the grid unit
+
+$spacing-list: (
+  ("eighth", $ca-grid/8),
```
ActuallyACat: I made sure the classnames these spits out are somewhat easy to reason about. 
Padding of 0 is `p-0`
Padding of 1/8 is `p-eighth`

mbylstra:
I love the `classNameAndIHaveSpokenToDST` name! I think you have one-upped the `dangerouslySetInnerHtml` name.

mbylstra:
I think the spacing in https://cultureamp.design/guidelines/spacing/ is out of date (it just reflects the current state, not Zen). I can't actually see it defined in the Figma file, but I'm of the understanding that the new base unit is intended to be `6px` (one unit would be 6px rather than 24px, and there wouldn't be any fractional values - 6px is the smallest you can go), but this is just from memory. I think it's worth speaking to @katsambarcus about.

packages/component-library/components/Box/Box.tsx
kemblesong: The plus side of short names only is that you won't get confused with actual `style` attribute keys, which use the long names, but expect css strings.

kemblesong: I don't think a `Box` should have negative values. It's usually a hack to add gutters to grids or add "aesthetic overlapping", and both of those cases are either rare or very specific.

ActuallyACat: @kemblesong Yeah, that's a really good call out. As much as I like the idea of allowing both naming conventions I think it may cause confusion.

packages/component-library/stories/Heading.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+  <Heading variant="display-0">This is a display heading</Heading>
+)
+
+export const Heading1 = () => <Heading variant="heading-1">Heading 1</Heading>
+
+export const Heading2 = () => <Heading variant="heading-2">Heading 2</Heading>
+
+export const Heading3 = () => <Heading variant="heading-3">Heading 3</Heading>
+
+export const Heading4 = () => <Heading variant="heading-4">Heading 4</Heading>
```
ActuallyACat: I really wanted to use levels here (i.e. `<Heading level={1}>` but this wouldn't translate well to the naming in Paragraphs (intro-lede, paragraph-body, as `<Paragraph variant="intro-lede">`, etc). Another alternative for `variant` might be `appearance` (Atlaskit), `type` (Ant Design), `intent` (Redbubble).

Whatever name we use here will be carried through to other typography components and buttons (to start with). 

@katsambarcus @didoesdigital Did you have any thoughts on this one?

ActuallyACat:
[[react] Heading and Box components for typography uplift](https://trello.com/c/MFENdAjG/784-react-heading-and-box-components-for-typography-uplift)

packages/component-library/components/Spacing/padding.ts
sebpearce: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/styles/layout.scss

Yep we actually have mixins for this:

image

How we do this in these components is up for debate.

packages/component-library/stories/Heading.stories.tsx
sebpearce: For what it's worth, my vote is for `variant` across the board.

packages/component-library/components/Box/Box.tsx
jasononeil: I might be old fashioned but I prefer the long names 😅 I find less thinking is involved when reading code, especially if I'm new to the API.

But this is total bikeshedding, feel free to disregard

packages/component-library/components/Spacing/padding.ts
ActuallyACat: Interesting... That throws a spanner in the works! So looking through the repo and this applied in certain contexts and it shouldn't be a global rule. With that in mind here are some initial thoughts 

Box could take a rtl prop. When this is applied all left and right paddings and margins will swap. Where this is slightly annoying is it will apply to both margin and padding, and developers may only want rtl on one of those. To mitigate they could use another box without `rtl`.
```
<Box px="2">
   <Box rtl={true} mr="4" ml="2>Content</Box>
</Box>
// outputs
margin-right: 2*$ca-grid;
margin-left: 4*$ca-grid;
padding: 2*$grid 0;
```

Thoughts?

packages/component-library/stories/Heading.stories.tsx
ActuallyACat: @sebpearce That's looking like it's going to be the most consistent to the rest of the Kaizen components, so i'm inclined to leave it.

ActuallyACat:
Screen Shot 2020-02-20 at 9 41 20 am

LHS the new Tile component in Zen
RHS building component with the new spacings

The spacings don't seem to be fit for purpose right now - I'm thinking we will need to support the half units above 1unit, like 1.5, 2.5 etc. Potentially more. These are the units in tile: 
- 42px (1.75 grid units)
- 24px (1 grid unit)
- 48px (2 grid units)
- 6px (0.25 grid units)
- 30px (1.25 grid units)
- 16px (0.6 grid units) -> potential unicorn

If we supported 0.25 and 0.5 increments the units that the Spacing module supports would also need rework. `<Box p="1 1/2">` doesn't read as well as `<Box p={1.5}>`. 

One thing to keep in mind is I'm not suggesting we shouldn't allow "unicorn" spacing. Sometimes it makes sense for something to be 2px to get it looking _just right_. My hope for spacing is that it would satisfy the vast majority of the needs of developers using Kaizen.

**Recommended reading:**
https://www.figma.com/blog/line-height-changes/
> In text boxes, Figma will now distribute line height just like web does. This should help most people with most tasks — not just designing for the web.

packages/component-library/components/Spacing/padding.ts
ActuallyACat: ^ This is what I've implemented. If someone thinks this is a bad idea let me know!

sebpearce: RTL prop sounds 👌 to me. We're doing this already in a few cases I think.

packages/component-library/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+  | "h1"
+  | "h2"
+  | "h3"
+  | "h4"
+  | "h5"
+  | "h6"
+
+export interface HeadingProps {
+  classNameAndIHaveSpokenToDST?: string
+  children: React.ReactNode
```
mbylstra: I'm sure I've seen headings in the wild that contain things like italic for just one word, so making this flexible witih `ReactNode` is a probably a good thing.

I'm not so familiar with the TS react typings - Does `React.ReactNode`  also permit a list of nodes?

Eg:
`<h1>The <em>final</em> countdown </h1>`
or would you have to do:
`<h1><div>The <em>final</em> countdown</div></h1>`

packages/component-library/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+  | "span"
+  | "h1"
+  | "h2"
+  | "h3"
+  | "h4"
+  | "h5"
+  | "h6"
+
+export interface HeadingProps {
+  classNameAndIHaveSpokenToDST?: string
```
mbylstra: Heavy Plus

packages/component-library/stories/Heading.stories.tsx
mbylstra: FWIW: I quite like `appearance` as it's very accurate even if it's a little long. I was thinking `style` which is common in things like Microsoft Word and Google Docs, but that's way overloaded with the meaning in css/stylesheets.

I seem to recall we were on a mission to remove `variant` from all the components because it's too vague, such as renaming `variant` to `mood` in the Notification and Modals.

packages/component-library/components/Spacing/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  | 0.75
+  | 1
+  | 1.25
+  | 1.5
+  | 1.75
+  | 2
+  | 2.5
+  | 3
+  | 3.5
+  | 4
```
sebpearce: I think we need to come up with a solid solution to this before we ship it. @ActuallyACat I vaguely remember you saying that we might confine this list to the Headings only, rather than having them available for other components – is that right?

packages/component-library/components/Spacing/types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type Breakpoint = {
+  mobile: "0"
+  tablet: "768"
+  desktop: "1024"
+}
```
sebpearce: I assumed from our chat the other day that we'd have a crack at one of the alternative approaches we talked about for doing breakpoints (i.e. allowing people to pass in custom breakpoints) before shipping this. Was that something you planned for a future PR after shipping this one?

packages/component-library/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+  | "heading-2"
+  | "heading-3"
+  | "heading-4"
+  | "heading-5"
+  | "heading-6"
+
+export type AllowedTags =
+  | "pre"
+  | "p"
+  | "a"
```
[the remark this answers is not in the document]
mbylstra: Also, the `div` allowed tag + `children: React.ReactNode` is a loop hole that let's you put anything in there. I think that's OK as it would be pretty unlikely, and at least this makes people thing twice about doing something super weird like using a `script` tag for a heading.

packages/component-library/components/Heading/Heading.tsx
ActuallyACat: Here's the definition for the types: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L203

So all of those examples you posted would be valid with this component.

packages/component-library/components/Heading/Heading.tsx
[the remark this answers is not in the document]
ActuallyACat: Yeah, this will be especially useful when the Paragraph component is built out. You'd be able to use the components in a way that feels closer to using plain ol' html.
```
<Paragraph>This is a paragraph with a <Link href="#">link</Link>.</Paragraph>
<Heading>You can also add <Link href="#">links</Links> within a heading - but why would you want to?</Heading>
```

packages/component-library/components/Spacing/types.ts
ActuallyACat: These have been moved from the "global types" (`packages/component-library/types.ts`) to be within the Spacing package. At this time, the spacing classes only support these units so I think it's fairly safe to keep it here.

packages/component-library/components/Spacing/types.ts
ActuallyACat: Just so we're both understanding how this works at the moment: 
https://github.com/cultureamp/kaizen-design-system/pull/287/files#diff-94e1cb36c969eb01867046d18c80f301R22-R29

This will generate classes like these ones: 
```
@media (min-width: 768px) {
   .m-0--tablet {
     margin: 0
   }
   .m-0-point-25--tablet {
     margin: 0
   }
   .m-0-point-5--tablet {
     margin: 0
   }
   ...etc
}
```

Depending on the props that are passed into the Spacing function, it will then generate the classes that should be applied to the component. So this is what's happening under the hood: 
```
// margin.ts
m({mobile: 0, tablet: 1.5, desktop: 3})

// array of classes generated from the above function
[ 'm-0--mobile', 'm-1-point-5--tablet', 'm-3--desktop']

// relevant css generated from margin.module.scss or padding.module.scss
@media (min-width: 0px) {
   .m-0--mobile {
     margin: 0
   }
}
@media (min-width: 768px) {
   .m-0-point-5--tablet {
     margin: 0
   }
}
@media (min-width: 1024px) {
   .m-3--desktop {
     margin: 0
   }
}

// react component
<Heading className="m-0--mobile m-1-point-5--tablet m-3--desktop>Stuff</Heading>
```

Because we're using Sass and want to reproduce this functionality for both React and Elm, we can't do this dynamically. Those CSS classes need to be determined ahead of time, they can't be dynamic and they have to be generated from a finite list of breakpoints. 

If people start using this I think we can put a little more consideration into it. Here's some pros/cons for some approaches from the top of my mind:

*This approach*
pro: 
* CSS-based solution, so it's easy to reproduce this feature in Elm.
con: 
* The spacing is restricted to these breakpoints

*Dynamic breakpoints*
pro: 
* Dynamic! Yay! 
con: 
* Has to be implemented in Elm/React. 
* You can't use media queries on inline styles in React. So you'd have to look to use Window.matchMedia() or something like that.

packages/component-library/components/Spacing/types.ts
sebpearce: Ah right. I assumed that the fact that Spacing was its own package meant that it was intended to be used with multiple components

packages/component-library/components/Spacing/types.ts
sebpearce: :thinking_very_much:

sebpearce:
```
The spacing is restricted to these breakpoints
```
☝️ I guess the critical question for me here is whether this is a dealbreaker. If we do it in such a way that we can _add_ custom breakpoints later I'm totally on board, but if we lock ourselves into an approach where we can't do that easily, we're going to run into that issue pretty quickly I think.

mbylstra: One idea I've had around being able to pass in custom breakpoints is using **CSS variables** for the breakpoints. **CSS variables are not supported in IE11, but that doesn't matter** as we only support desktop for IE11 (I've never heard of a mobile device or tablet that runs IE11). So as long as the default is desktop, it should continue to work in IE11.

I think it might be a very long time or potentially never until we move away from sharing SCSS between Elm and React (not even sure this is what we want to do - the whole reason we are using scss is to be to share styles between React and Elm so there's less code duplication. Still, it's a much bigger discussion). So, I think this is worth considering.

It's possible to set css variables from within Elm (it's a little wierd) as well as in JS.

ActuallyACat: @sebpearce I lifted those particular breakpoints from the current codebase. I'll get some feedback from the team whether these breakpoints will be okay to push forward with (keeping in mind the LHS navigation will be disappearing soon).

@mbylstra Interesting idea! Wouldn't you run into issues if you had a box with a certain set of breakpoints and another that needed to use another set of breakpoints?

mbylstra: I didn't think about that, but I thinking further realised you can scope CSS variables to a class. So, if you know the name of the CSS Modules classname of a component (not quite sure how you'd get that from outside the component), then you can set a general var name like `--mobile-breakpoint`  but scope it to the class. I feel like it's theoretically possible, but might need some trickery.

mbylstra: (this is something you can't do at all with scss variables I believe)

mbylstra: I made a proof of concept in Codepen: https://codepen.io/mbylstra/pen/GRJvpmP?editors=1111.

- The first box defaults to the global default.
- the second box is overridden in CSS in a local scope
- the third box is overriden in JS in a local scope

Looks like you need direct access to the DOM node to be able to do this. In React, you can get the ref, in Elm this would be tricky or annoying. I _think_ you could also dynamically add a stylesheet that adds the override, which would avoid needing the reference to the dom node (no need to to inline styles).

mbylstra: media queries might pose further complications, but still seems theoretically possible.

ActuallyACat: What I'm hoping to achieve from these breakpoints within Box: 
1 - Intrinsic documentation. These should be easy to reason about without needing to constantly refer to documentation. 
2 - Suitable for the _majority_ of our design use cases
3 - This should make it quicker for developers to add media queries for basic padding/margins while working in the context of a component
4 - Performant. Ideally not using JS events.

I think this solution meets those requirements (but I will be checking in with the broader team about #2).

The options for more bespoke use cases (IMO) are allowing people to write their own media queries (which we cannot restrict anyway), and/or bring in a package (or make one) for container queries. My preference is to do the former, and consider the latter when use cases arise. 

Here's some container query packages:
https://github.com/react-container-query/react-container-query
https://atlaskit.atlassian.com/packages/core/width-detector

Stuff other design systems are doing: 
**Braid Design System** 
Breakpoints - A static list: https://github.com/seek-oss/braid-design-system/blob/da1375e06171b27205c3f455b08c2fc406dc236c/lib/themes/baseTokens/seekAnz.ts#L137-L144
Container queries (sort of): https://github.com/seek-oss/braid-design-system/blob/bc8b6616949063197d475ac0b597736e544f6826/lib/components/ContentBlock/ContentBlock.treat.ts

Github is using Styled System: 
https://github.com/primer/components/blob/master/src/constants.js#L38
https://styled-system.com/#layout

ActuallyACat: @mbylstra Just had a little bit of a play with that idea: https://codepen.io/actuallyacat/pen/VwLzabP

Unfortunately, we can't use CSS variables in media queries

packages/component-library/stories/Heading.stories.tsx
didoesdigital: My previous notes on this:

- level
- hierarchy
- variant
- headingSize/paragraphSize
- headingSizing/paragraphSizing
- headingScale/paragraphScale
- scale

My concern with "variant" or "type" is that it is a generic catch all name that could mean basically anything. In a design system, you'd see that used to describe colors, length, "mood", spacing, layout/composition, anything. It's often already taken when you come to it, and means something different across every component. 

That said, "variant" when truly used to mean the main variants of this component, would be the one time I'd be comfortable with it. For Heading, I would expect the "variant" prop to describe the main thing that varies between the main uses of headings: its levels.

packages/component-library/components/Spacing/types.ts
mbylstra: Ah, you are right! That does well and truly kill that idea!

mbylstra: I found another way by creating a new stylesheet dynamically: https://codepen.io/mbylstra/pen/GRJvPJV?editors=1111. I'm not saying this isn't a massive hack, or that it wouldn't be bad for performance, but just food for thought. (also possible in Elm because you can just add `<style>` before the Heading)

mbylstra: I don't mean that we should add something like that to this PR, it's more that if need arises to provide custom breakpoint values, there are still options to explore if we are limited to scss+Elm+React (this can come later)

ActuallyACat: Thanks for looking into this, definitely something we should look into when we add support for dynamic breakpoints

PR #290, opened 2020-02-18
feat: set focus and blur attributes

Confidenceman02:
Set focus and blur attributes on the Button component.

Elm modal when handling focus-lock will need these attributes on the buttons that are used in Confirmation variants.

### DIFF
--- a/packages/component-library/components/Button/Button.elm
+++ b/packages/component-library/components/Button/Button.elm
@@ -18,7 +18,9 @@ module Button.Button exposing
     , iconPosition
     , id
     , newTabAndIUnderstandTheAccessibilityImplications
+    , onBlur
     , onClick
+    , onFocus
     , primary
     , reverseColor
     , reversed
@@ -27,10 +29,10 @@ module Button.Button exposing
     )
 
 import CssModules exposing (css)
-import Elm18Compatible.Html.Events as Events exposing (defaultOptions, onWithOptions)
 import Html exposing (Html, a, button, span, text)
 import Html.Attributes
 import Html.Attributes.Aria
+import Html.Events as HtmlEvents
 import Icon.Icon as Icon
 import Icon.SvgAsset exposing (SvgAsset)
 import Json.Decode as Json
@@ -105,10 +107,12 @@ view (Config config) label =
 
         attribs =
             buttonClass
-                ++ onClickAttribs (Config config)
+                ++ onClickAttribs config
+                ++ onFocusAttribs config
+                ++ onBlurAttribs config
                 ++ automationIdAttr
                 ++ titleAttr
-                ++ buttonTypeAttribs (Config config)
+                ++ buttonTypeAttribs config
                 ++ idAttr
     in
     span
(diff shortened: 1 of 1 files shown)


packages/component-library/components/Button/Button.elm
```diff
@@ -27,10 +29,10 @@ module Button.Button exposing
     )
 
 import CssModules exposing (css)
-import Elm18Compatible.Html.Events as Events exposing (defaultOptions, onWithOptions)
```
mbylstra: This is good to get rid of. We don't have any 0.18 code any more 🎉

packages/component-library/components/Button/Button.elm
```diff
@@ -145,19 +169,26 @@ onClickAttribs (Config config) =
...
             ]
 
         Nothing ->
             []
 
 
-buttonTypeAttribs : Config msg -> List (Html.Attribute msg)
-buttonTypeAttribs (Config config) =
+buttonTypeAttribs : ConfigValue msg -> List (Html.Attribute msg)
+buttonTypeAttribs config =
```
mbylstra: Is there a reason why you switched from the opaque type to the non-opaque record? This seems inconsistent with the rest of the configuration options. Although opaque types are a bit more of a pain to work with, they make it easier to add/remove attributes later without needing a major change. This is especially important in core components like Button, which is used in 100s of places.

packages/component-library/components/Button/Button.elm
```diff
@@ -145,19 +169,26 @@ onClickAttribs (Config config) =
...
+                            , preventDefault = preventDefault
+                            }
+                        )
+                        (Json.succeed msg)
             in
-            [ onWithOptions
-                "click"
-                { defaultOptions | preventDefault = preventDefault }
-                (Json.succeed msg)
+            [ HtmlEvents.custom "click" decoder
```
mbylstra: I really appreciate this change and removing the 0.18 compatibility dependency - thanks! Could I be a pain and ask you to split this into a separate PR though? It's unrelated to setFocus and setBlur. Just in case this change causes issues, we need to be able to roll it back easily (at the very least it should be a separate commit).

It also makes it easier to categorise the change. adding `setBlur` and `setFocus` is a `feat`, but changing the implementation of `onClick` is a `fix`. It makes it easier to release changes as it's clearer what's going to change or not.

packages/component-library/components/Button/Button.elm
Confidenceman02: Its only used internally and it only uses the ConfigValue. The thing that calls it builds a `Config configValue` then the function de-structures it with `Config configValue` but on uses the configValue properties. Its not setting any value in this function. 

I updated all the attrib functions to just receive ConfigValue.

packages/component-library/components/Button/Button.elm
Confidenceman02: 👍

packages/component-library/components/Button/Button.elm
mbylstra: Oh, I see now - it's not exported. I get it now - there aren't actually any configuration options for `onBlur`, `onClick` - the function takse a `msg`. Makes sense! No issue then.

Confidenceman02: 👍

PR #293, opened 2020-02-19
fix: remove elm18 module reference

Confidenceman02:
We are using elm 19 exclusively so this PR removes uses of Elm18 compatible modules for the Button component.


Confidenceman02:
Ah! Beat me to it I just realised!!

PR #295, opened 2020-02-19
feat: apply focus lock on modal

Confidenceman02:
# Background 
Elm modal was previously not handling focus. This meant when the modal opened the focus was left somewhere on the main page and not with the content within the modal.

Also, if focus did happen to be within the modal content, a user could essentially tab focus their way out of the open modal.

# Proposed Solution
In order to properly manage focus we create a focus-lock. Essentially this traps focus within the confines of the first and last focusable elements of the modal content.

If focus for example is on the last focusable element and the user "tabs", we reset focus onto the first focusable element. The same applies for focus on the first focusable element i.e. shift tabbing would set focus to the last focusable element.

# Improvements
Currently focus lock works for Confirmation variants only. This is because they are preset modal views that have a defined first and last focusable element. 

The Generic variant of the modal receives a custom view from the consumer so we cannot know the first and last focusable elements. An API will need to be developed to let consumers mark which elements of their custom views are the first and last focusable elements. This can then utilise the focus-lock implementation.

Focus is currently not being returned to the original trigger.

### DIFF
--- a/packages/component-library/draft/Kaizen/Events/Events.elm
+++ b/packages/component-library/draft/Kaizen/Events/Events.elm
@@ -6,7 +6,9 @@ module Kaizen.Events.Events exposing
     , isEnter
     , isEnterAndValueChange
     , isEscape
+    , isShift
     , isSpace
+    , isTab
     , isUpArrow
     , mapAt
     , onBlurAt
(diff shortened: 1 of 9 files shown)


mbylstra:
Would you be able to add a stateful story where you have to click on a button to open the modal first? The current stories open the modal immediately on page load which makes it hard to QA. It's raising the question to me whether the the first element should be focused only if you open the modal via keyboard, and it's hard to tell this without being able to run it - I'm afraid there's too much code for me to be able to tell just by looking at the code.

PR #298, opened 2020-02-21
feat: Uplift Title Block heading to Greycliff

sebpearce:
## Changes

This PR uplifts the heading of the Title Block from Ideal Sans to Greycliff.

It uses styles Heading 1, 2 and 3 for the desktop, tablet and mobile viewport widths respectively:

## Context

The Title Block is currently only used in Murmur, so it can be safely uplifted without causing issues in other repos. It is being uplifted as part of the work Design Systems Team is doing as part of our rollout of the typography uplift in Murmur.

### DIFF
--- a/packages/component-library/draft/Kaizen/TitleBlock/NavigationButton.scss
+++ b/packages/component-library/draft/Kaizen/TitleBlock/NavigationButton.scss
@@ -1,7 +1,7 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/deprecated-component-library-helpers/styles/type";
+@import "~@kaizen/deprecated-component-library-helpers/styles/color";
 @import "~@kaizen/component-library/styles/border";
-@import "~@kaizen/component-library/styles/color";
 @import "~@kaizen/component-library/styles/layout";
 @import "~@kaizen/component-library/styles/responsive";
 
--- a/packages/component-library/draft/Kaizen/TitleBlock/NavigationButton.scss
+++ b/packages/component-library/draft/Kaizen/TitleBlock/NavigationButton.scss
@@ -1,9 +1,9 @@
 @import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/deprecated-component-library-helpers/styles/type";
 @import "~@kaizen/component-library/styles/border";
 @import "~@kaizen/component-library/styles/color";
 @import "~@kaizen/component-library/styles/layout";
 @import "~@kaizen/component-library/styles/responsive";
-@import "~@kaizen/component-library/styles/type";
 
 .button {
   @include ca-type-ideal-button;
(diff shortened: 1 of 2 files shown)


sebpearce:
[Uplift Title Block with Zen type styles](https://trello.com/c/qtJN6lNL/795-uplift-title-block-with-zen-type-styles)

packages/component-library/draft/Kaizen/TitleBlock/NavigationButton.scss
```diff
@@ -27,6 +27,7 @@
 
   @include ca-media-tablet-and-up {
     color: $kz-color-wisteria-800;
+    padding-top: 17px;
```
sebpearce: This was to nudge the vertically-centred menu buttons into alignment with the new heading, so that they sit on the same baseline (previously they had both been vertically centred, but the new font being bigger caused that to look weird).

mbylstra: I might be nitpicking, but perhaps this should be in rems, in case people update their font size settings - the alignment would go out of whack. Also I'm curious about this value - did you visually align it? Why does only tablet and up need it?

I was always meaning to add a `px-to-rem` helper mixin in Kaizen - basically it's just the px value divided by root font size (which for most or all browsers 16px).

sebpearce: I think this'd be easier to talk about via a Zoom chat where I can show you the visual.

sebpearce: For posterity's sake:

> perhaps this should be in rems, in case people update their font size settings - the alignment would go out of whack.
You might be right, I'll switch it and play around with the font size settings.

> did you visually align it?
Yep. Also ran it past two designers and got the thumbs up.

> Why does only tablet and up need it?
Because only tablet and up shows the buttons at all. The mobile size hides them.

mbylstra: Regarding px/rem, I made this mixin a while back, which is currently in use in one of our private repos:

```
@function px2rem($pixels, $base-font-size: 16) {
  @return ($pixels / $base-font-size) * 1rem;
}
```
The `* 1rem` is due to scss wierdness when units are converted.

mbylstra: I imagine we would generally want to use this (or something like it) whenever we want sizes to scale with text, and only not use it when we explicitly don't (border widths for example)

mbylstra: (The answer to the other two questions are 👍 )

mbylstra: Changing base font sizes rather than simply zooming is a very obscure thing to do, but some people do do it (I remember looking into stats and it was less than 1%). In Firefox:
image

My hazy memory is that the zoom feature (probably needed more horsepower) was introduced into browsers a long time after the "change font size" feature.

PR #301, opened 2020-02-21
feat: Add prevent default on keydown to Button

Confidenceman02:
This adds the ability to prevent default behaviour on keydown when
interacting with the button.

The modifier function takes a list of decoders provided by the consumer.
The successful decoders will prevent default behaviour.

## Example
Say we want to prevent default behaviour when the user presses 'tab' on the button.

```
Button.view (Button.default |> Button.preventKeydownOn [isTabDecoder]) "Button"
```

## Why is this being added now?
The elm Modal component handles focus lock which is important for overall modal accessibility. It internally uses the Button component and needs to be able to prevent the default behaviour of the Tab key in specific situations 

e.g. When the user is focused on the first focusable element and shift/tabs. We prevent the native tab which will place the focus somewhere outside the modal and instead focus the last focusable element thus creating the "lock".

### DIFF
--- a/packages/component-library/components/Button/Button.elm
+++ b/packages/component-library/components/Button/Button.elm
@@ -35,7 +35,7 @@ import Html.Attributes.Aria
 import Html.Events as HtmlEvents
 import Icon.Icon as Icon
 import Icon.SvgAsset exposing (SvgAsset)
-import Json.Decode as Json
+import Json.Decode as Decode
 import Maybe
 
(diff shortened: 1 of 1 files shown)


sebpearce:
Looks good, although I'm curious as to the use case this is intended for where we're preventing default behaviour of keydowns on Buttons (it sounds like it's relevant to accessibility).

PR #302, opened 2020-02-24
feat: Uplift Empty State with Greycliff headings

sebpearce:
image

## Changes
- adds optional Zen headings (using the Greycliff font), controlled by a boolean `useZenStyles` prop that is false by default
- updates all Storybook stories to use this prop by default
- adds an extra Storybook story to demonstrate the legacy version without the prop

## Context
Design Systems Team is uplifting Murmur now, and Perform soon. Murmur uses the Empty State component consistently, but Perform uses a forked version of it _except in one particular case where it uses the Kaizen Empty State_.

Because of this, we want to uplift the Empty State as part of our Murmur work without making the single Empty State instance in Perform look strange. This prop means we can ship the Murmur uplift by using the prop without affecting the instance of it in Perform.

The Perform Empty State can then be uplifted as part of the Perform rollout.

### DIFF
--- a/packages/component-library/draft/Kaizen/EmptyState/EmptyState.tsx
+++ b/packages/component-library/draft/Kaizen/EmptyState/EmptyState.tsx
@@ -34,6 +34,7 @@ export type EmptyStateProps = {
   straightCorners?: boolean
   illustrationType?: IllustrationType
   layoutContext?: LayoutContextType
+  useZenStyles?: boolean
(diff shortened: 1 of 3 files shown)


sebpearce:
[Uplift Empty State component with optional Zen prop](https://trello.com/c/Vmzr7MnL/794-uplift-empty-state-component-with-optional-zen-prop)

PR #312, opened 2020-03-03
feat: Added Heading component for Elm

ActuallyACat:
Related to this one: https://github.com/cultureamp/kaizen-design-system/pull/287

The Elm Heading component matches the React Heading component.... with the exception of allowing consumers to pass in data attributes. I was completely stuck on this and I'm thinking that this is not possible unless the consumer wraps this component in another div.


mbylstra:
For the data attributes (and allowing other Html attributes), here's an idea :

```
type Attribute =
    Data { name : String, value : String }  -- allows any data-xxxx attribute
    | TabIndex Int
    | AriaRole AriaRole

type AriaRole =
    Button | Alert -- etc...

toUntypedAttribute : Attribute -> Html.Attribute
toUntypedAttribute typedAttribute =
    case typedAttribute of
       Data { name, value } ->
           Html.Attribute ("data-" ++ name) value
      TabIndex n ->
          Html.Attribute "tabindex" (toString n) -- I'm rusty on the toString function
       ... ->
          ...etc

```
then you can pass the untyped attribute to any element because they all take the type `Html.Attribute`

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,194 @@
...
+type alias ConfigValue msg =
+    { tag : Element msg
+    , variant : TypeVariant
+    , id : Maybe String
+    }
+
+
+type BaseConfig msg
+    = BaseConfig (ConfigValue msg)
+
```
mbylstra: Is `BaseConfig` dead code?

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,194 @@
...
+
+h5 : Config msg
+h5 =
+    Config { defaultConfig | tag = Html.h5 }
+
+
+h6 : Config msg
+h6 =
+    Config { defaultConfig | tag = Html.h6 }
```
mbylstra: Looks like if you forget to explicitly set a variant modifier `|> variant Heading6` then it will default to `Heading1` which seems like a strange default for a `h6`. Is this what we want? Is this also how the React component works?

packages/component-library/components/Heading/Heading.elm
ActuallyACat: oops, good catch!

packages/component-library/components/Heading/Heading.elm
ActuallyACat: I think this is now at feature parity with the React component. The only required prop is `variant`. If `tag` is not supplied, the component "guesses" the best tag match based on what the variant is in `inferTagFromVariant` (heading-1 will output h1, heading-2 will output h2, etc).

I've added `classNameAndIHaveSpokenToDST`. I notice that the custom classnames don't get hashed, so there might be some Elm wizardry that I don't quite understand yet.

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,270 @@
...
+                Nothing ->
+                    inferTagFromVariant config.variant
+
+        resolveCustomClass =
+            case config.classNameAndIHaveSpokenToDST of
+                Just classNameAndIHaveSpokenToDST_ ->
+                    classNameAndIHaveSpokenToDST_
+
+                Nothing ->
+                    ""
```
ActuallyACat: Surely there's an easier way to default a property that might not exist to an empty string.... right?

mbylstra:
[[Elm] Heading and Box components for typography uplift](https://trello.com/c/d5FJdg0b/785-elm-heading-and-box-components-for-typography-uplift)

mbylstra:
[[Elm] Heading and Box components for typography uplift](https://trello.com/c/d5FJdg0b/785-elm-heading-and-box-components-for-typography-uplift)

packages/component-library/components/Heading/Heading.elm
mbylstra: Oh, the classname hashing issue was for the `classNameAndIHaveSpokenToDST` parameter. The only answer to this I think is for the consumer to provide the already hashed string. There's an Elm function to do this in the Elm CSS Modules library that Kevin built, but people might not know about it.... but _they will speak to DST_, and we can tell them when they ask! (that name keeps on giving!) It's the `toString` function here: https://package.elm-lang.org/packages/cultureamp/elm-css-modules-loader/latest/CssModules#Helpers.

mbylstra: I wrote this a little prematurely - let me think about it some more.

packages/component-library/components/types.ts
```diff
@@ -0,0 +1,64 @@
...
+  | 2.5
+  | 3
+  | 3.5
+  | 4
+
+export type Breakpoint = {
+  mobile: "0"
+  tablet: "768"
+  desktop: "1024"
+}
```
sebpearce: It feels like we're in the same boat here as what we talked about in retro with the last PR. I want to approve this PR, but I don't feel confident yet that we came to a consensus on the breakpoints. This could just be me missing/forgetting the memo though, or not making sure I'm across the research you've done already. @ActuallyACat Do you feel like you have a solid idea of what people want and what the next steps are if we're going to modify these? Can we actually sign off on this as it is and stand behind it with good reasons?

packages/component-library/components/Heading/Heading.elm
sebpearce: I'm curious about this too – always do it the way you did here.

mbylstra: You can use `Maybe.withDefault`:

It would look like this:
```
resolveCustomClass =
    config.classNameAndIHaveSpokenToDST |> Maybe.withDefault ""
```

or 

```
resolveCustomClass =
    Maybe.withDefault "" config.classNameAndIHaveSpokenToDST
```
(I like the pipeline version better even though it's a couple of extra chars)

Either of these two are certainly a lot shorter. But are they better? I dunno - I really don't mind explicit case matches on `Maybe`s (I think it reads a little better),

packages/component-library/components/types.ts
mbylstra: The page I was on:
- no need to provide a way to do custom breakpoints. We can cross this bridge when people need it, and we have several remaining to explore (such as dynamic stylesheets, as P.O.C'd in the last PR)
- But, I was expecting that we would ditch the device specific breakpoints in favor or abstract ones, seeing as there was general consensus in the practice that this was better from the poll. We could change it later but that could be difficult if people are already relying on the device specific ones.

packages/component-library/components/Heading/Heading.elm
mbylstra: I'll update the Story to use CSSModules `toString` function and it should work.

packages/component-library/components/types.ts
sebpearce: Both good points.

As for the naming, why have names at all? If we're worried about it being difficult to change later, why not just refer to them by their pixel values instead of choosing a name? I mean, if TypeScript is going to show you what options you have to choose from, it sort of eliminates the problem of having to remember what they are, right?

mbylstra: This is my opinion:

- Heading component: **essential**
- Spacing component: **essential** (we will definitely need spacing between text, and we don't want people hacking it with custom div wrappers and custom classes)
- responsive spacing: **not essential**  (it's cool, but I never saw this as a requirement, and I never saw a design for it - should spacing necessarily get tighter on smaller screens? Perhaps we could just remove it in this PR, and do a future follow up PR?)

I guess what I'm saying

sebpearce: I think the issue is that if people use this spacing component as their spacing solution, they have no way to make it responsive because they can't use CSS anymore.

In other words, we still need to build that functionality before people can start using the spacing component (to avoid nasty surprises where they can't make it responsive), but if we just want to ship this to start working with it in the component library, I'd have no problem with removing it from this PR and postponing it until a future one.

mbylstra: Are there many examples of responsive spacing at present? If it's currently used often, then we should support it. If not, is this an imagined requirement?

ActuallyACat: > It feels like we're in the same boat here as what we talked about in retro with the last PR. I want to approve this PR, but I don't feel confident yet that we came to a consensus on the breakpoints. This could just be me missing/forgetting the memo though, or not making sure I'm across the research you've done already. @ActuallyACat Do you feel like you have a solid idea of what people want and what the next steps are if we're going to modify these? Can we actually sign off on this as it is and stand behind it with good reasons?

Going to rip this out of the PR and open up a Box PR separately... This PR was opened before I decided to remove spacing out of the other PR.

packages/component-library/components/Heading/Heading.elm
mbylstra: I added a crazy blinking text example in the Story: 
image

packages/component-library/components/Heading/Heading.elm
ActuallyACat: TIL, thanks @mbylstra. I've updated to use `Maybe.withDefault` - much cleaner

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,270 @@
...
+    , classNameAndIHaveSpokenToDST : Maybe String
+    }
+
+
+defaultConfig : ConfigValue msg
+defaultConfig =
+    { tag = Nothing
+    , variant = Display0
+    , id = Nothing
+    , data = []
```
mbylstra: I think in my head, rather than a `data` properties, there would be an `attributes` property, that could contain many different types of attributes: data, aria attributes (eventually). It's find if it's just data attributes for now, but some of the other names should reflect this. (wait, I'll link a comment)

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,270 @@
...
+id : String -> Config msg -> Config msg
+id id_ (Config config) =
+    Config { config | id = Just id_ }
+
+
+classNameAndIHaveSpokenToDST : String -> Config msg -> Config msg
+classNameAndIHaveSpokenToDST value (Config config) =
+    Config { config | classNameAndIHaveSpokenToDST = Just value }
+
+
```
mbylstra: I think this should be named `addDataAttribute`, if you can only add `data-` attributes.
OR, the `DataAttribute` type should be renamed `Attribute` and allow for other attributes other than just `data-` attributes (even if for now data attribute is the only possible one).

Eg:

```
type Attribute =
    DataAttribute { key : String, value : String }
```
This is a custom type with only one option, but you could add other options later.

I don't mind which way we go, but we should pick one.

mbylstra: (note: this was in reference to the `addAttribute` function but the Github UI is linking it to the `classNameAndIHaveSpokenToDST` for some reason.

ActuallyACat: Renaming to `addDataAttribute`, or `dataAttribute`, seems like the best approach. Seems like the `add.*` convention should only be used when records are being merged / you can pipe multiple options into a record... what do you think?:
```
(Heading.h1
    |> Heading.variant Heading6
    |> Heading.addAttribute { name = "automation-id", value = "inbox-heading" }
    |> Heading.addAttribute { name = "another-attr", value = "yo" }
)
```

I think we should consider aria and data attributes separately. I don't feel overly strongly about this though, but here's my rationale: 

In Typescript, data attributes aren't valid JS identifiers so you can get away without having types, which is what I've done in the React component (https://www.typescriptlang.org/docs/handbook/jsx.html#attribute-type-checking), whereas aria attributes are typed (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1516). 

E.g.
```
export const Heading = ({
  classNameAndIHaveSpokenToDST,
  children,
  tag,
  variant,
  ...otherProps // this picks up data-* 
}: HeadingProps & AriaAttributes) // adds types for aria-* attributes
```
I'm feeling that the most consistent approach is to consider them separate for both Elm and React. In Elm we might also want to be more explicit about the types of aria attributes instead of accepting any key, something like this: https://package.elm-lang.org/packages/fapian/elm-html-aria/latest/

Thoughts?

mbylstra: Yeah, I think it's fine to have separate params: dataAttributes, ariaAttributes, etc. Definitely having the ariaAttributes as a separate param makes it possible to use https://package.elm-lang.org/packages/fapian/elm-html-aria/latest/ (otherwise it'd have to be forked to make it work with a custom data type, as it's output type is Html.Attribute which is opaque).

On naming, my choice would be `addDataAttribute`. The `add` bit is to indicate that you can add more than one (I think you should be able to), unlike `id` or `variant` which can take only one value. `dataAttribute` sounds like you can only have one. Alternatively it could be `addDataAttributes` and take a list, but I think just having `addDataAttribute` is all that's necessary.

There's a couple of other places I noticed where the name is too generic. I'll add a comment on them.

packages/component-library/components/Heading/Heading.elm
```diff
@@ -0,0 +1,265 @@
...
+    { tag = Nothing
+    , variant = Display0
+    , id = Nothing
+    , data = []
+    , classNameAndIHaveSpokenToDST = Nothing
+    }
+
+
+toUntypedAttribute : DataAttribute -> Html.Attribute msg
+toUntypedAttribute typedAttribute =
```
mbylstra: this should be `toUntypedDataAttribute`

packages/component-library/components/Heading/Heading.elm
ActuallyACat: I'll change `data` to `dataAttribute` for clarity

PR #315, opened 2020-03-05
feat: Uplift Modal typography

sebpearce:
This uplift is part of the typography uplift work Design Systems Team is doing across the platform. It is intended to be shipped along with or shortly after Murmur is uplifted to use Greycliff.

## Changes

- Change headings to Greycliff
- Make `color` and `type` imports come from deprecated helpers package instead of component library `styles/` dir
- Use Wisteria 800 as a default text colour at top level (instead of inheriting it from the consumer – now Storybook stories will no longer have pure black text in Modals)
- Use antialiasing for all text

image
image
image


didoesdigital:
@sebpearce on design review, there's nothing show stopping (these all look out of scope for this PR):

- Firefox headings look heavier
- The modal colors could be darker to match the Zen UI Kit e.g. Coral 500 instead of Coral 400
- The Elm Cautionary icon is the "i" where it should be "!"
- We could remove the brushes now

sebpearce:
Thanks – I'll put these into a card. We could do all of them in a separate "Uplift Modals" PR 👍

PR #317, opened 2020-03-05
chore: Switch to dart-sass & break out $ca-grid into its own file

sebpearce:
# IMPORTANT ⚠️ 

Once this PR is merged, consuming repos that do their own compilation of Kaizen's Sass code **must** replace `node-sass` with `sass` (`dart-sass` under the hood).

**This PR therefore must not be shipped until all consuming repos have PRs ready to go and the change has been cleared with affected teams.**

## Changes

- Replace the `node-sass` package with `sass` (= `dart-sass`) from Kaizen's root `package.json`, as well as in Storybook's and Kaizen Site's `package.json`
- Break out `$ca-grid` into its own file called `styles/grid.scss`
- Add a `@use` statement to every file in Kaizen that references `$ca-grid`

## Background

The `@use` statement works differently to `@import`.

With `@import`, if I define `$ca-grid` in `type.scss`, and then for some reason `color.scss` imports `type.scss`, I can then access `$ca-grid` by importing `color.scss` (a nested import).

This is not only confusing – you can't see a definition of `$ca-grid` when you look at `color.scss` – it's also dangerous: if you ever remove `$ca-grid` from `type.scss`, you might only look for files that import `type.scss` to make sure they have the correct new import, and thus miss files that were depending on `color.scss` for access to `$ca-grid`. (This happened recently in a Murmur uplift that DST did, and it caused a silently-breaking visual bug, because the imported thing was a Sass function, and those don't break when they're missing.)

### A note on using `@use`

Unlike `@import`, `@use` automatically namespaces imports. If you want to use `$ca-grid` and you do this:

`@use "~@kaizen/component-library/styles/grid"`

...you have to access it by doing `grid.$ca-grid` (the `grid` at the start is the filename).

If you want `$ca-grid` to be available in the global namespace for the current file, you must do:

`@use "~@kaizen/component-library/styles/grid" as *;`

This allows you to use `$ca-grid` as normal.

⚠️ `@use` must also come at the start of the file, before any `@import` statements.


sebpearce:
[Break out $ca-grid into its own file](https://trello.com/c/0iUVe1Wm/813-break-out-ca-grid-into-its-own-file)

mbylstra:
The readme could point to https://github.com/cultureamp/kaizen-design-system/pull/317/files#diff-bf020fc461a3c23a3d6d6edfa0968d1eR33 and  https://github.com/cultureamp/kaizen-design-system/pull/317/files#diff-ccd726111bcc27a0c48b313489453816R31 as an example of how to modify webpack to use dart-sass.

jasononeil:
Is the `use "…" as *` syntax (removing the namespace) a thing you’re doing because you prefer the unnamespaced variable names, or just because it’s easier to implement the change this way?

I quite like that you need the namespace normally (`grid.$ca-grid`) as there's zero questions of where it's coming from.

sebpearce:
@jasononeil I prefer it too actually. I think I just assumed that everyone would complain about it being too cumbersome to type. Maybe it's fine though 🤔

mbylstra:
I imagine it would be either very difficult (to automate) or very time consuming to change every usage to the fully qualified syntax. Looks like there's consensus that the fully qualified syntax is better (I agree) - we could make that a further future cooldown task though (time permitting)

mbylstra:
At the very least, we could make a recommendation in the comms to use the fully qualified syntax, even if we don't update the "legacy" usages.

ActuallyACat:
My preference would be to see all the imports change to use `use` (where applicable). Is there a particular reason you're not going all in on it in this PR?

sebpearce:
> My preference would be to see all the imports change to use use (where applicable). Is there a particular reason you're not going all in on it in this PR?

Just that this PR was originally focused on `grid`. We could do it for sure though. I also want to convert everything to `@use`. It might be cleaner to do the rest in a follow-up PR though, since there could be unforeseen issues when we convert the rest of the imports 🤔

PR #318, opened 2020-03-09
feat: Paragraph component for React

ActuallyACat:
Paragraph only has one required prop, which is variant. The options are identical to the Zen UI Kit in Figma: `intro-lede`, `body`, `small`, `extra-small`. 

Usage is similar to the Heading component: 
```
<Paragraph variant="body">
   Some text, or nest other <Link>Components</Link>.
</Paragraph>
```

`data-*`, `tag` and `classNameAndIHaveSpokenToDST` are all optional props.

### DIFF
--- a/packages/component-library/components/Paragraph/Paragraph.tsx
+++ b/packages/component-library/components/Paragraph/Paragraph.tsx
@@ -1,0 +1,46 @@
+import classNames from "classnames"
+import { createElement } from "react"
+
+const styles = require("./Paragraph.module.scss")
+
+export type ParagraphVariants = "intro-lede" | "body" | "small" | "extra-small"
+
+export type AllowedTags =
+  | "pre"
+  | "p"
+  | "a"
+  | "div"
+  | "span"
+  | "h1"
+  | "h2"
+  | "h3"
+  | "h4"
+  | "h5"
+  | "h6"
+
+export interface ParagraphProps {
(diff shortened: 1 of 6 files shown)


mbylstra:
This might be a silly question, but why is there no `Paragraph.module.scss` in this PR. Is that already in master? I was going to make a note about needing to remove the dashes so it works in Gatsby, but maybe this isn't the right PR for that.

packages/component-library/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,46 @@
+import classNames from "classnames"
+import { createElement } from "react"
+
+const styles = require("./Paragraph.module.scss")
```
mbylstra: Where _is_ `Paragraph.module.scss`. It's not in master and I'm not seeing it in this PR (but then how is the build passing 🤔 )

ActuallyACat: The Elm component included the styles. Since that PR is now merged, the module should be in master and hence the file diff is removed in this PR. 

Elm PR: https://github.com/cultureamp/kaizen-design-system/pull/320/files
Paragraph.module.css in master: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Paragraph/Paragraph.module.scss

PR #320, opened 2020-03-11
feat: Paragraph component for Elm

ActuallyACat:
Usage is very similar to the Heading component for Elm. Variant is the only required property, and there's also support for tags, classNames (AndIHaveSpokenToDST), ids, and data attributes.

```
Paragraph.view
                (Paragraph.p |> Paragraph.variant IntroLede)
                [ Html.text "IntroLede Paragraph" ]
```

### DIFF
--- a/packages/component-library/components/Paragraph/Paragraph.elm
+++ b/packages/component-library/components/Paragraph/Paragraph.elm
@@ -1,0 +1,225 @@
+module Paragraph.Paragraph exposing
+    ( Config
+    , DataAttribute
+    , TypeVariant(..)
+    , a
+    , addDataAttribute
+    , classNameAndIHaveSpokenToDST
+    , div
+    , h1
+    , h2
+    , h3
+    , h4
+    , h5
+    , h6
+    , id
+    , p
+    , pre
+    , span
+    , tag
+    , variant
+    , view
+    )
+
+import CssModules exposing (css)
+import Debug exposing (log)
+import Html exposing (text)
+import Html.Attributes
+import List
+
+
+type Config msg
+    = Config (ConfigValue msg)
+
+
+type alias Element msg =
+    List (Html.Attribute msg) -> List (Html.Html msg) -> Html.Html msg
+
+
+type alias DataAttribute =
+    { name : String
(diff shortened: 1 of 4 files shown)


packages/component-library/stories/Paragraph.stories.elm
```diff
@@ -0,0 +1,27 @@
...
+                (Paragraph.p |> Paragraph.variant Body)
+                [ Html.text "Body Paragraph" ]
+        , statelessStoryOf "Small" <|
+            Paragraph.view
+                (Paragraph.p |> Paragraph.variant Small)
+                [ Html.text "Small Paragraph" ]
+        , statelessStoryOf "ExtraSmall" <|
+            Paragraph.view
+                (Paragraph.p |> Paragraph.variant ExtraSmall)
+                [ Html.text "ExtraSmall Paragraph" ]
```
sebpearce: Probably good to make these look exactly the same punctuation-wise as the Figma typography page, so:
```
Paragraph Intro Lede
Paragraph Body
Paragraph Small
Paragraph Extra Small
```

sebpearce:
This is looking 👨‍🍳💋👌

packages/component-library/components/Paragraph/Paragraph.elm
```diff
@@ -0,0 +1,225 @@
...
+defaultConfig : ConfigValue msg
+defaultConfig =
+    { tag = Nothing
+    , variant = Body
+    , id = Nothing
+    , dataAttribute = []
+    , classNameAndIHaveSpokenToDST = Nothing
+    }
+
+
```
mbylstra: We'd probably want to share this code eventually in some kind of helper file, but let's wait for the third occurrance in rule of three style.

mbylstra:
Nice! I would approve this, but I agree with Seb's suggestion on updating the story names.

packages/component-library/components/Paragraph/Paragraph.elm
ActuallyACat: Heavy plus. Let's consider this for the next cycle

PR #322, opened 2020-03-12
fix: remove top and bottom padding on the icon to make the block shorter

juliohr:
## Background

When applying the guidance block on the streamlined onboarding survey project, we noticed that the block was too tall. Exploring a few possibilities with designers, we found that the best compromise is to remove the top and bottom padding on the icon's wrapper. This makes the block 24px shorter. 

## Before

Survey_Index

## After 

Survey_Index

## Responsiveness

The text in the middle will take up to 650px of width. If the button on the right starts to take up space the text will shrink and we will prevent the button from wrapping. There is a design guideline that prevents call to action text from having more than three words. Below is an extreme case just for illustration.  (The current design was approved)

Slack___Marc_Katsambis__Mindy_Chang___Culture_Amp


ActuallyACat:
Just flagging this - mobile responsiveness should be explored when this component moves from draft to a core component.
Screen Shot 2020-03-16 at 2 11 02 pm

packages/component-library/stories/GuidanceBlock.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react"
 
 import { GuidanceBlock } from "@kaizen/component-library/draft"
-const bannerImg = require("@kaizen/component-library/stories/illustrations/@2x_single_point_onboard_survey.png")
+const bannerImg = require("@kaizen/component-library/stories/illustrations/team.png")
```
ActuallyACat: 👍 Nice to see the naming is consistent with Figma

packages/component-library/stories/GuidanceBlock.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react"
 
 import { GuidanceBlock } from "@kaizen/component-library/draft"
-const bannerImg = require("@kaizen/component-library/stories/illustrations/@2x_single_point_onboard_survey.png")
+const bannerImg = require("@kaizen/component-library/stories/illustrations/training.png")
```
ActuallyACat: This should be team.png - it's causing an error in your build:
> Module not found: Error: Can't resolve '@kaizen/component-library/stories/illustrations/training.png' in '/workspace/packages/component-library/stories'

PR #327, opened 2020-03-16
docs: uplift kaizen site headings

mbylstra:
Uplift headings to Zen styles.

I left some implementation notes in comments below.

Preview: https://dev.cultureamp.design/dst/uplift-kaizen-site-headings/

image


site/src/components/PageHeader.tsx
```diff
@@ -64,7 +65,15 @@ const PageHeader: React.SFC<PageHeaderProps> = ({
             </div>
           )}
           <div className={styles.mainSection}>
-            <h1 className={styles.headingText}>{headingText}</h1>
+            <div className={styles.headingTextContainer}>
+              <Heading
```
mbylstra: Zen Heading 1 was too small for the front page (`headingOnly`), but Display 0 was too big for non front page pages.

site/src/components/PageHeader.tsx
```diff
@@ -64,7 +65,15 @@ const PageHeader: React.SFC<PageHeaderProps> = ({
...
+                variant={headingOnly ? "display-0" : "heading-1"}
```
mbylstra: Zen Heading 1 was too small for the front page (headingOnly), but Display 0 was too big for non front page pages.

site/src/components/PageHeader.tsx
```diff
@@ -64,7 +65,15 @@ const PageHeader: React.SFC<PageHeaderProps> = ({
...
+                tag="h1"
+                classNameAndIHaveSpokenToDST={styles.headingTextOverrides}
```
mbylstra: This is because the color needed to be overridden to white. We are looking into extending the API during cooldown.

site/src/components/PageHeader.tsx
```diff
@@ -64,7 +65,15 @@ const PageHeader: React.SFC<PageHeaderProps> = ({
             </div>
           )}
           <div className={styles.mainSection}>
-            <h1 className={styles.headingText}>{headingText}</h1>
+            <div className={styles.headingTextContainer}>
```
mbylstra: `headingTextContainer` should be able to replaced with the upcoming `Spacing` component in future

site/src/pages/index.tsx
```diff
@@ -86,17 +87,21 @@ export default ({ location }) => {
               </a>
             </div>
             <div className={styles.guidelinesTextContainer}>
-              <div className={styles.heading}>
-                <a href={withPrefix("/guidelines/overview")}>Guidelines</a>
+              <div className={styles.headingContainer}>
```
mbylstra: `headingContainer should be able to replaced with the upcoming `Spacing` component in future

site/src/styles/markdown.scss
```diff
@@ -79,8 +80,7 @@ $link-icon-margin: 12px;
   }
 
   :global(.md-h2) {
-    font-size: 2rem;
-    line-height: 1.125;
+    @include kz-typography-heading-2;
```
mbylstra: I didn't use the Text component here because we need to be able to support Markdown which doesn't let you add css classes, so you need to do complicated CSS selectors, and you can't reach into CSS Modules obfuscated class names (which are used in our Text component). Eg:
https://github.com/cultureamp/kaizen-design-system/blob/master/site/src/styles/markdown.scss#L88

mbylstra: It's a bit of an issue that we can't use our Text components in Kaizen Site, but I.M.O Kaizen Site is not what the Design System was designed for, and Markdown comes with a lot of complications. It's relatively small and maintainable if we have to do some custom stuff to over come limitations of the Design System.

mbylstra:
[Uplift type on Kaizen Site (stretch goal)](https://trello.com/c/i2v3ySKS/752-uplift-type-on-kaizen-site-stretch-goal)

PR #329, opened 2020-03-17
feat: Add max-width to Paragraph styles

sebpearce:
## Changes

Add a `max-width` rule to each paragraph style, according to its design.

These max-width values were taken from the Typography page of the Zen UI Kit in Figma:

image

### DIFF
--- a/packages/component-library/components/Paragraph/Paragraph.module.scss
+++ b/packages/component-library/components/Paragraph/Paragraph.module.scss
@@ -14,6 +14,7 @@
   font-size: $kz-typography-paragraph-intro-lede-font-size;
   line-height: $kz-typography-paragraph-intro-lede-line-height;
   letter-spacing: $kz-typography-paragraph-intro-lede-letter-spacing;
+  max-width: 60.9375rem; // 975px
 }
 
 .body {
@@ -22,6 +23,7 @@
   font-size: $kz-typography-paragraph-body-font-size;
   line-height: $kz-typography-paragraph-body-line-height;
   letter-spacing: $kz-typography-paragraph-body-letter-spacing;
+  max-width: 48.75rem; // 780px
 }
 
 .small {
@@ -30,6 +32,7 @@
   font-size: $kz-typography-paragraph-small-font-size;
   line-height: $kz-typography-paragraph-small-line-height;
   letter-spacing: $kz-typography-paragraph-small-letter-spacing;
+  max-width: 42.5rem; // 680px
 }
 
 .extra-small {
@@ -38,4 +41,5 @@
   font-size: $kz-typography-paragraph-extra-small-font-size;
   line-height: $kz-typography-paragraph-extra-small-line-height;
   letter-spacing: $kz-typography-paragraph-extra-small-letter-spacing;
+  max-width: 37.5rem; // 600px
 }


ActuallyACat:
I'm still on the fence about whether this is needed given our layout will already constrain the Paragraph component.

sebpearce:
Could you elaborate on how that will happen? I'm curious as to how it will work given that each paragraph style has a different value for max width.

sebpearce:
[Add max-width to Paragraph component](https://trello.com/c/gqtUq5Fn/835-add-max-width-to-paragraph-component)

mbylstra:
Is there a reason to not also make max width a token?

ActuallyACat:
> Could you elaborate on how that will happen? I'm curious as to how it will work given that each paragraph style has a different value for max width.
> 
> Or will it be manual, where the engineer needs to know about the max-width and set it themselves?

This strays too far in the direction of the responsibility of a layout (or container) components for me. Does a Paragraph component _really care_ what it's max width is? Or rather, should it fill the entire width of whatever container it's in, and let the max width logic sit in a separate container? 

Further, I would hazard a guess that the max widths set in Figma are actually just reflecting the  intended max-width of the container the text should be placed in, i.e. _all_ the content on the page should sit in a  975px container.

mbylstra:
> Further, I would hazard a guess that the max widths set in Figma are actually just reflecting the intended max-width of the container the text should be placed in, i.e. all the content on the page should sit in a 975px container.

The max widths are representing usability requirements regarding readability - this is why it's different for different font sizes. So even if the container is 975px, the paragraphs should be max 780px as specified in the UI Kit (of course converted to rem units in code). Text becomes hard to read if it is too wide, and this depends on the font size, not the container.

We could get people to add wrappers to set the max width but then it becomes hard to enforce (and I do think max widths should be enforceable - they are not optional - I think there are zero good reasons why you'd want to break that)

ActuallyACat:
Adding a style rule that teams potentially have to override would be more of a burden. I can think of a couple of scenarios within the Design System itself where we would be breaking this rule: 
* Inline Notifications use Paragraph Small, which should have a max-width of 680px. In this component, the notification spans the whole width of whatever container it's placed in. When we remove the sidebar, this will be over 1080px. 
* Similarly, Global Notifications use Paragraph Body, which should have a max-width of 780px. This also spans the whole width of whatever container it's placed in. 
* Expanded popout in Table uses Paragraph Body, which has a max-width of 780px, which can also span the whole width of the container. 

This is just within the Design System. Are we _really sure_ that this is something we want to enforce in the component? @katsambarcus @didoesdigital

mbylstra:
I could be wrong about this, but I was under the impression that text should never be longer than the specified max widths. I stand corrected if this is the intention, but it sounds to me like Paragraph Small at 1080px in an inline notification would be really strange and hard to read (same for the two other examples).

sebpearce:
> Adding a style rule that teams potentially have to override would be more of a burden. 

The way I see it, this is exactly what we want from a design standpoint. The default is that paragraphs are no longer than their max width as shown in Figma, and in the cases *where we break with convention*, there is a workaround that's more work. The burden of overriding makes you aware that what you're doing breaks convention.

The alternative is to shift the burden of work to the _default_ case, i.e. a normal paragraph, where the developer needs to be aware of the ideal max-width of each container of each text style. This is more work for them. But it would be easier for them (less work, even) to break convention and use the wrong max width value.

ActuallyACat:
> The way I see it, this is exactly what we want from a design system standpoint. 
I think you misunderstand my argument, I am both challenging the premise of the actual rule (which is why I cc'd Marc and Di), and if we think the rule is still a good idea then I would like to see it implemented elsewhere. 

From what I see in Figma, and looking at the components in the DS, this rule should not exist at all. We're _already breaking convention in many places in the Design System_, in _every single scenario_ where a component can stretch the whole width of its parent. I could be wrong, of course, but this is not what I've seen in the DS. Show me otherwise.

Why would we place a restriction on consumers when it's something that we not adhering to?

I'd like more context for the reason for max-width. Is it just when this text is used within a large body of text? What happens when multiple text styles are used in a large container. Do we justify the paragraphs differently in those scenarios?

sebpearce:
Gotcha. Maybe it'd be faster to do this in a Zoom with Marc & Di then?

PR #331, opened 2020-03-18
docs: don't render the storybook iframe if it's the overview page

mbylstra:


site/src/templates/componentPage.tsx
```diff
@@ -44,7 +44,7 @@ export default ({ data, pageContext, location }) => {
   )
 
   const renderStorybookIFrame = () => {
-    if (!md.frontmatter.demoStoryId && md.frontmatter.title !== "Overview") {
```
mbylstra: There was a logic bug: the Overview does not have a demoStoryId, so it was not matching this line. I moved the logic for the overview outside the component, and I don't think this warning should show for the overview.

mbylstra:
[Don't show demo block in overview page of Kaizen Site](https://trello.com/c/me4DVhWt/837-dont-show-demo-block-in-overview-page-of-kaizen-site)

mbylstra:
This is just a comment to see if it shows up in the Slack integration I'm playing with.

site/src/templates/componentPage.tsx
```diff
@@ -83,7 +83,7 @@ export default ({ data, pageContext, location }) => {
         </Sidebar>
         <Content>
           <ContentNeedToKnowSection listOfTips={md.frontmatter.needToKnow} />
-          {renderStorybookIFrame()}
+          {md.frontmatter.title === "Overview" ? null : renderStorybookIFrame()}
```
sebpearce:
```suggestion
          {md.frontmatter.title === "Overview" && renderStorybookIFrame()}
```
I find this easier to read, but no biggie as both do the job.

site/src/templates/componentPage.tsx
sebpearce: This was my brainfart 🎉

site/src/templates/componentPage.tsx
mbylstra: The commit suggestion feature is pretty cool - I've never used that before. I like Ruby's `unless` and Elm's `if ... then ... else`, but yeah I think it's more idiomatic JS.

mbylstra: I just find it weird that boolean logic is used for control flow - seems to be just taking advantage of the fact that right side of the boolean operator happens to not get executed if the left one fails.

sebpearce: I agree, it is weird lol. It seems to be pretty common in React – I guess because people see the `else` bit as unnecessary.

Issue #225, opened 2020-01-16
There's no guidance on Avatars


didoesdigital:
We have avatars in use in Engagement and Performance. We need to consolidate the design direction and UI Kit assets and we need to draft guidelines for them so that we can build a shared Kaizen component.

Original:
![image](https://user-images.githubusercontent.com/2476974/72490563-9e7efd00-386b-11ea-924e-eeafbbcac64d.png)

Zen:
![image](https://user-images.githubusercontent.com/2476974/72490575-ab035580-386b-11ea-8bee-479a709caf36.png)

# Guidelines needed

## Summary paragraph

What _is_ an avatar?

## Need to know

What is the high-priority info people need to make a decision about avatars? e.g. 2 bullet points

## To keep in mind

Why do we need this? What user problem does it solve? 1 or 2 sentences

How to use this? What considerations are there for usage? (design or code)

## When to use and when not to use

When to use?

When not to use? What to use instead?

## External links

Examples of other existing design systems, blog posts, components.

PR #337, opened 2020-03-18
docs: add Avatar component docs

didoesdigital:
Closes: https://github.com/cultureamp/kaizen-design-system/issues/225


site/docs/components/avatar.mdx
```diff
@@ -0,0 +1,70 @@
...
+<Dont>
+
+- Avoid an avatar when most people or groups won't have an image to show.
+
+</Dont>
+</DosAndDonts>
+
+## See also
+
+- [Link vs button](/guidelines/link-vs-button/).
```
ActuallyACat: /meganit Are these two guidelines related?

didoesdigital: If an avatar is linked to a user profile page, there would be temptation to implement it as a button because it looks more buttony styled than link-ish.

PR #338, opened 2020-03-19
fix: Remove text margin in the button

mollyywang:
If a button has a icon inside, the text only has the right/left margin
image
image

If a button has no icon inside, the text has the both right left margin which makes the button bigger
image
image

### DIFF
--- a/packages/component-library/components/Button/styles.scss
+++ b/packages/component-library/components/Button/styles.scss
@@ -345,8 +345,6 @@ $caButton-verticalPaddingForm: calc(
   line-height: 1;
 
   &:only-child {
-    margin: 0 $ca-grid / 2;
-
     %caButtonSecondary & {
       margin: 0;
     }


ActuallyACat:
This looks good to me! Have you checked what these changes might look like in Murmur?

ActuallyACat:
Just noticed that your PR is failing because the commit body doesn't following conventional commits: https://github.com/cultureamp/kaizen-design-system/blob/master/CONTRIBUTING.md#conventional-commit

tl;dr, if you add another commit (this can be empty) with the appropriate commit prefix, the build will be happy: 
`git commit --allow-empty -m "fix: Remove the text margin in the button if there is no icon in the button."`

mbylstra:
I think it would be difficult for Molly to check Murmur as it's not something she works in, it's only possible to check at the moment with a dev setup, but I wouldn't expect Molly to have Murmur set up.

Even if we had a way to deploy to a test/dev farm, it would be really hard or impractical to QA every button usage. I think this is a pretty safe change - it now more closely matches the UI Kit, so I think it can only be an improvement. There's a very small chance that some alignment issue might occur as buttons will now be narrower, but that's just the risk we'll have to take until we have a comprehensive visual regression suite.

I'm happy to take care of the Murmur bump once this has been released. There is one thing we can do - the system tests in Murmur generate _some_ screenshots, so we can look at them (but this can only be done after the PR is merge), and also do a some spot check QAs. I don't think it's worth a QA swarm. We can always roll back the change if it causes issues in Murmur (extremely unlikely)

PR #339, opened 2020-03-19
fix: Add padding in dropdown

mollyywang:
There is a padding around all buttons:
image
But there is no padding around the dropdown:
image
So when we have button+dropdown, it looks like the edit icon is in the middle:
image
What's the new expected behaviour?
Add the same padding around the dropdown.
image
image


mbylstra:
Although this has been approved (I also approve of the code changge), let's hold off on releasing it until we have a good plan for QA. I will discuss it with Molly tomorrow morning how we can approach QA.

PR #343, opened 2020-03-20
feat: Add color to Heading

sebpearce:
## Changes

- added the `color` prop to both the React and Elm components
- make sure both components are correctly applying `.small` and `.large` classes, which affect the color output if the `color` prop is set to `positive` or `negative` (if the `variant` is a Heading 2 or larger, it should get `.large` and it will be a different colour to Heading 3 or smaller, which gets `.small`).
- updated both React and Elm stories
- updated the React spec
- removed the `a` tag as an available option for both the React and Elm components (see below)

### Note on removing `a` as an available tag

This was following up on a discussion that DST had where we agreed that anchor tags do not belong here. Links tends to wrap things completely, or wrap a piece of text inside a larger bit of text. That means that you would be more likely to **pass in a link as a child of the Heading component** or **wrap the entire Heading component inside a link** (which means you could also include other things in the link, like an icon next to the Heading) rather than make the Heading component itself manifest as an anchor tag. We may also create a `<Link>` component in the future, and if so it'd be good to restrict anchor tags to that component rather than have some in there and some in Heading or Paragraph that end up drifting from `<Link>`.

### DIFF
--- a/packages/component-library/components/Heading/Heading.module.scss
+++ b/packages/component-library/components/Heading/Heading.module.scss
@@ -63,3 +63,42 @@
   line-height: $kz-typography-heading-6-line-height;
   letter-spacing: $kz-typography-heading-6-letter-spacing;
 }
+
+.dark {
+  color: $kz-color-wisteria-800;
+  opacity: 1;
+}
+
+.dark-translucent {
+  color: $kz-color-wisteria-800;
+  opacity: 0.7;
+}
+
+.light {
+  color: $kz-color-white;
+  opacity: 1;
+}
+
+.light-translucent {
+  color: $kz-color-white;
+  opacity: 0.8;
+}
+
+.positive {
+  &.small {
+    color: $kz-color-seedling-600;
+  }
+  &.large {
+    color: $kz-color-seedling-500;
+  }
+}
+
+.negative {
+  &.small {
+    color: $kz-color-coral-600;
+  }
+
+  &.large {
+    color: $kz-color-coral-500;
+  }
+}
--- a/packages/component-library/components/Heading/Heading.module.scss
+++ b/packages/component-library/components/Heading/Heading.module.scss
@@ -69,17 +69,17 @@
   opacity: 1;
 }
 
-.dark-translucent {
+.dark-reduced-opacity {
   color: $kz-color-wisteria-800;
   opacity: 0.7;
 }
 
-.light {
+.white {
   color: $kz-color-white;
   opacity: 1;
 }
 
-.light-translucent {
+.white-reduced-opacity {
   color: $kz-color-white;
   opacity: 0.8;
 }
--- a/packages/component-library/components/Heading/Heading.module.scss
+++ b/packages/component-library/components/Heading/Heading.module.scss
@@ -4,7 +4,6 @@
 @import "~@kaizen/deprecated-component-library-helpers/styles/type";
 
 .heading {
-  color: $kz-color-wisteria-800;
   margin: 0;
 }
 
--- a/packages/component-library/components/Heading/Heading.spec.tsx
+++ b/packages/component-library/components/Heading/Heading.spec.tsx
@@ -10,6 +10,28 @@ describe("<Heading />", () => {
     const headingClasslist = headingMock.getByText("Example").classList
     expect(headingClasslist).toContain("heading")
     expect(headingClasslist).toContain("display-0")
+    expect(headingClasslist).toContain("large")
+  })
+
+  it("adds a .small class instead of .large if a Heading 3 is used", () => {
+    const headingMock = render(
+      <Heading variant="heading-3" tag="div">
+        Example
+      </Heading>
+    )
+    expect(headingMock.getByText("Example").classList).toContain("small")
+    expect(headingMock.getByText("Example").classList).not.toContain("large")
+  })
+
+  it("adds a .dark-reduced-opacity class if the color prop is set to that", () => {
+    const headingMock = render(
+      <Heading variant="heading-3" color="dark-reduced-opacity">
+        Example
+      </Heading>
+    )
+    expect(headingMock.getByText("Example").classList).toContain(
+      "dark-reduced-opacity"
+    )
   })
 
   it("changes rendered HTML element when passed tag", () => {
--- a/packages/component-library/components/Heading/Heading.tsx
+++ b/packages/component-library/components/Heading/Heading.tsx
@@ -15,7 +15,6 @@ export type HeadingVariants =
 export type AllowedTags =
   | "pre"
   | "p"
-  | "a"
   | "div"
   | "span"
   | "h1"
--- a/packages/component-library/components/Heading/Heading.tsx
+++ b/packages/component-library/components/Heading/Heading.tsx
@@ -50,7 +50,7 @@ export interface HeadingProps {
    * Allowed heading variants
    */
   variant: HeadingVariants
-  textColor?: AllowedColors
+  color?: AllowedColors
 }
 
 export const Heading = ({
@@ -58,7 +58,7 @@ export const Heading = ({
   children,
   tag,
   variant,
-  textColor = "dark",
+  color = "dark",
   ...otherProps
 }: HeadingProps) => {
   const inferredTag =
@@ -68,7 +68,7 @@ export const Heading = ({
     styles.heading,
     styles[variant],
     classNameAndIHaveSpokenToDST,
-    styles[textColor],
+    styles[color],
     VARIANTS_24PX_OR_GREATER.includes(variant) ? styles.large : styles.small,
   ]
 
--- a/packages/component-library/components/Heading/Heading.tsx
+++ b/packages/component-library/components/Heading/Heading.tsx
@@ -3,6 +3,8 @@ import { createElement } from "react"
 
 const styles = require("./Heading.module.scss")
 
+const VARIANTS_24PX_OR_GREATER = ["display-0", "heading-1", "heading-2"]
+
 export type HeadingVariants =
   | "display-0"
   | "heading-1"
@@ -24,6 +26,14 @@ export type AllowedTags =
   | "h5"
   | "h6"
 
+export type AllowedColors =
+  | "dark"
+  | "dark-translucent"
+  | "light"
+  | "light-translucent"
+  | "positive"
+  | "negative"
+
 export interface HeadingProps {
   /**
    * Not recommended. A short-circuit for overriding styles in a pinch
@@ -40,6 +50,7 @@ export interface HeadingProps {
    * Allowed heading variants
    */
   variant: HeadingVariants
+  color?: AllowedColors
 }
 
 export const Heading = ({
@@ -47,6 +58,7 @@ export const Heading = ({
   children,
   tag,
   variant,
+  color = "dark",
   ...otherProps
 }: HeadingProps) => {
   const inferredTag =
@@ -56,6 +68,8 @@ export const Heading = ({
     styles.heading,
     styles[variant],
     classNameAndIHaveSpokenToDST,
+    styles[color],
+    VARIANTS_24PX_OR_GREATER.includes(variant) ? styles.large : styles.small,
   ]
 
   return createElement(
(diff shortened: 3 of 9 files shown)


packages/component-library/components/Heading/Heading.tsx
```diff
@@ -25,18 +26,28 @@ export type AllowedTags =
...
   variant: HeadingVariants
+  textColor?: AllowedColors
 }
 
 export const Heading = ({
   classNameAndIHaveSpokenToDST,
   children,
   tag,
   variant,
+  textColor = "dark",
```
ActuallyACat: Have you thought about what this might mean for other components that accept colour? Will we end up with `textColor`, `backgroundColor`, `iconColor`, etc?

packages/component-library/components/Heading/Heading.module.scss
```diff
@@ -63,3 +63,42 @@
   line-height: $kz-typography-heading-6-line-height;
   letter-spacing: $kz-typography-heading-6-letter-spacing;
 }
+
```
ActuallyACat: Since the `color` property has been decoupled from the heading classes, my preference would be to remove `color` from `.heading` (line 7). If a user does not pass in a colour it defaults to `dark` so this is essentially dead code.

ActuallyACat:
Could you separate the `a` stuff to a different PR, or make it clearer that this is being removed in your current PR?

packages/component-library/components/Heading/Heading.tsx
sebpearce: Yes and yes – that's the intention. Do you see any problems arising from that? 
This came up in a meeting I had with Marc and Di where the prop used to be called `color`, but they were pretty keen on it being `textColor` to differentiate it from future stuff.

packages/component-library/components/Heading/Heading.module.scss
sebpearce: 👍

sebpearce:
> Could you separate the a stuff to a different PR, or make it clearer that this is being removed in your current PR?
Sure – this is still a draft PR and wasn't actually ready for review 😊  I'll have a proper PR summary ready to go when it's open.

packages/component-library/components/Heading/Heading.tsx
ActuallyACat: > Do you see any problems arising from that?
Yes. Following this will result in name collisions with CSS properties (`background-color` is `backgroundColor` when represented as an object in React). Will this make it confusing for developers? Will they (incorrectly) think that `backgroundColor` should take hex values? 

My 2c is we should always avoid names that could overload the meaning of our props, where possible. Foreground/background is my preference for this reason.

ActuallyACat: It's also slightly more succinct - 2 color props to remember instead of many (.*Color could result in SO MANY combinations).

mbylstra: I like it that if you have an icon with text, you could just specify `foregroundColor` rather than having to specify `textColor` and `iconColor`. It's less flexible if you did actually want a different text color to icon colour, but that seems like an edge case we shouldn't have to worry about just yet?

mbylstra: `backgroundColor` is a tricky one - I can't imagine what name you could come up with that wasn't the same as the `background-color`. But the TS types should be enough to make it clear that it isn't a hex value? (Would it even compile if you gave it a hex value?). I don't think it makes sense for headings to have background colors, so we don't have to worry about background color just yet.

mbylstra:
Oops, I thought this was ready for review too. Make sure to name it "Add color to _React_ Heading and Paragraph", if this is only React.

packages/component-library/components/Heading/Heading.tsx
ActuallyACat: I'm thinking to drop the `color` from the name with the `foreground/background` suggestion. In your edge case, that could be solved if we lean more heavily on composition rather than overloading components with props. As an example:
```
<Tab icon="help" iconColor="dark" backgroundColor="light" textColor="wisteria-800">
   Example
</Tab>
```

Better utilisation of composition:
```
<Tab fg="wisteria-800" bg="light">
  <Icon variant="help" fg="dark"/>
  Example
</Tab>
```

sebpearce:
It'd be good for us to sync on this issue of when PRs are ready. I've been using Draft PRs as: "I want to see this on GitHub so I can keep track of it visually, but I don't want anyone else to look at it yet unless I ask them for feedback".

packages/component-library/components/Heading/Heading.tsx
sebpearce: @ActuallyACat That last example you posted resolves the only qualm I had with the fg/bg idea (what if the icon is a different colour to the text?). I'll go with fg/bg, although I wonder if it could be confusing to a developer if they find a component with a type `fg: string`, and there's no `bg`, which will be the case here. I can imagine some people going "wtf is `fg`?" For that reason I'm leaning towards `foregroundColor`, even if it's clumsier.

sebpearce: Maybe `fgColor` (and `bgColor`)?

mbylstra:
I also use github Drafts purely as a backup mechanism, so my draft's might an extremely rough state that I happened to be in at the end of the day (definitely not meant for anyone to see!), seeing as we don't have any backup system in place for our laptops (I've lost work before, so I'm very wary of this!)

Related is that when you get a request for review using the Github UI interface, it's not very clear whether this means "request for early feedback" or "request for final review". Both are valuable. I think the way we track comments in Slack is working pretty well, so rather than relying on the Github UI features, we could just comment more: eg "This is not finished, but I'd like to get some feedback on the direction" vs "This is ready for final review" (and then "ready for follow up review" if there were changes requested).

packages/component-library/components/Heading/Heading.tsx
sebpearce: > Yes. Following this will result in name collisions with CSS properties (background-color is backgroundColor when represented as an object in React). Will this make it confusing for developers? Will they (incorrectly) think that backgroundColor should take hex values?

I just had a thought – wouldn't TypeScript and Elm completely eliminate this confusion, since it shows you what you can pass in?

mbylstra: If I were to vote, I'd just use the unabbreviated words as much as possible (this is specifically in regards to abbreviated vs unabbreviated, not `textColor` vs `foregroundColor`). I don't think there's a big enough advantage to abbreviating things. Saves a tiny about of typing time and yeah a little easier to ready once you're experienced, but then forces you to have remember whether something was abbreviated or not and it's extra friction for. Clearly we won't be abbreviating everything, "CCIIHSTODST" (see if you can work out that one), so it's not super clear what the rules are for when to abbreviate and when not to. `fgColor` seems like a half measure to me!

Also, what's obvious to you might not be obvious to someone else. Let me recount my recent experience when I honestly thought `md` stood for "middle" and it was to do with responsitivity, not "markdown" (there was a CSS class something like `mdHeading`. I wasted at least an hour because of this thinking that we had a responsive system in place for Kaizen Site.

sebpearce: Totally agree with you @mbylstra, but in this particular case `backgroundColor` has this problem:

> Following this will result in name collisions with CSS properties (background-color is backgroundColor when represented as an object in React).

Do you know how we could get around that here?

mbylstra: It's not a _technical_ limitation though right? One is a css property, and one is a prop, so they wouldn't technically clash. It's slightly confusing, but I think the type system should make it pretty clear. It is a bit of a tricky one, but I'd still lean towards no abbreviations.

sebpearce: Also, I'm assuming that if we reject `textColor` and `backgroundColor` on the basis that one of them conflicts with a CSS property, then we'll make this a rule going forward and consistently reject every prop name that conflicts with a CSS property, yeah?

sebpearce: @mbylstra I agree, TS and Elm will clear up the confusion.

mbylstra: But I'm happy to step back and trust you to make a decision on this one (this kind of thing can easily go around and circles!)

sebpearce: This is where it gets complicated. Is it my decision to make because I started the PR? If @ActuallyACat disagrees, should I trust her to make the decision, since she mentioned she has strong opinions about how our component APIs should be structured? I feel pressure not to drag the conversation on, since this is a thing we've been doing a lot recently, but I also feel like it'd be irresponsible of me to not mention to you two when I spot risks.

mbylstra: Hehe, yeah it is tricky. What I really meant to say is that my comments are my 2c, but I don't feel super strongly about it, and it's harder with 3 people, so I'm stepping out so that at least there's only two people that have to agree, which should be easier than 3!

sebpearce: Alright. I'm going to ask that @ActuallyACat has the final say on this, and I'll lay out the arguments against her idea for her to consider before she makes the call:

• `fg` and `bg` could be unintuitive and confusing for first-time users, particularly if there is no `bg` and they just see `fg` on its own and don't know what it is. Worst-case scenario could be that someone uses the Heading or Paragraph component, wants to change the colour but doesn't realise that `fg` does this, but they don't ask us, and instead create a container div that reaches in and changes the colour from outside.
• `fg` and `bg` are not part of any system that our designers use to talk about colour, so this arguably goes against our efforts to make the design system feel as one (subcategory: "designers and engineers speak the same language").
• even though `backgroundColor` looks like a CSS property, TypeScript and Elm would prevent anyone from passing in hex values or the like, and would likely clear up confusion about what they can pass in. `textColor` isn't a CSS property so it doesn't have the issue at all. Both of these terms match the way that designers talk about colour and are intuitive to read at first glance.

As long as I've clearly articulated all those, I'm happy to go with whatever @ActuallyACat chooses since I'm happy she has all the info that I could have provided. 👍

ActuallyACat: These are all 💯:
`fgColor` and `bgColor`
`foregroundColor` and `backgroundColor`
`fg` and `bg`

Speaking generally, I think overloading language should be something we're mindful of and call out when we see it. It's all about tradeoffs, after all. You're both right that TS types should make this easy to reason about in this case, so this trade-off seems like a reasonable one. 

`textColor` particularly irks me because it's a distinction we shouldn't need to make (referencing my composition example).

sebpearce: To summarise the trade-offs:

- **designers and developers speak the same language** (`textColor`, `color` ✅ `foreground`, `fg`, `fgColor` ❌)
- **props don’t look like CSS properties and thus potentially overload the terms** (`textColor`, `foreground`, `fg`, `fgColor` ✅, `color`, `backgroundColor` ❌)
- **having unnecessary words in prop names is unnecessary, and would mean we’d have to make further unnecessary distictions later** (`color` ✅ , `textColor` ❌)

After chatting to @ActuallyACat about this, I'm going to go with `backgroundColor` and `color` since they seem to both work for 2 out of 3 trade-offs and the one they don't satisfy isn't that bad we think.

mbylstra: One thing I just want a little clarity on: will the heading and paragraph components take `backgroundColor` props (regardless of whether it's named `bg`, `bgColor`, etc), and if so, why?

ActuallyACat: Nice summary, thanks @sebpearce 

@mbylstra I don't think there's a compelling use case for that, no. But these colour props I would consider to be "common props" (like `tag`) that will most likely be used across other components. I could see these being used on icons, cards, and other more atomic components.

packages/component-library/components/Heading/Heading.spec.tsx
```diff
@@ -10,6 +10,28 @@ describe("<Heading />", () => {
     const headingClasslist = headingMock.getByText("Example").classList
     expect(headingClasslist).toContain("heading")
     expect(headingClasslist).toContain("display-0")
+    expect(headingClasslist).toContain("large")
+  })
+
+  it("adds a .small class instead of .large if a Heading 3 is used", () => {
```
ActuallyACat: Noice 👍

PR #344, opened 2020-03-20
Ally/box component

ActuallyACat:
It's a box. It doesn't really do much. It's job is to be a componentised representation of the spacing units we use in Kaizen.


ActuallyACat:
Old PR, doing some admin before reopening

PR #345, opened 2020-03-20
feat: Adding Box component for React

ActuallyACat:
It's a box - It doesn't really _do_ much. Its job is to be a componentised representation of the spacing units that are recommended for use in Kaizen.


packages/component-library/components/Spacing/padding.ts
```diff
@@ -0,0 +1,85 @@
...
+  } else {
+    Object.keys(unit).forEach(key => {
+      classes.push(styles[`p-${convertFractionToString(unit[key])}--${key}`])
+    })
+  }
+
+  return classes
+}
+
+export const responsivePaddingClasses = ({
```
mbylstra: A bit of a nitpick, but perhaps this should be renamed to just `paddingClasses` until we tackle responsitivity again, to avoid confusion.

packages/component-library/components/Spacing/margin.ts
```diff
@@ -0,0 +1,86 @@
...
+  } else {
+    Object.keys(unit).forEach(key => {
+      classes.push(styles[`m-${convertFractionToString(unit[key])}--${key}`])
+    })
+  }
+
+  return classes
+}
+
+export const responsiveMarginClasses = ({
```
mbylstra: same with this (remove the `responsive`)

PR #348, opened 2020-03-23
docs: Using the Docs Addon to generate better stories

ActuallyACat:
Before 
Screen Shot 2020-03-23 at 1 51 30 pm

After
Screen Shot 2020-03-23 at 1 53 32 pm

Note that not all of our stories have the prop table functionality turned on.


packages/component-library/components/Paragraph/Paragraph.tsx
```diff
@@ -19,9 +19,19 @@ export type AllowedTags =
   | "h6"
 
 export interface ParagraphProps {
+  /**
+   * Not recommended. A short-circuit for overriding styles in a pinch
+   * @default ""
+   */
```
ActuallyACat: These are what's displayed in the prop table on the docs tab. The table will still display if these props aren't commented, but it will be missing the description in the props table.

package.json
```diff
@@ -103,6 +105,7 @@
     "webpack": "^4.40.2"
   },
   "devDependencies": {
+    "@storybook/addon-docs": "^5.3.17",
```
ActuallyACat: If you're interested: https://github.com/storybookjs/storybook/tree/871a4ed7bacf546f17379451a9449d6a16d3b3ff/addons/docs/react

site/tsconfig.dist.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "extends": "../tsconfig.json",
+    "noEmit": false
```
ActuallyACat: The rationale here is that the site just needs to be type-checked (so don't emit output, i.e don't compile to javascript). The Gatsby config should take care of the rest...

tsconfig.json
```diff
@@ -10,6 +10,6 @@
     "noImplicitAny": false,
     "strict": true,
     "allowJs": true,
-    "noEmit": true
+    "noEmit": false
```
ActuallyACat: Ideally, I would have left this as `true` and then used the custom config for running storybook in ts-node (in packages/component-library/tsconfig.storybook.json) that has this set to false. For some reason this isn't working.

This is annoying because we don't actually care if it compiles the site files or not.

packages/component-library/stories/Modal.stories.tsx
```diff
@@ -47,7 +47,7 @@ export default {
 }
 
 export const ConfirmationPositiveKaizenSiteDemo = () => (
-  <ModalStateContainer isInitiallyOpen={true}>
+  <ModalStateContainer isInitiallyOpen={false}>
```
mbylstra: These were intentionally open to make it easier to view the modals without having to open them. Personally I'm fine with it as I think it more accurately represents the standard modal bevahiour, but it does make it annoying if you want to quickly check the design of the opened modal. Worth checking with @didoesdigital . We had a discussion about this recently. If you look at the Elm ones, there are some stories that have "user initiated" in the name, and some that are "open by default".

ActuallyACat: I can fix this up with a bit of refactoring. There's an experimental feature that will help us. Slot functions https://github.com/storybookjs/storybook/blob/871a4ed7bacf546f17379451a9449d6a16d3b3ff/addons/docs/docs/docspage.md#slot-functions

mbylstra: Cool!

ActuallyACat: Just spent the last little while looking into this further. It's possible, but we would kinda need to maintain two separate stories, one for the story panel that defaults to being open, and another for the docs panel that defaults to closed. The code looked like it would be confusing for someone who would be adding new stories. 

WDYT if we leave this off, and look into alternatives if we get a feature request?

mbylstra: Yeah, sounds like it's not worth the complexity and maintenance burden. Sounds good to me - I think the trade-off of having this new feature is totally worth it!

PR #350, opened 2020-03-25
feat: Elm table component

melm73:
We currently already have a pretty cool React table component. This PR is to add an Elm version.


melm73:
[Add elm table to kaizen](https://trello.com/c/HPfTSgbJ/3016-add-elm-table-to-kaizen)

PR #352, opened 2020-03-25
feat: add disabled confirmButton feature to elm modal

cset00:
BREAKING CHANGE: everywhere the modal is used will need to now have the 'onConfirmDisabled' in the config

Context:

- I'm looking to create something like an ApplyChanges confirmation modal in Elm - where the user has to input a number that matches for example the number of ignored users, before going ahead with their employee data import and applying the changes. 
- If the user hasn't entered any numbers or entered a number that doesn't match, the confirm button should be disabled.

Example screenshot of current react modal I'm trying to have in Elm.
image

### DIFF
--- a/packages/component-library/draft/Kaizen/Modal/Modal.elm
+++ b/packages/component-library/draft/Kaizen/Modal/Modal.elm
@@ -139,6 +139,7 @@ type alias ConfirmationConfig msg =
     , onConfirm : Maybe msg
     , confirmLabel : String
     , dismissLabel : String
+    , onConfirmDisabled : Bool
     }
 
 
@@ -297,6 +298,13 @@ viewContent (Config config) =
                             Nothing ->
                                 confirmationConfig
 
+                    withOnConfirmDisabled confirmationConfig =
+                        if configs.onConfirmDisabled then
+                            ConfirmationModal.onConfirmDisabled True confirmationConfig
+
+                        else
+                            ConfirmationModal.onConfirmDisabled False confirmationConfig
+
                     withBodySubtext confirmationConfig =
                         case configs.bodySubtext of
                             Just subtext ->
@@ -340,6 +348,7 @@ viewContent (Config config) =
                     commonConfirmationConfig confirmationConfig =
                         withOnDismiss confirmationConfig
                             |> withOnConfirm
+                            |> withOnConfirmDisabled
                             |> withBodySubtext
                             |> withFocusableIds
                             |> withFocusLockAttribs
--- a/packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.elm
+++ b/packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.elm
@@ -10,6 +10,7 @@ module Kaizen.Modal.Presets.ConfirmationModal exposing
     , negative
     , onConfirm
     , onConfirmBlur
+    , onConfirmDisabled
     , onConfirmFocus
     , onDismiss
     , onHeaderDismissBlur
@@ -59,6 +60,7 @@ type alias Configuration msg =
     , onConfirmFocus : Maybe msg
     , onConfirmBlur : Maybe msg
     , confirmId : Maybe String
+    , onConfirmDisabled : Bool
     }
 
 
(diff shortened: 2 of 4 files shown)


packages/component-library/draft/Kaizen/Modal/Modal.elm
```diff
@@ -139,6 +139,7 @@ type alias ConfirmationConfig msg =
     , onConfirm : Maybe msg
     , confirmLabel : String
     , dismissLabel : String
+    , onConfirmDisabled : Maybe Bool
```
Confidenceman02: This can be changed to just `Bool` no need for the Maybe.

packages/component-library/draft/Kaizen/Modal/Modal.elm
```diff
@@ -297,6 +298,19 @@ viewContent (Config config) =
                             Nothing ->
                                 confirmationConfig
 
+                    withOnConfirmDisabled confirmationConfig =
+                        case configs.onConfirm of
+                            Just confirmMsg ->
+                                case configs.onConfirmDisabled of
```
Confidenceman02: Using just a Bool for this type removes the need for this case

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.elm
```diff
@@ -59,6 +60,7 @@ type alias Configuration msg =
     , onConfirmFocus : Maybe msg
     , onConfirmBlur : Maybe msg
     , confirmId : Maybe String
+    , onConfirmDisabled : Maybe Bool
```
Confidenceman02: Just `Bool` works fine.

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.elm
```diff
@@ -242,6 +245,19 @@ footer config =
                 Nothing ->
                     buttonConfig
 
+        withOnConfirmDisabled buttonConfig =
+            case config.onConfirm of
```
Confidenceman02: type `Bool` removes this case

packages/component-library/draft/Kaizen/Modal/Modal.elm
cset00: Yup! I can remove the Maybe. But just confirming.. it will still have the `case configs.onConfirm` as it was already there... and for `configs.onConfirmDisabled` instead of having 2 cases of `Just True and _`, it'll be `True` and `False` ?

packages/component-library/draft/Kaizen/Modal/Modal.elm
```diff
@@ -297,6 +298,19 @@ viewContent (Config config) =
                             Nothing ->
                                 confirmationConfig
 
+                    withOnConfirmDisabled confirmationConfig =
+                        case configs.onConfirm of
```
Confidenceman02: Hmm... Do we need to couple the onConfirm with the disabled state? In my mind the button can be disabled without there being an onConfirm msg. Also if I change the order of withOnConfirm below to be after `onConfirmDisabled` then this will always result in `False`.

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.elm
```diff
@@ -242,6 +245,19 @@ footer config =
...
+                Just confirmMsg ->
+                    case config.onConfirmDisabled of
```
Confidenceman02: Could we use an `if` instead of a case?

cset00: yeahh is there a best practice for when to use ifs vs case? I always assumed in Elm it seems to be preferred to use case, and I've just defaulted to that without looking into it more.

Confidenceman02: case works fine but its mostly about staying consistent with the code style of Modal.

packages/component-library/draft/Kaizen/Modal/Modal.elm
cset00: Thankssss! This looks way better now 🙏 :D

PR #354, opened 2020-03-25
fix: update minimist transitive dependency to fix vulnerability

mbylstra:
`minimist` was a transitive dependency of `lerna`. Upgrading `lerna` to the latest version did not fix the vulernatbility (it appears that `lerna` has not upgraded it yet), so I used the `resolutions` feature of yarn/npm to upgrade the transitive dependency.

I left in the lerna upgrade, because why not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,5 +112,7 @@
     "@storybook/node-logger": "^5.3.9",
     "jscodeshift": "^0.7.0"
   },
-  "resolutions": { "**/**/minimist": "^1.2.3" }
+  "resolutions": {
+    "**/**/minimist": "^1.2.3"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "jest": "^24.9.0",
     "jest-circus": "^24.9.0",
     "jest-static-stubs": "^0.0.1",
-    "lerna": "^3.16.4",
+    "lerna": "3.20.2",
     "node-sass": "^4.12.0",
     "normalize.css": "^8.0.1",
     "postcss-flexbugs-fixes": "^4.1.0",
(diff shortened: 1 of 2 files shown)


mbylstra:
[security vulnerability minimist](https://trello.com/c/i7tA2sNY/847-security-vulnerability-minimist)

package.json
```diff
@@ -81,7 +81,7 @@
     "jest": "^24.9.0",
     "jest-circus": "^24.9.0",
     "jest-static-stubs": "^0.0.1",
-    "lerna": "^3.16.4",
+    "lerna": "3.20.2",
```
ActuallyACat: Missing a hat here `"lerna": "^3.20.2",`

mbylstra: I can manually add the hat, but I'm curious what command I should have used.

I tried `yarn upgrade lerna --latest`, which didn't make any changes (I don't understand why)

So, instead I found the latest version of lerna from npm, and used `yarn add lerna@3.16.4`. How do you get the hat in there without having to manually modify package.json?

I think it's good to match the currently established convention, but I'm wondering why the hat is considered best practice? Seems more repeatable to have a specific version specified, seeing as we manually do upgrades.

ActuallyACat: When you upgrade to a specific version, you can indicate it as you would see in the package: `yarn upgrade left-pad@^1.0.0`. TIL you can also pass in caret: https://classic.yarnpkg.com/en/docs/cli/upgrade/#toc-yarn-upgrade-package-latest-l-caret-tilde-exact-pattern

There's a couple of different takes on why caret is best practice. In theory, it allows you to accept future bug fixes and features that shouldn't break things. Of course, sometimes shit happens and something intended to be a bug fix could be a breaking change. Or sometimes what you've written with only work with a specific version of a package. This why some people like to pin their dependencies. 

IMO, lock files do a pretty good job at mitigating against that.

mbylstra: Thanks for this!

mbylstra:
The commit history is a bit off a mess (I didn't realise you can add the hat in `yarn add`), but I think has has go there in the end!

package.json
mbylstra: I added the hat - ready for re-review.

PR #355, opened 2020-03-26
fix: title block breadcrumb z index

mbylstra:
There were some incompatible z-indexes between `NavigationBar` and `TitleBlock` causing the TitleBlock's back button eclipse the navigation bar when scrolling down.

Fixed by decreasing the z-index of the the back-button/breadcrumb.

This PR also adds a story that integrates the NavigationBar and TitleBlock so that they can be tested together. Storybook: https://dev.cultureamp.design/dst/fix-title-block-breadcrumb-z-index/storybook/. The story is `Layout (React)` > `With TitleBlock`. This story reveals another unrelated issue - there is no way to make the TitleBlock span the entire width of the screen like it does in product. The product is able to do this by not actually using the Layout component, but instead selectively using scss from it. I'm not sure if the Layout component is actually used anywhere (other than the styles) - it'd be good to get stats on this kind of thing!

### Before:

image

### After:

image


mbylstra:
[Fix title block breadcrumb eclipsing the navigation bar](https://trello.com/c/o5KmHh4Z/850-fix-title-block-breadcrumb-eclipsing-the-navigation-bar)

mbylstra:
I believe there are forks of the TitleBlock that may or may not have the same problem. My plan is to get this released, bump it, and if there are still issues in places pass it on to the relevant teams (and show them the solution)

packages/component-library/draft/Kaizen/TitleBlock/TitleBlock.scss
```diff
@@ -46,7 +46,7 @@
   @include ca-type-ideal-small;
   display: flex;
   position: absolute;
-  z-index: $ca-z-index-popover;
+  z-index: $ca-z-index-sticky;
```
mbylstra: The reason I chose `$ca-z-index-sticky` is that it's `1020`, to make it lower than `$ca-z-index-fixed`, which is `1030` and is what the navigation bar uses.

ActuallyACat: I might be missing context here, but why does the title block need an elevated z-index?

PR #356, opened 2020-03-27
fix: informative tag icon

mbylstra:
## Background

The informative mood for the Tag component uses the "!" instead of the  "i" icon.

## Why it matters

It makes the informative mood seem more alarming than it should be.

Before:
image

After:
image


mbylstra:
[Informative Tag should use "i" icon, not "!" (Elm + React)](https://trello.com/c/HPabdgWz/851-informative-tag-should-use-i-icon-not-elm-react)

PR #357, opened 2020-03-27
fix: add missing font-family fallbacks to zen heading mixins

mbylstra:
The greycliff heading mixins in `@kaizen/deprecated-component-library-helpers` did not have any `font-family` fallbacks, and they were overriding any font-family fallbacks set in `body` tags.

**Note:** the storybook book stories are just temporary for QA. They are necessary to be able to QA this, but we don't want them on the public site. The plan is that once this is approved, I'll remove the commit with the stories before releasing (I cherry picked the commit from a previous PR)

image
This shows the font falling back to Helvetica (I just changed the Greycliff name in dev tools). Previously it would fall back to the default system font (a serif font)

### DIFF
--- a/packages/deprecated-component-library-helpers/styles/type.scss
+++ b/packages/deprecated-component-library-helpers/styles/type.scss
@@ -287,7 +287,9 @@ $ca-ideal-locale-he-font-family: "Ideal Sans A", "Ideal Sans B", Tahoma,
   @include ca-type(
     $size-ratio: $size / 16,
     $family:
-      "#{$kz-typography-heading-1-font-family}, Helvetica, Arial, sans-serif",
+      unquote(
+        "#{$kz-typography-heading-1-font-family}, Helvetica, Arial, sans-serif"
+      ),
     $base-size: $ca-greycliff-font-base-size,
     $descender-height: $ca-greycliff-font-descender-height,
     $line-height-in-grid-units: $line-height-in-grid-units,
--- a/packages/deprecated-component-library-helpers/styles/type.scss
+++ b/packages/deprecated-component-library-helpers/styles/type.scss
@@ -286,7 +286,8 @@ $ca-ideal-locale-he-font-family: "Ideal Sans A", "Ideal Sans B", Tahoma,
 ) {
   @include ca-type(
     $size-ratio: $size / 16,
-    $family: $kz-typography-heading-1-font-family,
+    $family:
+      "#{$kz-typography-heading-1-font-family}, Helvetica, Arial, sans-serif",
     $base-size: $ca-greycliff-font-base-size,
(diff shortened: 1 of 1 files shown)


packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -286,7 +286,8 @@ $ca-ideal-locale-he-font-family: "Ideal Sans A", "Ideal Sans B", Tahoma,
 ) {
   @include ca-type(
     $size-ratio: $size / 16,
-    $family: $kz-typography-heading-1-font-family,
+    $family:
+      "#{$kz-typography-heading-1-font-family}, Helvetica, Arial, sans-serif",
```
mbylstra: I don't know if `Helvetica, Arial, sans-serif` is the latest best practice for fallbacks (I copied it from above in this file, which was probably written several years ago). THere might be more recent system fonts (Avenir?) which are closer to Greycliff,  but it's much better than nothing and should be able to test our theory that the lack of fallbacks is causing PDF rendering issues.

mbylstra:
[Add font-family fallbacks to scss zen heading mixins](https://trello.com/c/afNRd0wm/853-add-font-family-fallbacks-to-scss-zen-heading-mixins)

packages/deprecated-component-library-helpers/styles/type.scss
mbylstra: The rest of the PR will be be removed once approved (might need a re-tick)

mbylstra: _Update_: gah, this is totally wrong (good thing I QA'ed it!)

It ends up as `font-family: "Greycliff, Helvetica, Arial, sans-serif"`, but we want `font-family: Greycliff, Helvetica, Arial, sans-serif`

I'm not sure how to do that without SCSS thinking `Helvetica, Arial, sans-serif` are extra function arguments 🤔

sebpearce: Could you alternate between double and single quotes? Like have one delimiting the string and the others surrounding the font families?

mbylstra: Ah, I missed this. I pushed a change thought that fixes it. I think that might have the same problem though, where scss gets confused between whether something is a string or a list of function arguments.

ActuallyACat: You can still use interpolation here, it's just being sensitive about your quotes. 

```
$font: 'Greycliff CF';

.thing { 
  font-family: "#{$font}", Helvetica, Arial, sans-serif;
}
```

mbylstra:
Cool, I dropped the stories commit. I'm just about to duck into a meeting. I'm happy for you to merge this as soon as it builds.

PR #361, opened 2020-03-31
feat: Add border tokens

sebpearce:
The structure here seems like what we want for now, since we are only currently supporting one border width (2px) and this may continue indefinitely.

In future, we may want to consider moving to a sizing system like [this](https://github.com/cultureamp/kaizen-design-system/issues/373).


ActuallyACat:
[Border and shadow tokens](https://trello.com/c/w8ftYg9w/869-border-and-shadow-tokens)

PR #362, opened 2020-03-31
feat: Add shadow tokens

sebpearce:


ActuallyACat:
[Border and shadow tokens](https://trello.com/c/w8ftYg9w/869-border-and-shadow-tokens)

packages/design-tokens/tokens/shadow.json
```diff
@@ -0,0 +1,12 @@
+{
+  "kz": {
+    "shadow": {
+      "small": {
```
ActuallyACat, line 4: Calling this out because it's now the second time I've seen this (the other time in the Data component). Should our convention for referring to sizes is to use the long name (small, medium, large) or the short name (xxs/xs/sm/md/lg/xl/xxl)? Thoughts?

mbylstra: hmm, I think we definitely need consistency throughout the Design System, so we should stick to one or the other (same goes with things like `mt`). I think it's a pretty big decision which way we go (abbreviated vs non-abbreviated) - dare I say DACI worthy?

My 2c:

tokens - I really don't think it's necessary to abbreviate token names, as consumers rarely use them directly (there's not a lot of code that uses tokens) so all it does is harm readability with little benefit to "speed".

component APIs (React and Elm) - I can see it being useful here for component that you use multiple times per day, but I still lean towards unabbreviated across the board. I know we are doing this in the Spacing and Box components, but seeing as they aren't really publicised (or used) yet (that I know of), it wouldn't be completely too late to change. I realise the spacing/Box PR did pass code review, but I still feel it could have been discussed further, as this was brought up multiple times in the comments.

TLDR:
- If we stick with the convention set by the Spacing and Box components, we should definitely use the abbreviations
- But are we definitely locked into this convention?

ActuallyACat: >  dare I say DACI worthy?
👀 Definitely interested in how we can communicate and get feedback from other engineers from the practice about naming. I wonder if one big DACI with an example component (like a bitton) will be more effective than a series of smaller DACI.

> tokens - I really don't think it's necessary to abbreviate token names
Sounds good. Since this is Seb's work I'm guessing he would tend to agree with you. There's a level of nuance we can exercise here - it would be acceptable (IMO) to use one convention for tokens and another for components. 

Let's keep long names in tokens and discuss short names when Seb is back.

mbylstra: Yes, I was thinking of a DACI for naming conventions in general, rather than for this specific component (yeah, would be good to have one or two examples in there). Let's revisit when Seb's back.

PR #368, opened 2020-04-02
feat: Create TextArea & TextAreaField components

dougmacknz:
The long-awaited `textarea` component for Kaizen.

Storybook:
https://dev.cultureamp.design/dug/textarea/storybook-static/?path=/story/textareafield-react--default-kaizen-site-demo

image

I've kept this mainly consistent with how `Input` and `TextField` have been built, but left out a lot of the extra functionality for now (disabled, reversed, onFocus, etc etc).

Needs documentation (separate PR preferably) and to be added to the UI Kit in Figma.


dougmacknz:
Firstly, thank you all of that @mbylstra!

1. I have tested in Edge and IE11. The only issue seems to be with the uncontrolled versions in Storybook with Edge, which is super laggy to type in. I tried doing uncontrolled in Murmur with Edge and it's all working fine so it's just something specific with Storybook. Otherwise everything seems to work fine in Edge & IE11.
2. I have tested with Grammarly on Chrome for both the uncontrolled and controlled versions and things seem to be fine. I am able to use Grammarly features with no adverse effects.
3. As discussed I think you mistakenly thought this component was always in controlled mode.
4. I have created a `defaultValue` prop for prefilling in uncontrolled mode. It turns out react supports this prop out of the box on `textarea`, so I didn't have to do much here.
I've also created a controlled example story as you may have seen.
5. I haven't added that timeout test sorry, I should probably get a move on with this so maybe we'll have to come back to this if or when it becomes an issue.
6. Tested in Safari and I did see the auto correct functionality. It didn't seem to have any bad effects on the component.

I've also added `onBlur` and `onFocus` props :)

PR #369, opened 2020-04-03
feat: Publish the draft well component

ActuallyACat:
## About 
The intention of this PR is to check that the new draft components will:
1. Publish as expected 
2. Be consumed as expected

I've made all the changes we'll need for Storybook to find the location of the draft stories, but I should note that the current Well story is coming from `@kaizen/component-library`, not `@kaizen/draft-well`. I did this because it's too premature to remove Well from `component-library` before verifying there's no hiccups with this approach. 

### Things to note
There is now duplication of the Well component. There hasn't been much work on Well, so hopefully, these two don't get out of sync.


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
   "version": "independent",
   "npmClient": "yarn",
-  "packages": ["packages/*"]
+  "packages": ["packages/*", "packages/draft/*"]
```
ActuallyACat, line 4: New structure
```
- packages
  - component-library
  - draft
    - well
    - other draft components
```
IMO I think it's clearer keeping `draft` within packages instead of creating another root level folder. Thoughts?

sebpearce: Agree, I think it's a good idea to imagine what it'll look like when it scales. Without your proposed structure it'll be `component-library` with like 50 siblings.

storybook/tsconfig.storybook.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../tsconfig.json",
```
ActuallyACat, line 2: I've moved all storybook related config to `/storybook`. The alternative was to duplicate this config in each new draft package and that felt icky to me.

ActuallyACat:
[Separate one component out into a separate package to test import issues / build issues etc.](https://trello.com/c/GZH8nOV0/859-separate-one-component-out-into-a-separate-package-to-test-import-issues-build-issues-etc)

lerna.json
zioroboco: Another option would be a sibling of packages, e.g.:

```yaml
# packages which define individual components
- components:
  - draft-...
  - draft-...
  - maybe-non-draft-also-at-some-point

# tokens, libraries of components, utilities, etc.
- packages:
  - component-library
  - tokens
```

ActuallyACat: 🤔 That might cause some confusion for newcomers looking for what components they should be using now. How about:
```
# Legacy or non-standard packages, and draft components
- packages:
  - deprecated-component-library-helpers
  - draft-well
  - maybe-non-draft-also-at-some-point

# Source of truth and current tokens, libraries of components, utilities, etc.
- core:
  - component-library
  - tokens
```

zioroboco:
Thinking out loud...

It's a shame that we needed so much boilerplate in the component package's `package.json`! if we weren't compiling to typescript, a lot of that could go away — e.g. the prepublish scripts, ignore rules, clean script...

All of our product repos are able to build from typescript source code in their dependencies, and this is how we've started publishing other internal packages. It makes linking and debugging much easier, but it might cause confusion if we did this with a public package...

ActuallyACat:
@zioroboco https://github.com/lerna/lerna#lernajson
The first example here leads me to believe that >1 directory deep is okay, as long as it's been defined in `lerna.json`. I wouldn't mind moving it up one directory though, but if I do that I would be rejigging some names of the other directories. I'll chat with the team and see what their thoughts are. 

>It's a shame that we needed so much boilerplate in the component package's package.json! 
I think this is a good thing. Some packages (like design-tokens) have slightly different prepublish and clean scripts. If it gets a little too hefty or confusing for newcomers we could add use yeoman to quickly bootstrap new draft components.

lerna.json
mbylstra: I like the idea of separating the core packages from the "non core" packages.

Yeah, as Ally mentions, the non-core package aren't always components (like the  `deprecated-component-library-helpers`)

The only thing that’s slightly confusing to me is that `component-library` and `tokens` are packages, but the directory names makes it seem like they are not (might be confusing to people that aren’t familiar with the monorepo). I read "package" as "NPM packages" - they are all packages in that sense.

At the risk of being wordy, could it be `draft-packages` and `core-packages`?
(deprecated-component-library-helpers is not a draft per say, but it's in same spirit and I can't think of a better name).

Or is there something special about the name `packages` that Lerna requries?

mbylstra:
> The first example here leads me to believe that >1 directory deep is okay, as long as it's been defined in lerna.json. I wouldn't mind moving it up one directory though, but if I do that I would be rejigging some names of the other directories. I'll chat with the team and see what their thoughts are.

Even if it's possible, I do kinda prefer the sibling directories, because the nested package kind of makes it look like there are only two packages, one named `component-library` and one named `draft` (but draft is just a directory not a package).

ActuallyACat:
> Otherwise LGTM for the most part, and I'm happy to approve when it's ready!
All ready now :D The draft packages are now a sibling to the existing packages directory. 

> Btw, note that this change will require explicitly including a path to the @kaizen/draft-well package in the elm.json files for all Elm apps which will require access to this package.
Good callout, thanks :D I'll make sure I include this in some release notes when the drafts are all migrated.

draft-packages/well/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "prepublish": "tsc --project tsconfig.dist.json",
+    "build": "yarn clean && yarn prepublish",
+    "build:watch": "yarn clean && yarn prepublish --watch",
+    "clean": "rimraf '*.d.ts' '*.js' '*.map'"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cultureamp/kaizen-design-system.git",
+    "directory": "packages/draft/Well"
```
zioroboco, line 14:
```suggestion
    "directory": "draft-packages/well"
```

draft-packages/well/Well.elm
```diff
@@ -0,0 +1,158 @@
...
+            maybeIdAttr
+                ++ variantClass
+                ++ borderStyleClass
+                ++ noMarginClass
+    in
+    div ([ styles.class .container ] ++ attribs) children
+
+
+styles =
+    css "@kaizen/component-library/draft/Kaizen/Well/styles.scss"
```
zioroboco, line 147:
```suggestion
    css "@kaizen/draft-well/styles.scss"
```

zioroboco:
@ActuallyACat speaking of Elm and storybook, the new package will need to be added to https://github.com/cultureamp/kaizen-design-system/blob/ally/draft-packages/elm.json#L3-L10

PR #372, opened 2020-04-05
(WIP draft) Add Data component

sebpearce:
## Changes

This PR adds a Data component in React and Elm.

image

## Supported colors

As per the Color sticker sheet in the UI Kit, this component must support:

- Wisteria 800 ("dark")
- Seedling 500 ("positive")
- Coral 500 ("negative")

All typography styles used in the Data component are 24px or greater, so there is no need to handle different shades of Seedling or Coral.

image

## Potential gotchas

This component must:

- display correctly when used in an RTL container
- allow for smaller units to be passed in **before** the number (e.g. `$363`)
- allow for smaller units to be passed in **after** the number (e.g. `75%`)
- allow for a unit to come after the number, but the same size as the number itself, e.g. `363k`, and this must work in RTL as well if the `k` needs to change places
- not restrict the user to a set of strings like `%` and `$`, since e.g. other locales have different currencies, so it must be flexible here

A possible solution to the above is to have props like:
- `children: React.ReactNode` (for the main number and full-size units like `k`; consumer can handle RTL by putting the `k` in a separate span)
- `prefixUnits?: string` (not "left" or similar because RTL)
- `suffixUnits?: string`


mbylstra:
I think it will be easier to add the Elm version to this PR (as it depends on the scss), rather than make a separate PR. I will try to do that today.

ActuallyACat:
> I think it will be easier to add the Elm version to this PR (as it depends on the scss), rather than make a separate PR. I will try to do that today.

I'll take the React component. I don't think there needs to be any changes to the Sass file so we  shouldn't conflict.

packages/component-library/components/Data/Data.tsx
```diff
@@ -0,0 +1,88 @@
+import classNames from "classnames"
```
ActuallyACat, line 1: IMO this component got _pretty close_. Outstanding items in the React component: 
* Tests
* Accessibility checks 
* Explore all use cases +, -, $, %, different currencies, etc etc

PR #375, opened 2020-04-07
very WIP attempt at Elm version of box component

mbylstra:
note: I was trying something new here: instead of our standard "config pipeline" approach, I've gone for something more like elm-css, where you pass in attributes/props as a list of function calls (behind the scenes it turns that into a custom type which then gets converted to a config record, but that's just implementation details). The reason was to make for a more concise API for the consumer.


mbylstra:
Will re-open when/if resuming work on this

PR #378, opened 2020-04-08
feat: Use the new border radius token

ActuallyACat:
Components affected: 
* Button - https://cultureamp.design/storybook-static/iframe.html?id=button-react--primary
* Navigation, the RHS button links - https://cultureamp.design/storybook-static/iframe.html?id=navigationbar-react--default
* Toast notification - https://cultureamp.design/storybook-static/iframe.html?id=toastnotification-react--cautionary
* Inline notification - https://cultureamp.design/storybook-static/iframe.html?id=inlinenotification-react--dismissible-cautionary
* Collapsable Groups - Also fixed a small bug that became more pronounced with a larger border radius. https://cultureamp.design/storybook-static/iframe.html?id=collapsible-react--collapsible-group
* Splitbutton - Button itself, as well as the dropdown menu. The same bug affecting collapsable groups. https://cultureamp.design/storybook-static/iframe.html?id=splitbutton-react--primary-disabled
* Empty State - https://cultureamp.design/storybook-static/iframe.html?id=emptystate-elm--default
* Checkbox https://cultureamp.design/storybook-static/iframe.html?id=checkboxgroup-react--with-bottom-margin
* Textfield https://cultureamp.design/storybook-static/iframe.html?id=textfield-elm--default*
* Text area https://cultureamp.design/storybook-static/iframe.html?id=textareafield-react--default-controlled
* Loading placeholder - https://cultureamp.design/storybook-static/iframe.html?id=loadingplaceholder-react--default-multiple-combined-block-and-inline
* Modal - Same bug as collapsable and split button https://cultureamp.design/storybook-static/iframe.html?id=modal-react--confirmation-informative
* Popover - https://cultureamp.design/storybook-static/iframe.html?id=popover-react--cautionary
* Select - https://cultureamp.design/storybook-static/iframe.html?id=select-react--single
* Table - https://cultureamp.design/storybook-static/iframe.html?id=table-react--expanded-popout


ActuallyACat:
[Uplift border-radius in Kaizen](https://trello.com/c/8UJbdJ6z/875-uplift-border-radius-in-kaizen)

mbylstra:
Thinking ahead to the upcoming policy: "engineers may bump fix and feat releases without doing any manual QA". Would you feel comfortable rolling these out to both our main without doing any QA? I'd lean towards it being fine (even just doing a quick QAing _after_ is released and doing a rollback if necessary is probably fine). Just wanted to make sure, or whether we wanted to consider this to be a major release, which would encourage some manual QA.

ActuallyACat:
@mbylstra Good question! Seb and I brain dumped some thoughts about how we could start applying semver to design changes here: https://docs.google.com/document/d/1Gx2JYeg6DlNFxMYlPHLT2AqzUXGj4Fq00_kclDJ-N8w/edit#

To answer your question - yup, I would be happy to release this without manual QA

packages/component-library/components/MenuList/Menu.module.scss
```diff
@@ -13,6 +14,8 @@ $side-padding: 3/4 * $ca-grid;
   border: 1px solid $kz-color-wisteria-200;
   box-shadow: $ca-box-shadow;
   line-height: 40px;
+  border-radius: $kz-border-solid-border-radius;
+  overflow: hidden; /* clip the background colors on child hover */
```
mbylstra, line 18: Something to watch out for with `overflow: hidden` is that it can potentially clip tooltips and anything else that might appear outside the menu (whether having a tooltip in a menu is a good pattern or not, we definitely have examples of this). I don't _think_ our Kaizen tooltips are absolutely positioned to the top of the document, so I think they'd get clipped (I think they _should_ be absolutely positioned to the top of the doc to avoid these issues, but that that's a separate issue). I could be wrong, but it's worth a check.

One thing I've been thinking about are "integration stories". I need to do this recently to solve z-index issues (the two components were fine by themselves, but not when integrated). Perhaps we could have a story with a tooltip inside one of items to see if there's an issue. If there's an issue, I'd suggest leaving this out and circle back to it later.

Another example of something that could be clipped would sub menus, but that would come down to implementation of that (I don't think this component has ever been used to create fly out sub menus)

ActuallyACat: > integration stories

💯 I can't wait until we have visual regression tests. 

I'm leaning towards removing this and coming back to it later.

mbylstra:
I found a few more `border-radius` instances that seem like they could use the standard token (I've checked the designs in Figma - they have 7px radius):

- HeroCard. https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/draft/Kaizen/HeroCard/HeroCard.scss#L15 (by way of it being a card, and card having a 7px border)

- Tooltip. https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/draft/Kaizen/Tooltip/Tooltip.scss#L10

- Field Message: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/draft/Kaizen/Form/Primitives/FieldMessage/styles.scss#L6

packages/component-library/components/MenuList/Menu.module.scss
```diff
@@ -13,6 +14,7 @@ $side-padding: 3/4 * $ca-grid;
   border: 1px solid $kz-color-wisteria-200;
   box-shadow: $ca-box-shadow;
   line-height: 40px;
+  border-radius: $kz-border-solid-border-radius;
```
mbylstra, line 17: I think it might be better to just leave it is it was with the square border. It looks a bit glitchy when hovering because the border dissapears, at least until we figure out a better way to handle this (perhaps without needing to do `overflow: hidden` which might cause clipping issues)

image

mbylstra: Another example I saw (DropDown uses Menu):

image

packages/component-library/draft/Kaizen/Collapsible/styles.scss
```diff
@@ -61,6 +66,8 @@ $heading-active-color: $kz-color-wisteria-100;
 
 .section {
   background-color: white;
+  border-radius: 0 0 $kz-border-solid-border-radius
+    $kz-border-solid-border-radius;
```
mbylstra, line 70: image

Tricky one, but I think this needs to be removed once it becomes sticky. Or if it's sticky mode, maybe it's easier to leave the round borders out for now (like previous). I'm not sure I'm if point at the right line of code here.

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.scss
```diff
@@ -19,6 +20,8 @@
   padding: $ca-grid $ca-grid $ca-grid * 1.5;
   text-align: center;
   background-repeat: no-repeat;
+  border-radius: $kz-border-solid-border-radius $kz-border-solid-border-radius 0
+    0;
```
mbylstra, line 24: image

It's a bit glitchy becuase of the double border. `overflow-hidden` can be used, but again, it's risky that it might clip tooltips, etc.

I wonder if there's a better way to solve the double border border problem? One thing I tried is making the outer border radius much bigger (eg `15px` so that it's hidden). This looks better, but it's a bit of a hack and not maintainable. Maybe there's a good CSS trick someone's figured out. Maybe it's better to leave this square like before for now (the second image below) - could revisit when we do a full uplift and refactor of modals.

packages/component-library/components/MenuList/Menu.module.scss
didoesdigital: With a quick look I've found another example that needs to be fixed. Let's do a design review next week.
 
image

ActuallyACat: Menulist is now back to 90deg corners: https://dev.cultureamp.design/ally/border-radius/storybook-static/iframe.html?id=menulist-react--default

Another thing to note, PEX might be looking at `MenuList` when they uplift the navigation. Something to keep an eye on.

Navbar is also fixed: https://dev.cultureamp.design/ally/border-radius/storybook-static/iframe.html?id=navigationbar-react--default

packages/component-library/draft/Kaizen/Collapsible/styles.scss
ActuallyACat: Back to square borders now: https://dev.cultureamp.design/ally/border-radius/storybook-static/iframe.html?id=collapsible-react--collapsible-group

packages/component-library/draft/Kaizen/Modal/Presets/ConfirmationModal.scss
ActuallyACat: Noting that bleed is acceptable in this scenario: https://trello.com/c/8UJbdJ6z/875-uplift-border-radius-in-kaizen

packages/component-library/components/NavigationBar/components/Menu.module.scss
```diff
@@ -66,9 +67,9 @@ $menu__square-size: $ca-grid * 2;
 
   > * {
     color: #fff;
-    background-color: #35374a; // DST-TODO: Replace this with $kz-color-wisteria-800 once kaizen-design-tokens is installed
-    border-radius: $ca-border-radius;
-    padding: $ca-border-radius 0;
+    background-color: $kz-color-wisteria-800;
+    border-radius: $kz-border-solid-border-radius;
+    padding: 3px 0;
```
didoesdigital, line 72: Leave a TODO comment here?

PR #381, opened 2020-04-09
fix: Export Paragraph and Heading from components folder

okeeffed:
Ensures `Paragraph` and `Heading` are exported from the components folder.


packages/component-library/components/index.ts
```diff
@@ -10,3 +10,5 @@ export * from "./Text"
 export * from "./Spacing"
 export * from "./Heading"
 export * from "./Box"
+export { Paragraph } from "./Paragraph"
```
ActuallyACat, line 13: There's a bunch of types that are also exported from these components, could you instead: 
```
export * from "./Paragraph"
export * from "./Heading"
```

okeeffed: Ahhh that failed before in testing - namespace clashing

PR #382, opened 2020-04-15
docs: add search field docs

didoesdigital:
To help designers and engineers make decisions about how to use search, this PR adds docs for Search Field guidelines.

"Search field" as the Component name leaves it open for us to add a "Search" page later for broader guidelines about search behaviour as a user.

Branch preview: 

- New search page: https://dev.cultureamp.design/dst/add-docs-for-search-field/components/search-field/
- Minor updates on table page: https://dev.cultureamp.design/dst/add-docs-for-search-field/components/table/
- Minor updates on select page: https://dev.cultureamp.design/dst/add-docs-for-search-field/components/select/


site/docs/components/search-field.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+import WhenToUseAndWhenNotToUse from "docs-components/WhenToUseAndWhenNotToUse"
+import WhenToUse from "docs-components/WhenToUse"
+import WhenNotToUse from "docs-components/WhenNotToUse"
+
+## Visuals
+
+### UI Kit
+
+<iframe style="border: none;" width="744" height="450" src="https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2Fe3TyAOZKfLg9cA01Iy1Mgm%2FKaizen-Site-embed%3Fnode-id%3D19823%253A84" allowfullscreen></iframe>
```
ActuallyACat, line 22: Not to be fixed in this PR, but this is missing "populated result" and "empty result" search states.

katsambarcus:
**- Search icon: Use the search icon to help users identify search fields. Either display the search icon as a presentional icon inside the search input or as an Icon Button to submit.** 👈 Should we be more prescriptive about where the search icon lives? E.g. on the left. I worry giving designers the options will create unnecessary variants of the search field. Also there's a typo for  'presentational'.

**- By showing the number of results available, people may choose how many results to inspect and whether they need to amend their search query.** 👈 I'd be curious to know what this would look like. Is the intent to start exploring the Search Results (e.g. Combine search with Autocomplete where possible.) components?

**- Global search: try to use global search that searches all of the site instead of scoped search, according to newer studies on search usability.** 👈 Will we also offer a guideline for search that only searches the immediate page/content/section? (e.g. if a search bar sits top right of a table half-way down a page?)

- If search results are likely to take longer than 1 second to show for the majority of users, display a loading indicator. For example, display a loading spinner in place of the search icon. See **loading guidelines**. 👈 Needs a link to Loading Guidelines page.

**- Empty state: Before a user has searched, consider giving them an indication of what to expect from a search. See Empty State component.** 👈 I'm not clear what this means. Is it suggesting to use an Empty State in the Search field or in proximity to the Search field? I think for total clarity, we should give some more thought as to what an empty state should look for space where Search field results _will_ be displayed. Also need to link to the Empty State component page, but I wonder if we remove this recommendation altogether for the time being until we are in agreeance on what an Empty State looks like for Search.

**- Highlight matching text: Clearly identify matching and non-matching text in results. For example, in a data table, consider showing matching text in stronger font weight.** 👈 …stronger font weight _and text highlight formatting 💡 We should defs consider putting together some guidance around how to do text highlighting. I know Patsy had experimented with this and I've seen other designers (Lindsey?) also play with this stylistic idea. Worth exploring!  

- Consider if **“debouncing”** is necessary for rapidly changing search queries to maintain performance and responsive interaction. 👈 What does debouncing mean? I think I know what you're suggesting, but I think this could be clearer. My understanding is you don't want the drop-down to appear/reappear as the user is typing a search. 

⚠️ The current text styling for 'When to use and when not to use' is definitely too small. Hoping with the incoming changes these will inherit body copy sized text.

didoesdigital:
@katsambarcus thanks for all the feedback. I've addressed most of it in the updated docs, and made notes for iterative improvements to follow up on the rest after this PR.

Issue #388, opened 2020-04-16
There's no guidance block documentation


didoesdigital:
We have a guidance block draft component thanks to https://github.com/cultureamp/kaizen-design-system/pull/280 but no guidance on how/when to use it.

![image](https://user-images.githubusercontent.com/2476974/79412507-079c1b00-7fe9-11ea-81f9-767182633d83.png)

![image](https://user-images.githubusercontent.com/2476974/79412521-0c60cf00-7fe9-11ea-86e1-72b11880152b.png)

![image](https://user-images.githubusercontent.com/2476974/79412541-1551a080-7fe9-11ea-9004-291f555707e9.png)

![image](https://user-images.githubusercontent.com/2476974/79412544-17b3fa80-7fe9-11ea-9f94-37a0bbd6fcd5.png)

![image](https://user-images.githubusercontent.com/2476974/79412547-1a165480-7fe9-11ea-96d9-3294059b8475.png)

didoesdigital:
![image](https://user-images.githubusercontent.com/2476974/79412855-c6583b00-7fe9-11ea-9d13-1f10bdcedfb8.png)

dlimiter:
Some examples from Trello:
<img width="525" alt="trello1@2x" src="https://user-images.githubusercontent.com/152166/80167914-f3799e80-8624-11ea-955d-6a1b15bc1bf7.png">
<img width="435" alt="trello2@2x" src="https://user-images.githubusercontent.com/152166/80167920-f83e5280-8624-11ea-8c33-076ad7f8f98d.png">
<img width="437" alt="trello3@2x" src="https://user-images.githubusercontent.com/152166/80167926-fb394300-8624-11ea-8fea-5d9f2dd2aa47.png">

PR #390, opened 2020-04-16
docs: add guidance block docs

didoesdigital:
Closes: #388

Branch preview links:

* New Guidance Block page: https://dev.cultureamp.design/dst/add-guidance-block-docs/components/guidance-block/
* Updated Hero Card page: https://dev.cultureamp.design/dst/add-guidance-block-docs/components/hero-card/
* Updated Tile page: https://dev.cultureamp.design/dst/add-guidance-block-docs/components/tile/
* Update icons guidelines page: https://dev.cultureamp.design/dst/add-guidance-block-docs/guidelines/icons/


packages/component-library/draft/Kaizen/GuidanceBlock/README.mdx
```diff
@@ -0,0 +1,87 @@
...
+* Provide a summary view of longer form content contained on another page. This is similar to a content application of [tile](/components/tile) but with more prominence/hierarchy (e.g Manager learning)
+* Cross promote relevant content or functionality in a different part of the platform.
+
+### Anatomy
+
+* **Card container:** The Guidance Block content appears inside a card.
+* **Spot illustration:** Provide an illustration that assist the user in the task the Guidance Block describes.
+* **Title:** Use the title to introduce the action we want people to take.
+* **Description:** Use the description to give detail about what the user can expect about taking action.
+* **Action button:** Use a link styled as a primary action button (if there are no others on the page) or a default action button to navigate to the page with the action that the Guidance Block describes.
```
juliohr, line 38: It doesn't necessarily have to link to a page. The button can execute an action then render a notification saying things went well. For example, if we have a guidance block saying "We saw that you have a draft survey hanging for 5 weeks. Would you like to launch now? (LAUNCH)". This could do the action without any navigation.

juliohr:
### In regards to the guidance block:
- Is there any guideline about stacking guidance blocks? We faced this issue on survey index. There was a scenario where two guidance blocks could be rendered at the same time. We discussed making one a priority and showing just that one.

- Another issue that we faced was having a slightly long action name that broke the button layout. We made a fix to make the button wider if the text is long (not wrapping), but if the text is too long, it wouldn't look nice. Perhaps some guidelines on the wording of the call to action button.

packages/component-library/draft/Kaizen/GuidanceBlock/README.mdx
didoesdigital: @juliohr I'm surprised by this. I would have thought the user would need more information about the draft survey in question to launch the survey than to launch it from the guidance block. If this Guidance Block implements an action immediately, while all others navigate somewhere, each might incorrectly set user expectations about how these work. Can you tell me more about that use case?

dlimiter: Hmm, we have a use case where the action button doesn't navigate away, but rather launches a modal (as the context of the action doesn't require a full page form). 

@didoesdigital I hear what you're saying about user expectations. In this case I'm personally a bit more more concerned about how the the UX of safe exploration/forgiveness interacts with this.   Ideally the content and button labelling would do a good job of describing what selecting the button will do, but if that is difficult, or the action is not easily reversible then I'd expect at least a confirmation interaction to protect the user.

didoesdigital: Ah! I wonder if we then update the guidelines to say something like "at a minimum, it must have a confirmation modal"?

dlimiter: I wonder if Lindsey's use case in comment reporting breaks this?

PR #391, opened 2020-04-16
docs: rename do to whenToUse and similar

didoesdigital:


ActuallyACat:
Just so I'm understanding correctly, the scope of this PR is to change the DoAndDont component to the WhenToUse (since that's how it's currently being used). 

Then a future PR will re-create a DoAndDont component?

didoesdigital:
> Just so I'm understanding correctly, the scope of this PR is to change the DoAndDont component to the WhenToUse (since that's how it's currently being used).
> 
> Then a future PR will re-create a DoAndDont component?

@ActuallyACat sure, we could do that. I was going to continue on this PR but shipping this first might be better, then I can rebase/merge and update other open PRs using the component before continuing.

PR #396, opened 2020-04-19
feat: set position NavigationBar refactor

caseysinclair:
- Use TS types to ensure correct children
- Re-configure how links are positioned, by using their respective key (primary, secondary, final) to 
- Mobile tweaks


caseysinclair:
[Create an explict way to pass primary, secondary and final links in the nav-bar.](https://trello.com/c/wRgvxk5L/24-create-an-explict-way-to-pass-primary-secondary-and-final-links-in-the-nav-bar)

packages/component-library/components/OffCanvas/components/Menu.tsx
```diff
@@ -1,34 +1,34 @@
...
+  links: MenuProps
 }
 
-const Menu = ({ links }: Props) => {
-  const indexOfFirstSecondaryLink = links.findIndex(
-    link => link.props.secondary
+const Menu = ({ links, section }: Props) => {
+  const renderNavItem = (link: MenuProps, index) => (
+    <li
+      key={index}
```
sampnorris, line 15: I know this was part of the package before, but index as keys can cause unnecessary re-renders, I see that the `uuid` package is a dependency. Could we use this + an 'almost-unique' sting to render a key.

E.G. the `url` for links, and `heading` for menu items

caseysinclair: Good pickup! I'll change those up

caseysinclair:
[Dropdown menu closes when you click outside of it](https://trello.com/c/mLdfX0wR/27-dropdown-menu-closes-when-you-click-outside-of-it)

PR #399, opened 2020-04-20
feat: Draft components change scope

ActuallyACat:
BREAKING CHANGE: Draft components move to be under a different scope, previously Kaizen, now KaizenDraft


mbylstra:
I like that Draft is in the name too. I wasn't always obvious before whether something was draft or not 👍

PR #400, opened 2020-04-20
feat: Replace ca-box-shadow with Zen shadow tokens

sebpearce:
## Changes

All instances of `$ca-box-shadow` along with hard-coded box-shadows (where appropriate) have been replaced with either the `$kz-shadow-small-box-shadow` or `$kz-shadow-large-box-shadow` token from the `design-tokens` package.

## Not changes

- Didn't update inset shadow on Checkbox, since that's part of border work
- Didn't update inset shadow on Radio, since that's part of border work
- Changed `$ca-box-shadow` on Toggle Switch "thumb" to hard-coded, since it's particular to that component and `shadow-small` is too big
- Left FieldMessage with `box-shadow: 0 0 0 1px $ca-palette-input-validation-negative;` because it's particular to that component

### DIFF
--- a/draft-packages/well/CHANGELOG.md
+++ b/draft-packages/well/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.1.6](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/draft-well@1.1.5...@kaizen/draft-well@1.1.6) (2020-04-17)
+
+**Note:** Version bump only for package @kaizen/draft-well
+
+
+
+
+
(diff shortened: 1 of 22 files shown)


sebpearce:
Question: do we remove `$ca-box-shadow` from `border.scss` and potentially break consuming repos?

mbylstra:
> Question: do we remove $ca-box-shadow from border.scss and potentially break consuming repos?

that would be a pretty big thing to do. I don't think we have any scope in this cycle to update consuming repos. If anything that would be a stretch goal. We might just have to live with tech debt like we have with the old color vars.

sebpearce:
Yeah maybe it's easier to just get everyone on the new import rather than break the old ones.

mbylstra:
I'm just pasting screenshots as I go. Might be useful for design review:

(some of them are too narrow just because I had two screens next to each other)

image

image

image

image

image

image

image

image

image

image

image

image


I couldn't figure out how to see the `compact` version of TitleBlock (something to do with scrolling? I didn't quite understand). I think it's low risk enough.

PR #401, opened 2020-04-20
fix: update caButtonContainer display type to avoid IE11 issue

vicchirino:
## Objective
At this moment we are facing an issue with the button component on `IE`. 
The button is not adjusting to the parent size in this browser.

**This is IE11**
the yellow background is the parent

image

**This is Chrome** 
image

Look that the button is resizing to fit in his parent.

### Causes
it seems that `inline-block` is not playing nice with `Ie11` 
Updating the `display` prop to be `inline-felx` solve the issue. 

## Changes
replace 
```
%caButtonContainer {
  @include ca-type-block(inline-block);
}
```
to be 
```
%caButtonContainer {
  @include ca-type-block(inline-flex);
}
```

### DIFF
--- a/draft-packages/well/CHANGELOG.md
+++ b/draft-packages/well/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.1.7](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/draft-well@1.1.6...@kaizen/draft-well@1.1.7) (2020-04-20)
+
+**Note:** Version bump only for package @kaizen/draft-well
+
+
+
+
+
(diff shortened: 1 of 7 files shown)


mbylstra:
Hi Victor, because there's a single commit, you need the PR title to match the commit message to make Moai happy:

image

more info here: https://github.com/cultureamp/kaizen-design-system/blob/master/CONTRIBUTING.md

I will continue to review this PR, I just noticed this.

mbylstra:
I was curious and looked into this more:

inline-flex

image

inline-block

image

So, it seems with inline-block, childen will not fill the height of the container, but with inline-flex they will. So there is _some_ different. I don't think it matters for this PR as it's not possible for consumers to set a height on the container. It is possible if people do something crazy like `button > div { height: 100px }`but we really want to disuade people from "reaching inside with CSS selectors" to customise things. I still think this PR is safe, as the chance of anyone doing that is extremely low.

vicchirino:
awesome! Thanks for taking a look a this @mbylstra. Yes, i was a little concerned too, at the beginning thought it was the `width` prop that wasn't playing good on IE11. 
Then i couldn't find a difference between `inline-box` & `inline-flex` but yeah... didn't try with the height. 
Thanks again for going deep in this 👍

PR #402, opened 2020-04-21
feat: desktop navigation bar zen styles uplift

ellisjas:
Zen uplift for the desktop version of the navigation bar. This includes uplift to the `Menu` and `Link` components. Mobile not included.

## Design
https://www.figma.com/file/KR5oFFi8ahPzJGhthw7mLs/Navigation-UI-Consolidation?node-id=1072%3A5244


packages/component-library/components/NavigationBar/components/Dropdown.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useLayoutEffect, useRef, useState } from "react"
+
+import classNames from "classnames"
+import { MenuGroup, MenuItemProps } from "../types"
+import MenuItem from "./MenuItem"
+
+const styles = require("./Menu.module.scss")
```
caseysinclair, line 7: Does the `Dropdown` component make exclusive use of the `Menu` styles? 

Wonder if it's worth moving that style code to `Dropdown.module.tsx`. This would prevent downstream changes if things were changed in the `Menu.module.scss` 

Another option might be to compose them from a central spot where `Menu.module.scss` and `Dropdown.module.scss` could `compose` / `@include` them.

ellisjas: I was going to separate it out when I do mobile uplift.

packages/component-library/components/NavigationBar/components/Link.module.scss
```diff
@@ -37,22 +31,22 @@ $link-margin: $ca-grid / 4;
   position: relative;
   display: flex;
   align-items: center;
-  height: 100%;
-  padding: $ca-grid;
+  padding: 0 $ca-grid/1.5;
```
caseysinclair, line 34: Just missing parentheses `padding: 0 ($ca-grid / 1.5);`

packages/component-library/components/NavigationBar/types.ts
```diff
@@ -27,21 +28,22 @@ export type LinkProps = {
 
 export type MenuProps = {
   header?: React.ReactElement<any>
-  items: Array<MenuItem | MenuGroup>
+  items: Array<MenuItemProps | MenuGroup>
```
caseysinclair, line 31: Might be good to namespace menu group with props too. `MenuGroupProps`

PR #406, opened 2020-04-22
feat: Use ports to dynamically adjust the select input on change

Confidenceman02:
# Background
When a user inputs text into the select input the input's width widens and contracts dynamically to accomodate the text. To achieve this in pure elm there are two obvious approaches, that being query the DOM on keydown or use the size attribute.

## Query the DOM
This was the very first approach that was taken on elm select. Querying the DOM has the issue that it takes some time to find the node in the tree. The more branches the DOM tree has the more time it will take to find that node. For most things the time taken is completely reasonable but when adjusting something based on user input there can be no perceivable lag. 

Employing the DOM query method meant that as the user typed the input width would ever so slightly lag behind and sometimes get stuck into in between widths that partially hid text. Performance on heavy pages made this effect even worse.

## Size attribute
The `size` attribute is passed a character count and it sets the input width to accomodate the amount of characters that you set. This definitely worked better than the DOM query method however when the platform started using a new font the `size` attribute needed some tuning. 

Because different fonts have different font weights the `size` attribute can't accomodate all cases. It is most inconsistent when handling capital letters, for example if we say there are 4 characters in the input `size 4` but 3 of those are capitalised, the width of the input ends up being way off which leads to hidden text.

# Solution
Using ports we are able to set up a mutation observer on a sacrificial sizer div, query its width and set the select inputs width to match. This means that no matter how fast the user types the change to the width is instant. 

The elm component dispatches a port to connect the observer when the input receives focus and then again when it is blurred to disconnect the observer and reset the input width.

# Things to consider
Using ports for the component at the moment is opt in and set to false by default. This buys us some time to switch it on for projects. This is extremely easy to do and only requires setting the `usePorts True` in the Select state and invoking the Select ports function in Elm set up code.

```
import selectPorts from "@kaizen/component-library/draft/Kaizen/Select/ports"

setupPort(ports) {
  selectPorts(ports)
}

-- elm init code here --
```

When not using ports the component will function as before.


mbylstra:
Hey Jaime. Sorry, we made a change to how we Stories are structured so that we could get them to work with Chromatic - that explains the conflict. In the new structure, rather than doing the `loadElmStories` inside the existing stories.tsx file, there needs to be a second .tsx - the new one is named ElmSelect.stories.stories. So now this `Select.stories.tsx`, `ElmSelect.stories.tsx`, and `Select.stories.elm`. I'm happy to resolve the conflict to save you the effort.

Confidenceman02:
Not a problem @mbylstra !! Cheers for resolving!

PR #408, opened 2020-04-23
feat: Update border width and color + make borders invisible where appropriate

sebpearce:
## Changes

- Update border widths on all relevant components to use the ones from the`design-tokens` package.

## Risks & things to keep in mind

- Buttons' border width will change from 1px to 2px in this PR. That means that all buttons in consuming repos that are not Kaizen buttons will stand out as looking wrong once this ships.


mbylstra:
[Use the new border width tokens in the Component Library](https://trello.com/c/zWvp5OAm/888-use-the-new-border-width-tokens-in-the-component-library)

mbylstra:
[Use the new border width tokens & make some borders transparent](https://trello.com/c/zWvp5OAm/888-use-the-new-border-width-tokens-make-some-borders-transparent)

mbylstra:
When I yarn link this is Murmur it won't compile because Murmur is using this variable that was remove: https://github.com/cultureamp/kaizen-design-system/pull/408/files#diff-5e77baee7dee0b91c4ef99b35bbc011bL8.

I think we might have to consider any variable removal a breaking change (I suggest we just leave them in for this PR). You're not really supposed to reach in like that, but we also don't enforce it - it's perfectly valid from a technical perspective. This is a bit of an issue with scss in my opinion - no way to stop people from reaching into implementation.

sebpearce:
@mbylstra Thanks for doing that and good catch! I think a solution here might be to keep the var but update it to 2px so that it the name of the var accurately represents the new border width. Hopefully that'll mean we'll have fewer non-Kaizen Buttons with the wrong border width.

sebpearce:
> You're not really supposed to reach in like that, but we also don't enforce it

BTW, I wonder if there is a way to enforce it. I always assumed there wasn't. I would start lobbying for that tomorrow if it were possible!

mbylstra:
I wonder if the latest version of scss with it's new module system has a way 🤔 I might suss it out in cooldown.

I think it might be safer to leave the deprecated variable as it was (`1px`). I think whatever is using it probably looks nothing like Zen any way. This way we don't have to worry.

mbylstra:
Otherwise, if scss doesn't support it, I've always been thinking about some kind of Kaizen specific linter that could pick up on using variables that aren't from the design-tokens repo.

mbylstra:
https://sass-lang.com/documentation/at-rules/use#private-members 🎉

sebpearce:
image

sebpearce:
glass

mbylstra:
image
(this is a very obscure reference by the way)

packages/component-library/draft/Kaizen/Select/styles.elm.scss
```diff
@@ -8,7 +8,7 @@
 
 @mixin select-border($color: $color) {
   border-color: $color;
-  box-shadow: 0 0 0 1px $color;
+  // box-shadow: 0 0 0 1px $color;
```
mbylstra, line 11: intentional?

PR #409, opened 2020-04-23
feat: Content Tabs Zen uplift 50%

mbylstra:
This is a purely visual, superficial uplift. This component has been de-prioritised, so this PR just has some basic visual improvements (it's not a complete uplift and refactor or promotion to core).

### out of scope for this PR:

- focus outline (did not previously exist, was not removed in this PR)
- any refactors such as removing unused imports, or updating usages of legacy tokens
- animation for hover states
- reversed version
- badge option
- rename to "Content Tabs". This PR has kept the same name it originally had - just "Tabs". It doesn't conflict with the Header tabs, so it's not so bad, and would mean it's easier for consumers to upgrade to the new Zen styles.
- responsive (mobile) 

### QA
I had a look in all of our repos, and I couldn't find a single one that uses this component (!). Perhaps there are some new WIP branches that are using it. Still, it makes QA easy because we don't have to worry. The only things I might have been worried about are:
- it's taller than before
- there's a min width on the tabs now (potentially wider overall)
- there background is colored in when active (might jut against something above in an ugly way)


packages/component-library/draft/Kaizen/Tabs/styles.scss
```diff
@@ -1,52 +1,75 @@
...
 .tab {
-  $underline-height: 4px;
-
-  @include ca-padding($top: $ca-grid * 0.5, $bottom: $ca-grid * 0.75);
   @include ca-type-ideal-button;
   @include ca-inherit-baseline;
+
+  position: relative;
+  top: 0; // this overrides baseline shift. @include ca-inherit-baseline does not work because
+  // we need to make this position: relative, for the absolutely positioned underline.
```
mbylstra, line 20: I've left using the paragraph component or Inter out of scope for this, so there's some required hackery

packages/component-library/draft/Kaizen/Tabs/styles.scss
```diff
@@ -1,52 +1,75 @@
...
+    right: 0;
+    bottom: 0;
+  }
 }
 
 .disabledTab {
   composes: tab;
-  color: $kz-color-wisteria-300;
   cursor: inherit;
+  opacity: 0.3;
```
mbylstra, line 69: I think this generally a good way to make things disabled visually, and the Zen UI Kit seems to do it like this everywhere. We might want to consider adding a token for it somewhere. Out of scope for this though.

packages/component-library/draft/Kaizen/Tabs/styles.scss
```diff
@@ -1,52 +1,75 @@
...
 
   &:hover {
-    color: $kz-color-wisteria-300;
+    color: $kz-color-wisteria-800;
+    background-color: transparent;
```
mbylstra, line 73: Don't show the hover state when disabled

packages/component-library/draft/Kaizen/Tabs/styles.scss
```diff
@@ -1,52 +1,75 @@
...
   @include ca-margin($bottom: $ca-grid);
-  border-bottom: 1px solid $ca-border-color;
+  border-bottom: 1px solid add-alpha($kz-color-wisteria-700, 10%);
   display: flex;
 }
 
 .tab {
-  $underline-height: 4px;
-
-  @include ca-padding($top: $ca-grid * 0.5, $bottom: $ca-grid * 0.75);
```
mbylstra, line 16: No need for top/bottom padding as the text is now centered and the component has a fixed height

mbylstra:
[Section Tabs - basic Zen uplift](https://trello.com/c/nYVcsUcG/894-section-tabs-basic-zen-uplift)

mbylstra:
Something happened since merging master (they are now super tall for some reason). I'm putting this back in development as it's not our highest priority.

ActuallyACat:
This is now ready for design review. 

I found a few tiny bugs when poking about that I fixed since I was in the area. 

When a tab is both active and disabled it had a hover state which I have removed (I also added a story for this):
https://github.com/cultureamp/kaizen-design-system/pull/409/files#diff-2b02ee06827ba872e96085ae648ac299R113-R114

I changed the active tab border from `after` to `before` as I needed to make way for the focus state. Also not required - but a good win for a11y. 
https://github.com/cultureamp/kaizen-design-system/pull/409/files#diff-2b02ee06827ba872e96085ae648ac299R53

I've also removed the bottom border (the light wisteria one). I was unsure about this, so I'd like confirmation from Marc this is expected.

PR #416, opened 2020-04-26
fix: split Elm tsx stories into separate files

mbylstra:
This is so that there is not an illegal mix of storiesOf and CSF format stories in any single file. This is required for Chromatic. It will also remove console log error noise.

I wrote this python script to automate this:

```
import os


def split_elm_stories(lines):
    for n, line in enumerate(lines):
        if line.startswith("loadElmStories"):
            tsx_react_file = remove_load_elm_stories_import(lines[0:n])
            tsx_elm_file = add_elm_stories_import(lines[n:])

            return (tsx_react_file, tsx_elm_file)


def remove_load_elm_stories_import(lines):
    for n, line in enumerate(lines):
        if "loadElmStories" in line:
            lines.pop(n)
            return lines


def add_elm_stories_import(lines):
    the_import = 'import { loadElmStories } from "@cultureamp/elm-storybook"'
    return [the_import, ""] + lines


files = [f for f in os.listdir(".") if os.path.isfile(f)]
for filename in files:
    if filename.endswith("stories.tsx"):
        with open(filename, "r") as f:
            file_contents = f.read()
            if "loadElmStories" in file_contents:
                lines = file_contents.splitlines()
                (tsx_react_file, tsx_elm_file) = split_elm_stories(lines)
                name = filename.replace(".stories.tsx", "")
                tsx_elm_filename = "Elm" + name + ".stories.tsx"
                tsx_elm_contents = "\n".join(tsx_elm_file)
                tsx_react_contents = "\n".join(tsx_react_file)

                with open(filename, "w") as f:
                    f.write(tsx_react_contents)

                with open(tsx_elm_filename, "w") as f:
                    f.write(tsx_elm_contents)
```


ActuallyACat:
My only nitpick would be to update the README in `component-library`. We've had a few people confused about the Elm/React stories and this could confuse people: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/README.md

mbylstra:
Thanks, I was going to do a quick update to the README.md just now but I realised it's a bit harder to explain with the components being moved into draft. I'll make a follow up PR that explains how to write stories in general.

PR #420, opened 2020-04-27
feat: Use new focus ring & size for checkbox & radio

sebpearce:
## Changes

- Increase the size of the checkbox and radio from 20px to 24px
- Add the Zen focus ring for both components with the correct colour, width, border-radius etc.
- Fix any resulting alignment issues


mbylstra:
This will QA'ed alongside a future in-product QA Storm (which is why is is remaining draft for the moment)

mbylstra:
I got the go-ahead from @tamoore from Perform camp to release this (once confict has been fixed). We had a look at the usages of radio and checkbox in Performance. There were very few, and the release looked very safe.

PR #421, opened 2020-04-27
feat: Added CLI tool to run codemods

ActuallyACat:
This CLI tool should make it easier for teams to change all imports from the old draft location to the new location. To keep it simple, this is _just for consumers of Kaizen_ to change imports. We (or Ally) still needs to move all draft components within Kaizen to separate packages later this week.

It's better to pair review this, I'll book in time with @mbylstra and @sebpearce this week to go through it.


.buildkite/scripts/compile.sh
```diff
@@ -2,4 +2,4 @@
 set -e
 
 yarn install --frozen-lockfile
-yarn tsc
+lerna run build
```
ActuallyACat, line 5: I've changed this because each project may have different requirements to be built. As an example, `design-tokens` doesn't need typescript to compile its assets.

ActuallyACat:
This is now super out of date and didn't quite meet the needs we had at the time. A future variation of this tool should use AST parser.

PR #434, opened 2020-04-29
feat: Lifted all form components to the draft-packages folder

ActuallyACat:
I'm relatively happy with the status of this PR. 

Notable call outs: 
1. Where reasonable, similar components (such as checkboxgroup and checkbox) are packaged together. 
2. The file structure for forms is flatter. 
```
- text-area
   - KaizenDraft <- elm lives here
   - text-area-field
     - TextAreaField.tsx
     - style.scss
   - index.ts <- barrel file
   - package.json
   - tsconfig.dist.json
   - style.scss
```


draft-packages/checkbox/KaizenDraft/Checkbox.elm
```diff
@@ -0,0 +1,198 @@
+module KaizenDraft.Checkbox.Checkbox exposing
```
mbylstra, line 1: `KaizenDraft.Checkbox.Checkbox` means the directory structure should be `KaizenDraft/Checkbox/Checkbox.elm`

The build is passing though, I wonder if that's because there aren't any stories for the Elm checkbox, and Storybook is the only thing that tries to build elm code 🤔 . I will look into.

mbylstra: I think I see what's going on - there is a story for the elm Checkbox, but it's in the old draft dir and it's pointing at the old module.

I think this is a pretty big limitation of how Kaizen is set up - it only tries to compile lint) Elm code that's used in stories. We could the elm compiler as a linter even if the output isn't used.

Should be pretty easy to fix up with a regex. The new folder structure means the imports will now be `KaizenDraft.Checkbox` (which I much prefer! Thanks for doing this!) but I'm not sure what the CLI tool is expecting. I will dig in.

mbylstra: I'm 100% sure this is an issue - perhaps there's a loophole with how we use NPM for Elm.
I will create a branch locally that merges this branch with the codemod branch, run it on Murmur with yarn link and see what happens (I'm not sure if you have already tried that).

mbylstra: I was having trouble running the CLI tool locally. I would be surprised if the resulting Elm code would compile, so I thought I'd create a branch that moves the Elm files to the place that the CLI tool expects. There's another issue though:

in this PR:

`KaizenDraft/Checkbox.elm`

the CLI tool wants:

`KaizenDraft/Checkbox/Checkbox.elm`

but the current folder structure is:

`checkbox/KaizenDraft/checkbox`

The problem is the `checkbox` dir is lower case, and the Elm compiler expects it to be title case. So we'd have to rename `checkbox/KaizenDraft/checkbox` to `checkbox/KaizenDraft/Checkbox` which I think would break a lot of other things.

mbylstra: If this _is_ an issue, it might be easier to update the CLI tool from import `KaizenDraft.Checkbox.Checkbox` to `KaizenDraft.Checkbox` and leave this PR as is.

PR #438, opened 2020-04-30
feat: copy draft Form  into single draft-packages package

mbylstra:
This is just a backup plan in case we decide to do this.
I did quick mod of the python script and ran it on just the Form directory.

I believe this will work with and would mean teams wouldn't need to have another code freeze. However, we would be stuck with the nested Form structure (and would require a future code freeze and codemods to split things out into separate packages). It does have the benefit that the draft form components would be in a separate package so there'd be less bumps to core Kaizen.


mbylstra:
[Copy Drafts **except Form** Into Draft Packages](https://trello.com/c/S8f0qlYe/904-copy-drafts-except-form-into-draft-packages)

ActuallyACat:
No more gotchas with the script. Nice - this will make things easier :D

PR #443, opened 2020-05-01
fix: add dummy index.ts files to elm only packages

mbylstra:
The build is breaking in master because `tsc` was being run on some elm only packages that have no Typescript files and was throwing an error.


mbylstra:
I'm on the lint issue

PR #448, opened 2020-05-04
feat: [SUP-40] add Elm Table

lijuenc:
- this simply allows rendering of an Elm table with the same structure and styles as the React Table
- this does not have all the same features as the React Table. What's not included:
  - table row expanded states
  - linkable table rows
  - linkable table cells

Live demo:

https://dev.cultureamp.design/lijuenc/add-elm-table/storybook-static/?path=/story/table-elm--default

Screen Shot 2020-05-04 at 3 16 08 pm


lijuenc:
>Maybe it is too complicated to change, but was there a reason for putting the children as config instead of view receiving it as the second argument?


It was a pattern that I found from another component in this repository. Do you think it's better to just make children the second argument?

juliohr:
It feels more natural. The table component is basically a custom HTML element. The convention for HTML elements in elm is a function with two arguments: a list of attribute and a list of children. 
If it is too complicated you can leave as is.

ActuallyACat:
Would it be possible to move these changes to the new draft package: https://github.com/cultureamp/kaizen-design-system/tree/master/draft-packages/modal

We will shortly be removing all drafts in the core component library in favour of separate packages

PR #452, opened 2020-05-04
ci: Added pre-commit and pre-push hooks

ActuallyACat:
I've added two hooks: 
* pre-commit - run prettier on _staged_ files only. Super quick since it's only checking what is staged to be commited. 
* pre-push - check linting across the _whole_ repo. Slower, probably ~15s to run linting across everything.


ActuallyACat:
One _slightly_ annoying this is if you forget to update the title of your PR (like I just did) and want to force Moai to re-check, you need to supply a commit message: 
`git commit -m "force maoi to update" --allow-empty` 

Not the end of the world, but an important gotcha.

PR #458, opened 2020-05-05
chore: moved kaizen draft component changes to cofrrect place

sampnorris:


draft-packages/zen-off-canvas/package.json
```diff
@@ -42,6 +42,7 @@
   },
   "peerDependencies": {
     "@kaizen/design-tokens": "1.2.0",
+    "@kaizen/draft-zen-navigation-bar": "^1.3.3",
```
ActuallyACat, line 45: Why a peer dependency and not a dependency?

sampnorris: No specific reason, was following patterns in node-packages, can update this!

Just a Q, with component library as a dependency, couldn't this cause many versions of component library in each repo? 

Should component-library be a peer dependency, assuming that every consuming repo has a version installed?

ActuallyACat: > Just a Q, with component library as a dependency, couldn't this cause many versions of component library in each repo?

It can, and currently does. Some of the stuff we'll be working on is to break the some of the shared CSS out of Kaizen so we don't rely on `component-library` as much. 

My thinking for keeping `component-library` as a dependency rather than a peer dependency is to avoid a situation where devs can't use certain draft components because they haven't updated their version of the component library. In an ideal world, we would all be up to date so this isn't a concern, but that doesn't always happen 😬 If people haven't updated their version of `component-library`, but want to use a draft component that requires a different version they can take that performance hit of duplicate dependencies until they upgrade. 

Feel free to disagree - I'm working from a hunch!

sampnorris: Not at all an issue for me at the moment, I am sure you've given it more thought! Sounds good

PR #459, opened 2020-05-06
feat: pass arbitrary children to input edit model instead of text fields

rasoliveira:
We extend the input edit modal to be able to receive generic children as opposed to be scoped only to the input text fields as before.

The driver decision for this was when we wanted a modal containing a dropdown. We realised that the current implementation would accept only text fields and instructional text. There was no flexibility.

We combed the code looking for usages. Seems like only the story board was actually using it.


sebpearce:
@rasoliveira LGTM. Could I suggest adding a bit more context to the PR summary for changes like this so future archaeologists can understand why this change was made (e.g. what kind of stuff are you planning to pass in instead of text fields?)

PR #460, opened 2020-05-06
feat: Create new Button package

sebpearce:
## Changes

- Copies the Button component from the component library to a draft package (original is still intact)
- Updates the Button styles to the new Zen theme

image
image
image
image
image
image

These changes are based on the current state of the UI Kit:

image


packages/component-library/components/Button/styles.scss
```diff
@@ -440,4 +441,11 @@ $caButton-verticalPaddingForm: calc(
   &:active span {
     color: $kz-color-coral-600;
   }
+
+  &:hover,
+  &:active,
+  &:global(.focus-visible) {
+    color: $kz-color-coral-600 !important;
+    background: $kz-color-coral-100 !important;
```
ActuallyACat, line 449: Curious if there's something else you can do here rather that `important`. Could you increase the specicifity instead?

packages/component-library/components/Button/styles.scss
```diff
@@ -77,19 +76,34 @@ $caButton-verticalPaddingForm: calc(
...
-      // show custom focus ring when :focus-visible
       &:global(.focus-visible) {
-        border: $kz-border-focus-ring-border-width $kz-border-solid-border-style
-          $kz-color-cluny-500;
+        background: $kz-color-wisteria-100;
+        border-color: $kz-color-wisteria-700;
+      }
+
+      // show custom focus ring when :focus-visible
+      &:global(.focus-visible)::after {
```
ActuallyACat, line 85: :chef-kiss:

draft-packages/button/KaizenDraft/styles.scss
```diff
@@ -0,0 +1,451 @@
+@import "~@kaizen/deprecated-component-library-helpers/styles/color";
```
ActuallyACat, line 1: I know we briefly spoke about this but is severing this color dependency too ambitious for this PR?

draft-packages/button/KaizenDraft/styles.scss
```diff
@@ -0,0 +1,451 @@
...
+
+// reset user agent syles for button elment type
+@mixin button-reset {
+  appearance: none;
+  background: transparent;
+  border: none;
+  font: inherit;
+  margin: 0;
+  padding: 0;
+  transition: none; // override Murmur global styles :(
```
ActuallyACat, line 27: :((((

draft-packages/button/KaizenDraft/styles.scss
sebpearce: I would like to, but I'm not comfortable with the risk of silently introducing visual bugs. Would prefer to do this after we have dart-sass in Murmur.

packages/component-library/components/Button/styles.scss
sebpearce: Ah, I had forgotten about this. I think I did it because I couldn't find a way to increase the specificity. I'll have a look now

sebpearce:
Update: although I've added the draft package Button to `elm.json`, the path is looking for it as a package in `node_modules` – where it clearly can't be, since this PR hasn't merged and published the Button draft package yet.

image

If these Elm stories do need to come from an installed `node_modules` package, then that means that contributors of new draft packages will not be able to add a story (or have the story work in Storybook) for a new Elm component until that component has been published.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -0,0 +1,175 @@
...
+  fullWidth?: boolean
+  disableTabFocusAndIUnderstandTheAccessibilityImplications?: boolean
+  analytics?: Analytics
+}
+
+type LabelProps = {
+  iconPosition?: "start" | "end"
+  primary?: boolean
+  secondary?: boolean
+  reverseColor?: "lapis" | "ocean" | "peach" | "seedling" | "wisteria" | "yuzu"
```
didoesdigital, line 29: Are we in a position to deprecate lapis here?

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const variantClass =
+    (props.destructive && styles.destructive) ||
+    (props.primary && styles.primary) ||
+    (props.secondary && styles.secondary)
+
+  return classNames(styles.button, variantClass, {
+    [styles.form]: props.form,
+    [styles.reversed]: props.reversed,
+    [styles.iconButton]: props.iconButton,
+    [styles.reverseColorLapis]: props.reverseColor === "lapis",
```
didoesdigital, line 149: … and here

packages/component-library/stories/Button.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { action } from "@storybook/addon-actions"
 import * as React from "react"
 
 export default {
-  title: "Button (React)",
+  title: "Button (deprecated) (React)",
```
didoesdigital, line 8: Nice 👍

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
sebpearce: Oh good call! Yes we can, looks like it was just using Wisteria anyway: https://github.com/cultureamp/kaizen-design-system/pull/460/files?file-filters%5B%5D=.json&file-filters%5B%5D=.scss&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-3b161e88246c9ecb44a018720c570bc3R247

I'll do that in a sec.

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -0,0 +1,467 @@
...
+
+%caButtonContainer {
+  display: inline-block;
+}
+
+%caButton {
+  @include button-reset;
+
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
```
mbylstra, line 39: Something for the back burner, but we should probably centralise these text rendering decisions somehow as they don't seem specific to Button - github issue perhaps?

sebpearce: Agreed. I'll make an issue 👍

sebpearce: Btw they should probably go in the front end ops CSS reset, come to think of it.

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -0,0 +1,467 @@
...
+  &:hover span,
+  &:active span {
+    color: $kz-color-coral-600;
+  }
+
+  &:hover,
+  &:active,
+  &:global(.focus-visible) {
+    // These are only used by consumers, so the !important is there to make sure these
+    // override any local competing styles that we can't foresee
```
mbylstra, line 463: Should this have `:global(.js-focus-visible)` as well? - to follow the pattern of the other user `.focus-visible` usages (I guess it's trying to restrict that it's trying to restrict it so that it only applies if the plugin is actually used). I'm having a hard time understanding "These are only used by consumers" - is this something that should be deprecated? Is there anything special about this rule such that "to make sure these override any local competing styles that we can't foresee" only applies here and not everywhere else? I'm sure there's a reason, I'm just having a hard time following it.

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -0,0 +1,467 @@
...
+    color: $kz-color-coral-600 !important;
```
mbylstra, line 464: Should this have :global(.js-focus-visible) as well? - to follow the pattern of the other user .focus-visible usages (I guess it's trying to restrict that it's trying to restrict it so that it only applies if the plugin is actually used). I'm having a hard time understanding "These are only used by consumers" - is this something that should be deprecated? Is it due to a "reacher" somewhere? How did you discover this? Is there anything special about this rule such that "to make sure these override any local competing styles that we can't foresee" only applies here and not everywhere else? I'm sure there's a reason, I'm just having a hard time following it.

sebpearce: Fair, it is a pretty crappy comment. I'll document stuff here as well and maybe you can tell me how much of this you think should go in the comment:

Here's a link to [the current Button version of this on master, for comparison](https://github.com/cultureamp/kaizen-design-system/blob/c04eb2440e4166e699afe8d84ae10b30ab438e9e/packages/component-library/components/Button/styles.scss#L430).

> is this something that should be deprecated?

No. It's currently being used only by the Elm button – someone added it in the past as a hack to create a Button variant that was both `secondary` and `destructive`. There is no secondary destructive Button in React yet. As for the Elm one, there is only 1 usage of it in Murmur:

image

There does exist a Zen design for a Secondary Destructive Button, but since A) it's not implemented in React yet and thus not used anywhere, and B) we want to redesign the API anyway, I thought I'd leave the React implementation for later.

image

> Is it due to a "reacher" somewhere?

Sort of. My memory is vague but I remember when I tried to apply the focus ring code to this variant, it was losing the specificity battle due to something that was particular to the Elm component and so I assumed there might be reachers out there that were depending on things like the reached-into `span` (see lines above the comment), so I didn't want to remove it.

To be honest I don't have a clear understanding of why this hack was done the way it was, and once I realised that the Elm implementation was quite different to the React one, this seemed like the safer choice at the time to make sure any reachers we might not know about didn't win out. Now I can't even look at the Elm story in Storybook to see what this variant looks like, because this PR hasn't been merged yet which means the Elm package isn't published yet, so I had to remove the Elm story for it to work. I can follow up with a second PR after this to bring the Elm story back and check it.

> Is there anything special about this rule such that "to make sure these override any local competing styles that we can't foresee" only applies here and not everywhere else?

Just that `%caButtonDestructiveModifier` only exists in the hack used solely in the Elm implementation of the Button.

mbylstra: Ah, this does bring back some memories now. It might worth just mentioning that it's only used in the Elm version and that the Elm version of the Button has a slightly different Html structure (as far as I understand) - as a warning if anyone desperately needs to make a React destructive button before the API gets rewritten.

Yes, good chance we'll want to completely rewrite that. There so many scss mixins and overrides going on, it's quite hard to follow - hopefully there's a simpler way.

sebpearce: Roger, I'll reword it to include those details 👍 Cheers

PR #468, opened 2020-05-07
fix: navbar events - use href to detect if trigger or not

sampnorris:
- sorry, oversight, one line change to correct behaviour


sampnorris:
@ActuallyACat if you have another minute, I refined this even further to use the actual href attribute. Much more reliable aha

PR #470, opened 2020-05-08
feat: elm draft split button

juliohr:
This is the elm implementation of the split button. It is very limited at the moment and does only the necessary for its first usage ([survey plans app](https://github.com/cultureamp/big-frontend-repo/tree/master/packages/survey-plans)). 

Functionalities and limitations:
- it has only default variant
- it expects only an href as the primary action
- menu items don't support icons, or variant (active, destructive)
- any click outside of the split button will close the menu
- escape key press will close the menu

Storybook


sebpearce:
I was thinking the same thing.

> This is copied from MenuList kaizen component. This is being used only by the Elm
version since the React version relies on kaizen's MenuList

I'm not sure I understand here. Is there a reason why this component can't also rely on Kaizen's MenuList? There is an Elm one as @lijuenc mentioned. It's at `draft-packages/menu-list/KaizenDraft/MenuList/MenuList.elm`.

sebpearce:
Aaaaand I just noticed there's no Storybook story for the Elm MenuList 🤦‍♂️  which could be misleading to someone building a component like this. I'll look into that.

The Elm DropDown imports the Elm MenuList so it might be a place to look for an example of that.

draft-packages/split-button/KaizenDraft/SplitButton/SplitButton.elm
```diff
@@ -0,0 +1,229 @@
...
+
+
+subscriptions : Sub (Msg menuItemMsg)
+subscriptions =
+    let
+        keyDecoder : Decoder (Msg menuItemMsg)
+        keyDecoder =
+            Decode.map toMsg (Decode.field "key" Decode.string)
+
+        toMsg : String -> Msg menuItemMsg
```
Confidenceman02, line 58: There is a key helper in Kaizen.Events that can do this

draft-packages/split-button/KaizenDraft/SplitButton/SplitButton.elm
```diff
@@ -0,0 +1,229 @@
...
+        toMsg string =
+            case string of
+                "Escape" ->
+                    CloseDropdownMenu
+
+                _ ->
+                    NoOp
+    in
+    Sub.batch
```
Confidenceman02, line 67: These subscriptions will be firing for every onClick and keydown on the page even if the split button is not being used. Shouldn't they only fire when a SplitButtonState is `open`?

juliohr:
We wanted to avoid depending on draft things (we wanted to depend only on stable components). If the menuList draft has a major change in the future our darft split button won't block its release. If we need to change the things for this particular menu, we don't need test backwards compatibility with the other areas using the menulist.

PR #472, opened 2020-05-08
feat: React illustrations

ActuallyACat:
## What's new / different
Illustrations! Due to time constraints, we're hosting these in Kaizen Assets (separate repo) and using the SVG as an image source. _IDEALLY_, we would inline these SVGs so we an add animations and other niceness in the future, but this was made difficult by needing to support Elm.

### Different treatment of scene and spot
In the future when the SVG are inline to the component, we will want to take advantage of treeshaking. Proactively splitting these illustrations into separate components now means we will not have to release a breaking change in the future.

With compression/running SVGs through SVGO the sizes are:
* Scene illustrations is 1.1 MB (which is why these are separate components for treeshaking and adding animations in the future)
* Spot illustrations is 451 KB (non treeshaked)

### String literal types vs enums for Scene/Spot names. 

~This could get annoying:~ Confirmed - it's annoying. Changed to string types. Please let me know if you prefer the former
```
import { Spot } from "../illustration"
import { SpotName } from "../illustration/types"

const SpotBase = () => <Spot alt="" name={SpotName.Action} />
```

### Better approach to types
I'm playing around with keeping any additional types (ie. anything that's not related to the props or state) in a separate `types.ts` file. Since we use barrel files, we _typically_ only export the component, and if consumers need types they have to reach into the package to find them. 
pro 
1. Consumers can remain relatively ignorant of how we structure our components. 
2. Compatible with our usage of barrel files

con 
1. ??

In a non-draft world, this could look something like this:
```
import { Button } from "@kaizen/component-library"
import { OtherThingsTheButtonMightNeed } from "@kaizen/component-library/button"
```
In draft-world, this looks like this:
```
import { Spot } from "@kaizen/draft-illustration"
import { SpotName } from ".@kaizen/draft-illustration/types"

const SpotBase = () => <Spot alt="" name={SpotName.Action} />
```

## Why this is a draft
* There are no tests
* I haven't verified this with the a11y brains trust
* Haven't tested this with developers from other camps


ActuallyACat:
[React Illustration component](https://trello.com/c/e7XKV3Oy/918-react-illustration-component)

draft-packages/illustration/KaizenDraft/Illustration/style.module.scss
```diff
@@ -0,0 +1,3 @@
+.wrapper {
+  width: 100%;
```
ActuallyACat, line 2: Look at scenarios where people might accidently mess with ratio

draft-packages/illustration/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@kaizen/draft-illustration",
+  "version": "1.3.4",
```
ActuallyACat, line 3: ??? `1.0.0`

draft-packages/illustration/types.ts
```diff
@@ -0,0 +1,113 @@
+export type SceneName =
```
ActuallyACat, line 1: Remove or add comms to the readme?

ActuallyACat: Remove

draft-packages/stories/IllustrationScene.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react"
```
ActuallyACat, line 1: Not for this PR - but we should look at a page on the site that lists all the icons

ActuallyACat:
Todo - 

1. Add all spot and scene illustrations as separate stories

draft-packages/stories/IllustrationScene.stories.tsx
ActuallyACat: Mitigated by having a story for every illustration

draft-packages/illustration/KaizenDraft/Illustration/Base.tsx
```diff
@@ -0,0 +1,38 @@
...
+export type BaseProps = {
+  /**
+   *  Refer to the Base Illustration Sticker Sheet in Zen UI Kit
+   */
+  name: string
+
+  /**
+   * If there is context/text surrounding this illustration that provides alt text,
+   * provide an empty string
+   */
```
sebpearce, line 14:
```suggestion
   * If there is context/text surrounding this illustration that serves
   * the function of alt text, provide an empty string so that any
   * accessibility checks know the alt text has been intentionally left out
   */
```

ActuallyACat: Much clearer, thanks!

sebpearce:
This all looks good, I think once the design review is sorted then it's 🚢  time.

PR #474, opened 2020-05-11
feat: Responsive adjustments

caseysinclair:
…ry button display inline with primary


sebpearce:
@caseysinclair We try to discourage people from pushing images to the `component-library` repo since it permanently bloats the repo size (lives forever in git history).

Any chance you would consider uploading the image to `kaizen-design-system-assets` and then accessing it with the `hosted-assets` function? There's a quick explanation in the readme here:

https://github.com/cultureamp/kaizen-design-system/tree/master/packages/hosted-assets

If you're OK with that, DST can help with getting that extra PR through.

PR #475, opened 2020-05-12
feat: Ensure focus-lock for modals when using iOS VoiceOver [CAP-54]

makenosound:
The current `<GenericModal>` implementation has an issue with iOS VoiceOver where the focus lock implementation allows focus to escape if someone uses the swipe left/right gestures to move focus.

This PR fixes that issue by adding `aria-hidden="true"` attributes to every direct descendant of the `<body>` that _isn’t_ a parent of the passed reference modal node, and then undo-ing that change once the modal is closed.

The implementation is cribbed directly from the [reach-ui implementation solving the same issue](https://github.com/reach/reach-ui/blob/de0655b2de8698f5e7add2e4507dee5062750b36/packages/dialog/src/index.tsx#L320-L362).

You should be able to confirm the existing behaviour/new on these pages with VoiceOver on:

* [Broken focus lock in Modal](https://cultureamp.design/storybook-static/iframe.html?id=modal-react--confirmation-positive-kaizen-site-demo)
* [Fixed focus lock in Modal](https://dev.cultureamp.design/fix-focus-lock-ios/storybook-static/iframe.html?id=modal-react--confirmation-positive-kaizen-site-demo)

Swiping left once the modal is open in the broken one should move the focus back to the "Open modal" button eventually.

Questions:

~- [x] Are there any likely issues this will cause within apps?~
~- [x] Is there a sensible place for the utils here to live? I’ve dumped them into the component for now until you let me know :)~

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
@@ -42,11 +42,15 @@ class GenericModal extends React.Component<Props> {
     this.preventBodyScroll()
     this.ensureAccessiblityIsMet()
     this.scrollModalToTop()
+    if (this.modalLayer) {
+      this.removeAriaHider = createAriaHider(this.modalLayer)
+    }
   }
 
   onClose() {
     this.removeEventHandlers()
     this.restoreBodyScroll()
+    this.removeAriaHider()
   }
 
   addEventHandlers() {
@@ -58,6 +62,9 @@ class GenericModal extends React.Component<Props> {
     this.props.onEscapeKeyup &&
       document.removeEventListener("keyup", this.escapeKeyHandler)
   }
+
+  // tslint:disable-next-line: no-empty
+  removeAriaHider(): void {}
 
   preventBodyScroll() {
     const hasScrollbar =
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
@@ -150,4 +150,70 @@ class GenericModal extends React.Component<Props> {
   }
 }
 
+/**
+ * Get an element's owner document. Useful when components are used in iframes
+ * or other environments like dev tools.
+ *
+ * @param element
+ */
+function getOwnerDocument<T extends HTMLElement = HTMLElement>(
+  element: T | null
+) {
+  return element && element.ownerDocument
+    ? element.ownerDocument
+    : canUseDOM()
+    ? document
+    : null
+}
+
+function canUseDOM(): boolean {
+  return (
+    typeof window !== "undefined" &&
+    typeof window.document !== "undefined" &&
+    typeof window.document.createElement !== "undefined"
+  )
+}
+
+// tslint:disable-next-line: no-empty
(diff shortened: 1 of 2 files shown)


sebpearce:
@makenosound This is a pretty creative way to solve this problem!

> Are there any likely issues this will call within apps?

It's honestly hard to say. I can't think of any, but I've never tried this – a thorough QA on a farm might be the only way to gain any certainty. It's probably also worth asking this question to the Front End Practice as a whole in case anyone in a product team recognises it as a potential hazard.

> Is there a sensible place for the utils here to live? I’ve dumped them into the component for now until you let me know :)

I personally am fine with them going in there 👍 If we want to break them out later we can.

jeremy8883:
Yeah I can't think of any problems myself. You're doing a check if the dom element already had `aria-hidden` set, so I think that's the main part. Just some things to point out on the perform side of things:

* For our modals, we have an `Inert` component, placed at the root `App.jsx` level, and we toggle it based on if we have a modal visible. Not sure if that's of any value. It currently isn't hooked up to the kaizen modals, so there shouldn't be any conflicts here.
* One thing to stress test this solution would be to go to the Admin => Self Reflections => Create => User Selection, and Admin => Evaluation Cycles (requires the `mrf_cycles` feature flag) => Create => User Selection pages. Then click on "import csv". Here we show a modal on-top of a modal.

makenosound:
> Here we show a modal on-top of a modal.

Hah, excellent stress test @jeremy8883! I don’t yet have performance running, but I’ve added a Storybook in this PR that tests the same thing. It’s at https://dev.cultureamp.design/fix-focus-lock-ios/storybook-static/iframe.html?id=modal-react--nested-modal&viewMode=story

***

I’ve done a bunch of local testing and I’m pretty comfortable that this isn’t going to cause any issues unless something very weird happens.

@sebpearce — would you mind doing a proper review?

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
```diff
@@ -150,4 +157,70 @@ class GenericModal extends React.Component<Props> {
...
+    typeof window !== "undefined" &&
+    typeof window.document !== "undefined" &&
+    typeof window.document.createElement !== "undefined"
+  )
+}
+
+// tslint:disable-next-line: no-empty
+function noop(): void {}
+
+function createAriaHider(dialogNode: HTMLElement) {
```
dougmacknz, line 187: Would be quite helpful to have a comment at the top of this to briefly explaining what this does and why (same with the `canUseDom` function above).

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
```diff
@@ -150,4 +157,70 @@ class GenericModal extends React.Component<Props> {
   }
 }
 
+/**
+ * Get an element's owner document. Useful when components are used in iframes
+ * or other environments like dev tools.
+ *
+ * @param element
```
dougmacknz, line 164: (Nitpick) JSDoc @param seems a bit pointless when we have TS here

makenosound:
Ta, @dougmacknz. Have updated those comments now!

makenosound:
Ta, @dougmacknz. Have updated those comments now!

PR #476, opened 2020-05-12
feat: navigation responsive overhaul

sampnorris:
Why? 

All our performance items don't fit - so spacing is now responsive to give the nav enough room.
Labels with icons also disappear on size changes to give us more room to work with.

Screen Shot 2020-05-12 at 12 38 59 pm
Screen Shot 2020-05-12 at 12 38 48 pm


draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/_styles.scss
```diff
@@ -17,6 +17,78 @@ $ca-navigation-bar__link-item-indicator-height: 5px;
 $ca-navigation-bar__link-item-margin: 18px;
 $ca-navigation-bar__link-item-margin-mobile: 10px;
 
+$ca-breakpoint-desktop-small-up: 1280px;
+$ca-breakpoint-desktop-large-up: 1440px;
```
mbylstra, line 21: variables in this version of scss are essentially global, so there's a chance this might override or be overridden if the same name is added to the generic `responsive.scss` file. I'd suggest artificially namespacing it with `$navbar-breakpoint-desktop-small-up` or something similar, and to indicate that this is really meant to be private to this file (you really wouldn't want another component reaching in and being dependent on this var, but this version of scss has no way to stop that).

We're hoping to introduce the latest version of scss soon, which actually has a module/namespacing feature, so you wouldn't need to worry about clashes or super long names!

draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/_styles.scss
```diff
@@ -17,6 +17,78 @@ $ca-navigation-bar__link-item-indicator-height: 5px;
...
+
+:export {
+  caBreakpointMobileMax: #{$ca-breakpoint-tablet - 1};
+}
+
+@mixin navbar-mobile-only {
```
mbylstra, line 27: exactly like what you did here!

draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/_styles.scss
sampnorris: Noted!

mbylstra:
This is a completely public repo, so some care is required in showing product features that we wouldn't want the public to see (perhaps the Performance navigation is already public knowledge, but I'd veer on the side of caution). I'd suggest removing the Peformance specific stuff from the story and the screenshots (but I realise that's tricky as you need to test those specific word sizes)

I'm guessing you want a fix out pretty quickly, so I'm happy to approve this to get the fix out, but could I ask for a follow up PR that removes the (perhaps use random characters that are the same size - eg "kjsdhfkjshdf pdfgdfgj lkjdfglkdjfg".... looks weird but better than leaking stuff)

PR #478, opened 2020-05-12
docs: new copy section added to Toggle Switch

berry-thanki:
- New copy guidelines added
- Point removed from "Need to know"
- Link added to "External links"

"Do and Don't" block pending


didoesdigital:
Subject to visual confirmation, content Looks Good To Me

PR #483, opened 2020-05-13
docs: New copy guidelines added to Textfield

berry-thanki:
- Copy guidelines section added
- Point removed from "Need to know"
- Link added to "External links"

"Do and Don't" blocks pending


didoesdigital:
Subject to visual confirmation, content Looks Good To Me

PR #489, opened 2020-05-14
experiment: Component coordination with global state

makenosound:
I mused with @mbylstra in a convo the other day that I thought it would be interesting for the design system to do some "orchestration" of components under its purview.

This is a very quick experiment intended to start a discussion around that idea. What I’ve done is:

* Create a singleton "store" within the component library that contains some state and actions for affecting that state.
* Put a value into that state object for noting whether the navigation bar has a menu that’s open or not.
* Use that value to hide the `ToastNotification` that overlaps the menu when its open.

You can see the behaviour here: https://dev.cultureamp.design/orchestration-experiment/storybook-static/iframe.html?id=navigationbar-react--with-toast&viewMode=story

My thought is that something like could mean apps consuming the design system don’t have to worry about certain aspects of coordinating components: they’re handled OOTB. Which in turn hopefully mean less work implementing, and less fragmentation of implementation. The design system could also expose the store _to_ consuming apps, which would allow consuming apps to do their own coordination with the values it contains.

There’s a large rabbit-hole we could fall into with this sort of thing, and I haven’t deeply explored the pain points, but I think there’s value in _some_ variation of it.

Would be interested to hear thoughts!

### DIFF
--- a/packages/component-library/components/NavigationBar/components/Menu.tsx
+++ b/packages/component-library/components/NavigationBar/components/Menu.tsx
@@ -12,6 +12,7 @@ const chevronDownIcon = require("@kaizen/component-library/icons/chevron-down.ic
 import classNames from "classnames"
 import * as React from "react"
 import Media from "react-media"
+import store from "../../../state"
 import { MOBILE_QUERY } from "../constants"
 import { MenuGroup, MenuItem, MenuProps } from "../types"
 import Link from "./Link"
@@ -96,6 +97,7 @@ export default class Menu extends React.Component<MenuProps, State> {
   ) => {
     const open = !this.state.open
     this.setState({ open })
+    store.actions.setNavigationBarMenuOpen(open)
   }
 
   renderMenu() {
--- a/packages/component-library/components/Notification/ToastNotification.tsx
+++ b/packages/component-library/components/Notification/ToastNotification.tsx
@@ -1,4 +1,6 @@
 import * as React from "react"
+import store from "../../state"
+import useGlobalUIState from "../../state/useGlobalUIState"
 import GenericNotification, {
   NotificationType,
 } from "./components/GenericNotification"
@@ -15,14 +17,26 @@ type Props = {
 }
 
 const ToastNotification = ({ hideCloseIcon, ...otherProps }: Props) => {
+  const [state] = useGlobalUIState()
   const persistent = otherProps.autohide && hideCloseIcon
+  const opacity = state.navigationBar.menuOpen ? 0 : 1
 
   return (
-    <GenericNotification
-      style="toast"
-      persistent={persistent}
-      {...otherProps}
-    />
+    <div
+      style={{
+        position: "fixed",
+        right: 6,
+        top: 70,
+        opacity,
+        transition: "opacity 200ms",
+      }}
+    >
+      <GenericNotification
+        style="toast"
+        persistent={persistent}
+        {...otherProps}
+      />
+    </div>
   )
 }
 
--- a/packages/component-library/state/index.ts
+++ b/packages/component-library/state/index.ts
@@ -1,0 +1,26 @@
+import createSimpleStore, {
+  Actions,
+  SimpleStore,
+  State,
+} from "../util/simpleStore"
+
+const initialState: State = {
+  navigationBar: {
+    menuOpen: false,
+  },
+}
+
+const actions: Actions = {
+  setNavigationBarMenuOpen: menuOpen => state => ({
+    navigationBar: {
+      ...state.navigationBar,
+      menuOpen,
+    },
+  }),
+}
+
+function createGlobalUIStore(): SimpleStore {
+  return createSimpleStore(initialState, actions)
+}
+
+export default createGlobalUIStore()
--- a/packages/component-library/state/useGlobalUIState.ts
+++ b/packages/component-library/state/useGlobalUIState.ts
@@ -1,0 +1,14 @@
+import React, { useEffect, useState } from "react"
+import store from "./"
+
+export default function useGlobalUIState(slicer = state => state) {
+  const [slice, setSlice] = useState(slicer(store.getState()))
+
+  useEffect(() => {
+    store.subscribe(() => {
+      setSlice(store.getState())
+    })
+  }, [store, slicer, setSlice])
+
+  return [slice]
+}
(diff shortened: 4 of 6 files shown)


sentience:
This is cool. As I think I heard someone else say earlier this week, this feels slightly “smelly” in that it breaks my mental model of a component library as a set of independently usable and composable modules, and moves us into a world where there are some high-level components that provide more of a UI framework. There are some assumptions baked into components about how they will relate to other components, that they implicitly assume will be there.

Used sparingly, this sort of thing feels powerful to me. It's the sort of thing we get to do because we own the system and can choose to make assumptions about how it will be used.

Over-used, it feels like we're at risk of creating a global namespace problem, where component-library/state becomes a dumping ground that grows into a god object in our system.

Rather than a generic component-library/state that invites unrelated concerns to be mixed into it, I wonder if we might create a single-purpose state container (i.e. `useNavbarMenuState()` instead of `useGlobalUIState()`)?

makenosound:
> ... this feels slightly “smelly” in that it breaks my mental model of a component library as a set of independently usable and composable modules, and moves us into a world where there are some high-level components that provide more of a UI framework.

Yes, that’s where I have a little hesitation too. And I think "used sparingly" is the right approach — if we can find the right slices of state to expose then we’ll be able to do interesting things.

> Over-used, it feels like we're at risk of creating a global namespace problem, where component-library/state becomes a dumping ground that grows into a god object in our system.
>
> Rather than a generic component-library/state that invites unrelated concerns to be mixed into it, I wonder if we might create a single-purpose state container (i.e. useNavbarMenuState() instead of useGlobalUIState())?

Agree, keeping it contained until it’s proven useful seems like a good approach.

To an extent this isn’t that dissimilar to having a `<ThemeProvider>` or something, though it’s crossing beyond presentation boundaries.

makenosound:
@didoesdigital / @ActuallyACat — Would love to get your thoughts here whenever you have time.

didoesdigital:
This is neat!

> moves us into a world where there are some high-level components that provide more of a UI framework. There are some assumptions baked into components about how they will relate to other components, that they implicitly assume will be there.

This popped into my head too. I'm also wondering if it will lead to unpredictability in the design system from having too much magic. Would it be possible to separate the "UI framework" parts from the "component library" parts? e.g. another package?

@makenosound have you thought about how this might relate to Elm?

makenosound:
> This popped into my head too. I'm also wondering if it will lead to unpredictability in the design system from having too much magic.

Yeah, there’s some danger there that I think is hard to play out in the abstract. Finding a good place to test it in a limited-but-pushing-enough-boundaries way would be useful.

> Would it be possible to separate the "UI framework" parts from the "component library" parts? e.g. another package?

I don’t see why not. There’d necessarily be some interaction between those packages, but it would make the conceptual boundaries a little clearer.

> ... have you thought about how this might relate to Elm?

A little! The code I’ve implemented is React-y in example but the "store" however it manifests should be agnostic to its usage. The model I’ve implemented so far could (based on my limited understanding of Elm) broadcast state changes over an Elm port which would allow any Elm app that cared to to consume those events. And we could use another port to do the opposite: broadcast state changes up from an Elm component and propagate them out. We could also build the same idea using an event bus instead, which would align the two implementations a little more.

ActuallyACat:
I would be hesitant to use this as the default behaviour for presentational components, which I believe should be as simple as possible. Another consideration is it's not possible to short circuit this behaviour if it's built into the component.

My gut-feel is to provide people with tooling (decoupled from presentational components) that makes this easier if this is a common problem that teams experience. There's a couple of ways to cut this, here's what feels the most compelling to me with the example given: 

```
// Since this is consumes the navigationBar state, NavigationBar.tsx might be the best place for this wrapper?
export const ToastNotificationWrapper = (children) => {
  const [state] = useGlobalUIState()
  const opacity = state.navigationBar.menuOpen ? 0 : 1

  return (
    <div
      style={{
        position: "fixed",
        right: 6,
        top: 70,
        opacity,
        transition: "opacity 200ms",
      }}
    >
      {children}
    </div>
  )
}
```

```
// Navigationbar.stories.tsx
export const WithToast = () => {
  const [state] = useGlobalUIState()
  const opacity = state.navigationBar.menuOpen ? 0 : 1

  return (
    <>
      <Default />
      <ToastNotificationWrapper>
        <ToastNotification
          type="informative"
          title="Informative"
          automationId="notification1"
        >
          New user data is currently being processed. We'll let you know when
          the process is completed. <a href="/">Manage users</a>
        </ToastNotification>
      </ToastNotificationWrapper>
    </>
  )
}
```

makenosound:
> I would be hesitant to use this as the default behaviour for presentational components, which I believe should be as simple as possible. Another consideration is it's not possible to short circuit this behaviour if it's built into the component.

Yes, this is totally fair, @ActuallyACat. I should have prefaced my introduction with "this is not how it should be implemented". Rather it’s intended only to show the _possibilities_ having some notion of shared state among UI components might unlock.

I think ideally the presentational components are as blind to the implementation of global UI state — like in your example — they’d either consume a prop from a wrapper that was state-aware or the wrapper would handle the behaviour entirely.

The `ToastNotification` is kind of an interesting test case for the broader question that I was planning to bring up as part of this convo: how much coordinated control should the design system exert? In my mind the design system would automatically handle the display and coordination of multiple renderings to `ToastNotification` without the consuming app having to worry about any of that implementation. Stuff like:

1. Stacking multiple notifications correctly
2. Handling any scroll overflow
3. Handling 1 and 2 when the browser is resized
4. "Dismiss all" behaviour

This sort of coordination isn’t as necessary for most of the other components (perhaps with the exception of modals in some cases).

ActuallyACat:
> I think ideally the presentational components are as blind to the implementation of global UI state — like in your example — they’d either consume a prop from a wrapper that was state-aware or the wrapper would handle the behaviour entirely.

Ace - my thoughts exactly :chef-kiss:

> In my mind the design system would automatically handle the display and coordination of multiple renderings to ToastNotification without the consuming app having to worry about any of that implementation. 

This could sit within DST's responsibilities, but I don't necessarily think it should be the responsibility of the design system (to be annoyingly nuanced). The Design System is for presentational and layout components and any statefulness in the system should be related to UI state, and only when necessary (autocomplete is the first one that comes to mind). Atlaskit does an alright job of categorising all their components, check out the left navigation: https://atlaskit.atlassian.com/packages/navigation/atlassian-navigation

I don't know if this is a problem that can be generalised at this point. I'd love to chat more about the ToastNotification in particular, seems like it's something that's currently quite painful and I'd like to get a better understanding of what teams are doing now.

PR #501, opened 2020-05-20
feat: Add script to check for accessibility violations

JakePitman:
This script:
  - gets a list of all storybook example urls
  - runs axe-puppeteer against each url to retrieve accessibility violations
  - returns a non-zero exit code when violations are found


.buildkite/scripts/check-storybook-accessibility.ts
```diff
@@ -0,0 +1,80 @@
...
+import puppeteer from "puppeteer"
+
+const getExamples = async page => {
+  const handle = await page.evaluateHandle(() => ({ window, document }))
+  const properties = await handle.getProperties()
+  const windowHandle = properties.get("window")
+  const clientApiHandle = await (await windowHandle.getProperties()).get(
+    "__STORYBOOK_CLIENT_API__"
+  )
+  const getStorybook = await clientApiHandle.evaluate(clientApi =>
```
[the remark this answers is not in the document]
JakePitman: Thanks
I'll go with `storybook`, because otherwise we end up with `stories.story.stories` two lines down

.buildkite/scripts/check-storybook-accessibility.ts
```diff
@@ -0,0 +1,80 @@
...
+      analyses.push({ url, violations: analysis.violations })
+    }
+  })
+  return analyses
+}
+
+const main = async () => {
+  const browser = await puppeteer.launch()
+  const page = await browser.newPage()
+  await page.setBypassCSP(true)
```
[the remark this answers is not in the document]
JakePitman: Seems to work okay without it 👍

.buildkite/scripts/check-storybook-accessibility.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const baseIframeUrl = "http://localhost:57357/iframe.html"
+  await page.goto(`${baseIframeUrl}?id=table-elm--default`)
```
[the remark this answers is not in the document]
JakePitman: Thanks I was meaning to test that - it does work without a specified example.

.buildkite/scripts/check-storybook-accessibility.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+  return analyses
+}
+
+const main = async () => {
+  const browser = await puppeteer.launch()
+  const page = await browser.newPage()
+  await page.setBypassCSP(true)
+
+  const baseIframeUrl = "http://localhost:57357/iframe.html"
```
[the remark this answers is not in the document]
JakePitman: Yeah I'm not sure either... 🤔

.buildkite/scripts/run-accessibility-check.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e
+
+yarn install --frozen-lockfile
```
JakePitman, line 4: I was hoping this would install ts-node but it isn't... Do I need to install it in the dockerfile?

.buildkite/scripts/check-storybook-accessibility.ts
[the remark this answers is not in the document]
ActuallyACat: I think there are two options here:
1. You don't need an http server for the static build of Storybook. I don't know much about Puppeteer, but if you can run it over a completely static html file then this should work: `const baseIframeUrl = "./storybook/public/index.html"`. 
Failing that, a super lightweight server, like http-server, could be setup for the pipeline. https://www.npmjs.com/package/http-server
2. Wait until the artifacts have been uploaded to s3 (publish step) and run this script over the production assets/branch preview. It would be great to have the build fail before assets are published so this is really, really not ideal .`${KAIZEN_DOMAIN_NAME}${KAIZEN_BASE_PATH}/storybook-static`

.buildkite/scripts/run-accessibility-check.sh
```diff
@@ -0,0 +1,6 @@
...
+
+ts-node ./check-storybook-accessibility.ts
```
JakePitman, line 6:
```suggestion
yarn ts-node ./check-storybook-accessibility.ts
```

.buildkite/scripts/check-storybook-accessibility.ts
```diff
@@ -60,20 +62,35 @@ const main = async () => {
...
-    violations.forEach(analysis => {
-      console.log(analysis)
+    examples.forEach(example => {
+      console.log(example)
     })
+    console.log(
+      `More accessibility violations were found than the current allowable limit of ${passableViolationCount}. This likely means that a new accessibility violation has been added to the code base.`
+    )
     process.exit(1)
   } else {
```
JakePitman, line 89: Fail here too (give a nice 🌈  error message, asking engineer to reduce passableViolationCount)

.buildkite/scripts/run-accessibility-check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -e
+
+yarn install --frozen-lockfile
+
+yarn storybook:build
+yarn ts-node storybook/bin/check-storybook-accessibility.ts
```
JakePitman, line 7: TODO
This still needs a base storybook url passed in as an argument.
Need to experiment with passing in a file path, so that `public/storybook` works.

.buildkite/scripts/check-storybook-accessibility.ts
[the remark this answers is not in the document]
jasononeil: It looks like a fully qualified file URL is fine, so `file://$(pwd)/storybook/public`

.buildkite/scripts/run-accessibility-check.sh
jasononeil: Not sure what you changed in the last 16 days but this is working fine now

.buildkite/scripts/run-accessibility-check.sh
jasononeil: Done in https://github.com/cultureamp/kaizen-design-system/pull/501/commits/d8482f13eb9a32ab60465f5c5c0a27199bd0088e

package.json
```diff
@@ -10,6 +10,7 @@
     "node": "^12.16.3"
   },
   "scripts": {
+    "storybook-axe": "yarn ts-node storybook/bin/check-storybook-accessibility",
```
ActuallyACat, line 13: Meganit, take it or leave it
```suggestion
    "storybook:axe": "yarn ts-node storybook/bin/check-storybook-accessibility",
```

PR #502, opened 2020-05-20
chore: Use node v12.16.3 LTS

JakePitman:
This change is necessary to run puppeteer in this repo, which requires node 10.18.0 or higher.


zioroboco:
Any reason not to go up to the latest LTS? Most of our other repos are on `v12.16.3`.

PR #504, opened 2020-05-20
feat: removed references to the old draft packages

ActuallyACat:
I've removed all references to `component-library/draft` in our draft components. This was causing issues in other repos.


ActuallyACat:
I unlocked Chromatic for this build: https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=161

sebpearce:
I've scoured this and can't see anything that stands out, so if @mbylstra is happy to ✅ it then I am too.

PR #509, opened 2020-05-25
feat: Added a generator for draft components

ActuallyACat:
## About 
This is a yeoman generator to quickly generate draft components. It scaffolds all the annoying things like package, tsconfig and setting up the folder structure for draft components, so people don't need to copy + paste when contributing to Kaizen. 

We can extend on this at a later time for creating core components.


packages/generator/generators/app/templates/package.txt
```diff
@@ -0,0 +1,38 @@
...
+    "!draft-templates",
+    "!stories",
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {},
+  "peerDependencies": {
+    "@kaizen/design-tokens": "^1.2.0",
```
mbylstra, line 35: Should we update this to the lastest (`^1.6.0`), or is this intentional to loosen the restrictions for the consuming package?

packages/generator/README.md
```diff
@@ -0,0 +1,17 @@
...
+## Install
+Install Yeoman first: 
+`yarn global add yo`
+
+Then install this generator: 
+`yarn global add @kaizen/generator-create`
+
+## Usage
+`yo create` 
+`yo create:draft`
```
mbylstra, line 13: I think these instructions are out of date:

These are the only commands that worked for me:

`yo` (and then choose @kaizen/create from available generators)
`yo @kaizen/create`

`yo @kaizen/create:draft` or `yo create:draft` did not exist

I think `@kaizen/create:draft` or `@kaizen/create-draft` is better than just `@kaizen/create` (what if we want to create legacy packages), but I'm not familiar with yeoman to know how to do this.

I tested this by using `yarn link`.

mbylstra:
This is really nice 👍 and I like that it also generates some component scaffolding for both languages.

packages/generator/README.md
ActuallyACat: They are, good catch!

ActuallyACat:
I've added a placeholder default generator (the thing you see when you run `yo @kaizen/create`) so we can keep this generator generic. For now, we only have one subgenerator, `yo @kaizen/create:draft`. I think we'll need this in the future, so generic is good.

mbylstra:
Example run:

image

ActuallyACat:
LGTM - thanks for fixing up the Elm stories and components 👌

PR #511, opened 2020-05-26
feat: Create separate stylesheet for ca-grid var

sebpearce:
## Changes

This simply copies the `$ca-grid` variable into its own stylesheet, `grid.scss`, which lives in the component library.

This will allow us to import from this file instead of what everyone currently has to do, which is to get `$ca-grid` from `type.scss`.


sebpearce:
@ActuallyACat All good, this is a copy and the original is still intact 👍

PR #515, opened 2020-05-28
docs: update Illustration navTitle and demoStoryId

didoesdigital:
- Remove "(not built)"
- Add Illustration demo to Illustration page

Branch preview: https://dev.cultureamp.design/dst/update-illustration-docs/components/illustration/


sebpearce:
@didoesdigital Probably worth us adding a default height here. I'll add it now (I remember where the others were used so will be quicker to copy + paste)

image

PR #527, opened 2020-06-03
feat: Change default behaviour for modal focus [CAP-67]

makenosound:
[🎫 ](https://cultureamp.atlassian.net/jira/software/projects/CAP/boards/80?selectedIssue=CAP-67)

The existing behaviour for modals is that when they open we let the `<FocusLock>` component from the `react-focus-lock` package handle auto-focusing whatever it thinks is the "right" element to focus on (usually the first button or input element).

This isn’t ideal for a couple of reasons:

1. From an accessibility perspective, as a default, it means we can skip over framing content that is important for screen readers
2. It causes blocking issues on iOS using VoiceOver when it tries to auto-focus on a textarea — the keyboard isn’t invoked when the input is focusesd and is impossible to invoke.

After much discussion about the right default behaviour, we’ve decided that the best default rule would be to move focus to whatever the `<ModalAccessibleLabel>` is. This should be the primary label for the modal (and in the preset implementations is always the title of the modal).

While I was going through the code I realised that there was a potential ID clashing issue with the way we’re using constants for modal labels: there’s a place in Performance where we’re using two modals nested on top of one another, and so using the same ID for both modals means we could select and focus the wrong element.

To avoid that I’ve added a context that wraps the `GenericModal` that passes through dynamically generated IDs for `labelledby` and `describedby` that is consumed in `ModalAccessibleLabel` and `ModalAccessibleDescription`.

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
@@ -24,7 +24,7 @@ interface Props extends ContainerProps, ModalAccessibleContextType {}
 
 const MODAL_TRANSITION_TIMEOUT = 350
 
-function GenericModalContainer(props: ContainerProps): React.ReactNode {
+function GenericModalContainer(props: ContainerProps) {
   const labelledByID = uuid()
   const describedByID = uuid()
   return (
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
@@ -1,14 +1,18 @@
 import * as React from "react"
 import { createPortal } from "react-dom"
 import FocusLock from "react-focus-lock"
+import uuid from "uuid/v4"
 const { CSSTransition } = require("react-transition-group")
+import {
+  ModalAccessibleContext,
+  ModalAccessibleContextType,
+} from "./ModalAccessibleContext"
 
 import { warn } from "@kaizen/component-library/util/console"
-import { ID_DESCRIBEDBY, ID_LABELLEDBY } from "./constants"
 
 const styles = require("./GenericModal.scss")
 
-interface Props {
+interface ContainerProps {
   readonly isOpen: boolean
   readonly children: React.ReactNode
   readonly focusLockDisabled?: boolean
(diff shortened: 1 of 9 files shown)


draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
```diff
@@ -1,23 +1,48 @@
...
-interface Props {
+interface ContainerProps {
   readonly isOpen: boolean
   readonly children: React.ReactNode
   readonly focusLockDisabled?: boolean
   readonly onEscapeKeyup?: (event: KeyboardEvent) => void
   readonly onOutsideModalClick?: (event: React.MouseEvent) => void
 }
 
+interface Props extends ContainerProps, ModalAccessibleContextType {}
```
ActuallyACat, line 23: /nitpick I couldn't grok what this was for on a first parse. Could you rename to something like `GenericModalProps`?

ActuallyACat:
Just want to flag these two screenshots with @katsambarcus 

Screen Shot 2020-06-03 at 5 09 30 pm
Screen Shot 2020-06-03 at 5 09 22 pm

ActuallyACat:
LGTM - pending a check from Marc

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
makenosound: Ah, yep! Can do.

makenosound:
Thanks for the review, @ActuallyACat (and for pulling out the screenies I should’ve done that before).

@katsambarcus — FYI those focus rings should only appear when the user has invoked a modal from the keyboard. If it’s invoke via mouse click/tap then the element will be focused still, but the visual indicator won’t appear. This is a poor quality gif, but you should be able to see the different behaviour:

output

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
makenosound: That’s updated now, @ActuallyACat. Mind ✅ ing it again?

katsambarcus: @ActuallyACat @makenosound Thanks for sharing. That all makes sense ;-) 

The only other thing worth noting is the colour of the Primary button on these models needs to be upgraded to the Cluny 500 Primary Action.

makenosound: Thanks, @katsambarcus! I _think_ there’s a separate WIP branch to update those buttons everywhere, but perhaps @ActuallyACat can confirm.

If we need to do some work here I’m happy to adjust all the modals while I’m in this zone.

PR #532, opened 2020-06-05
feat: add temporary Inter helper mixins to deprecated-component-library-helpers

mbylstra:
This PR adds mixins for the Inter font to @kaizen/deprecated-helper-mixins.

It should be noted that these are intended to be _temporary_ mixins for a gradual and easy rollout of the Inter font in consuming packages. They are not "fully Zen" mixins - they merely swap the font but do not change any font sizes or typography heirarchy. They do not adhere to the Zen UI Kit. Mixins that adhere to the Zen UI Kit will come in a separate PR and can be used alternatively.

The  `ca-` prefix has been intentionally left in, as they don't really get the Kaizen stamp of approval. They are more of a stepping stone from the old very legacy `ca-xxxxx` mixins


packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -94,9 +101,9 @@ $ca-ideal-locale-he-font-family: "Ideal Sans A", "Ideal Sans B", Tahoma,
   );
 }
 
-@mixin ca-type-ideal-small-bold($args...) {
-  @include ca-type-ideal-small;
-  font-weight: $ca-weight-semibold;
+@mixin ca-type-inter-small-bold($args...) {
```
ActuallyACat, line 104: Could you duplicate this mixin too? Other wise this will have to be a breaking change

mbylstra: Yes, that was a mistake, I have pushed a commit.

mbylstra: fixed

PR #533, opened 2020-06-05
feat: Added inter mixins that map 1-1 with Figma typography

ActuallyACat:
I should note that these are for internal use only, we would hope that developers can use our components instead of these mixins (in the future)


ActuallyACat:
Consolidated the file as discussed

PR #536, opened 2020-06-09
fix: re-add @kaizen/draft-button Elm stories

mbylstra:
These can be added now that @kaizen/draft-button has been published

image


mbylstra:
forgot to update the new story names... commit incoming

ActuallyACat:
It wasn't introduced in this PR so it doesn't require a fix now, but I've noticed the Elm button is slightly different to the React one.

Screen Shot 2020-06-10 at 9 42 01 am

mbylstra:
> It wasn't introduced in this PR so it doesn't require a fix now, but I've noticed the Elm button is slightly different to the React one.

Good catch. I think Button is a really good candidate for Web Components. No internal state or complex events, but complexity in the number of variations and chance for drift. Very important piece of UI where consistency really matters.

PR #539, opened 2020-06-10
docs: Added addon-info plugin to Storybook

ActuallyACat:
## About
Now that we've split out draft components into separate components, we need better documentation for developers around where to import components from. This PR adds a plugin, `addon-info` to storybook so we can quickly add some inline MDX in our stories. 

**Pros**
- Can quickly write inline MDX. addon-docs would have required us to write stories in MDX to get customised comments in there
- addon-docs was janky with lots (>20) of stories

**Cons**
* ~This tab will appear in the non-storybook parts of the website. I _think_ we can disable this, reading more about it now.~ Fixed. I've moved the tab to the bottom of the page. 

Old: 
Screen Shot 2020-06-10 at 5 40 24 pm

New: 
Screen Shot 2020-06-11 at 10 32 25 am

I actually kinda like the new one. If a developer is looking at these docs they can toggle Show Info to view the props table etc


draft-packages/stories/Dropdown.stories.tsx
```diff
@@ -62,6 +62,14 @@ const Menu: React.FunctionComponent = () => (
```
ActuallyACat, line 62: I think this file should instead be in the component library stories.. I'm a little confused about this one. Looks like the React component is a core component, but the Elm components are draft.

mbylstra: Ah, this rings a bell - I remember coming across a component that was core in React and draft in Elm (or vice versa)

PR #545, opened 2020-06-15
ci: Add stylelint to lint styles

ActuallyACat:
## About
Kaizen accepts contributions from developers within Culture Amp. We would like to maintain a  standard for contributions in Sass, and where possible avoid some errors. 

## Rules
Related to our existing coding conventions. These should be a surprise to no one: 
```
"string-quotes": "double", 
"no-duplicate-at-import-rules": true,
"block-no-empty": true,
"block-opening-brace-space-before": "always",
"declaration-bang-space-after": "never",
```

Rules related to Stylelint's [possible errors](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules/list.md#possible-errors):
```
"selector-pseudo-class-no-unknown": [
   true,
   {
      ignorePseudoClasses: ["global", "export"],
   },
 ],
"no-duplicate-selectors": true,
"font-family-no-duplicate-names": true,
"color-named": [
   "always-where-possible",
   {
      severity: "warning",
   },
],
```

I've intentionally left this very barebones. It's more important to start linting styles than to create a set of rules that are too restrictive.


sebpearce:
> I've intentionally left this very barebones. It's more important to start linting styles than to create a set of rules that are too restrictive.

👨‍🍳 💋👌 

I'm a bit puzzled over what the `/* stylelint-disable-next-line no-duplicate-selectors */` stuff is there for though.

ActuallyACat:
> I'm a bit puzzled over what the /* stylelint-disable-next-line no-duplicate-selectors */ stuff is there for though in the Input styles.

Earlier above in the Sass file there's a declaration for `.withStartIconAdornment`, which also nests `.startIconAdornment`, and `.withDisabled`. These classes are repeated in `L136`. 
https://github.com/cultureamp/kaizen-design-system/blob/180ef629780b51b25d957f9e3b26d07fc082daac/draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss#L85

sebpearce:
Oh right! I think it's too late in the day for my brain, but that just looks like an error to me. If the classes are repeated, isn't the linter correct in telling us to get rid of whichever ones aren't being overridden?

I'm guessing you added those comments because this PR is just supposed to get the linter working and leave everything else as is, in which case I'm fine with it 👍

packages/component-library/styles/reset.scss
```diff
@@ -39,13 +41,6 @@
     height: auto;
   }
 
-  body {
-    margin: 0;
-    padding: 0;
-    -webkit-font-smoothing: antialiased;
-    -moz-osx-font-smoothing: grayscale;
-  }
-
```
mbylstra, line 48: nice!

PR #548, opened 2020-06-17
feat: Secondary Single Select Reversed (React)

mbylstra:
image

Branch preview:
https://dev.cultureamp.design/dst/secondary-select-react/storybook-static/?path=/story/select-react--single-secondary-with-ellipsis

Figma:
https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A25645

### In scope
- Secondary Single Select Reversed variant
- Visually meet the Zen design 80%, but not full Zen uplift (see out of scope)
- separate prop: `selectControlWidth`: `fillContainer` | `containSelection`. The default variant defaults to `fillContainer` and the secondary variant defaults to `containSelection`. When `containSelection` is true, the size of the "control" is shrink wrapped to the selected option text .
- truncate with ellipsis when the size of the control is bigger than a fixed width container

### Out of scope (these can be followed up in separate PRs):
- Secondary Single Select Non-Reversed (white) variant (should be easy, but descoping in the interest of scope hammering)
- the label (the current component doesn't support a label - it's currently BYO label)
- Full Zen uplift of related variants and where code (css) will shared between variants (eg: new focus rings). See https://github.com/cultureamp/kaizen-design-system/pull/548#issuecomment-647857515 for more details.

- Accessibility fixes to existing Select component (if there are any)

- this pre-existing bug in the searchable variant:
image
- Animation (possibly the control size should animate when it suddenly changes size to make it less jarring?)
- Searchable variant
- Secondary Multi Select and other more complicated versions
- Elm versions

### TODO
- [x] reversed secondary props and styling
- [x] `selectControlWidth` prop and basic implementation for `secondary` (not implemented for default)
- [x] `containSelection` should ellipsize when the control is wider than a fixed width container 
- [x] `selectControlWidth` should default to `containSelection` for `secondary`
- [x] make stories that exercise edge cases
- [ ] TS docs for storybook

### TBC

- [x] naming things
- [x] should the control we the width of the selected option, or be the width of the widest option, like a native `select`?

### Notes for the Code Reviewer

- I tried to add JS Docs, but they don't seem to be appearing in Storybook. Any tips to get this working would be appreciated? At least the docs exist in the source code for now.
- is `variant` the best prop name? Is something like `prominence` better? Acording to the Kaizen docs (https://cultureamp.design/components/select/), variant is `single` or `multi`, but that's not true in the code API, as the API is inherited from `react-select`.

### QA
- make sure to thoroughly regression test all of the other existing Select component variations, both in Chromatic and Storybook(functional). Shall we get Dylan? I feel we could make better use of our QA staff
- regression test the Elm select (I don't think it uses any shared style, but just in case)

### Design review screenshots (IE11)
Ignore the missing pixels: that is just because of the emulator (graphics card emulation issue?)
image
image
image

Just use Chromatic/Storybook for Non IE11


mbylstra:
Design review with @katsambarcus :

-  label has changed recently (which brought up that the current component doesn't even support labels)
- padding at top of dropdown is too small
- inside dropdown, currently selected option text colour should be cluny (not just the background)
- spacing between chevron to selected option should be 6px

Each of these are existing issues (they also exist for the default and non-reversed components). They will be made out of scope for this PR, and will be picked up in a second PR where we can do a more thorough design review.

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -14,7 +14,71 @@ const styles = require("./styles.react.scss")
...
+  }
+  const { variant = "default", reversed = false } = props
+
+  // the default for selectControlWidth depends on the variant
+  const selectControlWidth =
+    props.selectControlWidth != null
+      ? props.selectControlWidth
+      : variant === "secondary"
+      ? "containSelection"
+      : "fillContainer"
```
mbylstra, line 63: I hate this code, but I'm not sure what the better approach is? Prettier strips away brackets I tried to add to make it more readable (because of the gross nested ternaries). The ternary `?`/`:` is the only "expression operator" available in JS, otherwise I'd have to use `let` and `if` statements (which aren't expressions) which feels wrong. Any better ideas?

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -50,10 +115,11 @@ export const AsyncSelect = (props: AsyncProps) => {
         NoOptionsMessage,
         SingleValue,
         MultiValue,
+        IndicatorsContainer,
         ClearIndicator: null,
         IndicatorSeparator: null,
       }}
-      className={classNames(styles.container, props.className)}
+      className={classNames(styles.specificityIncreaser, props.className)}
```
mbylstra, line 122: This rename is a refactor. I was wondering why there was a `container` class wrapping everything even though it wasn't being used for any scss customisation. CSS Modules already scopes things, so it seemed unecessary. It also makes using the scss `&` operator more awkward, because you now have a grandparent selector. My guess is that it was added to increase specificity to win over react-select styles. I think this makes it more explicit why it exists.

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -63,6 +129,7 @@ const Control: typeof components.Control = props => (
     {...props}
     className={classNames(styles.control, {
       [styles.focusedControl]: props.isFocused,
+      [styles.disabled]: props.isDisabled,
```
mbylstra, line 132: disabled was previously unimplemented - it came for free from react-select, which I suppose was close enough to the Kaizen design.

draft-packages/select/KaizenDraft/Select/styles.react.scss
```diff
@@ -12,7 +12,7 @@
   color: $kz-color-wisteria-800;
 }
 
-.container {
+.specificityIncreaser {
```
mbylstra, line 15: renamed, as this seemed to be the only reason for it's existance

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -14,7 +14,71 @@ const styles = require("./styles.react.scss")
...
+  /**
+   * Whether the "control" (the button you click to open the menu) width fills the
+   * container or is as wide as the selected option text. Note that the control text
+   * will ellipsize if it is wider than the parent container.
+   * `variant="default" selectControlWidth="containSelection"` is not implemented and
+   * will throw a "not implemented" error
+   * @default "containSelection" if variant is "secondary", otherwise "fillContainer"
+   */
+  selectControlWidth?: SelectControlWidthType
+}
```
ActuallyACat, line 41: Nice - not sure why the info addon isn't picking these up though :/

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -14,7 +14,71 @@ const styles = require("./styles.react.scss")
...
+
+export type VariantType = "default" | "secondary"
+export type SelectControlWidthType = "fillContainer" | "containSelection"
```
ActuallyACat, line 44: Since the options here are binary, I'd prefer to use `fullWidth?: boolean`. 
fullWidth = true -> fillContainer
fullWidth = false -> containSelection

https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Button/Button.tsx#L9

draft-packages/select/KaizenDraft/Select/Select.tsx
ActuallyACat: No better way that I'm aware of. Splitting out to separate statements isn't too terrible. I don't have any preference to either approach 🤷‍♀️
```
let secondaryWidth = variant === "secondary"
      ? "containSelection"
      : "fillContainer";

  const selectControlWidth =
    props.selectControlWidth != null
      ? props.selectControlWidth
      : secondaryWidth
```

draft-packages/select/KaizenDraft/Select/Select.tsx
mbylstra: We arrived at this in this thread: https://cultureamp.slack.com/archives/GF02FQA30/p1592366131161600

We didn't consider "full width" true/false, but I guess we were trying to make it as self documentating as possible.
There's a potential third option: "widthOfTheWidestOption", so making it non-binary gives flexibility to add other options later. Happy to delay this PR until @didoesdigital is back.

PR #549, opened 2020-06-17
feat: Added spacing tokens

ActuallyACat:
## About 
There's still no consistent representation of a generic size scale in the Design System. We have spent a lot of time bikeshedding about this when discussing size in typography: 
https://github.com/cultureamp/kaizen-design-system/pull/268

In other conversations it seems like t-shirt sizes are the preferred approach:

1. In which borders were recommended to use xs, sm, md, etc https://cultureamp.slack.com/archives/GF02FQA30/p1585723122061200?thread_ts=1585720978.059000&cid=GF02FQA30
2. Generic spacing names in other parts of our platform: https://cultureamp.slack.com/archives/GF02FQA30/p1581981671002400 / [code example](https://github.com/cultureamp/big-frontend-repo/blob/7dece14f8ccd19288535f629d26dff93aae006d8/packages/content-hub/src/ContentHub/Component/Stack.elm#L19-L39)
3. T-shirt sizes: https://cultureamp.slack.com/archives/GF02FQA30/p1573089666005700?thread_ts=1573088762.000700&cid=GF02FQA30

I propose we adopt t-shirt conventions for size as: 
* It's used elsewhere in the platform
* It's short and easy to remember and type (compared to the longer notation `$kz-spacing-extra-extra-large`)
* We already use it in Kaizen props (in the long format) [here](https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/tag/KaizenDraft/Tag/Tag.elm#L45) and [here](https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/popover/KaizenDraft/Popover/Popover.tsx#L42)


sebpearce:
I'd like to revisit this during/after our breakpoint chat on Thurs.

sebpearce:
Something I'm not clear on: if I'm an engineer and my designer tells me that there should be a one grid unit gutter between two things, how would I do that with these tokens, given that there's no token for a single grid unit?

ActuallyACat:
You raise a good point though @sebpearce - for completeness and to bridge the gap between language between devs and designers it would be good to add another column here (i'll do in a follow up PR): https://cultureamp.design/guidelines/spacing/#to-keep-in-mind

PR #550, opened 2020-06-17
fix: Checkbox and radio button inconsistent behaviour on touch devices

makenosound:
We’ve had an issue come up where on touch devices (iOS specifically) the checkbox/radio fields don’t allow for selection, or at least behave inconsistently. They’ll _sometimes_ check easily, but them sometimes only after multiple taps, and then sometimes not really at all.

The root cause of this is the way that iOS handles `:hover` behaviour. Because you can’t really hover over an element on a touch device, iOS tries to create an intermediary state for these (to handle situations where `:hover` would result in content being made visible).

Our usage here with the `label:hover` rule though means that it intercepts the user tapping on the label and then stops the event propagating down to the contained `<input>`. The same is true for all the other hover states within the checkbox/radio primitives.

What I’ve done is separate all the `:hover` effects for these elements by using Level 4 Media Queries, so instead of this:

```
label:hover { ...
```

We have:

```
@media (hover: hover) and (pointer: fine) {
  label:hover { ...
}
```

... which ensures that the hover styles aren’t enabled for touch devices. This syntax is still an unfinalised spec, but it’s well supported across browsers:

* https://caniuse.com/#search=media%20hover
* https://caniuse.com/#search=media%20pointer

And in IE11 where it unsupported we’ll just get slightly less nicer hover styles but the fields will work.

There’s also a change here to set `pointer-events: none` to everything nested within the `<label>`s for each of `CheckboxField` and `Radio`. I was seeing very inconsistent behaviour tapping the individual "input" controls for each of those on iOS and it appears to be a similar-but-obviously-not-identical issue as the `:hover` problem. We don’t actually need anything inside the `<label>` to be interact-able because the label is wrapping the controls, and so this

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/CheckboxField/CheckboxField.tsx
+++ b/draft-packages/form/KaizenDraft/Form/CheckboxField/CheckboxField.tsx
@@ -1,4 +1,5 @@
 import { Checkbox, CheckedStatus, Label } from "@kaizen/draft-form"
+import { touchEnabled } from "@kaizen/component-library/util/browserFeatures"
 import classnames from "classnames"
 import * as React from "react"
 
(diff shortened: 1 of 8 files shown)


ActuallyACat:
Had to do a bit of reading up on this one. It LGTM, thanks 🎉

makenosound:
> Had to do a bit of reading up on this one

Yes, me too! 😅 Thanks for looking it over.

PR #553, opened 2020-06-17
fix: Update <Well> to latest colours/tokens

makenosound:
Does what it says on the tin.

Note: the way we’re setting no-border has changed to use a transparent border rather than removing the border so we need to change the order of all the rules to ensure the colour rules are last.


makenosound:
Thanks, @sebpearce! Ran this by @katsambarcus and he’s given the 👍 so I’ve pressed the button.

PR #554, opened 2020-06-17
fix: add pointer-events none to .box element

Confidenceman02:
The input used for checkbox was sitting behind styling elements and not initiating the onCheck event. This PR adds`pointer-events: none;' to the `.box` element to ensure the input is clickable.

The checkbox width and height is also conforming to the relevant sizes for the styled element to ensure the clickable area is maximised. Previously the input was a tiny square in the top left corner, it now is sized according to `$checkbox-size`.

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/Checkbox/Checkbox.elm
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/Checkbox/Checkbox.elm
@@ -175,7 +175,7 @@ view (Config config) =
                 ++ maybeOnCheckAttr
                 ++ maybeNameAttr
     in
-    div [ styles.class .container ]
+    div []
         [ input
             (attribs
                 ++ [ type_ "checkbox"
--- a/draft-packages/form/KaizenDraft/Form/Primitives/Checkbox/styles.scss
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/Checkbox/styles.scss
@@ -10,6 +10,8 @@ $focus-ring-offset: 2px;
 
 .checkbox {
   @include form-input-visually-hidden();
+  width: $checkbox-size;
+  height: $checkbox-size;
 }
 
 .icon {
@@ -26,6 +28,7 @@ $focus-ring-offset: 2px;
 
 .box {
   position: relative;
+  pointer-events: none;
   background: $kz-color-white;
   height: $checkbox-size;
   width: $checkbox-size;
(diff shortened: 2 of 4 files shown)


sebpearce:
LGTM. I guess you got the giant `yarn.lock` because running yarn wasn't part of our PR to add stylelint 🤷

sebpearce:
Lucky I checked. There's a bug when hovering over the label that I can't reproduce on master.

sentience:
Can you explain the use case here a bit more? I'm not a fan of starting a `z-index` arms race. If there was an invisible element in front of your checkbox blocking its interactivity, would it make more sense to use `pointer-events: none` on that invisible element?

Confidenceman02:
Looks like this is an issue with label. The padding-left is at fault here.

Confidenceman02:
> Can you explain the use case here a bit more? I'm not a fan of starting a `z-index` arms race. If there was an invisible element in front of your checkbox blocking its interactivity, would it make more sense to use `pointer-events: none` on that invisible element?

It's not really invisible. Its the box element which provides the actual style of the tick. Using `pointer-events: none;` does indeed work! Thanks for that.

Confidenceman02:
> Oh I see, so we've got a box sitting over the actual checkbox? Sounds like using `pointer-events: none` to make the visible checkbox (box) allow clicks through to the “real” checkbox may be a bit fragile. Do we have any guarantee that the real checkbox's clickable area will always match the visible checkbox?
> 
> I'm thinking we probably want to find a way to make the box receive the click event and have it toggle the checkbox.
> 
> Do we already have a working strategy for this in the React checkbox, and could we adopt that?

Hmm.. This would mean making it a stateful component if its handling toggling the checkbox. 
I'm not sure about the styled box handling the click considering its the actual input that receives focus and fires the onCheck event. This would mean we have an event handler on both the input and the styled box. 

The input is sized identically to the styled box so we can be pretty sure it's always going to match it. If we want to be super duper sure we can put the size on the parent instead then size the both the input and visible box to 100% i.e.

```
div [  width: 24px, height: 24px, position: relative. ]
      [  input  [width: 100%, height 100%]  []
      ,  visibleBox. [width: 100%, height: 100%]. []
      ]
```

sentience:
The way I've seen this done elsewhere is to put the visible checkbox (the box) inside the checkbox's `label`, so that just like the rest of the `label`, clicking it causes the checkbox to be toggled. No additional state or event listeners needed.

That may create new issues for you, though. If you're confident that the read and fake controls are being sized the same, I'm OK with it – especially if this is the same approach that our React component is using.

Confidenceman02:
> The way I've seen this done elsewhere is to put the visible checkbox (the box) inside the checkbox's `label`, so that just like the rest of the `label`, clicking it causes the checkbox to be toggled. No additional state or event listeners needed.
> 
> That may create new issues for you, though. If you're confident that the read and fake controls are being sized the same, I'm OK with it – especially if this is the same approach that our React component is using.

The box is actually inside the checkboxes label already. The div after the input is the box.
Screen Shot 2020-06-18 at 10 03 55 am

PR #558, opened 2020-06-18
fix: Add fallback font stacks

makenosound:
Our current `fontFamily` stack has no fallback stack, so if the font fails to load we can get sadness.


didoesdigital:
@makenosound I think we're serving Greycliff as "Greycliff CF", so you may need to restore the " CF"

makenosound:
@didoesdigital Ack, yes. I’ve fixed that now.

sebpearce:
Suggest running `yarn build` in the `design-tokens/` dir to generate the Sass and Less tokens, just to see if they turn out OK.

makenosound:
Ah, thanks @sebpearce! I didn’t realise that was a manual task.

sebpearce:
@makenosound It's actually automated, just useful to do it manually in situations like this 👍

ActuallyACat:
I keep thinking I've fixed this (build tokens before publishing the package) since the TS files are compiled and published but this package isn't :/ I'll put it back on my to do list.

makenosound:
Thanks for looking it over and handling the compilation, @ActuallyACat!

Issue #552, opened 2020-06-17
Missing font-weight styles for Kaizen Site strong elements


didoesdigital:
**Current:**

![image](https://user-images.githubusercontent.com/2476974/84864500-1c189480-b0ba-11ea-9bb7-ac5babdf59d9.png)

**Desired:**

`<strong>` elements should have a heavier weight:

![image](https://user-images.githubusercontent.com/2476974/84864654-508c5080-b0ba-11ea-88ab-b3f80c8112bd.png)

```
/* Paragraphs / Body (Bold) */

font-family: Inter;
font-style: normal;
font-weight: 600;
font-size: 16px;
line-height: 24px;
```

PR #560, opened 2020-06-19
fix: correct font weights for legacy ca-type paragraph bold mixins

mbylstra:
These were incorrectly set to `400` which as the same as the regular styles.

FYI, the reason that there are two copies of these mixins (`packages/component-library/styles/type.scss` and `packages/deprecated-component-library-helpers/styles/type.scss`), is that originally we deleted the ones in `packages/component-library/styles/type.scss` but ran into a problem where a consuming package was using an older version of a package that was expecting them due to peer dependencies (it's confusing). 

Once all consuming repos and packages have been upgraded, we will be able to delete `packages/component-library/styles/type.scss` which should be considered deprecated+++

### DIFF
--- a/packages/component-library/styles/type.scss
+++ b/packages/component-library/styles/type.scss
@@ -313,7 +313,7 @@ $ca-inter-locale-he-font-family: "Inter", Tahoma, sans-serif;
 }
 
 @mixin ca-type-inter-body-bold($size: 16, $args...) {
-  @include ca-type-inter($size: $size, $weight: 400, $args...);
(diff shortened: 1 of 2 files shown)


mbylstra:
...waiting on confirmation on whether the correct weight is 600 or 700

mbylstra:
Confirmed that the correct weight is 600 as defined in the Figma UI KIT (more details here: https://github.com/cultureamp/kaizen-design-system/issues/552). I think this will not fix that issue specifically (that seems to be related to a style for the <strong> tag not being defined, or something else, as the Kaizen site is not using these bold mixins)

PR #561, opened 2020-06-19
docs: define weight for .md-strong <strong> tags in Kaizen site

mbylstra:
This should hopefully fix https://github.com/cultureamp/kaizen-design-system/issues/552. I'm just waiting on the build to confirm.

(aside: I think there is a cooler way to link PRs to issues by putting the issue number in the PR title. I will look into that)


mbylstra:
Looks like it worked:

PR #563, opened 2020-06-19
ci: Added DangerJS to CI to find missing dependencies

ActuallyACat:
## About 
Sass imports that aren't in the package's dependency are resolved in the monorepo, but fail when packaged and distributed. We've run into scenarios when splitting out components into separate packages and modifying existing draft components this becomes apparent. 

Dangerfile is a tool that can help us automate checks that are usually done in PR. 

## What this does
The script checks for changes in Sass files and then finds the closest `package.json` and it's changes. The imports in the sass file are then matched to the changes in the package, and if there's an import in Sass that is missing in package, CI fails. 

## So... How do I know this will work?
There's a couple of options: 

**Check locally**
1. Branch off `ally/dangerfile`, and commit some code in `*.scss` and/or `package.json`. Find something obvious that should fail. As an example, `draft-packages/illustration/KaizenDraft/Illustration/style.module.scss` is one of our packages that doesn't depend on tokens. Add an import to that file`@import "~@kaizen/design-tokens";`. Do not add this as a dependency of illustration - we want to see this fail. 
2. Commit this change
3. Run `yarn danger local`. You should see the following:
`This pull request is missing a dependency for @kaizen/design-tokens in draft-packages/illustration/package.json`

**Create a PR**
Ally finish this section

**TODO** 

- [ ] Set up token in Buildkite
- [ ] Finish details on this PR


ActuallyACat:
Closing. Will use one of the out-of-the-box rules that ESLint makes available: https://github.com/cultureamp/kaizen-design-system/pull/621

PR #564, opened 2020-06-19
feat: Add tokens for breakpoints

sebpearce:
## Changes

- Adds layout tokens to represent global layout breakpoints.
- Updates TitleBlockZen (and stories) to use these breakpoints.
- Adds a content container placeholder to the TitleBlockZen story so that the breakpoints can be seen functioning correctly in two places at once.

## Background

Having these breakpoints is important for consistency across separate components relating to page layout – for example, in the work that PEX is currently doing.

One of the tricky parts of naming breakpoints is that although people tend to name them things like "tablet" and "desktop", or "small" and "medium", these words refer to _ranges between breakpoints_ and not breakpoints themselves.

I tried to have a balance of:
- simple
- intuitive (takes advantage of what's normal/common)
- clear

I also considered other naming conventions, such as:
```
"breakpoints": {
  "smallToMedium": "768px",
  "mediumToLarge": "1080px"
}
```
☝️  This has the advantage of not being tied to traditional names for device types, but it has two drawbacks:
1. If you see one of these in code, it's not clear if what the full system is – is there an `extraSmall` or an `extraLarge`? People are already familiar with the mobile-tablet-desktop framework and can more easily guess what's available by seeing one of these words.
2. The wording of "smallToMedium" etc. could be confusing because they could be interpreted as referring to a range rather than a single point, whereas with something like "desktopMinimum" and "tabletMinimum", it's obvious that these refer to single point.


ActuallyACat:
My position on device-agnostic naming is clear at this point. Probably moot given that it is actually coupled to devices. Please get additional thoughts from developers (in this PR), because this poll has been a big driver for my strong feelings on this and the deviation has not been documented anywhere. https://cultureamp.slack.com/archives/C2ZNME08P/p1583275901059800 
This decision should be properly documented in this pull request for future reference.

--- 
I like that you've considered clarity and intuition here. Overall I like the direction. 

However, I don't think the `minimum` suffix gives the clarity you're hoping for. If we're talking about "mobile-tablet-desktop framework" then "mobile", "tablet", and "desktop" are the accepted ranges and it's pretty well understood how to work with them. For example, when targetting _just_ tablets, `@media (min-width: $tablet) and (max-width: $desktop - 1)` feels like something most web developers would understand.

It follows that this feels verbose: 
```
@media (min-width: $kz-layout-breakpoints-tablet-minimum) and 
(max-width: $kz-layout-breakpoints-desktop-minimum - 1) { /* stuff*/ }
```
Building mobile first also makes these feel redundant: 
```
.someClass { 
   background: red: 
   @media (min-width: #{$kz-layout-breakpoints-tablet-minimum}) {
      background: pink;
   }
   @media (min-width: #{$kz-layout-breakpoints-desktop-minimum}) {
      background: blue;
   }
}
```

Another consideration is how we would use these breakpoints components. For example, it feels extra verbose for `Box`: `<Box m={{ tabletMinimum: 1, desktopMinimum: 4 }} />`.

sebpearce:
The poll in question for readers:

image

sebpearce:
I'll try and avoid sentences to keep it TLDR:

- The `minimum` thing was partially influenced by the reasoning here: https://www.freecodecamp.org/news/the-100-correct-way-to-do-css-breakpoints-88d6a5ba1862/
- I haven't yet found an accurate way to use words like `small` and `medium` or `tablet` and `desktop`, which actually refer to **ranges between breakpoints**, not breakpoints.
- The clarity problem with using a _range_ word to refer to a _single point value_ is that it's unclear which side of the range the value is on. Is `medium` the minimum value in the `medium` range, or is it the maximum? Given that a lot of our codebase is desktop-first, it's conceivable to me that someone might think "OK this is our desktop view, _and when we switch to a tablet view_, I'll use the `tablet` media query"... they would be assuming that they're accessing the *maximum* value in the tablet range, but the token will be giving them the minimum. I.e. what they really want is `max-width: $desktop-minimum - 1px` and not `max-width: $tablet-minimum`.
- If you strip away the `-minimum`, `max-width: $tablet` looks like it could be correct even though it isn't, whereas `max-width: $tablet-minimum` obviously looks wrong, as it should.
- I'll check on Slack. If the majority of CA engineers who answer think this level of clarity is unimportant, I'll remove `-minimum`.
- Agree that building mobile-first would make the `-minimum` thing redundant! I just imagine that for a lot of pre-existing code where people want to get the breakpoints from here, they're not going to totally restructure their code to make it mobile-first. Realistically that'll only happen on new pieces of work (or heavily uplifted).
- I'm having a hard time getting feedback on this, but I'll keep pushing for it.
- `<Box m={{ tabletMinimum: 1, desktopMinimum: 4 }} />` – in this case, I would assume as an engineer that the values here actually **do** refer to ranges rather than single points, so "minimum" would actually be confusing here and I wouldn't use for these keys.

sebpearce:
Our poll yielded a majority preference for `medium` and `large`, so we're going with those:
image

PR #565, opened 2020-06-22
fix: Specify box-sizing so layout is consistent without depending on the consuming app

cset00:
Using titleblock in a new repo causes some layout inconsistencies due to box-sizing. Ideally the component should not need to rely on the consuming app's style to have consistent layout.

Local verification using yarn link confirms the layout inconsistency issue is fixed.


sebpearce:
LGTM. Just have to make sure no consumers are depending on the lack of a `box-sizing` value. A bit of QA and visual regressions will hopefully surface that sort of thing.

PR #567, opened 2020-06-22
feat: Add automation ids to modal container and actions

dougmacknz:
Allows us to put automation ids on all types of modals. It will add the automation id to the container and then on each button with a `-action-[index]` suffix.

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/ModalFooter.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/ModalFooter.tsx
@@ -14,12 +14,13 @@ interface Props {
     disabled?: boolean
   }>
   readonly appearance?: "primary" | "destructive"
+  readonly automationId?: string
 }
 
 type ModalFooter = React.FunctionComponent<Props>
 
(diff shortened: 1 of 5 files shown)


draft-packages/modal/KaizenDraft/Modal/Primitives/ModalFooter.tsx
```diff
@@ -35,6 +36,7 @@ const ModalFooter: ModalFooter = props => {
               destructive={index === 0 && appearance === "destructive"}
               secondary={index > 0}
               disabled={a.disabled}
+              automationId={`${automationId}-action-${index}`}
```
ActuallyACat, line 39: Question, and not a blocker - but I'm curious about the name. I guess since you're setting the data attributes on the wrapping div around this so you need a different name, but I'm not sure if we have a precedent/pattern for the convention yet. Why `action`?

dougmacknz: I used `action` just because that's what they're called in the ModalFooter props: `actions`. Would you prefer `button`?

PR #573, opened 2020-06-24
docs: add Menu docs

didoesdigital:
- Add docs for Menu

Branch preview: https://dev.cultureamp.design/di/add-menu-page/components/menu/


draft-packages/menu/KaizenDraft/Menu/README.mdx
```diff
@@ -0,0 +1,63 @@
+---
+title: Menu
+navTitle: Menu
+summaryParagraph: A menu contains links to places or button actions. It does NOT show a selected item at all and the menu button text never changes.
```
mbylstra, line 4: Nice - the most important information is super clear and prominent!

mbylstra: The only kind of exception/edge based on today's discussion: "menu button text never changes". If the page is refreshed, and then the button text changes, did the button text actually change, or is it just a different page with a different button 🤔. This is relevant to whether or not we did actually implement the "heading 1 title block menu" as a Menu instead of a Select, or is this just an edge case where we're willing to break the rule for expediency? We can continue to discuss this outside this PR, but thought it's worth mentioning.

draft-packages/menu/KaizenDraft/Menu/README.mdx
```diff
@@ -0,0 +1,63 @@
...
+<WhenToUse>
+
+*   Use a menu when there are link and button options available that are not used enough to show by themselves.
+
+</WhenToUse>
+<WhenNotToUse>
+
+*   When an option from the list becomes "selected", use a [Select](/components/select) instead.
+*   If there are only 2 list items, show them immediately inline instead of using a menu.
+*   If you think you need 10 options, reconsider information architecture, page structure, and task flows to avoid including that many items in a menu.
```
mbylstra, line 49: I only just read this, which gives more weight to the idea that the "bug" where the dropdown breaks if it's taller than the viewport is not really a bug (definitely not worth worrying about).

PR #577, opened 2020-06-24
fix: upgrade to Elm 0.19.1

mbylstra:
This PR upgrades from 0.19.0 to 0.19.1. The main change here is that the Elm compiler is much stricter with how the elm module names like `module ElmStories.TagStories exposing (main)` match the directory structure. We can no longer use the existing `docs/Thing.stories.elm` structure, so each package has a new directory named `ElmStories` at the root of the package (it can't be put inside `docs` any more). Also this uses an upgraded version of `cultureamp/elm-storybook` that's compatible with Elm 0.19.1 (and requires a bit more boilerplate to use)

This is waiting on https://github.com/cultureamp/kaizen-design-system/pull/990 to be merged first.

Here is the Chromatic build: https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=234. I'm pretty sure the changes there are just the baseline snapshots are a little behind master.


mbylstra:
Closed in favor of https://github.com/cultureamp/kaizen-design-system/pull/1038

PR #579, opened 2020-06-25
feat: Added support for vertical tabs

ActuallyACat:
## About
This PR adds support for vertical tabs to the existing tab component. For backwards compatibility, the default direction will be row, to match the existing usage of tabs.

Visuals: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A28887

- [x] Clean up work
- [x] Design review
- [x] debug focus visible
- [x] is badge in scope?


ActuallyACat:
Comments from review with Di:

- disabled isn't actually disabled for anchor tags
- change to default cursor on disabled
- RTL support
- columns right align text
- row reverse the order of the tabs
- put a fake sidebar in the story to make the RTL stuff clear

ActuallyACat:
Comments from review with Di:

- disabled isn't actually disabled for anchor tags
- change to default cursor on disabled
- RTL support
- columns right align text
- row reverse the order of the tabs
- put a fake sidebar in the story to make the RTL stuff clear

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
```diff
@@ -69,18 +60,34 @@
...
   }
-
-  &:hover {
-    @extend %tabHover;
-  }
 }
 
-.tab + .tab {
-  @include ca-margin($start: $ca-grid * 0.5);
+.row {
```
mbylstra, line 65: the naming threw me off a bit. I thought it was the entire row for row-style-tabs, but it's actually a single tab. Suggest `rowTab` and `columnTab`?

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
```diff
@@ -69,18 +60,34 @@
...
+  min-width: 4.5rem - ($ca-grid * 0.75) * 2; // give tabs with short text some breathing room
+
+  :global(.js-focus-visible) &:global(.focus-visible):after {
+    border-radius: $kz-border-borderless-border-radius
+      $kz-border-borderless-border-radius 0 0;
+  }
 }
 
-.activeTab {
+.activeTabRow {
```
mbylstra, line 87: If `.row` is renamed to `.rowTab` then this would be better as `activeRowTab`.

Otherwise, here's a completely different suggestion.  I think you've got a bit of a BEM thing going one where the last word is the "modifier", but that's hard to grok that without any separators - without it you tend to read it as plain english, which is where it gets confusing. I believe `_`'s are valid - maybe that could be used as a separator?
So, what about `.tab_row` and `.tab_column`, `.activeTab_row`, `.activeTab_column`. I find that aesthetically ugly (that really doesn't matter), but a lot easier to grok.

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
```diff
@@ -99,17 +106,67 @@
...
+  justify-content: left;
+  padding-left: $kz-spacing-sm;
+  border-radius: $kz-border-solid-border-radius;
+
+  :global(.js-focus-visible) &:global(.focus-visible):after {
+    border-radius: $kz-border-borderless-border-radius;
+    z-index: 1; // pop the border out so sibling tabs don't overlap
+  }
+
+  &.rtl {
```
mbylstra, line 145: @sebpearce 's work on the the RTL stuff, but I believe the recommended a approach is to use a html attribute selector like `[dir="rtl"]` like in https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/split-button/KaizenDraft/SplitButton/styles.scss#L151". Otherwise you'd have to prop drill the rtl prop down every component, which is impractical, and would have the likely outcome that no one would do it.

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
```diff
@@ -99,17 +106,67 @@
...
+    justify-content: right;
```
mbylstra, line 146: Actually, it's better to just use `justify-content: start` and then there's no need to explicitly support RTL - it comes for free!
image

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
sebpearce: Just noticed this in the GitHub Slack updates 😎  Yeah, the RTL prop is just for stuff that can't be done in CSS, like using a different icon. `dir="rtl"` works even if they don't use the prop 👌

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
sebpearce: `justify-content: flex-start`

ActuallyACat: TIL - thanks!

mbylstra: Spoke to @sebpearce - `justify-content: flex-start` is better as `justify-content: start` is really new (I thought it was the other way around) and might not work in IE11.

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
ActuallyACat: I'd like to eventually use `context` for stuff like this.

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
ActuallyACat: Yup, managing these modifiers are going to start grinding away at me until we're a bit more consistent. 

I don't think i've made this any _worse_, but I definitely haven't made it _better_. I didn't change it too much from what it was - it was previously `stateThing` (like`activeTab`, `disabledTab`) now it's `thingState`, like `rowTabDisabled` )

draft-packages/tabs/KaizenDraft/Tabs/styles.scss
ActuallyACat: I'm kinda in favour of the components having an optional `dir` prop, so they can override ltr/rtl behaviour for specific components if needed. I've kept the `dir` prop here, but this would also work if `dir` wasn't passed to the component and instead the layout set the direction.

ActuallyACat: > but this would also work if dir wasn't passed to the component and instead the layout set the direction.

This is incorrect. I just tested it. Without touching any CSS: 
Layout => RTL = true
Tabs => RTL = false
Screen Shot 2020-06-29 at 4 09 40 pm

Layout => RTL = true
Tabs => RTL = true
This is what it _should_ look like for RTL languages
Screen Shot 2020-06-29 at 4 09 29 pm

sebpearce: Having the base `[dir="rtl"]` selector as a safety net along with an RTL override class sounds good to me 👍

sebpearce: BTW, in the past people have called that prop `textDirection` which is a bit more intuitive to people unfamiliar with the `dir` prop.

ActuallyACat: Ah 😬 `Box` uses `rtl?: boolean`. I'm going to switch to `textDirection?: "ltr" | "rtl"` since that has way more usages.

PR #582, opened 2020-06-26
feat: Add layout token for nav bar height

sebpearce:
## Changes

• Add a new layout token for navigation bar height

## Background

This is currently in more than one place, and could conceivably pop up in more places (e.g. setting the top margin on a page underneath the nav bar) so it would be good to have a single implementation of it.

### DIFF
--- a/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/components/Badge.module.scss
+++ b/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/components/Badge.module.scss
@@ -1,7 +1,7 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/component-library/styles/color";
-@import "~@kaizen/component-library/styles/type";
 @import "~@kaizen/component-library/styles/layout";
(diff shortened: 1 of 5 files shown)


sampnorris:
🙏🏻

draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/components/Badge.module.scss
```diff
@@ -1,7 +1,7 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/component-library/styles/color";
-@import "~@kaizen/component-library/styles/type";
 @import "~@kaizen/component-library/styles/layout";
+@import "~@kaizen/component-library/styles/type";
```
sebpearce, line 4: While we're here, this `type.scss` is actually deprecated.

For new typography stuff, there's `~@kaizen/design-tokens/sass/typography`.

For `$ca-grid`, there's `~@kaizen/component-library/styles/grid`.

I see there are things like `$ca-weight-medium` in here that still depend on `type` so no need to get rid of them now, but just an FYI to spread awareness 👍

PR #586, opened 2020-06-26
fix: avoid reliance on Elm18Compatible in UserInteractions/Keyboard

mbylstra:
This upgrades some code in `UserInteractions/Keyboard.elm` to Elm 0.19 so that it no longer relies on a Elm 0.18 shim. This is part of a bigger effort to remove all usages of cultureamp/elm-upgrade-shims (https://github.com/cultureamp/kaizen-design-system/pull/577)

Ideally this would be using `event.key` (which is a string) instead of `event.keyCode` (which is an int), as that is the official recommendation for Elm 0.19, but that's out of scope for this PR.

### DIFF
--- a/draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
+++ b/draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
@@ -1,6 +1,7 @@
 module KaizenDraft.UserInteractions.Keyboard exposing (Key(..), decoder, subscriptions)
 
 import Browser.Events
+import Html.Events
 import Json.Decode as Decode exposing (Decoder)
 import KaizenDraft.UserInteractions.KeyCodes as KeyCodes
 
@@ -49,7 +50,7 @@ keyCodeToKey keyCode =
 
 keyDownDecoder : Decoder Key
 keyDownDecoder =
-    Decode.field "keyCode" Decode.int |> Decode.map keyCodeToKey
+    Decode.map keyCodeToKey Html.Events.keyCode
 
 
 subscriptions : (Key -> msg) -> Sub msg
--- a/draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
+++ b/draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
@@ -60,5 +60,6 @@ subscriptions toMsg =
 
 {-| deprecated: this has been left in to avoid a breaking API change. This function is no longer used in this module. The name is confusing as it's not actually a decoder (it's just a transformation function)
 -}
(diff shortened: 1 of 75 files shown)


draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
```diff
@@ -46,6 +47,17 @@ decoder keyCode =
...
+
 subscriptions : (Key -> msg) -> Sub msg
 subscriptions toMsg =
-    Keyboard18.downs (decoder >> toMsg)
+    Browser.Events.onKeyDown (keyDownDecoder |> Decode.map toMsg)
+
+
+{-| deprecated: this has been left in to avoid a breaking API change. This function is no longer used in this module. The name is confusing as it's not actually a decoder (it's just a transformation function)
+-}
+decoder =
```
Confidenceman02, line 62: Could we type def this?

draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
```diff
@@ -46,6 +47,17 @@ decoder keyCode =
         Other
 
 
+keyDownDecoder : Decoder Key
+keyDownDecoder =
```
Confidenceman02, line 51: We can use Html.Events.keyCode instead!

mbylstra: sweet, didn't know that was there!

Confidenceman02: There is even a mention about the deprecation of keyCode which is pretty neat.

mbylstra: It should mention this in the docs for `Browser.Events.onKeyDown` though! (it tells you to write your own decoder)

draft-packages/user-interactions/KaizenDraft/UserInteractions/Keyboard.elm
```diff
@@ -46,6 +48,18 @@ decoder keyCode =
...
+    Decode.map keyCodeToKey Html.Events.keyCode
```
Confidenceman02, line 53: This is nit picky, could we expose `keyCode` please? :)

mbylstra: This was intentional. I wanted to be clear that it's not something from this module.

Confidenceman02: Ah!!!! no probs!

PR #587, opened 2020-06-26
docs: add Filter Drawer docs

didoesdigital:
First draft docs for Filter Drawer component

PR #590, opened 2020-06-26
feat: @kaizen/draft-filter-drawer (React, new package)

mbylstra:
# FilterDrawer

A filter drawer contains filters. It controls filtering of an entire page.

filterdrawer

### Docs 
(new docs introduced in this PR, which was rebased on this PR: https://github.com/cultureamp/kaizen-design-system/pull/587/files)
https://dev.cultureamp.design/dst/filter-drawer/components/filter-drawer/
https://dev.cultureamp.design/dst/filter-drawer/guidelines/filtering/

### Figma design
https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=9473%3A0

### Storybook stories
https://dev.cultureamp.design/dst/filter-drawer/storybook-static/?path=/story/filterdrawer-react--default-story

### Chromatic build
https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=193

### Accessibility

- the button has `aria-expanded` true|false, which causes screen readers to announce to the user whether the dropdown is open or not when they click the button
- the button also has `aria-controls` linked to the drawer element. This has no effect in VoiceOver, but in JAWS it will anounce a keyboard shortcut that can be used to navigate to the open drawer.

### IE11
It's a little fugly, but usable. (ignore the missing pixels, that's a VM issue):
image


draft-packages/filter-drawer/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@kaizen/draft-filter-drawer",
+  "version": "1.4.1",
```
ActuallyACat, line 3: This is my bad since it's in the generator, but this should start at v1.0.0

draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
```diff
@@ -0,0 +1,98 @@
...
+  hideDropdown: () => void
+
+  /**
+   * If at least one filter is enabled, then a filter badge will show.
+   * The filter button badge shows the number of filter fields with applied filters.
+   */
+  numberOfFiltersEnabled?: number
+}
+
+const dropdownId = "filter-drawer-dropdown"
```
ActuallyACat, line 51: Consider when there are multiple dropdowns on the same page - this would result in multiple identical ids. This will potentially mess with the ariaControls in this component as it won't know which component the id is in reference to. 

I would suggest looking into something like [cuid](https://github.com/ericelliott/cuid) to generate unique ids for this component.

draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
```diff
@@ -0,0 +1,98 @@
...
+export interface FilterDrawerProps {
+  /**
+   * The text that goes inside the filter button
+   */
+  labelText: string
+
+  /**
+   * Metadata shows relevant info about the content being filtered
+   */
+  metadata?: React.ReactNode
```
ActuallyACat, line 19: Nit/ I wasn't quite sure what metadata referred to on first read through. It wasn't until I looked at the story it clicked that this is the stuff that displays beside the filter drawer. That being said, I can't think of a better name 🤷

mbylstra: I got the name from the Figma, and I think it makes sense to use the same anme.

draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
mbylstra: That's a good idea.

draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
ActuallyACat: That sounds reasonable to me

draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
mbylstra: Good call - this was a lazy oversight on my behalf. I decided to go with having the consumer provide the id. I think there is some benefit in human readable ids, or at least letting the consumer decide.

mbylstra:
Sweet, thanks!

PR #591, opened 2020-06-29
fix: reverse badge colour for admin theme

sampnorris:
- transition background color on theme change
- added story for theme change


sampnorris:
Screen Shot 2020-06-29 at 3 43 23 pm

PR #594, opened 2020-06-29
feat: Added card package

ActuallyACat:
# About
> Cards are a structural component to contain primary content.

The space _around_ card and the space _within_ card is without margins or padding, so that the consumer has more flexibility to use this component however they need. More atomic/building block components, like `Box`, can be used to space content within `Card` if needed. I've added an example to Storybook with `Box` and `Paragraph`.

## References
Docs: https://cultureamp.design/components/card/
Figma: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A14085

## The component
https://dev.cultureamp.design/ally/card/storybook-static/iframe.html?id=card-react--default-story

## Todo
- [x] Design review
- [x] Does the card need to have formatting in children by default?
- [x] Discuss + understand use cases


draft-packages/card/KaizenDraft/Card/Card.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const Card = ({
+  children,
+  tag = "div",
+  classNameAndIHaveSpokenToDST,
+  ...otherProps
+}: CardProps) => {
+  const className =
+    (classNameAndIHaveSpokenToDST ? classNameAndIHaveSpokenToDST : "") +
+    " " +
+    styles.wrapper
```
ActuallyACat, line 28: No point adding classnames as a dep if this is all it's needed for.

draft-packages/card/KaizenDraft/Card/Card.tsx
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Not recommended. A short-circuit for overriding styles in a pinch
+   * @default ""
+   */
+  classNameAndIHaveSpokenToDST?: string
+
+  /**
+   * HTML elements that are allowed on Card.
+   */
+  tag?: "pre" | "div" | "article" | "header" | "main" | "section"
```
ActuallyACat, line 16: These are the only tags I'm thinking we need right now. It will be easy to expand on this list later

ActuallyACat:
Review with Marc 
* We need a Divider to be used within Cards. I'll create a separate PR for this Use case: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=9487%3A263 
* I forgot the white background on card 😅 Now fixed.

sebpearce:
Pending design sign-off, looks good to me 👌  The lack of built-in margin/padding is the key bit I think.

PR #595, opened 2020-06-29
fix: Add clear button to React Select implementation

makenosound:
The implementation of draft-select rendered `null` for the `ClearIndicator` component. This meant you could use `isClearable` as a prop but it would not actually render the button to allow you to clear it.

clear


makenosound:
image

Updated with the correct icon 😬 . Mind checking again, @mbylstra?

PR #604, opened 2020-07-01
fix: Changed box-sizing in tag so the height matches Figma

ActuallyACat:
## About 
This PR (at this stage) fixes the height of the Tag and removes focus states for non-dismissible tags.

I've been looking at this so long that i can't tell whether it's now aligned correctly: 
Screen Shot 2020-07-02 at 3 28 49 pm

Since I had the time, I also increased the tap target of the dismissable icon. It's now sitting at 26.4px x 26.4px (previously 14.4px x 14.4px). What this means is you can now dismiss the tag by tapping/clicking anywhere within the black box in the screenshot.

Still not quite where it should be, but any larger and it could overlap with the targets of components sitting next to it.
Screen Shot 2020-07-02 at 3 21 17 pm

Link: https://dev.cultureamp.design/ally/tag-fix/storybook-static/iframe.html?id=tag-react--default-medium-kaizen-site-demo

Not sure if these things are in scope: 
- [x] Focus ring on the dismissable button
- [x] UI Kit doesn't have border on hover state - should I remove?

### DIFF
--- a/draft-packages/tag/KaizenDraft/Tag/Tag.scss
+++ b/draft-packages/tag/KaizenDraft/Tag/Tag.scss
@@ -42,6 +42,8 @@ $small: $ca-grid;
 }
 
 .dismissIcon {
+  position: relative;
+  display: flex;
   padding: $kz-spacing-xs;
   margin-top: -$kz-spacing-xs;
   margin-right: -0.5375rem;
@@ -58,6 +60,24 @@ $small: $ca-grid;
   &:active {
     color: $kz-color-wisteria-800;
   }
+
+  svg {
+    position: relative;
+  }
+}
+
+.background {
+  // As our icons are only one-tone, this span places
+  // a white background under the dismissible icon to
+  // give the illusion of a background colour that is
+  // different to the rest of the tag
+  position: absolute;
+  display: inline-block;
+  width: $kz-spacing-xs;
+  height: $kz-spacing-xs;
+  background-color: white;
+  left: 10px;
+  top: 10px;
 }
 
 .truncate {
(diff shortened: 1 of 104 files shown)


ActuallyACat:
Since I had the time, I also increased the tap target of the dismissable icon. It's now sitting at 26.4px x 26.4px (previously 14.4px x 14.4px). Still not quite where it should be, but any larger and it could overlap with the targets of components sitting next to it. 

Screen Shot 2020-07-02 at 3 21 17 pm

didoesdigital:
Design review: we can ship these improvements, and follow up any additional improvements. The increased touch targets are fantastic, great stuff 👏

Example of follow ups, there's a little bit of weirdness with zoom levels. 100% zoom:
 
image

110%:

image

Compared to Figma:

image

(The text is a teensy bit lower in the 100% zoom shot)

draft-packages/tag/KaizenDraft/Tag/Tag.scss
```diff
@@ -73,8 +73,8 @@ $small: $ca-grid;
   // different to the rest of the tag
   position: absolute;
   display: inline-block;
-  width: $kz-spacing-xs;
-  height: $kz-spacing-xs;
+  width: 8px;
+  height: 8px;
```
didoesdigital, line 77: For optical adjustments like this, I would usually do something like `$kz-spacing-xs + 2px; // optical adjustment`

ActuallyACat: Good call - fixed.

PR #605, opened 2020-07-02
feat: Menu is now left aligned by default

mbylstra:
⚠️ This PR contains a BREAKING CHANGE ⚠️ 

Storybook previews: https://dev.cultureamp.design/dst/menu-align-left-right/storybook-static/?path=/story/menu-react--label-and-icon

#### `align="left"` (this is now the default)
image

#### optional `align="right"`
image

# Changes:

### BREAKING CHANGE: Menu is now left aligned by default 
adds an `align` prop to Menu and MenuDropdown with options `"left"` and `"right"`. Previously rigtht-aligned
was the only option. Now left aligned is an option and it becomes the default. ⚠️ **Consumers must add `align="right"`
to any existing usages to retain identical functionality.** ⚠️ 

### feat: further Zenify the MenuDropdown component
- there is now a small gap between the button and the dropdown
- the dropdown is flush to the left/right of the button

### docs: add story for edge case where button is at bottom of screen
The Menu had some functionality so that it would still appear on screen if the bottom of the dropdown dissapears below
the bottom of the screen. I added this story to make sure I hadn't broken that functionality. This edge case isn't styled perfectly, but it seemed not worth worrying about as it's such an extreme edge case.
https://dev.cultureamp.design/dst/menu-align-left-right/storybook-static/?path=/story/menu-react--label-and-icon-bottom

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
@@ -9,6 +9,7 @@ type MenuState = {
 }
 
 export type MenuProps = {
+  align?: "left" | "right"
   button: React.ReactElement<ButtonProps>
   menuVisible?: boolean
   automationId?: string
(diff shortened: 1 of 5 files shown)


sebpearce:
Love it. Checked in Storybook and looks great. Nice mobile work too 👌

PR #606, opened 2020-07-02
feat: Add missing paragraphBodyBold typography token

makenosound:
Adds a `paragraphBodyBold` token to match the small/extra-small versions.

I’ve run the build script locally to check the output and diff gives me these additional lines (no removals):

Sass:

```scss
$kz-typography-paragraph-body-bold-font-family: "Inter", "Noto Sans", Helvetica,
  Arial, sans-serif;
$kz-typography-paragraph-body-bold-font-weight: 600;
$kz-typography-paragraph-body-bold-font-size: 1rem;
$kz-typography-paragraph-body-bold-line-height: 1.5rem;
$kz-typography-paragraph-body-bold-letter-spacing: normal;
```

Less:

```less
@kz-typography-paragraph-body-bold-font-family: "Inter", "Noto Sans", Helvetica,
  Arial, sans-serif;
@kz-typography-paragraph-body-bold-font-weight: 600;
@kz-typography-paragraph-body-bold-font-size: 1rem;
@kz-typography-paragraph-body-bold-line-height: 1.5rem;
@kz-typography-paragraph-body-bold-letter-spacing: normal;
```

### DIFF
--- a/packages/design-tokens/tokens/typography.json
+++ b/packages/design-tokens/tokens/typography.json
@@ -106,23 +106,9 @@
         "lineHeight": "1.5rem",
         "letterSpacing": "normal"
       },
-      "paragraphBodyBold": {
-        "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
-        "fontWeight": "600",
-        "fontSize": "1rem",
-        "lineHeight": "1.5rem",
-        "letterSpacing": "normal"
-      },
       "paragraphSmall": {
         "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
         "fontWeight": "400",
-        "fontSize": "0.875rem",
-        "lineHeight": "1.125rem",
-        "letterSpacing": "normal"
-      },
-      "paragraphSmallBold": {
-        "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
-        "fontWeight": "600",
         "fontSize": "0.875rem",
         "lineHeight": "1.125rem",
         "letterSpacing": "normal"
@@ -134,12 +120,8 @@
         "lineHeight": "1.125rem",
         "letterSpacing": "normal"
       },
-      "paragraphExtraSmallBold": {
-        "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
-        "fontWeight": "600",
-        "fontSize": "0.75rem",
-        "lineHeight": "1.125rem",
-        "letterSpacing": "normal"
+      "paragraphBold": {
+        "fontWeight": "600"
       },
       "buttonPrimary": {
         "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
--- a/packages/design-tokens/tokens/typography.json
+++ b/packages/design-tokens/tokens/typography.json
@@ -106,6 +106,13 @@
         "lineHeight": "1.5rem",
         "letterSpacing": "normal"
       },
+      "paragraphBodyBold": {
+        "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
+        "fontWeight": "600",
+        "fontSize": "1rem",
+        "lineHeight": "1.5rem",
+        "letterSpacing": "normal"
+      },
       "paragraphSmall": {
         "fontFamily": "\"Inter\", \"Noto Sans\", Helvetica, Arial, sans-serif;",
         "fontWeight": "400",
--- a/packages/design-tokens/tokens/typography.ts
+++ b/packages/design-tokens/tokens/typography.ts
@@ -1,15 +1,21 @@
 import * as typographyTokens from "./typography.json"
 
+type TypeDefinition = {
+  fontFamily: string
+  fontWeight: string
+  fontSize: string
+  lineHeight: string
+  letterSpacing: string
+}
+
+type TypeWeightVariant = {
+  fontWeight: string
+}
+
 export interface Typography {
   kz: {
     typography: {
-      [key: string]: {
-        fontFamily: string
-        fontWeight: string
-        fontSize: string
-        lineHeight: string
-        letterSpacing: string
-      }
+      [key: string]: TypeDefinition | TypeWeightVariant
     }
   }
 }


sebpearce:
@didoesdigital @makenosound Somehow I missed the conversation around this yesterday, but a few weeks ago I was chatting to someone and agreeing that since all of these rules are the same except for the font weight, rather than create an entirely new style just for bold paragraph text (which we don't do for other type styles), it would probably make more sense to just include an extra variable on the type styles that need a bold weight, like:

```
...
$kz-typography-paragraph-body-font-weight: 400;
$kz-typography-paragraph-body-bold-font-weight: 600;
```

I don't have strong opinions on this – just thought it would save engineers the hassle of duplicating all these double rules for bold text. Throwing it out there in case you find this persuasive.

makenosound:
Ah, yes! I think that was me 🙊 (too many things happening at the moment). I do think that, unless we think there are likely to be modifiers to spacing etc (and I can only imagine that’s letter-spacing) we’d be better having small modifiers for weight only.

didoesdigital:
Great point, @sebpearce . 

When I was using the bold token for the Badge component, it felt strange (inconsistent and potentially error prone with the hard-to-notice differences) to write:

```
  font-family: $kz-typography-paragraph-extra-small-font-family; // not bold
  font-size: $kz-typography-paragraph-extra-small-font-size; // not bold
  font-weight: $kz-typography-paragraph-extra-small-bold-font-weight; // bold
```

Do we want that?

I don't feel strongly either way, but if we take the approach of fewer tokens, we'll need to remove the extra tokens for `extra-small-bold` and `small-bold`, and update their use in the Badge component.

makenosound:
What do y’all reckon @didoesdigital / @sebpearce? I’m happy to do the tidy up if we have a decision.

sebpearce:
Also a good point, although I think the issue of it not being easily visible whether the word `bold` is in there probably remains in either case.

For engineers I would imagine often the way people would use these would be something like:
```
.myClass & {
  font-family: $kz-typography-paragraph-body-font-family;
  font-weight: $kz-typography-paragraph-body-font-weight;
  font-size: $kz-typography-paragraph-body-font-size;
  line-height: $kz-typography-paragraph-body-line-height;
  letter-spacing: $kz-typography-paragraph-body-letter-spacing;

  strong {
    font-weight: $kz-typography-paragraph-body-bold-font-weight;
  }
}
```
...which would make it clearer that there's a difference.

makenosound:
An alternative might be to name these explicitly as variants:;

```
font-family: $kz-typography-paragraph-extra-small-font-family;
font-size: $kz-typography-paragraph-extra-small-font-size;
font-weight: $kz-typography-paragraph-extra-small-font-weight-variant-bold;
```

sebpearce:
I like that idea if we did go down that route @makenosound 👍

sebpearce:
This discussion also ties in to the question of whether *links* are separate type styles too – I know they are in the UI Kit (for some type styles), but that generally isn't how they work in code...

makenosound:
Yeah, I think it probably does make some practical sense for links. Usually you’re going to want to have them be things that only vary a couple of properties to the right values, rather than overriding all of family/weight etc.

sebpearce:
> I do think that, unless we think there are likely to be modifiers to spacing etc (and I can only imagine that’s letter-spacing) we’d be better having small modifiers for weight only.

☝️  This seems to me to be the most relevant argument here in favour of separate type styles. If we go down the `-variant` road, we're stuck with the current letter-spacing etc. for bold text.

didoesdigital:
For reference, some other design systems with typography tokens:

- https://www.lightningdesignsystem.com/design-tokens/#category-font `$font-weight-bold`
- https://kalo.design/foundations/design-tokens/ `fontWeightBold`
- https://github.com/opentable/design-tokens/blob/master/OTKit/otkit-typography-desktop/token.yml `font-weight-bold`
- https://github.com/envato/foundation-design-system-tokens/tree/master/packages/typography `font-weight--bold`
- https://github.com/seek-oss/braid-design-system/blob/master/lib/themes/baseTokens/seekAnz.ts#L44-L48 `fontWeight` `strong`

Some design systems with typography don't mention weight e.g. https://github.com/bitnami/hex/blob/master/packages/hex-core/src/variables/_typography.scss

sebpearce:
Thanks @didoesdigital. I notice in (almost?) all of those they have font weights defined individually, which would mean that bold is always 700 for example. If our type styles are such that bold is always 700, we could just do that any have a separate token for it. But my understanding was that _some_ of our type styles use 700 as their bold, and some might use 500 or 600. Having a look now.

sebpearce:
@makenosound Just to close this off, adding a single `paragraph-bold-weight: 600` token sounds like the way to go here. This way it doesn't prevent us from adding a full type style for bold paragraphs later 👍

makenosound:
Yep, cool. Happy for me to add that and remove the new `somethingBold` tokens (including @didoesdigital’s from the other day)?

sebpearce:
Thanks, yeah that's fine – it looks like 3 paragraph type styles that have the `Bold` suffix.

makenosound:
🤙 Thanks for hashing it out! I will make those tweaks and push it up for review.

sebpearce:
`Design-tokens`' first breaking change!

makenosound:
I’ve made that change now @didoesdigital / @sebpearce. Running `yarn build` locally gives me this diff for `less` and `sass` respectively:

```diff
 @kz-typography-paragraph-small-font-size: 0.875rem;
 @kz-typography-paragraph-small-line-height: 1.125rem;
 @kz-typography-paragraph-small-letter-spacing: normal;
-@kz-typography-paragraph-small-bold-font-family: "Inter", "Noto Sans", Helvetica,
-  Arial, sans-serif;
-@kz-typography-paragraph-small-bold-font-weight: 600;
-@kz-typography-paragraph-small-bold-font-size: 0.875rem;
-@kz-typography-paragraph-small-bold-line-height: 1.125rem;
-@kz-typography-paragraph-small-bold-letter-spacing: normal;
 @kz-typography-paragraph-extra-small-font-family: "Inter", "Noto Sans",
   Helvetica, Arial, sans-serif;
 @kz-typography-paragraph-extra-small-font-weight: 400;
 @kz-typography-paragraph-extra-small-font-size: 0.75rem;
 @kz-typography-paragraph-extra-small-line-height: 1.125rem;
 @kz-typography-paragraph-extra-small-letter-spacing: normal;
-@kz-typography-paragraph-extra-small-bold-font-family: "Inter", "Noto Sans",
-  Helvetica, Arial, sans-serif;
-@kz-typography-paragraph-extra-small-bold-font-weight: 600;
-@kz-typography-paragraph-extra-small-bold-font-size: 0.75rem;
-@kz-typography-paragraph-extra-small-bold-line-height: 1.125rem;
-@kz-typography-paragraph-extra-small-bold-letter-spacing: normal;
+@kz-typography-paragraph-bold-font-weight: 600;
 @kz-typography-button-primary-font-family: "Inter", "Noto Sans", Helvetica,
   Arial, sans-serif;
 @kz-typography-button-primary-font-weight: 700;
```

```diff
 $kz-typography-paragraph-small-font-weight: 400;
 $kz-typography-paragraph-small-font-size: 0.875rem;
 $kz-typography-paragraph-small-line-height: 1.125rem;
 $kz-typography-paragraph-small-letter-spacing: normal;
-$kz-typography-paragraph-small-bold-font-family: "Inter", "Noto Sans", Helvetica,
-  Arial, sans-serif;
-$kz-typography-paragraph-small-bold-font-weight: 600;
-$kz-typography-paragraph-small-bold-font-size: 0.875rem;
-$kz-typography-paragraph-small-bold-line-height: 1.125rem;
-$kz-typography-paragraph-small-bold-letter-spacing: normal;
 $kz-typography-paragraph-extra-small-font-family: "Inter", "Noto Sans",
   Helvetica, Arial, sans-serif;
 $kz-typography-paragraph-extra-small-font-weight: 400;
 $kz-typography-paragraph-extra-small-font-size: 0.75rem;
 $kz-typography-paragraph-extra-small-line-height: 1.125rem;
 $kz-typography-paragraph-extra-small-letter-spacing: normal;
-$kz-typography-paragraph-extra-small-bold-font-family: "Inter", "Noto Sans",
-  Helvetica, Arial, sans-serif;
-$kz-typography-paragraph-extra-small-bold-font-weight: 600;
-$kz-typography-paragraph-extra-small-bold-font-size: 0.75rem;
-$kz-typography-paragraph-extra-small-bold-line-height: 1.125rem;
-$kz-typography-paragraph-extra-small-bold-letter-spacing: normal;
+$kz-typography-paragraph-bold-font-weight: 600;
 $kz-typography-button-primary-font-family: "Inter", "Noto Sans", Helvetica,
   Arial, sans-serif;
 $kz-typography-button-primary-font-weight: 700;
```

makenosound:
@sebpearce No worries! Only thing I need to work out is the conventional PR title for this. The contributing doc had me thinking I needed to add `BREAKING CHANGE: ` to the PR title but it seems unhappy with that. Do you know what the correct incantation is for a breaking change?

sebpearce:
Yep just needs to be in the commit message, not the title – see example here:
https://github.com/cultureamp/kaizen-design-system/blob/master/CONTRIBUTING.md#breaking-changes

makenosound:
Oh, also realised I needed to update the type definitions. Have done so now in a way that preserves the strictness.

sebpearce:
Perfect, just had a look. 👌

makenosound:
@sebpearce Are there any implications we need to think about in terms of updating other packages with this change, or will those just tease out over time?

sebpearce:
The latter. Generally we don't proactively do those sorts of updates unless there's a deadline – trying to have a scalable approach where we don't need to think about it. (If I turn out to be wrong here, it means there's a problem with the current approach)

makenosound:
Nice one. I will press the big green button then!

PR #607, opened 2020-07-02
feat: StatelessMenu

mbylstra:
## Overview

`StatelessMenu` is the same as `Menu` but it does not have any internal state. The consumer holds the state and is responsible for passing the state to the component, and handling user interactions that change that state. This provides much more flexibility than `Menu`, but comes at the cost of needing more setup code.

## Storybook Story
Story for `StatelessMenu` demonstrating some things you can do: https://dev.cultureamp.design/dst/stateless-menu/storybook-static/?path=/story/menu-react--default-stateless-menu

## Summary of changes:

- New module `StatelessMenu` which tries to re-use as much code from
- In order to achieve code reuse, Menu was significantly refactored to use hooks, so that re-usable view functions could be extracted and re-used between Menu and StatelessMenu. The diff is huge and probably impossible to read - you're better off looking at the file directly for code review.
- Unlike `Menu`, `StatelessMenu` does not use React.cloneElement to add the button event handlers. Rather you provide afunction that renders a button and takes event handlers as an argument (which could be considered a "render prop" or more generally a higher-order function - a function that takes a function as an argument). I ran into a problem for FilterDrawer when I tried to pass in a custom component (a FunctionComponent), where the even listeners weren't added. I found a workaround by using a plain-old function rather than a FunctionComponent, but it was very unexpected. In any case, I think that a higher-order function is a better  approach - `cloneElement` can be very unpredictable and susceptible to race conditions as it mutates the actual DOM (which goes against the grain of React's design). I would have preferred to refactor `Menu` to also not use `cloneElement` but that would be a breaking change. This has also lead to the implementation being more complicated.

## Thoughts:

I think there's a big trade-off between these 3 options:

1. Only support the stateful component. This is easiest for consumers and keeps the code the simplest. The problem is, that it might make it impossible to do some things or you need to rely on massive hacks, such as using `Ref`'s to be able to influence the state from outside the component, or you need two sources of truth of the state - one inside the component and one outside and try to keep them in sync (which is never a good thing)
2. Only support the stateless component. This is  The Elm Way ™️ . I like it because it's the most flexible - you never have to "lift" state. The problem with lifting state is that it causes a major breaking change when you decide at some point that the component should have been stateless because it's too inflexible being stateful. However, I have to admit it does add a lot of boilerplate for simple cases where you can't imagine a need to have access to the state.
3. Provide stateful and stateless components. When you find a need to "lift" state you can add an extra "stateless" component, which is what this PR does. It does not require a breaking change, which is nice. Also, the stateless version can be considered the "expert" version for when you need to do custom things (while the stateful versions remains the easy to use one). The tradeoff is that I think the implementation becomes quite complex as you should share view code between the two versions (the views should be identical, and there's a huge potential for divergence and bugs and inconsistencies if you don't).  It's far simpler having one implementation (stateful or stateless) I think this makes the code pretty hard to read (that's my excuse for how hard to read the code now is!)

All in all,  I think Option 3 is the best set of tradeoffs for this component and situation.

I'm more than happy to got through this over Zoom to answer any questions!

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
@@ -13,6 +13,7 @@ export type MenuProps = {
   button: React.ReactElement<ButtonProps>
   menuVisible?: boolean
   automationId?: string
+  children: React.ReactNode
 }
 
 type Menu = React.FunctionComponent<MenuProps>
--- a/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
@@ -1,5 +1,5 @@
 import { ButtonProps } from "@kaizen/draft-button"
-import * as React from "react"
+import React, { useState } from "react"
 import MenuDropdown from "./MenuDropdown"
 
 const styles = require("./styles.scss")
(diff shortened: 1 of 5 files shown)


draft-packages/menu/KaizenDraft/Menu/Menu.tsx
```diff
@@ -1,84 +1,90 @@
...
+      onClick: e => {
+        e.preventDefault()
+        toggleMenuDropdown()
+      },
+      onMouseDown: e => e.preventDefault(),
+    }),
+  })
 }
+
+type RenderProps = {
```
mbylstra, line 55: Couldn't think of a better name for this type. It's the props for the `render` function - nothing to do with the "render prop" pattern.

draft-packages/menu/KaizenDraft/Menu/Menu.tsx
```diff
@@ -1,84 +1,90 @@
...
-  isMenuVisible: boolean
-}
-
-export type MenuProps = {
+export type GenericMenuProps = {
   align?: "left" | "right"
-  button: React.ReactElement<ButtonProps>
   menuVisible?: boolean
   automationId?: string
+  children: React.ReactNode
```
mbylstra, line 11: Was previously implicit

sebpearce:
This is epic. I love that you decided to export `StatelessMenu` as a separate component alongside `Menu` – an excellent API decision IMO.

draft-packages/menu/KaizenDraft/Menu/Menu.tsx
```diff
@@ -1,84 +1,90 @@
...
-  toggleMenu = (e: MouseEvent) => {
-    e.stopPropagation()
+  const dropdownButtonContainer = React.createRef<HTMLDivElement>()
 
-    const currentState = this.state.isMenuVisible
-    this.setState({
-      isMenuVisible: !currentState,
-    })
-  }
+  const [isMenuVisible, setIsMenuVisible] = useState<boolean>(menuVisible)
```
sebpearce, line 27: Nice to see hooks here – most of this code was ported from the very old Dropdown and needed some refreshment.

draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
```diff
@@ -0,0 +1,31 @@
...
+import MenuDropdown from "./MenuDropdown"
+const styles = require("./styles.scss")
+
+export type StatelessMenuProps = {
+  isMenuVisible: boolean
+  toggleMenuDropdown: () => void
+  hideMenuDropdown: () => void
+  renderButton: (args: {
+    onClick: () => void
+    onMouseDown: (e: any) => void
```
sebpearce, line 13: I'm not clear on why `onMouseDown` takes an event argument here but `onClick` doesn't.

sebpearce:
Checked the branch preview Storybook to see if TitleBlockZen's mobile actions drawer was affected (it uses MenuItems). No issues.

sebpearce:
⚠️ Warning: looks like the "bottom of screen" story has broken on this branch compared with master.

image

draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
mbylstra: Ah, now that you mention it, I think it's actually a potential bug. `e.stopPropagation()` was being called on `onClick` in the original version, but I didn't notice it as it was part of the `toggleDropdown` function. It doesn't seem to make any difference in practice, but I will update it - there must have been a reason for it. Good catch!

mbylstra:
> Warning: looks like the "bottom of screen" story has broken on this branch compared with master. 
Another great catch. I think I must have broke it when I moved the menuContainer refs around 😬

mbylstra:
>  I love that you decided to export StatelessMenu as a separate component alongside Menu – an excellent API decision IMO.
This was Ally's suggestion :)

PR #608, opened 2020-07-02
feat: abillity to add additional content to GenericButton

mbylstra:
## The main change: the additionalContent prop for Generic Button (React)

Adds an optional `additionalContent?: React.ReactNode` prop to `GenericButton`. Can be any React node. Additional content appears to right of the button. Note that this can only be applied to GenericButton, not `Button` or `IconButton`. What this means is that you can't add additionalContent to Button, or IconButton, but you can create a new button type that can take the additionalContent prop. The motivation for this is to be able to add a `Badge` to the `FilterDrawerButton`, which will be used in the `Filter Drawer` component (currently a separate PR that's waiting on this one).

This enables us to create a variant of Button, without needing to change the Button or IconButton APIs (the existing "concrete" buttons). This does require one **slightly controversional change**: `GenericProps` and `LabelProps` are now publicly exported, so consumer may construct new button components which we don't really want. That being said, there was nothing stopping people just copying the props, or using the `any` type, so 🤷‍♂️ .

Other approaches explored:
- Completely fork `Button` and `GenericButton` to make `FilterDrawerButton` ❌  (too much code duplication and too hard to maintain)
- add `additionalContent` to Button (too much flexibility - we don't want people putting random content inside the Button component)  ❌ 
- add a `badge` prop to Button (still too much flexibility, and needing to deal with unwanted prop combinations like `iconPosition="end"`) ❌ 
- add a `variant="filter-drawer-button"` prop to Button. It does not seem worth adding a lot of extra complexity to Button just to support a variant that's used internally inside one component and shouldn't be used anywhere else. ❌ 

## The FilterDrawerButton component

This PR adds a FilterDrawerButton component to `@kaizen/draft-button` (and adds stories) but the only reason it's in this PR, is to dogfood the change to the GenericButton API (to make sure the types are correct and that it works). I intend to move it out of this package into the future `@kaizen/draft-filter-drawer` package. I could have done this all inside the Filter Drawer PR to avoid the move, but that's already a very large PR, and this is fairly pretty big API design decision, so I wanted this PR to be small.

https://dev.cultureamp.design/dst/button-additional-content/storybook-static/?path=/story/button-zen-react--reversed-filter-drawer-button

image

### Chromatic
https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=192
You should see that this hasn't introduced any regressions to `Button` or `IconButton`

### DIFF
--- a/draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
+++ b/draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
@@ -12,7 +12,7 @@ const styles = require("./FilterDrawerButton.module.scss")
 
 type FilterButtonProps = {
   labelText: string
-  numFiltersEnabled: number
+  numberOfFiltersEnabled: number
   onClick: (e: any) => void
   onMouseDown: (e: any) => void
   reversed?: boolean
@@ -20,7 +20,7 @@ type FilterButtonProps = {
 
 export const FilterDrawerButton = ({
   labelText,
-  numFiltersEnabled,
+  numberOfFiltersEnabled,
   onClick,
   onMouseDown,
   reversed,
@@ -36,7 +36,7 @@ export const FilterDrawerButton = ({
       label={labelText}
       icon={filterIcon}
       iconPosition="start"
-      additionalContent={renderBadge(numFiltersEnabled)}
+      additionalContent={renderBadge(numberOfFiltersEnabled)}
       onClick={onClick}
       onMouseDown={onMouseDown}
     />
@@ -47,7 +47,7 @@ const ButtonAllowingAdditionalContent = (
   props: GenericProps & LabelProps & AdditionalContentProps
 ) => <GenericButton {...props} />
 
-const renderBadge = (numFiltersEnabled: number) =>
-  numFiltersEnabled > 0 ? (
-    <Badge variant="active">{String(numFiltersEnabled)}</Badge>
+const renderBadge = (numberOfFiltersEnabled: number) =>
+  numberOfFiltersEnabled > 0 ? (
+    <Badge variant="active">{String(numberOfFiltersEnabled)}</Badge>
   ) : null
--- a/draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
+++ b/draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
@@ -1,0 +1,53 @@
+import { Badge } from "@kaizen/draft-badge"
+import classnames from "classnames"
+import * as React from "react"
+import GenericButton, {
+  AdditionalContentProps,
+  GenericProps,
+  LabelProps,
+} from "./components/GenericButton"
+const filterIcon = require("@kaizen/component-library/icons/filter.icon.svg")
+  .default
+const styles = require("./FilterDrawerButton.module.scss")
+
+type FilterButtonProps = {
+  labelText: string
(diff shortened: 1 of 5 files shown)


draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -365,11 +365,11 @@ $caButton-verticalPaddingForm: calc(
   }
 
   &:not(:first-child) {
-    @include ca-margin($start: 0.5em);
+    @include ca-margin($start: 0.5em, $end: null);
```
mbylstra, line 368: I did this because there was an issue when a third node was added to Button: `ca-margin($start: 0.5em` adds `$end: 0` by default which as overriding the `not(:last-child)` declaration and causing the second middle element to lose it's padding (as it's neither the last nor the first child, so both rules apply) . This change does not add any regressions (see Chromatic build)

sebpearce: That's `null` working as intended there 👌

draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      iconPosition="start"
+      additionalContent={renderBadge(numFiltersEnabled)}
+      onClick={onClick}
+      onMouseDown={onMouseDown}
+    />
+  </div>
+)
+
+const ButtonAllowingAdditionalContent = (
+  props: GenericProps & LabelProps & AdditionalContentProps
```
ActuallyACat, line 47: This feels nice. Splitting out the `AdditionalContentProps`, `GenericProps` and `LabelProps` gives me a very good indication of the intention + allowable features of this wrapper without needing to dig much deeper in implementation 👍

draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  GenericProps,
+  LabelProps,
+} from "./components/GenericButton"
+const filterIcon = require("@kaizen/component-library/icons/filter.icon.svg")
+  .default
+const styles = require("./FilterDrawerButton.module.scss")
+
+type FilterButtonProps = {
+  labelText: string
+  numFiltersEnabled: number
```
ActuallyACat, line 15: Nitpicky, but I can't think of any other components that use shortened prop names... Maybe `numberOfFiltersEnabled` instead? Or even `badgeNumber`?

mbylstra: I really prefer `numberOfFiltersEnabled`, and am actually a big advocate for never abbreviating names. I always worry that people will freak out about long names, but I guess I should embrace it more! I prefer it to `badgeNumber` as it more clearly explains what you're meant to use it for.

ActuallyACat: Sounds good to me

mbylstra:
@ActuallyACat I've made that change, if you could pls give it a ✅

draft-packages/button/KaizenDraft/Button/FilterDrawerButton.tsx
mbylstra: I'm really liking the idea of making GenericProps permissive but have restrictive options for the various button variants. Thanks for the suggestion!

PR #610, opened 2020-07-03
fix: Ensure headers meet WCAG contrast

makenosound:
Some of the header colours in the ConfirmationModal do not meet WCAG contrast requirements. This nudges them up a shade to ensure they do.

The modal headers are intended to be uplifted in the not-too-distant-future to these colours anyway, but good to have them accessible in the interim.


didoesdigital:
🎉 Thanks, @makenosound

PR #615, opened 2020-07-03
feat: add React default secondary select

didoesdigital:
These:

image

- Adds default secondary single select
- Adds story for default secondary single select
- Adds story for default secondary single select disabled

Built:

image
image
image
image

Branch preview:

- https://dev.cultureamp.design/dst/update-secondary-select/storybook/:
    - https://dev.cultureamp.design/dst/update-secondary-select/storybook-static/?path=/story/select-react--single-secondary
    - https://dev.cultureamp.design/dst/update-secondary-select/storybook-static/?path=/story/select-react--single-secondary-disabled

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/styles.react.scss
+++ b/draft-packages/select/KaizenDraft/Select/styles.react.scss
@@ -90,34 +90,80 @@
     cursor: default;
   }
 
-  &.secondary.reversed {
-    .control {
-      background: add-alpha($kz-color-white, 10%);
-      color: $kz-color-white;
-      border: $kz-border-borderless-border-width
-        $kz-border-borderless-border-style $kz-border-borderless-border-color;
-      border-radius: $kz-border-borderless-border-radius;
+  &.secondary {
+    &.default {
+      .control {
+        background: transparent;
+        border: $kz-border-borderless-border-width
+          $kz-border-borderless-border-style $kz-border-borderless-border-color;
+        border-radius: $kz-border-borderless-border-radius;
 
-      &:hover {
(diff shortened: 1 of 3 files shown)


draft-packages/select/KaizenDraft/Select/styles.react.scss
```diff
@@ -90,34 +90,80 @@
...
+    &.reversed {
+      .control {
+        background: add-alpha($kz-color-white, 10%);
+        color: $kz-color-white;
+        border: $kz-border-borderless-border-width
+          $kz-border-borderless-border-style $kz-border-borderless-border-color;
+        border-radius: $kz-border-borderless-border-radius;
+
+        &:hover {
+          background: add-alpha($kz-color-white, 20%);
```
mbylstra, line 147: This sparked my question around how opacity should be achieved. The consensus is to use `add-alpha($kz-color-white, 20%)` (it's a the scss function). I realise this is just inherited code, so no need to update it here. We could do a big sweep across the codebase easily enough.

sebpearce: > The consensus is to use add-alpha($kz-color-white, 20%) (it's a the scss function)

The policy that we've been communicating/using for the last few months has been `rgba($kz-color-white, 20%).` We've been telling people to stop using `add-alpha` since it's a custom function that I built and we want to only use built-in SCSS functions.

sebpearce: (I agree it sucks that it's not real CSS – having unexpected bugs because of the build not breaking when custom Sass functions aren't present seems worse though!)

PR #621, opened 2020-07-07
fix: Switch from TSlint to ESlint

ActuallyACat:
## Why
1. tslint is deprecated
2. There’s better maintained plugins in eslint, mostly because of point 1 (**)

(**) If I focus on moving Kaizen to eslint, it also means I can trash the dangerfile work which I was planning to do next. There’s an eslint plugin to do what that script checks:
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md 

### Rocks
- [x] Prettier and Eslint conflict, especially around switch statements
- [x] Figure out which rules to ignore

### Resources
Q&A: https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/FAQs.md
The guide: https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration#eslint-installation

Here's the script log:
```
8 ESLint rules behave differently from their TSLint counterparts:
  * @typescript-eslint/no-unused-expressions:
    - The TSLint optional config "allow-new" is the default ESLint behavior and will no longer be ignored.
  * camelcase:
    - Leading and trailing underscores (_) in variable names will now be ignored.
    - ESLint's camel-case rule does not allow pascal or snake case variable names. Those cases are reserved for class names and static methods.
  * import/no-extraneous-dependencies:
    - ESLint does not support whitelisting modules
  * no-invalid-this:
    - Functions in methods will no longer be ignored.
  * no-redeclare:
    - ESLint does not support check-parameters.
  * no-underscore-dangle:
    - Leading or trailing underscores (_) on identifiers will now be forbidden.
  * prefer-arrow/prefer-arrow-functions:
    - ESLint does not support allowing standalone function declarations.
    - ESLint does not support allowing named functions defined with the function keyword.
  * space-before-function-paren:
    - Option "constructor" is not supported by ESLint.
    - Option "method" is not supported by ESLint.

3 rules are not known by tslint-to-eslint-config to have ESLint equivalents:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "import-spacing".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "prefer-conditional-expression".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "whitespace".

1 editor setting is not known by tslint-to-eslint-config to have an ESLint equivalent:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "editor.formatOnSave".
```


site/src/docs-components/icons/IconGrid.tsx
```diff
@@ -4,9 +4,7 @@ import IconTile from "./IconTile"
 
 const styles = require("./IconGrid.scss")
 
-type IconGridProps = {}
-
-const IconGrid: React.SFC<IconGridProps> = props => (
+const IconGrid: React.SFC = props => (
```
ActuallyACat, line 7: Flagging this as something to verify

tsconfig.json
```diff
@@ -1,6 +1,13 @@
 {
   "compilerOptions": {
-    "plugins": [{ "name": "typescript-tslint-plugin" }],
+    "plugins": [
```
ActuallyACat, line 3: `typescript-plugin-css-modules` config will be followed up in another PR (it's missing a declaration file and other setup)

site/src/docs-components/icons/IconGrid.tsx
ActuallyACat: I initially thought this was a component 🤦‍♀️ Since it's internal than this update is fine.

tsconfig.json
ActuallyACat: Decided to remove this to keep the PR clean and single-purpose.

PR #624, opened 2020-07-07
feat: Changed props from row/column to horizontal/vertical

ActuallyACat:
BREAKING CHANGE: Changed props from row/column to horizontal/vertical

- [x]  Check which consumers are using Tabs < v2

Since this is a breaking change, I've been explicit in which versions of tabs will cause issues when moving to draft-tabs@2.0 (anything >1.5.0)

Tests passed in chromatic: https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=190

### DIFF
--- a/draft-packages/tabs/KaizenDraft/Tabs/styles.scss
+++ b/draft-packages/tabs/KaizenDraft/Tabs/styles.scss
@@ -31,7 +31,7 @@
     font-weight: $kz-typography-heading-4-font-weight;
   }
 
-  // Use JS polyfill to simulate :focus-visible, not yet supported by browsers
+  // Use JS polyfill to simulate :focus-visible, not yet supported by bhorizontalsers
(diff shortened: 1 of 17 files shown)


draft-packages/tabs/KaizenDraft/Tabs/styles.scss
```diff
@@ -31,7 +31,7 @@
     font-weight: $kz-typography-heading-4-font-weight;
   }
 
-  // Use JS polyfill to simulate :focus-visible, not yet supported by browsers
+  // Use JS polyfill to simulate :focus-visible, not yet supported by bhorizontalsers
```
didoesdigital, line 34: :D Bulk find and replace, eh?

ActuallyACat: No, I meant to type `bhorizontalsers`

PR #627, opened 2020-07-07
fix: Replace the sized-bold values with the new variant

makenosound:
These were dropped in `@kaizen/design-tokens@2.0.0` in favour of a single bold variant for `paragraph` https://github.com/cultureamp/kaizen-design-system/pull/606

@didoesdigital — I’m not totally sure what the change version should be here with `peerDependencies`. Guess this should be a `BREAKING CHANGE: ` too?


didoesdigital:
@makenosound breaking change sounds right to me but I'm not confident about that CC @ActuallyACat

ActuallyACat:
We've started defining what a breaking change / major changes are in Kaizen, but it's still very much a draft: https://docs.google.com/document/d/1Gx2JYeg6DlNFxMYlPHLT2AqzUXGj4Fq00_kclDJ-N8w/edit#heading=h.1ttjbamtxy8o

At this point, major changes have been used in breaking API changes only. I'm keen to change this, and I think this is a good candidate for a major change.

makenosound:
@ActuallyACat — The build here appears to be failing because of missing values from https://github.com/cultureamp/kaizen-design-system/pull/606 which is a bit confusing. I am wondering if it’s because they need to be compiled for use within Storybook (which is using local versions of packages) versus in IRL usage where the Sass/Less files are compiled on release. Does that sound right to you?

ActuallyACat:
This is possibly the cause @makenosound. It feels like bad DX to expect devs to commit the transformed tokens when they make changes. An alternative might be to use a git hook and run `yarn prepublish && git add` before pushing...

makenosound:
I’ve built them into this PR now @ActuallyACat and it looks like that was the issue. Definitely feels like it’d be nicer DX to have that handled automatically.

makenosound:
I think I have satisfied the great and powerful Moai. Should be good to go now @ActuallyACat

PR #630, opened 2020-07-08
feat: prefix Menu handler functions with "on"

mbylstra:
The `hideMenuDropdown` and `toggleMenuDropdown` handler functions
have been renamed to `onHideMenuDropdown` and `onToggleMenuDropdown` to follow
React conventions (and plain good naming practice)

- This is a **breaking change for StatelessMenu**
  (which is currently unused in Kaizen and elsewhere)
- This is **not a breaking change for Menu** (only an internal refactor)


draft-packages/menu/KaizenDraft/Menu/Menu.tsx
```diff
@@ -45,11 +45,7 @@ export type MenuProps = GenericMenuProps & StatefulMenuProps
 type Menu = React.FunctionComponent<MenuProps>
 
 const Menu: Menu = props => {
-  const {
-    align = "left",
-    dropdownWidth = "default",
-    menuVisible = false,
-  } = props
```
mbylstra, line 52: This is an unrelated change, but I removed `dropdownWidth = "default"` as it was dead code (dropdownWidth is just passed straight through to `DropdownMenu`)

draft-packages/stories/Menu.stories.tsx
```diff
@@ -190,8 +190,8 @@ const StatelessMenuExample = props => {
       </Paragraph>
       <StatelessMenu
         isMenuVisible={isMenuVisible}
-        toggleMenuDropdown={toggleMenu}
-        hideMenuDropdown={hideMenu}
```
mbylstra, line 194: As you can see, only the StatelessMenu stories needed to be updated. The `Menu` stories are unchanged (were not affected by the breaking change)

makenosound:
Looks sensible to me, @mbylstra!

mbylstra:
Oops, I was too optimistic the build would pass, and actually it caught that I'd also accidentally updated `packages/component-library/components/Dropdown/DropdownMenu.tsx` which is the legacy dropdown, and out of scope for this PR 😅 . I removed the changes to that file and the build is passing now.

draft-packages/menu/KaizenDraft/Menu/Menu.tsx
ActuallyACat: I wonder if the prop should be deprecated too? 
https://github.com/cultureamp/kaizen-design-system/pull/630/files#diff-5233a0f50609d2bf1752cd5a7f97633d1d5efe4ef48c90b946ba9b48392d64c3L22 

This is what I've been doing in components - which gives a really nice visual indicator (strikethrough) in some editors: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Button/Button.tsx#L7-L9

ActuallyACat (approved):

PR #633, opened 2020-07-09
docs: add favicon to Kaizen Site

### LINKED ISSUE #96: Missing favicon on website.
If a favicon.ico is provided in the root of the site this problem disappears :) 

There will need to be a decision about whether there's a Kaizen-specific branding or whether it's okay to use the Enso icon from the main website.

didoesdigital:
Closes #96 

image
image


site/src/components/Head.tsx
```diff
@@ -2,6 +2,7 @@ import { graphql, useStaticQuery } from "gatsby"
 import "normalize.css"
 import * as React from "react"
 import { Helmet } from "react-helmet"
+const favicon = require("../favicon.favicon.png")
```
ActuallyACat, line 5: Should this be `../favicon/favicon.png` ?

didoesdigital: Yes. It is… unclear to me why this didn't break locally.

PR #634, opened 2020-07-09
fix: reversed filter button class was not being applied

mbylstra:
OLD
to
NEW

image


draft-packages/filter-drawer/KaizenDraft/FilterDrawer/styles.module.scss
```diff
@@ -24,21 +24,6 @@ $menuContentVerticalPadding: 6px; // this comes from the MenuContent component,
     calc(#{$kz-spacing-sm} - #{$menuContentHorizontalPadding});
 }
 
-// The button used for the Filter Drawer component is similar to the
```
mbylstra, line 27: This was dead duplicated code (duplicated in FilterDrawerButton.modules.scss)

PR #636, opened 2020-07-09
docs: add demo id to filter drawer docs

mbylstra:
image

Preview: https://dev.cultureamp.design/dst/filter-drawer-demo-story-id/components/filter-drawer/

The storybook embed doesn't seem to support the background colour supplied by the CSF config, so I added a wisteria div background to the story.


draft-packages/stories/FilterDrawer.stories.scss
```diff
@@ -9,3 +11,9 @@
     width: auto;
   }
 }
+
+.siteDemoWrapper {
+  background-color: $kz-color-wisteria-700;
+  height: 368px;
```
didoesdigital, line 17: Without the height specified, we could reuse this class for other components that want a Wisteria background. Perhaps just leave a comment that says this is for Filter Drawer so if we want to abstract it out later, it's easy?

PR #648, opened 2020-07-13
ci: Added css-module-plugin to add types to styles

ActuallyACat:


zioroboco:
Note that [`typescript-plugin-css-modules`](https://github.com/mrmckeb/typescript-plugin-css-modules) doesn’t actually create types for css modules.

It’s a language server plugin which gives you editor completions for classnames in local development… but nothing will break in CI, so you can’t use it to chase down type errors after a refactor.

PR #652, opened 2020-07-15
fix: remove minimist from package.json resolutions

mbylstra:
This was required in the past because a dependency had a transitive dependency on a version of minimist that had a security issue. This is no longer necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,9 +126,6 @@
     "lint-staged": "^10.2.2",
     "storybook-chromatic": "^3.4.1"
   },
-  "resolutions": {
-    "**/**/minimist": "^1.2.3"
-  },
   "lint-staged": {
     "*": [
       "prettier --write"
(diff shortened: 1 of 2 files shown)


mbylstra:
Comparing master with my branch with `yarn audit`, it seems they have the same level of vulnerabilities:

this branch
```
17327 vulnerabilities found - Packages audited: 4180
Severity: 17326 Low | 1 Moderate
:sparkles: Done in 48.21s.
```
master
```
17327 vulnerabilities found - Packages audited: 4179
Severity: 17326 Low | 1 Moderate
:sparkles: Done in 47.25s.
```

I'm not sure where the vulnerability alerts were coming from (was it Github, or dependabot, or now Github because dependabot acquired Dependabot?), but this PR doesn't appear to make it less vulnerable. I guess the alerts only come through for "High" vulnerabilities?

PR #655, opened 2020-07-15
docs: add data viz page

didoesdigital:
This PR adds a **data visualization guidelines** page.

This is intended to be a starting point only, and we add visuals and more details over time. Eventually, I expect the sections on charts could move to their own component pages for each chart type.


site/docs/guidelines/data-visualization.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+- [Data viz project: Bar Chart (Horizontal)](https://datavizproject.com/data-type/bar-chart-horizontal/)
+- [Data viz project: Stacked Bar Chart](https://datavizproject.com/data-type/stacked-bar-chart/)
+- [The Data Visualisation Catalogue: Bar Chart](https://datavizcatalogue.com/methods/bar_chart.html)
+- [Shopify Polaris: horizontal bar charts](https://polaris.shopify.com/design/data-visualizations#section-horizontal-bar-charts)
+- [Shopify Polaris: vertical column charts](https://polaris.shopify.com/design/data-visualizations#section-vertical-column-charts)
+- [Morningstar: horizontal bar chart](https://designsystem.morningstar.com/charts/horizontal-bar.html)
+- [Government of Canada, Aurora: Bar charts](https://design.gccollab.ca/data/column-and-bar-graphs)
+
+## Hero data
```
ActuallyACat, line 97: I think this one needs an example. I know the concept of heroing data, but not the noun "hero data"

didoesdigital: Updated explanation:

> Hero data refers to prominent figures on a page that draw attention to individual values using proportionally larger type sizes. Use hero data to highlight a single value when it does not need context or comparisons.

… and a picture!

PR #656, opened 2020-07-15
docs(site): Move storybook out of an iframe

ActuallyACat:
## About
The current Kaizen setup builds two main assets: Gatsby and Storybook (deployed to `http://www.cultureamp.design/storybook-static`). When on the `/storybook` route we [load Storybook into an iframe](https://github.com/cultureamp/kaizen-design-system/blob/master/site/src/pages/storybook.tsx). The biggest drawback of this approach is Storybook can't push changes to the URL. We can't easily share links to stories or get meaningful analytics on stories.

## What's different
Instead of serving Storybook from Gatsby, I have instead added a header and footer to the _existing_ `storybook-static`.  What this means is a user who goes to `storybook` will not go via Gatsby anymore.

Pros
- can see the URL change -> good for analytics and sharing stories

Cons 
- If we update the navigation links we now need to change it in two places (one in gatsby, the other in storybook)

### TODO
- [x] Add analytics to storybook
- [x] Should I move the static-storybook build to `/storybook`?


storybook/header-preset/header.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {
+  Link as NavLink,
+  NavigationBar,
+} from "../../packages/component-library/components/NavigationBar"
+
+import Footer from "../../site/src/components/Footer"
+
+const styles = require("./header.module.scss")
+
+export const SiteHeader = () => {
```
ActuallyACat, line 12: I considered moving this component to be inside the `./site` directory. If we ever need to update the navigation content we'll need to remember to update it here too. I used the existing Kaizen component, so we don't need to worry about syncing visual changes.

.buildkite/scripts/publish.sh
```diff
@@ -11,7 +11,7 @@ aws s3 sync --delete \
 
 aws s3 sync --delete \
     "storybook/public" \
-    "s3://${KAIZEN_DOMAIN_NAME}${KAIZEN_BASE_PATH}/storybook-static"
+    "s3://${KAIZEN_DOMAIN_NAME}${KAIZEN_BASE_PATH}/storybook"
```
ActuallyACat, line 14: I _think_ this will automagically change the route from `http://cultureamp.design/storybook-static` to `http://cultureamp.design/storybook` - is this correct @zioroboco ?

storybook/header-preset/preset.ts
```diff
@@ -0,0 +1,34 @@
+import path from "path"
+import { NormalModuleReplacementPlugin } from "webpack"
+const webpackConfig = require("../webpack.config")
+
+module.exports = {
+  managerWebpack: async config => {
+    const replaceLayoutPlugin = new NormalModuleReplacementPlugin(
```
ActuallyACat, line 7: We can check for development environment and only render for prod

storybook/header-preset/header.tsx
zioroboco: This changes pretty infrequently — maybe a big ole comment at the top of each file, warning of the duplication?

.buildkite/scripts/publish.sh
zioroboco: Yep. This will write the output of the storybook build to the new path in S3, including for dev deploys.

The s3/cloudfront stack takes urls like `cultureamp.com/some/path` and return whatever it finds at `s3://<branch-routing-bits>/some/path/index.html`. The line you've changed will deploy assets to the `/storybook/*` path.

zioroboco: Which, by the way, means if you ever wanted to deploy anything else (maybe an example page, like we were discussing), you only need to build it in CI and deploy it to whichever path you want, as above.

This will also work for branch deploys.

ActuallyACat: Neat, this is good to know.

storybook/header-preset/header.tsx
ActuallyACat: Updated :D

storybook/header-preset/header.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react"
+import { render } from "react-dom"
+import {
+  Link as NavLink,
+  NavigationBar,
+} from "../../packages/component-library/components/NavigationBar"
```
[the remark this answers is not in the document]
ActuallyACat: This change is on an outdated file - this has already been updated.

storybook/header-preset/header.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const SiteHeader = () => {
+  let branch = ""
+  if (window.location.hostname.includes("dev")) {
+    // the dev branch is appended before the rest of the pathname
+    // given this component will only ever mount on storybook-static
+    // strip that from the path
+    branch = window.location.pathname
+      .split("/")
+      .filter(curr => curr !== "" && curr !== "storybook-static")
```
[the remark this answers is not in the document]
ActuallyACat: Yeah, this was kinda gross to do. I'll update - thanks

package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "ts-node": "TS_NODE_TRANSPILE_ONLY=true ts-node --project storybook/tsconfig.storybook.json",
     "storybook": "yarn install && yarn ts-node node_modules/.bin/start-storybook -c storybook",
-    "storybook:build": "yarn ts-node node_modules/.bin/build-storybook -c storybook -o storybook/public",
+    "storybook:build": "NODE_ENV=production yarn ts-node node_modules/.bin/build-storybook -c storybook -o storybook/public",
```
zioroboco, line 15: Last question, mostly out of curiosity: is this fixing something? ☝️

ActuallyACat: Just making sure I don't add the storybook header and footer for local builds: https://github.com/cultureamp/kaizen-design-system/pull/656/files#diff-85d77a5014e178e22ecd89cd8866a047R8

PR #660, opened 2020-07-17
fix: Make TitleBlockZen title extend full width when no buttons

sebpearce:
## Changes

- Fixes a bug where the title was not extending to its full width even when there was space available.

Other small changes relating to horizontal positioning are also piggybacking on this PR:

- Fixes a title truncation issue where at smaller "desktop" widths the title was not getting an ellipsis but just being cut off abruptly.
- Adds a story for the case where there is only a long title plus a subtitle – not buttons.
- Fixes the bug where the top row of the Title Block was exceeding its max-width at the smallest "desktop" width when the content is cramped. This was fixed by setting `min-width: 0` on various elements to allow `flex-shrink` to work as intended.
- Puts `flex-shrink: 0` on the Main Actions (the 1 or 2 buttons in the top right corner). This is a trade-off that offers less glitchy behaviour at more cramped viewports at the cost of having less real estate for the other content.

### DIFF
--- a/draft-packages/stories/TitleBlockZen.stories.tsx
+++ b/draft-packages/stories/TitleBlockZen.stories.tsx
@@ -659,3 +659,24 @@ export const DefaultOnlyPrimary = () => (
 DefaultOnlyPrimary.story = {
   name: "Default (only primary action)",
 }
+
+export const DefaultOnlyLongTitle = () => (
+  <TitleBlockZen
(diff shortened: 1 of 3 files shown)


sebpearce:
[Titleblock truncation bug](https://trello.com/c/s7RhioID/4-titleblock-truncation-bug)

sampnorris:
No Comments - I even booted the branch preview in IE - ✅

PR #661, opened 2020-07-17
docs(site): Fix url from storybook on production

ActuallyACat:
On branch previews we returned a base URL including the forward slash, but on production we weren't. This patches that error.

### DIFF
--- a/storybook/header-preset/header.tsx
+++ b/storybook/header-preset/header.tsx
@@ -7,15 +7,24 @@ import { Link as NavLink, NavigationBar } from "@kaizen/component-library"
 const styles = require("./header.module.scss")
 
 const SiteHeader = () => {
-  let branch = ""
-  if (window.location.hostname.match(/^dev/)) {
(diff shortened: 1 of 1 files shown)


storybook/header-preset/header.tsx
```diff
@@ -7,15 +7,24 @@ import { Link as NavLink, NavigationBar } from "@kaizen/component-library"
...
-    // strip that from the path
-    const branchName = window.location.pathname.match(/(.+)storybook/)
-    branch = branchName && branchName.length > 1 ? branchName[1] : ""
+  const getBaseUrl = (pathname, origin, hostname) => {
+    let branch = ""
+    if (hostname.match(/^dev/)) {
+      // the dev branch is appended before the rest of the pathname
+      // given this component will only ever mount on storybook-static
+      // strip that from the path
+      const branchName = pathname.match(/(.+)?\/storybook/)
```
ActuallyACat, line 16: This regex was the culprit: `/(.+)?\/storybook/`

PR #662, opened 2020-07-20
feat: Add focus support to the Button components

jeremy8883:
* Give the ability to imperatively call `focus` the two `Button` components.

image

* Forward down any `aria-*` props down to the two `Button` components. Edit: support has been updated to allow for all kebab cased props.

### Motivation

I'm going through perf-ui at the moment, and making some a11y improvements. One of those improvements is for a dropdown menu, which uses this kaizen button. When the user presses escape, and we've closed the dropdown menu, we need to place the focus back on the original button.

The menu in perform, under "Goals":

image

I realised after the fact, that we already have a button dropdown in kaizen, and I should have just replaced the perform-ui code. But, I think the updates will still be handy at some point.

### DIFF
--- a/draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
+++ b/draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
@@ -53,8 +53,10 @@ type Props = ButtonProps & {
 
 export type ButtonFunctions = { focus: () => void }
 
-const getAriaProps = (props: object) => {
-  const keys = Object.keys(props).filter(k => k.startsWith("aria-"))
+// We're treating custom props as anything that is kebab cased.
+// This is so we can support properties like aria-* or data-*
+const getCustomProps = (props: object) => {
+  const keys = Object.keys(props).filter(k => k.indexOf("-") !== -1)
   return keys.reduce((acc, val) => {
     acc[val] = props[val]
     return acc
@@ -108,7 +110,7 @@ const renderButton = (props: Props, ref: Ref<HTMLButtonElement>) => {
     ...rest
   } = props
   const label = props.icon && props.iconButton ? props.label : undefined
-  const ariaProps = getAriaProps(rest)
+  const customProps = getCustomProps(rest)
 
   return (
     <button
@@ -141,7 +143,7 @@ const renderButton = (props: Props, ref: Ref<HTMLButtonElement>) => {
         props.analytics && JSON.stringify(props.analytics.properties)
       }
       ref={ref}
-      {...ariaProps}
+      {...customProps}
     >
       {renderContent(props)}
     </button>
@@ -158,7 +160,7 @@ const renderLink = (props: Props, ref: Ref<HTMLAnchorElement>) => {
     onBlur,
     ...rest
   } = props
-  const ariaProps = getAriaProps(rest)
+  const customProps = getCustomProps(rest)
 
   return (
     <a
@@ -182,7 +184,7 @@ const renderLink = (props: Props, ref: Ref<HTMLAnchorElement>) => {
         props.analytics && JSON.stringify(props.analytics.properties)
       }
       ref={ref}
-      {...ariaProps}
+      {...customProps}
     >
       {renderContent(props)}
     </a>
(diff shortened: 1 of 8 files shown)


draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -51,15 +51,36 @@ type Props = ButtonProps & {
...
-    className={classNames(styles.container, {
-      [styles.fullWidth]: props.fullWidth,
-    })}
-  >
-    {props.href && !props.disabled ? renderLink(props) : renderButton(props)}
-  </span>
-)
+export type ButtonFunctions = { focus: () => void }
+
+const getAriaProps = (props: object) => {
```
jeremy8883, line 56: This is copy/pasted in two places. Is there a recommended method for reusing this code? I suppose it's not a big deal, since the "component-library" version is deprecated.

jeremy8883:
> Forward down any aria-* props down to the two Button components

Shall I do this for `data-*` attributes also? I feel like this is likely to come up.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
ActuallyACat: It'd be nice to eventually have a package for miscellaneous utils like this. For now, I think this is fine

ActuallyACat:
> > Forward down any aria-* props down to the two Button components
> 
> Shall I do this for `data-*` attributes also? I feel like this is likely to come up.

I've been treating any kebab-case properties the same (aria and data, but there are some edge cases in there) and spreading them on the wrapper of the component I'm adding it to. One thing that comes to mind for me is figuring out which scenarios we want to treat aria/data attributes separately. In this scenario, probably not - the aria attributes belong on `button` which happens to be the root level element in this component. In other components, like the form components, the `input` isn't the root level element in those components, and hence we might want `data-*` on the wrapper, and different `aria-*` attributes elsewhere. 

Nothing for you to worry about now, just something that's been on my mind.

ActuallyACat:
Something is amiss with the example. I can tab focus, but clicking the button in the story isnt working for me: 
Storybook

jeremy8883:
So it should be focusing, but it won't visually show it as being focused, if it is being clicked with a mouse. This is due to the `.js-focus-visible` functionality. If you first tab into the second button, then press enter, it will show the focus outline.

Just confirming if this is the case for yourself also.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -51,14 +51,39 @@ type Props = ButtonProps & {
...
+
+// We're treating custom props as anything that is kebab cased.
+// This is so we can support properties like aria-* or data-*
+const getCustomProps = (props: object) => {
+  const keys = Object.keys(props).filter(k => k.indexOf("-") !== -1)
+  return keys.reduce((acc, val) => {
+    acc[val] = props[val]
+    return acc
+  }, {})
+}
```
okeeffed, line 64: Is this for kebab attributes other than `aria-*`and `data-*`? I would have thought native attributes like that would be supported by the `HTMLElement` types but not sure if this is for something else?

packages/component-library/components/Button/components/GenericButton.tsx
```diff
@@ -43,14 +43,39 @@ type Props = ButtonProps & {
...
+
+// We're treating custom props as anything that is kebab cased.
+// This is so we can support properties like aria-* or data-*
+const getCustomProps = (props: object) => {
+  const keys = Object.keys(props).filter(k => k.indexOf("-") !== -1)
+  return keys.reduce((acc, val) => {
+    acc[val] = props[val]
+    return acc
+  }, {})
+}
```
okeeffed, line 56: I have no idea of the current Kaizen structure, but is it possible to abstract this out?

jeremy8883: Yeah, I asked the same question 😛 

> It'd be nice to eventually have a package for miscellaneous utils like this. For now, I think this is fine

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
jeremy8883: So it's not quite related to the `HTMLElement` type. This is just to funnel any kebab props, including `aria-` and `data-` down to the actual `button` element (although I'm not sure if there are any other valid kebab attributes that we need to allow).

PR #665, opened 2020-07-21
feat: Add toggle switch with label after the toggle control

lijuenc:
Bring the toggle switch components for React and Elm in line with the UI kit by adding the ability to position the label after the control. Previously, the label was always before the control.

Screen Shot 2020-07-21 at 2 26 19 pm

Storybook: https://dev.cultureamp.design/lijuenc/add-label-after-toggle-switch-control/storybook/?path=/story/toggleswitchfield-react--label-at-end


katsambarcus:
Visually everything looks good, @lijuenc 

Looking at the naming in storybook, there is some misalignment with how they are referred to in the UI KIt. Particularly the conflation of RTL (localisation) with right and left (layout flexibility). The reason there are both options is to give designers the flexibility to position the toggle at either side of the screen, so naming the toggle with label to the right as 'Label at end (RTL) is confusing.

My recommendation would be to name them:

Label Left
Label Right

lijuenc:
Thanks @katsambarcus.

The reason that they're referred to using "start" and "end" is that that is the most commonly used terminology when handling LTR/RTL. Same as how we've avoided using margins and paddings with lefts and rights in the design system.

Using that naming:

Label Left would have the label on the left.
Label Left (RTL) would have the label on the right.

Would that be confusing?

makenosound:
Yeah, we do need the names to be abstracted from left/right dimensions. I think there’s a _bit_ of awkwardness about `start` and `end` in this context (`before/after` anchors around the label a little better), but as @lijuenc says it’s how we’ve referred to those sorts of configuration elsewhere and I don’t think there’s a super-obvious "winner" for other language so what’s there is OK.

draft-packages/form/KaizenDraft/Form/Primitives/Label/styles.scss
```diff
@@ -60,8 +60,12 @@ $label-start-margin: 10px;
   justify-content: space-between;
   align-items: center;
 
-  .labelText {
+  .prependedLabel {
     @include ca-margin($end: $label-start-margin);
```
dougmacknz, line 64: One small improvement that could be made: `$label-start-margin` is currently set to 10px, we could use `$ca-grid * 0.5` there instead.

dougmacknz (approved):
This is looking good to me 🚢

draft-packages/form/KaizenDraft/Form/Primitives/Label/styles.scss
lijuenc: That's a good point, however, I think `$ca-grid` has been deprecated in favour of `$kz-spacing` tokens and it would probably be better to change over all the form inputs to use those tokens at the same time as a task on its own.

sebpearce: I may not be up to date but the last I heard, the `$kz-spacing` tokens were more of an alternative option than a replacement for `$ca-grid`. @ActuallyACat might have a preference on this

Issue #673, opened 2020-07-27
The Prop table is not displaying for draft component 


ActuallyACat:
We use a plugin called `react-docgen-typescript-loader` to extract the props/state information from the components to be fed into a Storybook Plugin, `addon-info`. This is currently only working for "core" compoenents.

As an example, the Badge component does not have props. Click "show info" on this component page: https://cultureamp.design/storybook/?path=/story/badge-react--active

PR #674, opened 2020-07-27
ci: Add draft packages directories to react-docgen-typescript-loader (#673)

ActuallyACat:
Fixes: https://github.com/cultureamp/kaizen-design-system/issues/673

I had previously set the source directory to `../draft-packages` thinking that it would traverse the directory and find the files that it needs. This was not the case. 

I've updated the config so that it gets all the directories within draft-packages. 

Prev: 
Screen Shot 2020-07-27 at 6 23 05 pm

Now: 
Screen Shot 2020-07-27 at 6 24 29 pm

### DIFF
--- a/storybook/webpack.config.ts
+++ b/storybook/webpack.config.ts
@@ -2,6 +2,7 @@
 import "./pre-build"
 
 import { resolve } from "path"
+import { readdirSync } from "fs"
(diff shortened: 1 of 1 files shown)


ActuallyACat:
@zioroboco Has something changed recently in Moai? I think I used to be able to use `ci: fixes thing` in my commit messages.

PR #683, opened 2020-07-30
fix: Update react select version to fix scroll bug

dhotson:
This fixes a bug where the dropdown doesnt scroll to the selected option

Updated it in a couple of places. Only a minor update that fixes a few
bugs and adds a new feature. Should be backwards compatible.

## Before
before

.. see how when you re-open it's at the top?

## After
after

.. now when you open the select, the current selection is visible.

### DIFF
--- a/draft-packages/select/package.json
+++ b/draft-packages/select/package.json
@@ -36,7 +36,7 @@
     "@kaizen/draft-form": "^2.3.0",
     "@types/classnames": "^2.2.10",
     "classnames": "^2.2.6",
-    "react-select": "3.0.8"
+    "react-select": "^3.1.0"
   },
   "devDependencies": {
     "@cultureamp/elm-storybook": "cultureamp/elm-storybook#0.2.1",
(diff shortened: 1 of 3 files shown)


draft-packages/select/package.json
```diff
@@ -36,7 +36,7 @@
     "@kaizen/draft-form": "^2.3.0",
     "@types/classnames": "^2.2.10",
     "classnames": "^2.2.6",
-    "react-select": "3.0.8"
```
dhotson, line 39: This is the important version upgrade.

packages/component-library/package.json
```diff
@@ -43,7 +43,7 @@
     "react-animate-height": "^2.0.15",
     "react-focus-lock": "^1.19.1",
     "react-media": "^1.9.2",
-    "react-select": "^3.0.8",
```
dhotson, line 46: This upgrade may not have been necessary, but I didn't want us to have two slightly different versions of react-select.

mbylstra: This is for `@kaizen/component-library` and the other one is for `@kaizen/draft-select`. Ideally they would always be on the same version a regular refresh. Thanks for updating this! `react-select` is likely not even be used in `@kaizen/component-library` any more, and is probably a relic from when we moved components into draft packages. I'll create a github issue to remove unused packages from `packages/component-library/package.json`

PR #689, opened 2020-08-04
fix: Extend TitleBlockZen API so NavigationTabs can take an automation ID

okeeffed:
## Objective

See title

## Notes

I did see that the `div` was within the `a` tag but decided not to look into it.


caseysinclair:
[Add Select w/ label variant](https://trello.com/c/7qOy9PYK/25-add-select-w-label-variant)

PR #692, opened 2020-08-04
fix: add comments that Layout, Text and MenuList are deprecated

mbylstra:


packages/component-library/stories/Layout.stories.tsx
```diff
@@ -19,7 +19,7 @@ const supportIcon = require("@kaizen/component-library/icons/support.icon.svg")
 import { TitleBlock } from "@kaizen/draft-title-block"
 
 export default {
-  title: "Layout (React)",
+  title: "Layout (React, DEPRECATED)",
```
sebpearce, line 22:
```suggestion
  title: "Layout (deprecated) (React)",
```

just for @didoesdigital 's sake :D

image

site/src/components/Layout.tsx
```diff
@@ -1,3 +1,5 @@
+/* !!! This component is deprecated. Please do not use for new code  !!! */
```
sebpearce, line 1: 👌

packages/component-library/stories/Layout.stories.tsx
mbylstra: Haha, good eye, could I please get a retick?

PR #696, opened 2020-08-06
feat: add ability to include a select label

caseysinclair:
A small addition to add the ability to include a select label for Elm select

Screen Shot 2020-08-06 at 3 04 31 pm

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.elm
+++ b/draft-packages/select/KaizenDraft/Select/Select.elm
@@ -6,6 +6,7 @@ module KaizenDraft.Select.Select exposing
     , Style(..)
     , initState
     , isLoading
+    , label
     , menuItems
     , multi
     , placeholder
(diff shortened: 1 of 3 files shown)


caseysinclair:
[Add Select w/ label variant](https://trello.com/c/7qOy9PYK/25-add-select-w-label-variant)

mbylstra:
> Small change: I think there's meant to be a 1/4 grid gap between the label:
I realised that all of the components don't have this gap, so it's probably best to not put it in for now (even though it should), otherwise form elements will look very inconsistent on the page. I'll add git issue to update the styling across all the form elements (it should all happen at once). So, that doesn't need to be changed in this PR.

However, I realised there is an inconsistency between the naming whether form elements have `Field` in the name or not.
The convention looks like `XXXXField` is a component that wraps an input element and adds labels and descriptions. It's really hard to change these APIs later once the code is in use, so I think it's best to do it in this PR. I'd suggest making a `SelectField` component that wraps the existing `Select` component. It might make it a bit easier to BYO label if people need to do something custom with the Select component.

mbylstra:
We ended up creating a `SelectField` component, so closing this.

PR #698, opened 2020-08-06
feat: add textarea for Elm

caseysinclair:
Adds TextArea component for Elm

Screen Shot 2020-08-07 at 10 19 48 am

https://trello.com/c/31VPBD0n/28-build-elm-textareafield

### DIFF
--- a/draft-packages/stories/TextAreaField.stories.elm
+++ b/draft-packages/stories/TextAreaField.stories.elm
@@ -80,7 +80,7 @@ main =
                             |> TextAreaField.controlled True
                             |> TextAreaField.autoComplete False
                             |> TextAreaField.onBlurWithValue TextFieldMsg
-                            |> TextAreaField.textAreaValue "A prefilled value in uncontrolled mode"
+                            |> TextAreaField.textAreaValue "A prefilled value in controlled mode"
                             |> TextAreaField.onEnter TextFieldEnter
                             |> TextAreaField.rows 4
                         )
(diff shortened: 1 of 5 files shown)


caseysinclair:
[ee](https://trello.com/c/31VPBD0n/28-ee)

draft-packages/stories/TextAreaField.stories.elm
```diff
@@ -0,0 +1,221 @@
...
+            \m ->
+                storyContainer
+                    [ TextAreaField.view
+                        (TextAreaField.default
+                            |> TextAreaField.id "the-id"
+                            |> TextAreaField.labelText "Your reply"
+                            |> TextAreaField.controlled True
+                            |> TextAreaField.autoComplete False
+                            |> TextAreaField.onBlurWithValue TextFieldMsg
+                            |> TextAreaField.textAreaValue "A prefilled value in uncontrolled mode"
```
mbylstra, line 83: This is actually in controlled mode: `TextAreaField.controlled True`
```suggestion
                            |> TextAreaField.textAreaValue "A prefilled value in controlled mode"
```

There's not currently a way to prefill values in uncontrolled mode.
It's generally better to use uncontrolled unless you need controlled mode as typing can be more sluggish, and browser plugins like Grammarly will break (you need to add a `data-grammarly=false` or the field will be unusable for people using grammarly). This is fine for now, but I'd like us to make a follow up PR that allows you to prefill the element in uncontrolled mode (and we should favor the uncontrolled version in our app code)

PR #704, opened 2020-08-12
feat: Add data-automation-class to generic notification

okeeffed:
This will add a ` data-automation-class` to each notification with their `type` and `style` to find a generic match as opposed to specific notification matches. 

Reference conversation: https://cultureamp.slack.com/archives/C2ZNME08P/p1597132217066600

Screen Shot 2020-08-12 at 10 52 23 am

### DIFF
--- a/packages/component-library/components/Notification/__snapshots__/GlobalNotification.spec.tsx.snap
+++ b/packages/component-library/components/Notification/__snapshots__/GlobalNotification.spec.tsx.snap
@@ -3,6 +3,7 @@
 exports[`The basic notification renders correctly 1`] = `
 <div
   class="notification affirmative global hidden"
+  data-automation-class="generic-notification affirmative global"
   style="margin-top: 0px;"
 >
   <div
@@ -62,6 +63,7 @@ exports[`The basic notification renders correctly 1`] = `
 exports[`You can change the type 1`] = `
 <div
   class="notification informative global hidden"
+  data-automation-class="generic-notification informative global"
   style="margin-top: 0px;"
 >
   <div
--- a/packages/component-library/components/Notification/__snapshots__/InlineNotification.spec.tsx.snap
+++ b/packages/component-library/components/Notification/__snapshots__/InlineNotification.spec.tsx.snap
@@ -3,6 +3,7 @@
 exports[`Persistent versions of the notifications work 1`] = `
 <div
   class="notification negative inline hidden"
+  data-automation-class="generic-notification negative inline"
   style="margin-top: 0px;"
 >
   <div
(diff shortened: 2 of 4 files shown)


mbylstra:
I just realised that Toasts are notifications,  so that was a bad example, but the point was more that words like "affirmative" and "global" are very generic and might be on other components, making it hard to narrow down to notifications/toasts.

mbylstra:
Oh , I get it now, the type is either "toast" or "notification" - been a while since I looked at this component. Forget what I said. Approved!

mbylstra:
Actually..... I should have looked into this more before speaking.....

it works well for` toast`, but the other variants `global` an `inline` are very generic words. I don't think it'd hurt to put `notification` in there as well: eg `"notification toast cautionary"` and `"notification global cautionary"`

okeeffed:
> Actually..... I should have looked into this more before speaking.....
> 
> it works well for` toast`, but the other variants `global` an `inline` are very generic words. I don't think it'd hurt to put `notification` in there as well: eg `"notification toast cautionary"` and `"notification global cautionary"`

Yeah, easy! I was thinking of having `notification` in there too but I wasn't sure at what point the class names would become superfluous.

mbylstra:
On the one hand they add extra dom weight, but on the other hand I think they are really nice for developers to be able to find out what a component is just by doing inspect element. Sometimes all you see is divs and spans, and obfuscated class names, so it gives you something to search for in a codebase if you're trying to find where something is implemened. I doubt the dom weight is significant.

okeeffed:
Okies @mbylstra I've updated!

> I doubt the dom weight is significant.

I think Frontend Ops were considering removing `data-automation-*` during build down the track based on environment, right? Might not be fun for now, but I think that might happen in the not-so-distant future?

mbylstra:
> I think Frontend Ops were considering removing data-automation-* 
what does this mean exactly? Do you mean replacing with `data-testid`?

mbylstra:
I was thinking about that. Maybe `data-test-class` is a better name for consistency with react testing library  or  `data-testclass` 🙄. I think changing the name of `data-automation-class` is less important than `data-automation-id` as the concept doesn't exist anywhere else as far as I know.

okeeffed:
Oh as in stripping those tags out using Webpack based on environment during the build! So it could still be used for testing, but just wouldn’t be in the client-facing production bundles. Sorry I just saw this on my phone @mbylstra :(

mbylstra:
I could go either way - they add weight to the DOM, but they are also quite useful for debugging, even in production. Eg: you get a support ticket for a page in production, but you can't figure out what part of the code base is generating it or reproducing it locally - you can look at the attrs to help figure out whether it's coming from a Kaizen component or not, which can then help you figure out where the source code is that's generating the page/element. I think being able to debug is more important actually - I doubt that attrs would have much of an impact on page size. I've had to spend hours in the past trying to figure where the hell a div is coming from.

Issue #708, opened 2020-08-13
changes to Elm files are not being watched by Storybook dev server


mbylstra:
## To reproduce

1. `yarn storybook`
2. keep the `yarn storybook` console output in sight
3. Make a change on any Elm file that's used by a story (eg `packages/component-library/components/Paragraph/Paragraph.elm` or `draft-packages/popover/KaizenDraft/Popover/Popover.elm`, doesn't matter if it's draft or core)
4. Nothing happens in the console, and the corresponding story is not updated in Storybook.

Actually, you can just hit save on the Elm file, and see that webpack is not picking it up as nothing happens in the console.

**Note:** this does work for the _story_ Elm file itself, such as `draft-packages/stories/PopoverStories.elm`. You will see webpack do a recompile, but you won't see any changes in the browser (which is a different issue: TODO add link). This is because this file is directly referenced by the tsx file that includes the Elm story, so Webpack already has it in it's watch list. The problem is that all Elm files be referenced by the Story are not being added to dev server watchlist.

mbylstra:
## Workaround when re-compilation isn't even being triggered, to avoid having to restart Storybook

1. Make change to component such as `packages/component-library/components/Box/Box.elm` . No rebuild is triggered.
2. Save the file of the accompanying story. Eg `packages/component-library/stories/Box.stories.elm`. This will trigger a rebuild.
3. Look at the console and wait for compilation to finish.
4. refresh the browser. You should see the changes.

mbylstra:
It seems that CSS changes do get tracked, which might be why this issue seems intermittent, but my findings have been that it's not itermittent at all - referenced Elm files are never tracked, and probably haven't been for a very long time (if you are only working on CSS you might not notice any issues)

Issue #676, opened 2020-07-28
Changes to Elm storybook files are not triggering page refresh or hot module reloading


mbylstra:
The changes are triggering a webpack rebuild, but are not causing the browser to refresh.

### To replicate:
- Edit either an Elm module or a css module that's used by an elm module.
- Wait until webpack finishes building the change
- The browser is not refreshed, and you can't see the changes.

### Workaround:
- keep the storybook webpack build in a window that's close at hand. Wait until it's finished and manually refresh. No need to restart storybook at least.

### Notes:

- I believe that Kaizen's storybook webpack config will be replaced with the frontend-ops package (which will have Elm storybook in future), so there's no point trying to fix this now. I'm just creating this issue in case anyone else runs into it.
- There seemed to be an older issue where you needed to completely restart storybook every time. It seems you no longer need to do that, or that was another edge case I haven't run into. Please add comments if you run into this! _Update:_ this seems to be always happening: I've created a separate issue for this, it's a different problem: https://github.com/cultureamp/kaizen-design-system/issues/708

mbylstra:
> There seemed to be an older issue where you needed to completely restart storybook every time. It seems you no longer need to do that, or that was another edge case I haven't run into. Please add comments if you run into this!

I am running into this working on an Elm version of the story for the Box component. If I change the Story code, then the code is recompiled, but if I make a change to the component at `packages/component-library/components/Box/Box.elm` then nothing happens at all in the webpack console.

## Workaround when re-compilation isn't even being triggered, to avoid having to restart Storybook

1. Make change to component such as `packages/component-library/components/Box/Box.elm` . No rebuild is triggered.
2. Save the file of the accompanying story. Eg `packages/component-library/stories/Box.stories.elm`. This will trigger a rebuild.
3. Look at the console and wait for compilation to finish.
4. refresh the browser. You should see the changes.

It's a sucky workflow, but **much better** than having to restart storybook (which takes several minutes on my machine)


 Make change to core component such as `packages/component-library/components/Box/Box.elm`

mbylstra:
@caseysinclair I'm curious if you ran into this (requiring full storybook restart) when you were working on Elm components recently?

PR #710, opened 2020-08-14
fix: watch for changes to Elm files in webpack

mbylstra:
This PR fixes https://github.com/cultureamp/kaizen-design-system/issues/708 ("changes to Elm files are not being watched by Storybook dev server"), with three independent fixes:

Two in this repo:

- https://github.com/cultureamp/kaizen-design-system/pull/710/files#diff-0427c3e900d18b166932d7979cf273fdR14 (`forceWatch` is required in elm-webpack-loader config)
- https://github.com/cultureamp/kaizen-design-system/pull/710/files#diff-af640f91fe686ac1d41cf39269f952a9R6
(draft-packages/stories needs to be included in the elm.json)

And one in elm-webpack-loader, which has been fixed and forked into this repo:

- https://github.com/cultureamp/kaizen-design-system/pull/710/files#diff-0427c3e900d18b166932d7979cf273fdR1
(it's watching way too many files) I'm meaning to get that into the actual package, but that will likely take a while, so this is a temporary fix to a very bad bug.

Please follow the links for more details.

Note that this does not fix Elm Hot Module reloading (https://github.com/cultureamp/kaizen-design-system/issues/676) ,and live reload is not enabled, so the only way to view changes is to refresh the browser. I also have a fix for that (currently being used in a front end microservice), but it's a completely separate issue, and will come in a future PR.

## Before:

Changes to Elm files did not trigger a webpack rebuild of the relevant Elm app. The only way to see changes would be to restart webpack, which often takes longer than 5 minutes (**so we were looking at a 5 minute feedback loop per change!**) (unless you knew the workarounds, which were still a very annoying experience). See https://github.com/cultureamp/kaizen-design-system/issues/708 for more details. Something that was particularly frustrating about this was that it seemed to sometimes work (actually it works on the Main storybook files, just not the referenced ones), so you never knew whether you were seeing new or old code, making debugging a nightmore (I had a bug in some elm Code which prompted me to go on this deep dive as I simply could not debug it!)

## During:
Elm files were now being picked up, but changing any file in Kaizen (tsx, .md, anything) would cause every single Elm app to be recompiled (takes nearly a minute), making a terrible dev experience for everyone. This would not have been releasable. This is why multiple fixes need to be released in one PR.

## After:
Changing Elm files causes a recompile of only relevant Elm apps. Eg:
- Change `Button.elm`: Only `Button.elm` and `Button.stories.elm` are recompiled (less than 10 seconds)
- Change `Button.tsx`: Only  Button.tsx and `Button.stories.tsx` are recompiled.
- Change `README.md`: nothing needs to be recompiled.
- etc


package.json
```diff
@@ -125,7 +125,11 @@
     "husky": "^4.2.5",
     "jscodeshift": "^0.7.0",
     "lint-staged": "^10.2.2",
-    "storybook-chromatic": "^3.4.1"
+    "storybook-chromatic": "^3.4.1",
+    "glob": "^7.1.1",
+    "loader-utils": "^1.0.2",
+    "node-elm-compiler": "^5.0.0",
+    "yargs": "^6.5.0"
```
mbylstra, line 132: These are required for the forked version of `elm-wepack-loader`. Originally I tried to make a force a yarn workspace at `packages/elm-webpack-loader` (with `private: true` so it wouldn't get published), and this worked in local developement, but the CI publish script couldn't find it. I gave up and put it in `storybook/elm-webpack-loader-fix.js`

storybook/elm-webpack-loader-fix.js
```diff
@@ -0,0 +1,219 @@
...
+var loaderUtils = require("loader-utils")
+var elmCompiler = require("node-elm-compiler")
+var yargs = require("yargs")
+
+var runningInstances = 0
+var alreadyCompiledFiles = []
+
+var defaultOptions = {
+  cache: false,
+  forceWatch: false,
```
mbylstra, line 14: This was the first bug I found. Normally you don't need to include this, because `node-elm-compiler` infers that webpack is in watch mode by parsing the command line arguments (https://github.com/elm-community/elm-webpack-loader/blob/master/index.js#L79), but Storybook does not use the dev server, it has it's own custom server, presumably with watch mode configured somewhere. Fixing this caused the problem where too many files were being watched.

storybook/webpack.config.ts
```diff
@@ -112,11 +112,12 @@ export const elm: Rule = {
       loader: "elm-hot-webpack-loader",
     },
     {
-      loader: "elm-webpack-loader",
+      loader: resolve(__dirname, "elm-webpack-loader-fix.js"),
       options: {
         debug: false,
         cwd: resolve(__dirname, ".."),
         pathToElm: resolve(__dirname, "../node_modules/.bin/elm"),
+        forceWatch: true,
```
mbylstra, line 120: This was the first bug I found. Normally you don't need to include this, because node-elm-compiler infers that webpack is in watch mode by parsing the command line arguments (https://github.com/elm-community/elm-webpack-loader/blob/master/index.js#L79), but Storybook does not use the dev server, it has it's own custom server, presumably with watch mode configured somewhere. Fixing this caused the problem where too many files were being watched.

storybook/elm-webpack-loader-fix.js
```diff
@@ -0,0 +1,219 @@
+"use strict"
```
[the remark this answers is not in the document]
sentience: Could maybe use a comment in the file explaining that it's a fork, and pointing to a public GitHub issue on the loader explaining the need for it.

storybook/webpack.config.ts
sentience: We needed to `forceWatch` in Murmur for awhile, when we were using a custom programmatic invocation of webpack-dev-server rather than the standard NPM binary for this. Being able to remove `forceWatch` from our Elm loader was one of the main drivers for switching to "standard" webpack-dev-server. I understand this isn't an option for Kaizen's webpack config, since it's all wrapped in Storybook.

PR #713, opened 2020-08-17
feat: Elm Box component

mbylstra:
This is a the Elm version of the React Box component.

Example: heading with 1 grid unit margin bottom:

```elm
box [ marginBottom 1 ]
    [ Heading.view
        (Heading.h1 |> Heading.variant Display0)
        [ Html.text "Display0" ]
    ]
```

Storybook: https://dev.cultureamp.design/mab/elm-box-inline-styles/storybook/?path=/story/box-elm--box-default

_Notes:_

## It uses inline styles
Not even a CSS-in-JS or CSS-in-Elm framework, just good old fashioned inline styles! ☠️ ☠️ ☠️ 
I know!! But hear me out:
- We've decided to go with a "component only" approach for the Box (in TS and Elm), where you don't have to write a line of CSS. I think this is very good for decoupling the design system from the underlying technology (scss). But it has a problem:  css classes are finite, but we want to be able to create a near-infinite number of classes, one for each spacing property `padding-right`, `margin-left` times each possible value (0.25, 0.5...4.5, etc), so you get a huge number of classes that need to be manually mapped from class names to elm record fields. https://github.com/cultureamp/elm-css-modules-loader uses "records" for it's elm-css implementation, but in Elm there is no way to programmatically convert class names to record fields, like how it's done in the Typescript version (because objects in JS are more like `Dicts` than Elm records): In Elm you need a `Dict` for that. So, the amount of code required is huge. I started doing this in an earlier branch (https://github.com/cultureamp/kaizen-design-system/commit/71d2f1c5a3cadd4b1e95e231e4f12b4a17fc9b1f#diff-b13bd9f0979215a31560b234f1a77367R25), but it was a crazy amount of work and would be error prone and hard to maintain, and wonder if there's a better way we can do it in the future.

Although this is not perfect, I think it's better than _not_ having a Box component at all, and it's an implementation detail that can be changed without having to necessarily change the public API (much). Having to find scss and upgrade that to use a Box component instead is far harder - I really don't think it would be possible to automate that, but it'd be far more feasible to automatically upgrade pure Elm code.

## It uses a different style of API than the other components
Rather than "pipeline config" style:

```elm
Box.view 
    (Box.defaults |> marginTop 1 |> marginRight 0.5)
    [ text "content in the box" ]
```
it looks like this:
```elm
box 
    [ marginTop 1, marginRight 0.5] 
    [ text "content in the box" ]
``` 

I think the pipeline config style makes sense for more complex components, but this is far less verbose for simple components like box that should be quicker to use than having to write scss. It does requires very verbose implementation code, but I think we should optimise APIs for Kaizen consumers rather than Kaizen maintainers. The other place I've seen this pattern is https://github.com/rtfeldman/elm-css#elm-css (to be clear, this style of API has nothing to do witih css-in-Elm or inline styles, that's just a coincidence, you could use it for any API, but I think it works well as a DSL for html as it looks a lot like actual html)

## no ability to change the sizes responsively
- I think we should explore the potential of being able to use custom breakpoints using dynamic JS or Elm (could be dynamically generated `<style>` tags). Using classes limits us to a finite set of breakpoints. I didn't think it was worth doing this for this PR. You can always use scss instead of the Box component if you need responsive styles.

## No RTL support
- If RTL support is needed, let's open a ticket. I also think that use of `"[dir=rtl]"` css selectors is a better approach than having to prop drill a rtl/direction prop. Again, I thin it's something we could explore with CSS-in-Elm/JS solutions.

### DIFF
--- a/packages/component-library/components/Spacing/Spacing.elm
+++ b/packages/component-library/components/Spacing/Spacing.elm
@@ -1,0 +1,249 @@
+module Spacing.Spacing exposing (Config, GridFraction, Property, empty, inlineStyles, margin, marginBottom, marginLeft, marginRight, marginTop, marginX, marginY, padding, paddingBottom, paddingLeft, paddingRight, paddingTop, paddingX, paddingY, setProperty)
+
+import Dict exposing (Dict)
+import Html exposing (Attribute)
+import Html.Attributes exposing (style)
+
+
+type alias GridFraction =
+    Float
+
+
+gridInRem =
+    -- TODO: get this from @kaizen/design-tokens when it eventually supports Elm
+    1.5
+
+
+
+-- config
+
+
+type alias Config =
+    Dict String GridFraction
+
+
+empty =
+    Dict.empty
+
+
+padding : GridFraction -> Property
+padding n =
+    Padding n
+
+
+paddingTop : GridFraction -> Property
+paddingTop n =
+    PaddingTop n
+
+
+paddingRight : GridFraction -> Property
+paddingRight n =
+    PaddingRight n
+
+
+paddingBottom : GridFraction -> Property
+paddingBottom n =
+    PaddingBottom n
+
+
+paddingLeft : GridFraction -> Property
+paddingLeft n =
+    PaddingLeft n
+
+
+paddingX : GridFraction -> Property
+paddingX n =
+    PaddingX n
+
+
+paddingY : GridFraction -> Property
+paddingY n =
+    PaddingY n
+
+
+margin : GridFraction -> Property
+margin n =
+    Margin n
+
+
+marginTop : GridFraction -> Property
+marginTop n =
+    MarginTop n
+
+
+marginRight : GridFraction -> Property
+marginRight n =
+    MarginRight n
+
+
+marginBottom : GridFraction -> Property
+marginBottom n =
+    MarginBottom n
+
+
+marginLeft : GridFraction -> Property
+marginLeft n =
+    MarginLeft n
+
+
+marginX : GridFraction -> Property
+marginX n =
+    MarginX n
+
+
+marginY : GridFraction -> Property
+marginY n =
+    MarginY n
+
+
+setProperty : Property -> Config -> Config
+setProperty param config =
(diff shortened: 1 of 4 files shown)


mbylstra:
[Elm Box component](https://trello.com/c/a8IjZ7Nh/43-elm-box-component)

packages/component-library/components/Spacing/Spacing.elm
```diff
@@ -0,0 +1,249 @@
...
+                        "padding-y" ->
+                            [ style "padding-top" cssValue
+                            , style "padding-bottom" cssValue
+                            ]
+
+                        _ ->
+                            []
+
+                else
+                    [ style property cssValue ]
```
kemblesong, line 208: Are we deferring the decision to introduce https://github.com/rtfeldman/elm-css here?

mbylstra: Yeah, I certainly thought about, but this is the problem, you'd need to either:
A: have one elm-css instantation per fox instance (and make elm-css an implementation detail that consumers don't know about). Maybe it's fine having one `<style>` element per box, even if there are hundreds of boxes on the page, I'm really not sure, and it would take time to asses the performance of that.
B: get all consumers to buy into using elm-css, and having to do the elm-css boilerplate, just to use the box component which is mostly meant to speed you up. If decide to go all in on elm-css, it makes sense to do it this way.

mbylstra: Is there any elm-css code in production at Culture Amp?

kemblesong: Not that I'm aware of.
Those are good points though 👍

kemblesong (approved):
LGTM, but I'm not an advocate, I'm just passing by 😄

mbylstra:
Thanks, as you are heavy user of Elm and Kaizen at Culture Amp, I am very happy to accept your review!

ActuallyACat (approved):
Nice one 👌🏻

PR #715, opened 2020-08-18
feat: create Avatar draft component

sampnorris:
I've followed the directions from the Avatar page in the kaizen guide.

https://cultureamp.design/components/avatar/

Screenshot below, check out design preview branch!

Screen Shot 2020-08-18 at 1 41 45 pm
Screen Shot 2020-08-18 at 1 41 40 pm
Screen Shot 2020-08-18 at 1 41 36 pm
Screen Shot 2020-08-18 at 1 41 32 pm


sebpearce:
Looks excellent 👍 

It got me wondering about whether the initials logic should be a Kaizen thing or a consumer-side thing. Usually we want to keep components as dumb as possible, but in this case I can see a good argument for Kaizen making life easier for people if it takes care of the logic here.

katsambarcus:
Thanks, Sam. Looking good. A couple of questions:

**Sizes:** It looks like we've catered for the Large (60px) and Medium (48px) sizes. Is there a Small avatar (20px) in the works or consideration for it in the future? See [Avatar Sticker Sheet](https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A14305)

**Naming:** There are some inconsistencies in the story names when compared to the UI Kit. E.g. 'With Avatar Path' vs 'Photo Personal' or 'with broken avatar' vs 'Default User'. Is this the space where we should be trying to keep language aligned in an effort to 'speak the same language'?

**Inherit container:** I like that we have considered how the text can scale down to maintain a standard size of the avatar, instead of it blowing out. But what's the rationale for showing an avatar that has more than 2 characters? Doesn't the avatar pull the first letter from first and last name only? I wonder if this adds unnecessary complexity for engineers to make decisions?

**Initials Personal (Peach) vs Initials Others (Wisteria):** This pattern exists to help a user who doesn't have an avatar quickly scan the screen to see what things are relative to them versus other users. Has this been factored into the component?

sampnorris:
Thanks @sebpearce 
I hadn't even considered those, but glad they seem unnecessary, I added some stories to consider other symbolic languages like Japanese and Chinese Simplified.

sebpearce:
@sampnorris Ah sorry, I should've been clearer... initials don't exist in Japanese/Chinese etc, so it's not an issue. I was only thinking about those names being represented in English letters.

E.g. 村上春樹 = Murakami Haruki = MH. But actually it'd be HM, because those initials wouldn't be used outside a Western context as far as I know. So I think it's fine to just leave it.

sampnorris:
@katsambarcus 

Hey Marc, Happy to rename these to follow design language!

As for `small` I have been going off of this design file which looks the same but has no small option.
https://www.figma.com/file/e3TyAOZKfLg9cA01Iy1Mgm/Kaizen-Site-embed?node-id=17270%3A114
So I will update for a small version of the component easy peasy!

**Inherit**
I inherit container is to show a custom size, in a case where the user might want to stretch the container, the font size stays the same as the specified  but the circle gets bigger. This will help use 

You could set "inherit" and "small" where small will control the size of the font and inherit will size the container. Check out the design branch again as I have updated naming and stories to match designs so this will make more sense.

**3 or more letters**
As for initials - based off the design guidelines on cultureamp.design that reads:

```
We tend to use the first letter of each space-separated word for initials and capitalize them. You may need to resize the initials text for more than 4 names.
```
https://cultureamp.design/components/avatar/

See link above for where design logic was built from.

sampnorris:
@sebpearce Let's remove those stories and keep current behaviour! I'm sure someone else would be better tasked at investigating how this works. This logic is the same as the murmur avatars so this will be consistent.
Any changes can be added to this component and rolled out in the _future_

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,103 @@
...
+   * Src for the avatar image to load, if not passed we will derive initials from the full name.
+   */
+  avatarSrc?: string
+  /**
+   * Shows a different background colour if the avatar is the current user.
+   * @default "true"
+   */
+  isCurrentUser?: boolean
+  /**
+   * There are 2 fixed avatar sizes.
```
ActuallyACat, line 40: /nit `There are 3 fixed avatar sizes.`

sampnorris: Nice pick-up!

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type AvatarSizes = "small" | "medium" | "large"
+
+const getInitials: (fullName: string, max2Characters?: boolean) => string = (
+  fullName,
+  max2Characters = false
+) =>
+  fullName
+    .split(/\s/)
+    .reduce((acc, name) => `${acc}${name.slice(0, 1)}`, "")
```
sebpearce, line 19: Love it.

sebpearce:
@sampnorris ⚠️ I just noticed that in the story for this component, the avatar sizes are out of whack. They need `box-sizing: border-box` to look right.

in other words the white circle is part of the dimensions – this is also how it works in the Title Block avatar.

draft-packages/avatar/KaizenDraft/Avatar/styles.module.scss
```diff
@@ -0,0 +1,96 @@
...
+
+  &.otherUser {
+    background: $kz-color-wisteria-200;
+  }
+
+  // Sizes
+
+  &.small {
+    width: 1.25rem;
+    height: 1.25rem;
```
sebpearce, line 31: Suggest doing these in multiples of `$ca-grid` – that's the pattern that's encouraged in Kaizen. Makes it easier to reason about.

You can get it from `@import "~@kaizen/component-library/styles/grid";`.

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
sampnorris: Functional First <3

draft-packages/avatar/KaizenDraft/Avatar/styles.module.scss
sampnorris: Sweet, I thought that might have been exclusively for padding / margin but i'll update these

sampnorris:
@sebpearce  Border Box Added, thanks I'm sure that would have been annoying to run into when I went to use it hahaha

sebpearce (approved):
LGTM, as long as design review is happy then I am too 👍

PR #716, opened 2020-08-18
feat: Add report switcher to Title Block

sebpearce:
## Changes

- Adds a `secondary-small` variant to the Select component:

image

- Adds the report switcher to the Title Block:

- Restructures the breadcrumb so that it works the same way but doesn't cause exponentially-worse alignment issues as things get more complex. Without this improvement, the report switcher would have required this code that was already brittle to get even more brittle. Now the issue is moot.


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.scss
```diff
@@ -364,60 +374,13 @@ $tab-container-height-small: $ca-grid * 2.5;
...
   transition: opacity 0.3s ease;
   min-width: 0; // this is an important trick to prevent flexbox items from overflowing
+  align-items: center;
 
-  .breadcrumb:hover + &,
-  .breadcrumb:focus + & {
+  .breadcrumb:hover ~ &,
+  .breadcrumb:focus ~ &,
+  .breadcrumbTextLink:hover ~ &,
+  .breadcrumbTextLink:focus ~ & {
```
sentience, line 382: 👏

sampnorris (approved):
LGTM

PR #687, opened 2020-08-04
feat: Add hierarchical menu and hierarchical select components

lijuenc:
# Summary of changes

Adds two new components: `@kaizen/draft-hierarchical-menu` and `@kaizen/draft-hierarchical-select`.

The HierarchicalMenu component handles:
- layout
- animations and transitions
- keyboard navigation

The HierarchicalSelect is a trigger component that wraps the HierarchicalMenu and handles:
- opening/closing the menu
- data retrieval
- selection state

Screen Shot 2020-08-17 at 2 15 04 pm

HierarchicalMenu exists as its own component because it's not limited to being used as a dropdown menu and could theoretically be hooked up to any other type of trigger component in the future.

**Live storybooks**
- HierarchicalMenu
https://dev.cultureamp.design/lijuenc/hierarchical-menu/storybook/?path=/story/hierarchicalmenu-react--default-story
- HierarchicalSelect
https://dev.cultureamp.design/lijuenc/hierarchical-menu/storybook/?path=/story/hierarchicalselect-react--default-story

# Designs

- in Reporting context
https://www.figma.com/file/gKZo7BXAmV3A97GFF0ApyW/Leader-based-reporting?node-id=470%3A3
- prototypes
https://www.figma.com/proto/gKZo7BXAmV3A97GFF0ApyW/Leader-based-reporting?node-id=472%3A1199&viewport=765%2C-16466%2C1&scaling=min-zoom

# Not yet done but will be done in follow up patch releases
- Verify full compatibility on IE11 (functionality wise it's currently good, however there are some layout/styling issues to work through)
- Truncate long option names with ellipses. Show a tooltip with full value using the native HTML title attribute
- Improve animations to show new options transitioning in, instead of appearing in place
- Clean up focus styles

PR #718, opened 2020-08-19
feat: Hierarchical menu and hierarchical select improvements [PHRI-32]

lijuenc:
Follow up to https://github.com/cultureamp/kaizen-design-system/pull/687. Fixes the known issues from that PR and some more.

**Screenshot:**

Screen Shot 2020-08-19 at 11 46 46 am

**Storybook:**
https://dev.cultureamp.design/lijuenc/hierarchical-menu-improvements/storybook/?path=/story/hierarchicalselect-react--default-story

**Menu and select:**
- Make focus styles consistent with other Kaizen inputs and buttons

**Menu:**
- Truncate long option names with ellipses (screenshot above)
- Support disabled state for drill down button (<kbd>></kbd> in last option in screenshot above)
- Make current hierarchy node clickable ("Didier Elzinga" in screenshot above) so that the current level can selected without having to go back to the parent
- IE11 compatibility for:
  - keyboard navigation (<kbd>↑</kbd>, <kbd>↓</kbd>, <kbd>←</kbd>, <kbd>→</kbd>, <kbd>Space</kbd>, <kbd>Enter</kbd>)
  - some styling and layout issues

**Select:**
- Close menu when <kbd>Escape</kbd> is pressed
- Clear select value when <kbd>Backspace</kbd> or <kbd>Delete</kbd> is pressed
- Clear select value when cross icon button is clicked (screenshot above)
- IE11 compatibility for:
  - keyboard navigation (<kbd>↑</kbd>, <kbd>↓</kbd>, <kbd>←</kbd>, <kbd>→</kbd>, <kbd>Space</kbd>, <kbd>Enter</kbd>, <kbd>Backspace</kbd>, <kbd>Delete</kbd>)
  - some styling and layout issues

**Known issues:**
- The cross icon button does not work in IE11 (but works fine in Chrome, Firefox, Safari, IE Edge). The workaround for this for now is to use the <kbd>Backspace</kbd> or <kbd>Delete</kbd> key to clear the select value

### DIFF
--- a/draft-packages/hierarchical-menu/KaizenDraft/HierarchicalMenu/HierarchicalMenu.tsx
+++ b/draft-packages/hierarchical-menu/KaizenDraft/HierarchicalMenu/HierarchicalMenu.tsx
@@ -275,9 +275,14 @@ const Menu = (props: MenuProps) => {
             </button>
           </div>
           <div className={styles.current}>
-            <Text style="body-bold" tag="p" inheritBaseline>
-              {hierarchy.current.label}
-            </Text>
+            <button
+              className={styles.currentButton}
+              onClick={() => onSelect(hierarchy, hierarchy.current)}
+            >
+              <Text style="body-bold" tag="p" inheritBaseline>
+                {hierarchy.current.label}
+              </Text>
+            </button>
             <Text style="small" tag="p" inheritBaseline>
               Level {hierarchy.current.level}
             </Text>
--- a/draft-packages/hierarchical-menu/KaizenDraft/HierarchicalMenu/HierarchicalMenu.tsx
+++ b/draft-packages/hierarchical-menu/KaizenDraft/HierarchicalMenu/HierarchicalMenu.tsx
@@ -313,10 +313,9 @@ const Menu = (props: MenuProps) => {
                       <button
                         className={styles.childLabelButton}
                         onClick={() => onSelect(hierarchy, c)}
+                        title={c.label}
                       >
-                        <Text style="body" tag="p" inheritBaseline>
-                          {c.label}
-                        </Text>
+                        <div className={styles.childLabelText}>{c.label}</div>
                       </button>
                       {c.numberOfChildren > 0 && (
                         <button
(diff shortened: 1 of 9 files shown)


makenosound (commented):
This is feeling super tight, @lijuenc. The keyboard navigation is 😍 and all feels super natural, and the animations feel nice in all situations for me now.

Only thing that looks slightly awry is the padding of the elements:

image

I think the text within the Select and the Menu should probably align.

lijuenc:
> I think the text within the Select and the Menu should probably align.

I see what you mean @makenosound.

Does this look odd to you? The horizontal spacing is a lot less than the vertical spacing.

Screen Shot 2020-08-19 at 1 12 48 pm

If we equalise the horizontal and vertical spacing it looks like this which is actually a lot more condensed than say, the [Kaizen dropdown](https://cultureamp.design/storybook/?path=/story/dropdown-react--control-action).

Screen Shot 2020-08-19 at 1 20 41 pm

By the way, using the [Kaizen tokens for spacing](https://github.com/cultureamp/kaizen-design-system/blob/master/packages/design-tokens/tokens/spacing.json) means that there are pretty big differences between some levels (for example, `sm` to `md` is double).

```
"xs": "0.375rem",
"sm": "0.75rem",
"md": "1.5rem",
"lg": "2.25rem",
"xl": "3rem",
"xxl": "3.75rem",
"xxxl": "4.5rem",
"xxxxl": "5.25rem",
"xxxxxl": "6rem"
```

The screenshot in the PR description uses `xl` while the screenshots in this comment take it down one level to `lg` spacing.


(keeping in mind that we have to also maintain alignment with the Select Report control above).

makenosound:
I think the first one looks OK, though a bit hard to tell when there’s no selected row. It’s definitely fine for there to be a difference between horizontal/vertical padding.

The latter is definitely too compressed, yeah.

lijuenc:
Ok, adjusted, @makenosound.

makenosound (approved):

Issue #702, opened 2020-08-10
Update generator with automation-id


ActuallyACat:
We're currently adding `automation-id` to each of our components in Kaizen (thanks @okeeffed !). To ensure future components have this prop, we should add it to the generator package.

PR #720, opened 2020-08-19
fix: Added Readme and automation id to generator

ActuallyACat:
## About 
As much as possible, draft components should generate boilerplate code with sensible defaults for our components. In this PR I have added: 
* Automation id. This is a default we're moving to for all components so we can be confident we can target Kaizen components in test environments.
* READMEs for the cultureamp.design. This will (hopefully) prompt the PR-opener to consider if we need to update documentation, and if so, reach out to the DS channel for advice from an advocate. 
* Some custom markdown with information about imports to display on the Storybook info panel. 
* Information about how to update to the latest generator. 

This fixes https://github.com/cultureamp/kaizen-design-system/issues/702


packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
+---
```
ActuallyACat, line 1: @didoesdigital Not sure what a sensible default for this file should be. I'm hoping this will prompt the contributor to reach out to a designer for advice.

packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
...
+
+import WhenToUseAndWhenNotToUse from "docs-components/WhenToUseAndWhenNotToUse"
+import WhenToUse from "docs-components/WhenToUse"
+import WhenNotToUse from "docs-components/WhenNotToUse"
+
+<WhenToUseAndWhenNotToUse>
+
+<WhenToUse>
+* Use this for the thing
```
didoesdigital, line 16: From memory, I think these also need line breaks around them to work.

didoesdigital: `* Use <%= componentName %> when… `

packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
...
+</WhenToUse>
+
+<WhenNotToUse>
+* Don't use this for the other thing though!
```
didoesdigital, line 20: `When X, use Y component instead.`

packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
...
+* Use this for the thing
+</WhenToUse>
+
+<WhenNotToUse>
+* Don't use this for the other thing though! 
+</WhenNotToUse>
+</WhenToUseAndWhenNotToUse>
+
+## See also
+
```
didoesdigital, line 25: It's been useful to designers to see Markdown syntax examples. Perhaps show example link to commonly referenced component:

`* [Button](/components/button/)`

packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
...
+## External links
+
```
didoesdigital, line 27: As above, perhaps some examples:

```
- [Polaris: button](https://polaris.shopify.com/components/actions/button)
- [Atlassian: button](https://www.atlassian.design/components/button/examples)
- [Spectrum: button](https://spectrum.adobe.com/page/button/)
- [Lightning: button](https://www.lightningdesignsystem.com/components/buttons/#site-main-content)
```

… although there is a risk they won't be updated

packages/generator/generators/draft/templates/README.txt
```diff
@@ -0,0 +1,27 @@
+---
+title: "<%= componentName %>"
+navTitle: "<%= componentName %>"
+summaryParagraph: <%= componentName %>s are a neat component
+tags: [""]
+demoStoryId:  <%= componentName %>-react--default-kaizen-site-demo
+---
+
```
didoesdigital, line 8: Is it worth including DOs and DONTs imports for copy guidelines even without adding default sections for copy guidelines? Just so people can use them without needing to know about the imports? 

```
import DoAndDontContainer from "docs-components/DoAndDontContainer"
import Do from "docs-components/Do"
import Dont from "docs-components/Dont"
```

packages/generator/generators/draft/templates/README.txt
didoesdigital: @ActuallyACat Nice one. It's a good prompt.

Change `README.txt` to `README.md`?

ActuallyACat: This is just a template file - it will be copied, have the variables injected into it, and spit out as an `mdx` file: https://github.com/cultureamp/kaizen-design-system/pull/720/files#diff-53cc4ef3807f9e09838ac70f0218da41R49-R50

packages/generator/generators/draft/templates/README.txt
ActuallyACat: Will `TODOs` suffice?

PR #721, opened 2020-08-20
fix: Update SplitButton stories with draft-menu

makenosound:
The stories for the React `<SplitButton>` were using the Menu* components from the deprecated version. This updates the stories to use the equivalent components from `@kaizen/draft-menu`.


sebpearce (approved):
Love it.

PR #722, opened 2020-08-20
docs: Updated modal docs

ashleecharetteCA:
- Added roadblock
- Moved copy guidelines 
- Adjusted design guideline


caseysinclair:
[Fix TextAreaField so that it only shows the validation message when there is an actual error](https://trello.com/c/raq4GyBx/60-fix-textareafield-so-that-it-only-shows-the-validation-message-when-there-is-an-actual-error)

PR #725, opened 2020-08-24
feat: Add security-tip icon

lijuenc:
Adds the security-tip icon from the UI kit to the component library.

We'll be using this in report confidentiality and report filtering protections. Some contextual examples:

Screen Shot 2020-08-24 at 11 36 12 am

Screen Shot 2020-08-24 at 11 35 35 am


ActuallyACat (approved):
👍 Thanks for removing all the crud that sketch/figma generates too.

Issue #83, opened 2019-10-25
Site package contains untyped javascript


zioroboco:
There is still some javascript files left over from porting pages from `cultureamp-style-guide`.

After these have been upraded, `allowJs: false` can be set in the root `tsconfig.json`.

PR #728, opened 2020-08-24
fix: Set allowJs to false as we're now not using untyped javascript

ActuallyACat:
Fixes:  #83

Does what it says on the box. I've set the allowJs flag to false. 

Reference: https://www.typescriptlang.org/docs/handbook/compiler-options.html


zioroboco (approved):
Yay, nice.

This also (accidentally?) reformats a buuunch of template strings to double quotes, but it isn't touching any packages. Happy to approve the change to tsconfig — you can decide if you want to clean up.

ActuallyACat:
I've gone through and fixed up the main ones I could find in the site config. I'm hoping it catches everything going forward 🤞

PR #730, opened 2020-08-25
fix: ensure validation message is only shown when the actual status i…

caseysinclair:
Fixes an issue where the validation message was showing regardless of an error state


caseysinclair:
[Fix TextAreaField so that it only shows the validation message when there is an actual error](https://trello.com/c/raq4GyBx/60-fix-textareafield-so-that-it-only-shows-the-validation-message-when-there-is-an-actual-error)

mbylstra (approved):

PR #731, opened 2020-08-25
feat: Elm SelectField component

caseysinclair:
PR description by @mbylstra

This adds a new `SelectField` component for elm which wraps the existing `Select` component with standard field features including label, error message and description.

image

### Stories

single select (interactive demo with error state - it removes when you select an item): http://localhost:8081/?path=/story/selectfield-elm--single

multi select still works: https://dev.cultureamp.design/cs/kaizen/select-with-validation/storybook/?path=/story/selectfield-elm--multi-select

### Implementation notes

**the view type**
This was a tricky one, as the Select component view has the type `Html (Msg item)` (a custom msg type) and requires using `Html.map` to use. We could not for the life of us figure out a way to re-use the Msg type from the `Select` component and get our component to compile. Instead of having a custom msg type for the view, the type of the view of this component is simply `Html msg` (the consumer does not need to use `Html.map`). However, the consumer must pass in a `toMsg` function, of the type `Select.Msg item -> msg`, which is what you would pass in to `Html.map`. This is a pretty idiomatic way to create reusable interactive components in Elm, but it is now unfortunately inconsistent with the Select API (if you want to just use a Select component with no label/description/message)

**missing features and improvements to be made**

I'm afraid we had to drop a few things due to time constraints, and some things aren't as nice as they could be:

- we started implementing aria attributes for the description and message fields, but dropped it. At the very least the label hooked up to the hidden select input using `for`
- there is no Cautionary (yellow) status as per the Zen design
- As a consumer you need to import both `Select` and `SelectField` to use the `SelectField` component. It'd be nice if you only had to import `SelectField` and the use of `Select` was left as an implementation detail. It requires a lot of boilerplate to do this (eg: we'd need to create a new Msg type that just proxies the Select msg type). I'd like to do this in cool down before this component is used much (it will need to be a breaking change).
- there is code duplication in the `Configuration` record between `Select` and `SelectField`. This should be able to be cleaned up with an extensible record to share fields.

This PR was authored mostly by @caseysinclair with @mbylstra helping finish it off (which is not very apparent in the commits)


mbylstra (approved):

caseysinclair (commented):
All LGTM, good PR notes too!

PR #733, opened 2020-08-25
fix: title block secondary actions tablet view bugfix

sampnorris:
Previously there was a bug where the secondary actions did not show on a tablet view when the absence of either a primary or default action.

Image shows the secondary menu which was not shown when other 2 buttons were not present.

Screen Shot 2020-08-25 at 5 36 20 pm


sebpearce (approved):
LGTM

PR #739, opened 2020-08-28
fix: Change default label position to end

dougmacknz:
Small mistake was made in https://github.com/cultureamp/kaizen-design-system/pull/665 which has resulted in checkboxes and radios having no margin:
image

This is a guess at a fix, I'm just going to let this build so that I can check.


dougmacknz:
Superseded by https://github.com/cultureamp/kaizen-design-system/pull/740

PR #740, opened 2020-08-28
fix: Realign label margins for Checkbox and Radio

lijuenc:
Fixes an issue where label margins were not present for Checkbox and Radio.

Checked:
- CheckboxField (Elm)
- CheckboxField (React)
- CheckboxGroup (React)
- Radio (Elm)
- Radio (React)
- RadioGroup (React)


ActuallyACat:
I'm guessing this is related to this PR? Should the other one be closed in favour of this one? https://github.com/cultureamp/kaizen-design-system/pull/739

cc/ @dougmacknz

lijuenc:
Yup, I was just on a zoom with Doug, so this PR includes that too.

draft-packages/form/KaizenDraft/Form/Primitives/Label/styles.scss
```diff
@@ -41,9 +41,13 @@ $label-start-margin: 10px;
 
   display: flex;
 
-  .labelText {
+  .prependedLabel {
+    @include ca-margin($end: $label-start-margin);
+    order: -1; // place label before the control
+  }
+
+  .appendedLabel {
```
dougmacknz, line 49: Perhaps we should just bring these styles out of this nesting, and that way we should only need to have it written once.

If that doesn't work, I don't think there's any way to end up with a `prependedLabel` with a checkbox or radio, so we could delete that class from here.

lijuenc (commented):
Latest fix looks good 👍

dougmacknz (approved):

PR #741, opened 2020-08-28
fix: consider icon position end when additional content is used on generic button

sampnorris:
I have extended the functionality of the hidden `additionalContent` api, to position it before the icon when `iconPosition` is set to `end`. 
This considers the margin needed to space additional content and icon apart.

I have searched our major repos and there is currently no use case where `iconPosition` is set to `end` as well as `additionalContent` being provided. 

This produces a button where the additional content - in this case a badge - can live before the icon. And does not effect the filter drawer button.

Screen Shot 2020-08-28 at 3 49 28 pm

### DIFF
--- a/draft-packages/button/KaizenDraft/Button/components/GenericButton.module.scss
+++ b/draft-packages/button/KaizenDraft/Button/components/GenericButton.module.scss
@@ -77,6 +77,10 @@
   @extend %caButtonIconWrapper;
 }
 
+.additionalContentWrapper {
+  @extend %caButtonAdditionalContentWrapper;
+}
+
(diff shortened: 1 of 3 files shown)


mbylstra (approved):
Looks good! If you're wondering what it was created for initially, it was for the the additional icon in the filter drawer:

image.

 I've checked the branch preview, and it is rendering the same as before 👍 

I'm not aware of any other usages. We weren't sure if there would be a use case where the icon is at the end so we didn't design for that - now there is!

sampnorris:
@mbylstra  Yea I saw what it was used for before! Thanks for checking it out :)

PR #744, opened 2020-08-28
fix: Add border-box as a default style on new draft components 

ActuallyACat:
## About 
This will prevent future issues with border-box being inherited by whatever is set in the consuming repo. In the future, this can be expanded to a small utility function to share with other components.


sentience (approved):
Some whitespace changes in this PR. Are they wanted? Anything we should be doing to make these consistent?

ActuallyACat:
> Some whitespace changes in this PR. Are they wanted? Anything we should be doing to make these consistent?
Yup, they're intentional. The template generates code that's not formatted properly according to Prettier, so the whitespace changes address that.

Issue #701, opened 2020-08-10
Handling deprecation in Kaizen


ActuallyACat:
### Problem
The need to deprecate things is a side effect of a healthy design system. As we learn more about how our components are used we may want to make improvements, or move in another direction and remove the component from the system. This issue covers the latter case.

More specifically: 
* How do we indicate a prop is deprecated?
* How do we indicate a component is deprecated?
* How do we indicate a Sass mixins/variables are deprecated? 
* How do we show that a component is deprecated in Storybook? At what point do we remove the story? 

Extra discussion: 
* When should the deprecated thing be removed from the system?

ActuallyACat:
Here's my 2c: 

## How do we deprecate props?
### Deprecating a property of the type/interface

The interface for tabs previously contained the property `direction` (with the String Literal Type values of `"row" | "column" | undefined`) which controlled whether the tabs would render horizontally or vertically. We decided that a better name would instead be `orientation` with the values (`"horizontal" | "vertical" | undefined`). 
```
interface Props {
  /**
   * Renders the tab in a column or a row
   * @deprecated Use orientation property instead
   */ 
  direction?: "column" | "row"
  /**
   * Renders the tab in a horizontal or vertical orientation
   */
  orientation?: "horizontal" | "vertical"
}
```

### Deprecating some options of a property
Not too sure about this one. We could potentially still use the `@deprecated` annotation, but specify which options are no longer supported? 
```
interface Button { 
  /** 
   * Describes the appearance of a button
   * @deprecated The "yuzu" type is no longer supported
   */
  variant: "default" | "primary" | "yuzu"
}
```

## How do we deprecate components?
We can also use the `@deprecated` annotation for components too: 
```
/** 
 * @deprecated Use NewButton instead
 */ 
export const Button = () => {}
```
Depending on how yell-y we want to be, also pop a `console.warn` when the component is mounted. 

## How do we deprecate Sass mixins and variables?
### mixins
Use [warn](https://sass-lang.com/documentation/at-rules/warn) or [Sass Deprecate](https://github.com/salesforce-ux/sass-deprecate)
```
@mixin some-example {
   @warn "some-example is deprecated, use component instead"
}
```

### Variables
I haven't tried this, but [Sass Deprecate](https://github.com/salesforce-ux/sass-deprecate) might work for variables? 

## How do we show that a component is deprecated in Storybook? At what point do we remove the story?
[...]

PR #749, opened 2020-09-01
docs: Improve visibility of deprecated components

ActuallyACat:
# About
The deprecated components in Kaizen are now a little more explicit. I have: 
* Moved `menu-list` to legacy-packages. 
* Added a large `# Deprecated` warning within the stories of all deprecated components (React only - the Storybook addon doesn't work for Elm). Elm consumers will need to refer to the story name in storybook. 
* Deprecated `menu-list` and `title-block`  with `npm deprecate`. see here: 
Screen Shot 2020-09-01 at 11 13 41 am. This will also appear on the CLI when you install `@kaizen/draft-menu-list`.
* Added type hints for deprecated props and components.

## Questions 
What have I missed? 
Are there more deprecated components / props? 
Have I accidentally flagged something that's not deprecated?

## Not included 
Warnings when a deprecated component is mounted. I'd like to experiment with how noisy it makes CLI - I will handle this in a separate PR.

References https://github.com/cultureamp/kaizen-design-system/issues/701


ActuallyACat:
@sebpearce @sampnorris Is `@kaizen/draft-title-block` deprecated?

sebpearce:
@ActuallyACat Practically yes, but technically it will be in a couple of weeks once the Zen one has been rolled out everywhere. I think it's fine to add it to the list here 👍

sebpearce:
Just had a look through this and at the branch preview – great stuff 👌 

I realised I missed the Elm Text component `deprecated` label when I added them a few weeks ago 🤦  
image
Any chance we could piggyback that on this PR?

ActuallyACat:
Good catch @sebpearce - I've updated this now.

dougmacknz:
Is `NavigationBar` deprecated in favour of `ZenNavigationBar`?

Off topic from this PR, but we've got two different conventions for where to put `Zen`, for `TitleBlock` we've done a suffix `TitleBlockZen`, and for `NavigationBar` we've done a prefix `ZenNavigationBar`. I would vote for the former, so that they sit side by side, and it's easy to see that it's a replacement.

dougmacknz (approved):
Otherwise, everything here is looking good to me. We could do a follow up PR for the `NavigationBar` if we wanted. This will definitely help us avoid using deprecated components, thank you :)

ActuallyACat:
> I would vote for the former 
+++ for the reasons you mentioned - it's much better for visibility. Not sure what that's going to mean for the existing component though 🤔

PR #758, opened 2020-09-03
feat: Toast notification manager

makenosound:
This PR adds:

- A manager for toast notifications that handles their state/rendering them in a separate React tree
- An imperative API to allow users to add/remove/clear toast notifications without worrying about their rendering cycle

This fixes a couple of issues:

1. We no longer need every single consuming application to implement a (slightly different) positioning wrapper for toast notifications
2. Packages can potentially use toast notifications confidently (whereas before would’ve been awkward)
3. A separate rendering stack means they’re not subject to the rendering _context_ of the application, and so they can persist across routes within a SPA without the consumer jumping through hoops.

The new API looks like this:

```js
import { Button } from "@kaizen/draft-button"
import {
  addToastNotification,
  removeToastNotification,
  clearToastNotifications,
} from "@kaizen/component-library"

function App() {
  return (
    <Button
      label={"Show a notification"}
      onClick={() => {
        addToastNotification({
          type: "affirmative",
          title: "Success",
          automationId: "notification1",
          message: (
            <>
              New user data, imported by mackenzie@hooli.com has successfully
              uploaded. <a href="/">Manage users is now available</a>
            </>
          ),
        })
     }}
   />
  )
}
```

I’ve back-ported the imperative API to the existing `<ToastNotification>` component, so that existing instances within consuming applications automatically get the new behaviour. This means that their behaviour will change in line with the _intention_ above. I believe this will be a non-issue, but I’d appreciate input from others on whether they think there’ll be a problem.

## Limitations

1. The `removeToastNotification` and `clearToastNotifications` implementation is done through a declarative data change, and so doesn’t support animating notifications out properly. I did consider doing that, but it requires rethinking the animations for toast notifications more and they’re already a bit of a mess. When we eventually do that work we should probably rethink the animation approach for `GenericNotification` and perhaps look at adopting the approach we’re taking for modals (or settling on a new animation approach that handles declarative state).

  I feel we should probably separate `ToastNotification` (the component implementation) from `GenericNotification` so they’re no longer tied together, feels to me like we’re not getting value from that abstraction.

2. This change only affects the React version of `ToastNotification`, though it does make it possible for us to easily create a binding to Elm so that Elm doesn’t need to have a separate version of this component (or handle state with the same problems).

3. The solution I’ve implemented for overflowing numbers of `ToastNotification` is very naive/bad. It hard crops the list, and because we’re ordering the list from oldest to newest any new notifications will be cropped out and inaccessible. I’ve done this because our current usage of notifications means we’ll almost never hit this situation in product, and that a further refactoring of the transition animations structure would muddy this PR more. I did attempt to work around the issue by making the notifications list a scrolling overflow panel, but that is complicated because `pointer-events: none` does not play well with `overflow-y: scroll`.

## Transitions

As part of this work, I felt that the animations for `ToastNotification` components were very sluggish, so I’ve made them faster and smoother.

While doing that I couldn’t help but fix the `InlineNotification` too so it no longer does a weird margin-shift when animating in. Instead it takes layout immediately (and then animates itself in visually). When animating _out_ it’ll collapse smoothly to allow the content to position itself. Most of the time in use we render `InlineNotification` on a page and allow it to be removed, so I think this limitation is reasonable.

## Mobile display

I’ve also neated up the display of toast notifications on mobile so they aren’t totally broken. Hooray!


packages/component-library/stories/InlineNotification.stories.tsx
```diff
@@ -17,6 +17,12 @@ const multilineText = (
     this can fit into one line of text on an average screen...
   </>
 )
+const withContentBelow = (Story: React.FunctionComponent) => (
```
makenosound, line 20: I added this so it’s possible to see how `InlineNotification` animates in and affects other content.

ActuallyACat (dismissed):
I like this. I don't have any suggestions for implementation improvements. 

One use case that is kicking around in my mind is API errors that affect all endpoints (user session elapsed or something like that). I wonder if it's worth explicitly calling that out as a situation you might consider reusing `id` as a way to prevent multiple notifications containing the same error. Or be hands-off, and allow consumers to manage that use case themselves 🤷‍♀️

makenosound:
> One use case that is kicking around in my mind is API errors that affect all endpoints (user session elapsed or something like that). I wonder if it's worth explicitly calling that out as a situation you might consider reusing id as a way to prevent multiple notifications containing the same error.

Yeah, feels like that’d be a useful note in documentation, though I wonder exactly where that should live... I could create a story that has an explicit example with that in place perhaps.

makenosound:
Have added that additional story. Mind re-ticking @allybeaton ?

ActuallyACat (approved):
✨

PR #760, opened 2020-09-03
fix: Bump design-tokens peer dependency to the latest version

ActuallyACat:
This updates all packages to use the same version of `design-tokens` as a peer dependency across Kaizen. I've added a new version range for now, with the view to drop the ~v1 peer dependency in a couple of weeks when all dependants are on ~v2.

To see the breaking changes from 1.15.0 -> 2.0.0, [see this PR](https://github.com/cultureamp/kaizen-design-system/pull/606/files).

### DIFF
--- a/draft-packages/avatar/package.json
+++ b/draft-packages/avatar/package.json
@@ -37,7 +37,7 @@
     "react-textfit": "^1.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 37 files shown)


lijuenc (approved):
> added a new version range for now, with the view to drop the ~v1 peer dependency in a couple of weeks when all dependants are on ~v2

Sounds reasonable

PR #761, opened 2020-09-04
docs: move copy guidelines

didoesdigital:
Designers sometimes misunderstand copy guidelines because they appear before the introduction of key terms outlined in the Anatomy section. For consistency, we're moving all copy guidelines below the "To keep in mind" section.


mbylstra (approved):
👍 I was going to congratulate you on your last PR, but I have a feeling you're going to smash out a few more before the end of the day 😄

PR #763, opened 2020-09-04
feat: Allow routing libraries to be used with Button

dougmacknz:
There's a few things happening here on the new Kaizen button. I've bundled in a few breaking changes, so that we can sweep all the usages at once.

## Routing libraries
We currently have an awkward way of using react-router with Kaizen buttons. In most cases we are using the onClick handler, with react-router's `history.push()`. The issue with this is that there is no support for native link behaviour, such as opening in a new tab, and the element will not be a link as it semantically should be (it will be a button, unless an href prop is provided alongside the `onClick`)
Slack conversation about this here: https://cultureamp.slack.com/archives/C2ZNME08P/p1587088257293400

We have attempted to work around this in a couple of different ways, but none of them have been able to completely replicate the proper behaviour, and they have been quite messy to implement on each usage.

The ideal way to fix this is to allow us a clean way to use react-router's `Link` directly with Kaizen buttons. This can be achieved by allowing a `component` prop, which will replace the `<a>` or `<button>`, while still having the attributes (most importantly, the styling). This is the approach I have taken, the idea has come from [Material UI](https://material-ui.com/guides/composition/#routing-libraries) and [Atlas Kit](https://atlassian.design/components/button/code).

**BREAKING CHANGE**: In order to implement this - I had to fix some typing that we had on our `onClick` prop. The `event` was typed as `MouseEvent`, rather than a `React.MouseEvent`. This gave me some issues when putting `{...props}` onto the custom component. In order to fix this - we will need to update any places that have typed the `onClick` `event` from `MouseEvent` to `React.MouseEvent`.

## `preventDefault()` when specifying `onClick`

Currently on our buttons, whenever you specify an `onClick` prop - the Button will `preventDefault()` for you. This is something we ran into when attempting to work the issue above in Murmur. We will no longer require that solution with the above being implemented, however chances are we will run into an issue with this again, and it makes more sense to allow the consumer to have the choice on this.

@ActuallyACat did some digging and found that this came from way back when the button component was in `ca-ui`, so we reckon there wasn't any great reasoning behind this originally being implemented.

**BREAKING CHANGE**: To fix this, we will want to check which buttons are relying on the preventDefault behaviour, and if they are then we need to add the preventDefault to the usage. Alternatively we could just add `preventDefault` naively, with a comment, if we find that it will take too long to test every usage.

## Specific props for Aria, analytics and automationId

This is a smaller change in comparison to the above changes, however it is also a breaking change. @jeremy8883 did some work recently, to allow any arbitrary kebab case props to the Kaizen button, which means that we no longer have any need for the three aria props that currently exist, the `analytics` prop (which is deprecated now as well anyway with the introduction of Amplitude and deprecation of Treasure Data), and the `automationId` prop.

**BREAKING CHANGE**: We need to replace any usages of `ariaExpanded`, `ariaControls`, `ariaDescribedby`, `automationId` with kebab case equivalents (eg. `aria-expanded`). `analytics` can also be replaces with the `data-*` equivalents, but I doubt there will be any usages of the new button with this prop.


draft-packages/stories/Button.stories.tsx
```diff
@@ -331,3 +330,12 @@ export const MultipleButtons = () => (
     <Button label="Exit" automationId="demo-button-2" />
   </div>
 )
+
+export const CustomComponent = () => {
+  const customComponent = (props: CustomButtonProps) => (
+    <a href="#" {...props} />
```
jeremy8883, line 336: I was thinking that the risk here would be that if the props signature of the custom component doesn't match that of `CustomButtonProps`, some properties would get missed. But if all of our code is in in typescript, this error would be alerted, so I guess this would be fine.

We just need to hope that developers don't pass each prop one by one, otherwise if new props are added to `CustomButtonProps` may not get custom component instance.

jeremy8883 (commented):
Nice work! I'll leave it to someone from design systems to approve.

Not sure if it was mentioned in the slack convo (it looked quite long), but the way we go about this in perform, is creating a wrapper component, and using the `href` property, and replicating the `Link` functionality as a callback function. eg.

```
const isModifiedEvent = (event) =>
  !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)

export const isRegularMouseClick = (event) =>
  event.button === 0 && !isModifiedEvent(event)

const basedHref = href != null ? joinPath(router.location.basename, href) : href
const handleClick = (event) => {
      try {
        if (onClick) onClick(event)
      } catch (ex) {
        event.preventDefault()
        throw ex
      }

      if (basedHref != null) {
        if (
          !event.defaultPrevented &&
          isRegularMouseClick(event)
        ) {
          event.preventDefault()
          if (href) router.push(href)
        }
      }
    }
```

There's also the eslint rule `no-restricted-imports` to prevent developers from using the non-wrapper version of the component.

I'm not saying we should do this, just pointing it out. This custom component way also looks good to me!

makenosound (approved):
One stray variable, but otherwise looks fine!

ActuallyACat:
> Up for debate: This change could also be made for the automationId prop, however it might be best to leave it. It's possible in the future that we'll want to switch from using data-automation-id to data-testid. If we leave this prop, it might make our lives easier.

I personally have no qualms whether folks use `automation-id`, `testid`, or the recently discussed `automation-class`. The latest discussion we had yielded that automation-id is a pattern we want to continue for new components (at the very least). I think testid is for react-testing-library, but you can change that default [using configure](https://testing-library.com/docs/dom-testing-library/api-configuration). 

Some components, like Box, can accept any data attributes and spread it onto the wrapper of the component: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Box/Box.spec.tsx#L36-L43

I didn't get many strong signals from folks that this pattern was preferable though. 🤷‍♀️

ActuallyACat (approved):

PR #766, opened 2020-09-08
feat: Added helper to deprecate components

ActuallyACat:
# Objective
This package provides some HOC helpers to warn users about deprecated props and components.  

# Motivation and Context 
Maintaining a healthy codebase for Kaizen means we need a consistent way to deprecate components and props. 

This PR gives us a consistent way to deprecate components and works in tandem with the process outlined [in this document](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/1582563561/Deprecating+components).

## Q&A
**Why didn't you forward / map deprecate props to the new props (like [react-deprecate](https://www.npmjs.com/package/react-deprecate))?** 
There are too many considerations for this to be successful in Kaizen. Consider when both the object property and object values change, where there is no clear 1 to 1 mapping, and when one prop allows undefined / null values... so something like this: 
```
interface oldProps {
   type?: "default" | "prominent" | "warning" | "info"
} 

interface newProps {
   variant: "default" | "warning" | "info"
} 
```
type="default" would be fairly straightforward to map to variant="default". Where it gets complicated is the other use cases. What would type=undefined map to? What about type="prominent'? 

**What's next?**
I will use this helper on all our deprecated components (such as Text) as well as deprecated props. These will be released as canary releases. I'll then reach out to some folks to make sure this approach is effective and not ignored noise. 

# Checklist

# Additional Considerations
- ~Does there need to be right-to-left (RTL) options for localization and internationalization?~
- ~Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?~
- ~Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?~
- ~If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?~
- ~Have Storybook stories been updated?~


packages/component-life-span/index.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from "react"
+
+export const Counter = (() => {
```
ActuallyACat, line 3: Used a closure here so I can reset the counter in the tests, otherwise I would have used a var.

packages/react-deprecate-warning/src/WithDeprecatedComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const withDeprecatedComponent = <P extends object>(
+  WrappedComponent: React.ComponentType<P>,
+  { warning, name }: withDeprecatedComponentProps
+): React.ComponentType<P> =>
+  class extends React.Component<P> {
+    constructor(props) {
+      super(props)
+
+      log(name, warning)
```
lijuenc, line 17: We can probably use the `name` property on the `WrappedComponent` instead of relying on it being passed in all the time. That's if your intention was for `name` to always be the same as the component name, of course.

packages/react-deprecate-warning/src/index.spec.tsx
```diff
@@ -0,0 +1,217 @@
+// import "@testing-library/jest-dom/extend-expect"
```
lijuenc, line 1: Looks like a stray comment.

lijuenc (approved):
Nice, this feels quite intuitive to me. Would be keen to see what the other advocates think too.

PR #767, opened 2020-09-08
fix: added missing dependency

sampnorris:
Bugfix caused my missing dependency


sebpearce (approved):

ActuallyACat (approved):
Thanks for this - LGTM 🙏

PR #771, opened 2020-09-08
fix: expose public function in draft card

juliohr:
the draft card has a tag function to be piped when initialising a card
we only used the default configuration before
when creating a new card of type section I noticed that the tag function
and the Tag types were not exposed.

usage in murmur: https://github.com/cultureamp/murmur/pull/14416/files#diff-07e2d306a4bcf6ded329f8e77cc2cf1cR110
(search for Card.tag on the files)


ActuallyACat (approved):
Thanks @juliohr ! LGTM

PR #773, opened 2020-09-09
fix: Added missing dependencies for Modal, FilterDrawer, and Tile

ActuallyACat:
Does what it says on the box. To be released before https://github.com/cultureamp/kaizen-design-system/pull/772

### DIFF
--- a/draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
+++ b/draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
@@ -3,7 +3,7 @@ import { StatelessMenu } from "@kaizen/draft-menu/KaizenDraft/Menu/StatelessMenu
 import classnames from "classnames"
 import * as React from "react"
 /**
- * Eslint throws a false positive for modules that use require. Ensure you
+ * Eslint throws a false negative for modules that use require. Ensure you
  * are importing @kaizen/component-library into your package before turning
  * this rule off.
  */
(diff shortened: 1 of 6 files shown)


draft-packages/filter-drawer/KaizenDraft/FilterDrawer/FilterDrawer.tsx
```diff
@@ -2,6 +2,12 @@ import { MenuContent } from "@kaizen/draft-menu"
 import { StatelessMenu } from "@kaizen/draft-menu/KaizenDraft/Menu/StatelessMenu"
 import classnames from "classnames"
 import * as React from "react"
+/**
+ * Eslint throws a false positive for modules that use require. Ensure you
```
mbylstra, line 6: I'm a bit confused by the comment - should it be "Eslint throws a false **negative**"? As in, the rule thows an error, but it shouldn't?

"before turning this rule off.." Isn't that what `// eslint-disable-next-line` does - it's already turned off?

mbylstra: I think I might just be confused by the semantics of positive and negative 😆

mbylstra: positive as in in good, or positive as in this is "positively bad" or "this positively fails the check"?

mbylstra: maybe it could read "throws a false error" (if that's what it means)

ActuallyACat: D'oh, you're right. It should read `false negative`, as in, there is an error reported when there shouldn't be. Good catch, i've now fixed this up.

mbylstra (approved):
Sweet, I think I get it now.

- It was throwing warnings but us developers were ignoring them
- upgrading to error would break the build because of the require false negatives
- so we need to ignore the errors/warnings for the require statements
- after all that, we still need to add react-dom and the component library
- now all is good!

PR #775, opened 2020-09-10
ci: Upgrade react-testing-library to v11

ActuallyACat:
I noticed we're pretty far behind, this PR brings us to the latest version. The biggest change from jumping from v8 -> v11 was the new location of `configure` and  `fireEvent` (@testing-library/dom) and `wait` became `waitFor`. 

There's still some flakiness in `GenericNotification.spec.tsx` - TBH I'm not concerned about that as I think rewriting the animation implementation will make it easier to test. I wrapped `advanceTimersByTime` in a promise as that's what was recommended [here](https://github.com/testing-library/react-testing-library/issues/244)


Confidenceman02 (approved):
LGTM

Issue #668, opened 2020-07-23
Working with `Table` column widths


jeremy8883:
Following from our slack conversation:

> Question about the Table component. One of the problems that we’ve encountered often, is that the table uses display: flex, and either percentages or fixed pixel values to choose the widths of each column. The problem is that as soon as the text content changes, or the language changes to german, the cell widths become inconsistent.
> I understand why it was done this way, because it adds more flexibility for the expandable table row animations. However, 90% of the time, we don’t need this in perform.
> I was wondering as a solution, if kaizen could supply an alternative component, which explicitly uses `display: table`. That way, if there are any minor issues in our column sizing, the browser will make the appropriate adjustments.
> Or if there are any other ideas? I’m also good to create a PR. 

To sum up the responses so far from @mbylstra and @ActuallyACat :

* We could name the components `Table` and `ExpandableTable`. Or just have a single table with a prop that proxies to the appropriate component type.
* Using css grid would be nice, but we need to support ie11
* It has been attempted to do in `display: table`, and it worked, but the css complexity was high. The new zen designs are more simple, so maybe we could try and have the new zen components use `display: table`, with the expandable rows included.
* This is a hard problem, and we look at all of our current use cases to make an informed decision.

jeremy8883:
I've gone through perform, and listed all of the requirements that I can find. See the jira ticket PMC-428 for supporting screen shots.

### Currently supported (and will still be required)
* Row to be clickable, using `a`nchor or `button` elements, with `onClick` support. Note that every cell is clickable, except for the last split button column. This is because:
  1. nesting buttons is illegal
  2. it gives a little bit of click room around the split button, so if the user misses the click, they don't get taken to a completely different page.
* Some tables have table headers, some do not
* Mix between fixed column widths and dynamic widths
* Sortable column headers
* Checkbox fields, with "check all" in the table heading
* Expandable rows. An api call also needs to also be made upon expand.

### Not explicitly supplied by `Table`, but needs to compatible
* We have icon buttons to the right in a lot of tables. In the legacy non-kaizen tables, they show tooltips on hover. Shall we do the same here?
* Allow for "more" and "splitbutton" menus. Not that the table should support this explicitly of course, but I mention this because there was a previous issue where `overflow: hidden` was used, and was clipping off the dropdown menu. This is now fixed, but just to be aware.
* Infinite scroll

### Feature requests
[...]

PR #776, opened 2020-09-15
feat: Proof of concept for a flexible table wrapper over react-table

ActuallyACat:
Two options - 
1. **Build your own table, Kaizen gives you the building blocks** Each element of table is exposed as a react component and gives the consumer the ability to override styles, replace the HTMLElement tag, and supply whatever child elements they require. 
2. Give an array of data for heading and body. Component does everything else. 

Key call outs:
* Consumer needs to build the table row popout 
* Responsiveness to be implemented by consumer
* (intentionally) dumb as bricks - all other features will need to BYO (or lean on react-table hooks).

Related to: https://github.com/cultureamp/kaizen-design-system/issues/668

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,11 +24,7 @@ module.exports = {
     "sort-imports-es6-autofix",
     "prefer-arrow",
   ],
-  settings: {
-    "import/resolver": {
-      foo: { someConfig: value },
-    },
-  },
+
   rules: {
     "@typescript-eslint/adjacent-overload-signatures": "error",
     "@typescript-eslint/array-type": [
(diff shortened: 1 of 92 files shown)


ActuallyACat:
Another consideration: 

Looking at other design systems, the treatment of tables is purely for more atomic data; such as numbers, strings, etc. For Kaizen, we seem to allow more complex data in a table which certainly makes in harder to add additionaly functionality like sorting. Some examples: 

https://zendeskgarden.github.io/react-components/tables/
https://feelix.myob.com/#/Components/Tables/Table
https://ux.mailchimp.com/patterns/tables
https://www.carbondesignsystem.com/components/data-table/usage
https://clarity.design/documentation/tables
http://uspto.github.io/designpatterns/1.x/docs/components/tables.html
https://www.yelp.com/styleguide#section-tables
https://solid.buzzfeed.com/tables.html

PR #778, opened 2020-09-18
feat: Allow any Button props to be sent to GuidanceBlock actions

dougmacknz:
## Background

In a recent project using the GuidanceBlock component - I found that I needed to be able to use more than just the `label`, `onClick` and `disabled` props for the buttons. 

This PR allows you to send through any button props, rather than just the 3 that were specified previously. This also means that we won't need to adjust this package if the Button API changes.


juliohr (approved):
good job

ActuallyACat (approved):

PR #779, opened 2020-09-18
fix: Change all require to import (icons and styles) 

ActuallyACat:
**About** 
Previously, we used `require` for including any assets that don't come with typescript types, such as icons and styles. This PR add types for icons and styles so we can then import them like other assets. 

**Motivation**
We experienced some weird eslint errors (both false positives and false negatives) when using a mix of imports and require.

### DIFF
--- a/draft-packages/stories/ZenNavigationBar.stories.tsx
+++ b/draft-packages/stories/ZenNavigationBar.stories.tsx
@@ -3,15 +3,9 @@ import * as React from "react"
 import { Button, Icon } from "@kaizen/component-library"
 import { Link, Menu, ZenNavigationBar } from "@kaizen/draft-zen-navigation-bar"
 import { ColorScheme } from "@kaizen/draft-zen-navigation-bar/KaizenDraft/ZenNavigationBar/types"
-
-const caIcon = require("@kaizen/component-library/icons/ca-monogram.icon.svg")
-  .default
-const academyIcon = require("@kaizen/component-library/icons/academy.icon.svg")
-  .default
-const caMonogramIcon = require("@kaizen/component-library/icons/ca-monogram.icon.svg")
-  .default
-const supportIcon = require("@kaizen/component-library/icons/support.icon.svg")
-  .default
+import caIcon from "@kaizen/component-library/icons/ca-monogram.icon.svg"
+import academyIcon from "@kaizen/component-library/icons/academy.icon.svg"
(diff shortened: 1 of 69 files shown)


draft-packages/stories/ZenNavigationBar.stories.tsx
```diff
@@ -3,15 +3,9 @@ import * as React from "react"
...
-  .default
-const academyIcon = require("@kaizen/component-library/icons/academy.icon.svg")
-  .default
-const caMonogramIcon = require("@kaizen/component-library/icons/ca-monogram.icon.svg")
-  .default
-const supportIcon = require("@kaizen/component-library/icons/support.icon.svg")
-  .default
+import caIcon from "@kaizen/component-library/icons/ca-monogram.icon.svg"
+import academyIcon from "@kaizen/component-library/icons/academy.icon.svg"
+import supportIcon from "@kaizen/component-library/icons/support.icon.svg"
```
ActuallyACat, line 8: For some reason, `ca-monogram.icon.svg` was imported twice, once as `caMonogramIcon` and again as `caIcon`. I removed the duplicate and changed everything to use `caIcon`

package.json
```diff
@@ -21,7 +21,9 @@
     "lint": "elm-format --validate packages && yarn eslint && yarn prettier && stylelint '**/*.scss'",
     "lint:fix": "elm-format --yes packages && yarn eslint --fix && yarn prettier --write",
     "clean:storybook": "rimraf 'storybook/public'",
-    "clean": "lerna run clean && yarn gatsby clean && yarn clean:storybook && rimraf 'elm-stuff'",
+    "clean:stories": "rimraf './packages/component-library/stories/*.d.ts' './packages/component-library/stories/*.js.map' './packages/component-library/stories/*.js'",
+    "clean:draft-stories": "rimraf './draft-packages/stories/*.d.ts' './draft-packages/stories/*.js.map' './draft-packages/stories/*.js'",
```
ActuallyACat, line 25: You shouldn't need these two scripts unless you're running `yarn lerna run publish` locally (which a large majority of people will never need). I've added it here for convenience.

ActuallyACat: What should really happen is `yarn lerna run prepublish` shouldn't generate declaration files/source maps for assets that we don't actually publish 🙃 Will be tackled in a separate housekeeping PR.

PR #780, opened 2020-09-19
feat: textfield (elm) caution variation

juliohr:
The intention is to use it here: https://www.figma.com/file/OECVRbWEHCjgH4qLyqDlGK/Engagement-Quick-wins?node-id=1491%3A18607

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/styles.scss
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/styles.scss
@@ -3,6 +3,7 @@
 @import "~@kaizen/deprecated-component-library-helpers/styles/type";
 
 $message-error-color: $kz-color-wisteria-800;
+$message-caution-color: $kz-color-yuzu-300;
 $message-border-radius: $kz-border-solid-border-radius;
 
 // @TODO - taken from ~@kaizen/component-library/styles/color
@@ -22,6 +23,13 @@ $ca-palette-input-validation-negative: $kz-color-coral-600;
   &.reversed {
     color: $kz-color-white;
   }
+}
+
+.caution {
+  padding: $ca-grid / 2;
+  color: $kz-color-wisteria-800;
+  border-radius: $message-border-radius;
+  background: $kz-color-yuzu-300;
 }
 
 .error {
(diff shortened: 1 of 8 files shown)


draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/styles.scss
```diff
@@ -3,6 +3,7 @@
 @import "~@kaizen/deprecated-component-library-helpers/styles/type";
 
 $message-error-color: $kz-color-wisteria-800;
+$message-caution-color: $kz-color-yuzu-300;
```
dougmacknz, line 6: You didn't end up using this

draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/styles.scss
```diff
@@ -24,6 +25,13 @@ $ca-palette-input-validation-negative: $kz-color-coral-600;
   }
 }
 
+.caution {
+  padding: $ca-grid / 2;
+  color: $kz-color-wisteria-800;
+  border-radius: $message-border-radius;
+  background: $kz-color-yuzu-300;
+}
```
dougmacknz, line 33: Everything except for the `background` is the same as the error styling, so you could DRY this up with something like (and move it below the error styling).
```
.caution {
  composes: error;
  background: $kz-color-yuzu-300;
}
```

dougmacknz: Also, looking at Figma, I believe this is meant to be Yuzu 200 instead of 300

draft-packages/stories/ElmTextField.stories.tsx
```diff
@@ -5,4 +5,5 @@ loadElmStories("TextField (Elm)", module, require("./TextFieldStories.elm"), [
   "Default /w icon",
   "Default /w string description",
   "Default /w html description",
+  "Caution /w html description",
```
dougmacknz, line 8: (nit pick) This is still 'Default', w/ caution styling (see the naming on the React versions). The ones that aren't 'Default' are the reversed ones for dark backgrounds

dougmacknz (commented):
Nice!

Just to document - there's a few discrepancies from the UI Kit designs and this. Not your fault because the error styling has the same thing, feel free to ignore this.

UI Kit for hover styling:
image

Actual (no background colour):
image

UI Kit for focus styling:
image

Actual (note that the yellow border should remain, and just the blue focus ring is added)
image

PR #781, opened 2020-09-21
fix: Changed icons to use import, added types for icons

ActuallyACat:
**About**
Previously, we used require for including any assets that don't come with typescript types, such as icons and styles. This PR add types for icons and styles so we can then import them like other assets. To keep the PRs as lean as possible, I'll follow up another PR for changing styles from `require` to import`

**Motivation**
We experienced some weird eslint errors (both false positives and false negatives) when using a mix of imports and require.

**Details**
fix: Change imports in component library to other parts of the component library to relative
fix: added missing dependency in tile
fix: Formatting issues with import
fix: Updated tsconfig to use include and exclude properties
fix: Changed all icons to use import
fix: Added types for SVGs

Checked by diffing the js/d.ts/.map.js files of these packages to those against master - no changes. 
Ran Chromatic - no unexpected changes.


draft-packages/collapsible/KaizenDraft/Collapsible/__snapshots__/Collapsible.spec.tsx.snap
```diff
@@ -28,10 +28,9 @@ exports[`matches snapshot with everything enabled 1`] = `
             class="icon"
             focusable="false"
             role="presentation"
-            viewBox="0 0 0 0"
```
zioroboco, line 31: We could make these look more like our svg modules at runtime by using the jest transforms from frontend-ops:

https://github.com/cultureamp/frontend-ops/blob/master/packages/frontend-core/jest/transforms/icon.js

ActuallyACat: Neat, i like that idea. These snaps made me do a double-take. I'll add that now thanks!

ActuallyACat: Scratch that - I'll follow up in a separate PR so this one doesn't get too hairy.

zioroboco (approved):

PR #783, opened 2020-09-21
feat: allow uncontrolled Elm TextArea to be pre-fillable

mbylstra:
- it makes use of the defaultValue html property, to set
  a default value when the dom node is first created
- it's less risky than the controlled variant, which can have
  issues on old devices as the component isnt' updated quickly
  enough. Users can experience sluggish text entry or strange
  cursor behaviour.
- however, there is a big limitation: you cannot automatically change
  the value once the dom element has been created (for example, pressing
  button to fill the value). Most forms don't require this functionality.
- also do the same for `TextField`

Stories:

https://dev.cultureamp.design/mab/uncontrolled-elm-text-area-field/storybook/?path=/story/textfield-elm--default-controlled-prefilled-value
https://dev.cultureamp.design/mab/uncontrolled-elm-text-area-field/storybook/?path=/story/textfield-elm--default-uncontrolled-prefilled-value
https://dev.cultureamp.design/mab/uncontrolled-elm-text-area-field/storybook/?path=/story/textareafield-elm--default-controlled-prefilled-value
https://dev.cultureamp.design/mab/uncontrolled-elm-text-area-field/storybook/?path=/story/textareafield-elm--default-uncontrolled-prefilled-value

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.elm
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.elm
@@ -28,6 +28,7 @@ import Html exposing (..)
 import Html.Attributes
 import Html.Events exposing (onFocus, onInput)
 import Html.Extra exposing (nothing)
+import Json.Encode
 import KaizenDraft.Events.Events as Events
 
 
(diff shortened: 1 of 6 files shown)


mbylstra:
[Prefillable uncontrolled variant of Elm TextAreaField](https://trello.com/c/vqvYVfoR/37-prefillable-uncontrolled-variant-of-elm-textareafield)

caseysinclair (approved):
This looks good!

PR #784, opened 2020-09-21
fix: Apply spacing to labels when inline with form controls

ActuallyACat:
# About
Screen Shot 2020-09-21 at 2 29 06 pm
Fixes bug from #740  - where every label receives margin. We only want the additional spacing around labels when it is inline with form controls. When labels sit within a container with no other content around it, there should be no left padding when LTR, and no right padding when RTL.


ActuallyACat:
Chromatic build: https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=199

lijuenc (approved):

PR #785, opened 2020-09-21
feat: SplitButton and Menu updates

jeremy8883:
* Adjust the spacing between the `SplitButton` and the dropdown menu. In particular, fix the spacing for when the menu is displayed above the split button. A storybook story has also been created, as so this scenario can be tested.

Before:

image

After:

image

* Split the `MenuItem`'s `action` property into two props, `onClick` and `href`. The limitation with `action` was that it prevented us from having both a `href` and an `onClick` callback at the same time. In perform, we wanted to use `a`nchor elements for menu items, but I want to also link them up with react router, which requires an `onClick`.

After doing this update, I remembered that for buttons, we're passing down a `render` function to achieve the same outcome. LMK if you'd rather me change the implementation in this PR.

* Change the menu item disabled state to be a button element with a `disabled` prop, rather than simply reading the item as if it were ordinary text. This is so a screen reader will make it clear that the menu item is disabled.

* Make the dropdown z-index above the sticky and fixed layers. I'm not sure who to ask about this one. The problem is in perform, for instances where the menu is positioned above the split button, it shows below the top navigation header, making some menu items unclickable.


packages/component-library/styles/layers.scss
```diff
@@ -1,8 +1,8 @@
 $ca-z-index-body: 0;
-$ca-z-index-dropdown: 1000;
 $ca-z-index-sticky: 1020;
 $ca-z-index-fixed: 1030;
 $ca-z-index-modal-backdrop: 1040;
 $ca-z-index-modal: 1050;
+$ca-z-index-dropdown: 1055;
```
jeremy8883, line 6: LMK if this is alright. I'm not sure what all the situations are, but I feel like a `dropdown` is more closely related to a `popover`, in terms of layer positioning.

The main thing we need in perform is to have the z-index > 1030 (above the `fixed` layer`).

draft-packages/menu/KaizenDraft/Menu/components/MenuContent.module.scss
```diff
@@ -42,6 +42,10 @@ $side-padding: 1/2 * $ca-grid;
   align-items: center;
   flex: 0 0 100%;
   box-sizing: border-box;
+  background: none;
+  border: none;
+  width: 100%;
+  text-align: left;
```
sebpearce, line 48: The `text-align` was previously being inherited from `.menuContent`, which handles RTL. If we're adding it here could we do the same?

See https://github.com/cultureamp/kaizen-design-system/pull/785/files#diff-270ef25388a53fab2ab6511fe5ea3f26R21

draft-packages/menu/KaizenDraft/Menu/components/MenuItem.tsx
```diff
@@ -6,15 +6,36 @@ const styles = require("./MenuContent.module.scss")
...
-  action: string | ((e: React.MouseEvent<HTMLAnchorElement>) => void)
+  /**
+   * Historically, we only had `action`, which would accept either a string for
+   * the href, or a function for an onClick callback.
+   * The limitation here was that it prevented us from having
+   * both a href and an onClick callback at the same time.
+   * This field has been deprecated in favour of `onClick` and `href`.
+   * @deprecated
+   */
+  action?: string | ((e: React.MouseEvent<HTMLButtonElement>) => void)
```
sebpearce, line 17: Great to see this! This union type has been a real pain to deal with in the Title Block.

draft-packages/menu/KaizenDraft/Menu/components/MenuItem.tsx
```diff
@@ -28,34 +49,54 @@ const MenuItem = (props: MenuItemProps) => {
...
         {wrappedLabel}
       </a>
     )
   }
 
   return (
-    <a
-      href="#"
-      onClick={action}
+    <button
```
sebpearce, line 89: 👌

packages/component-library/styles/layers.scss
sebpearce: @ActuallyACat and @sentience may be interested in this too.

sebpearce:
LGTM, pending other's thoughts on z-index.

I checked TitleBlockZen on the branch preview since it depends on Menu heavily, and couldn't find anything untoward. 👍

packages/component-library/styles/layers.scss
jeremy8883: I got the ok from @ActuallyACat while we were on slack. Confirming if this update is alright on your end @sentience

sentience: Makes sense to me. As we've discussed before (I think), a global z-index scale is not a panacea, as stacking contexts often mean that `z-index` becomes locally scoped, at which point the scale falls apart.

With that caveat, I think that putting dropdown explicitly on top of fixed elements and modals makes perfect sense, where a global scale can be used.

sebpearce (approved):

PR #786, opened 2020-09-21
feat: basic Zen uplift for Select border and hover state

mbylstra:
there is still more to do but this makes it gel much better with
existing components like text inputs

**new**

image

image

**old**

image
image


mbylstra:
[basic Elm Select Zen uplift](https://trello.com/c/wQt7U4TO/84-basic-elm-select-zen-uplift)

caseysinclair (approved):

PR #782, opened 2020-09-21
fix: Change styles from require to use import

ActuallyACat:
Follows up https://github.com/cultureamp/kaizen-design-system/pull/781

Updates including of styles to use `imports`.

PR #794, opened 2020-09-28
ci: Create a dist config, added types to files in config

ActuallyACat:
## About 
When updating the types for SVGs and styles last week, the `prepublish` script wasn't checked. An assumption was made that the options set for `include/exclude` would be merged when it's extended, but it was overridden. This means the types that are needed to compile individual packages (the types for icons and styles) were missing. 

What this means:
```
// The "root" tsconfig that all other configs extend from 
{
  "compilerOptions": {
    /* stuff */
  },
  "include": ["./packages/**/*", "./draft-packages/**/*", "./types.d.ts" ],
  "exclude": ["**/node_modules"]
}
```

```
// an example from kaizen/component-library
 /* stuff */
"compilerOptions": {
    /* stuff */
 },
"include": [
  "components/**/*",
  "draft/**/*",
  "util/**/*",
  "index.ts",
  // NO types.d.ts file here. This property overrides what we set in the root config, not merged
 ],
}
```

Related to #781 and #782

### DIFF
--- a/draft-packages/avatar/tsconfig.dist.json
+++ b/draft-packages/avatar/tsconfig.dist.json
@@ -1,5 +1,5 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.dist.json",
   "compilerOptions": {
     "allowJs": false,
     "declaration": true,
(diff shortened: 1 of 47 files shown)


zioroboco (approved):
LGTM*

_* with the caveat that this has gotten messy, and we need to give the compiler config in this repo some love — specifically, removing the base tsconfig and properly separating the packages into individual projects_

ActuallyACat:
Yup - this is definitely in "needs more love" territory.

Issue #792, opened 2020-09-23
Roadblock modal missing in component library


ActuallyACat:
Storybook: https://cultureamp.design/storybook/?path=/story/modal-react--input-edit-positive
UI Kit: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A35439

![Screen Shot 2020-09-23 at 11 29 12 am](https://user-images.githubusercontent.com/1621182/93954030-06d59500-fd90-11ea-89c3-6971f6a277a0.png)

ActuallyACat:
FIxed in https://github.com/cultureamp/kaizen-design-system/pull/800

PR #800, opened 2020-10-01
feat: Add Roadblock modal (React), update confirmation modals to Zen designs

lijuenc:
Addresses: #792

Live Storybook: https://dev.cultureamp.design/lijuenc/zen-modals/storybook/?path=/story/modal-react--roadblock

Add Roadblock modal (React):

Screen Shot 2020-10-01 at 12 50 18 pm

Update the existing confirmation modals to the most recent Zen designs:

Screen Shot 2020-10-01 at 12 50 44 pm


draft-packages/modal/KaizenDraft/Modal/illustrations/cautionary.icon.svg
```diff
@@ -0,0 +1,34 @@
+<svg width="154" height="154" viewBox="0 0 154 154" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ActuallyACat, line 1: These are spot illustrations https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A4212

They're typically hosted in [kaizen-design-system-assets](https://github.com/cultureamp/kaizen-design-system-assets/blob/master/assets/illustrations/spot/moods-cautionary.svg) with a React wrapper around them [here](https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx)

lijuenc: oh, wasn't aware that that was a thing. Will update soon

katsambarcus:
Looking great, @lijuenc . Thanks for jumping on this. The only small detail is the width of div for the body of text feels a bit too wide in comparison to what exists in the UI Kit.

draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.scss
```diff
@@ -17,40 +13,27 @@
...
 .header {
   color: $kz-color-white;
-  padding: $ca-grid $ca-grid $ca-grid * 1.5;
+  padding: $kz-spacing-lg $kz-spacing-md $kz-spacing-lg;
   text-align: center;
-  background-repeat: no-repeat;
   border-radius: $kz-border-solid-border-radius $kz-border-solid-border-radius 0
     0;
+
+  // override Murmur global styles :(
```
ActuallyACat, line 21: RIP 😓

draft-packages/stories/Modal.stories.tsx
```diff
@@ -677,3 +683,29 @@ export const NestedModal = () => (
...
+            </Paragraph>
+          </div>
+        </RoadblockModal>
+      </div>
+    )}
+  </ModalStateContainer>
+)
+
+Roadblock.story = {
+  name: "Roadblock",
```
ActuallyACat, line 710: Calling this story `Confirmation (roadblock)` will put it next to the other confirmation modals and will be a little nicer for discoverability for this component. 

Screen Shot 2020-10-01 at 3 47 01 pm


```suggestion
  name: "Confirmation (roadblock)",
```

ActuallyACat (changes requested):
Once those illustrations are moved to use the `draft-illustration` that will be a ✅ from me

draft-packages/stories/Modal.stories.tsx
lijuenc: Technically it's not a Confirmation modal (hence why it's not just a variant of the existing Confirmation modals). Do you think labelling it as a Confirmation modal in storybook is still a good idea?

lijuenc:
> Looking great, @lijuenc . Thanks for jumping on this. The only small detail is the width of div for the body of text feels a bit too wide in comparison to what exists in the UI Kit.

Fixed in the latest commits

lijuenc:
> Once those illustrations are moved to use the `draft-illustration` that will be a ✅ from me

I've moved the React ones over to use `draft-illustration`, but we don't have the same illustration and hosted assets infrastructure for Elm. The Elm modal still uses the SVGs for now, is that ok?

draft-packages/stories/Modal.stories.tsx
ActuallyACat: Right you are! My mistake - what you have now is consistent with the UI kit so no change required.

draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.elm
```diff
@@ -207,24 +207,29 @@ header config =
     let
         resolveIcon =
             case config.variant of
-                Positive ->
-                    svgAsset "@kaizen/component-library/icons/success.icon.svg"
+                Cautionary ->
+                    svgAsset "@kaizen/draft-modal/KaizenDraft/Modal/illustrations/cautionary.icon.svg"
```
ActuallyACat, line 211: I'm kinda surprised there's no `hosted-assets` function for Elm here... I wonder if there's a reason for this? It would be a little nicer than duplicating assets. cc/ My go-to Elm person @mbylstra 

https://github.com/cultureamp/kaizen-design-system/blob/master/packages/hosted-assets/index.ts#L12

mbylstra: Ah, I did actually make a hosted-asset Elm function in our repo last cycle which I intended to move into the Kaizen in cooldown, but it didn't happen. I'm planning on dedicating some weekly time to Kaizen - this would be a good task. So, I can refactor this later in a separate PR once a hosted asset function exists.

ActuallyACat (approved):
LGTM, thanks Li!

draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.elm
ActuallyACat: I knew you would know :D That would be a huge win.

mbylstra: Just good timing really (or bad timing because i didn't get it in in time)

PR #801, opened 2020-10-01
feat: Remove need to specify primary & reversed

sebpearce:
## Changes

This PR affects TitleBlockZen.

Previously, consumers had to specify
```
primary: true,
reversed: true
```
for both primary (when not a menu) and default actions.

This PR removes that requirement.

Users can still override these defaults if they want, e.g. `primary: false`.


sampnorris (approved):
Looking good

Issue #791, opened 2020-09-23
Guidance block is missing "prominent" variant


ActuallyACat:
Storybook: https://cultureamp.design/storybook/?path=/story/guidanceblock-react--default
UI Kit: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=4513%3A0

![image (29)](https://user-images.githubusercontent.com/1621182/93952786-0e476f00-fd8d-11ea-8e55-ce8a0afcd28d.png)

ActuallyACat:
Closed in #804

PR #804, opened 2020-10-01
feat: GuidanceBlock Prominent Variant

pandurijal:
--- This PR is linked with this issue #791 ---

Hello, I created this GuidanceBlock "prominent" variant. I simply add the `prominent:boolean` property into the `GuidanceBlock` component and use a ternary operator to determine the backgroundColor used by the user. And I also using kaizen's `colorToken` to avoid hardcoding the colors.

Open to any feedback. Thank you!

image


draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -28,6 +29,7 @@ type Props = {
     }
   }
   persistent?: boolean
+  prominent?: boolean
```
ActuallyACat, line 32: One pattern that we typically lean on when there could be future visual variants to components is to use a `variant` prop, rather than booleans. This [Paragraph component](https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Paragraph/Paragraph.tsx#L43) is a great example of how we've done this in the past. 

The main motivation of doing it this way is to allow our components be flexible to future changes, and also making the component API a little easier to understand should more visual variants be added on in the future. Consider a future with multiple different visual styles of this block: 
```
type Props = {
   default?: boolean, 
   prominent?: boolean, 
   cautionary?: boolean, 
   positive?: boolean,
  /* other props */ 
} 
```

Each of these (theoretical) visual variants are exclusive, that is, you cannot have a GuidanceBlock that is a mixture of these props. So a `default` and `prominent`, or `positive` and `prominent` should not be mixed - however this approach allows consumers to mix. Having variant as a string forces the consumer to pick one visual variant. 

--- 

To be more specific, this could be approached like so:
```
type Props = {
   // we can add more variants in the future and it will be backwards compatible
   variant?: "default" | "prominent"
   /* other props */ 
} 
```

The existing users of the Guideance block will still receive the design they're expecting: 
```
const {
      actions: { primary, secondary },
      img,
      text,
      variant = "default", // 
      withActionButtonArrow
} = this.props
```

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -76,13 +78,19 @@ class GuidanceBlock extends React.Component<Props, State> {
...
     return (
       <div
         className={this.bannerClassName()}
-        style={{ marginTop: this.marginTop() }}
+        style={{
+          marginTop: this.marginTop(),
+          backgroundColor: prominent
+            ? colorTokens.kz.color.cluny[100]
+            : colorTokens.kz.color.white,
+        }}
```
ActuallyACat, line 93: For styles that don't require the javascript runtime, we typically put them in Sass. This helps us avoid using inline styles. Something like this is preferred: 
```
className={classnames(this.bannerClassName(), {
  [styles.prominent]: variant === "prominent"
}}
```

ActuallyACat (changes requested):
Hey @pandurijal 

Firstly, thank you so much for taking the time to make a contribution to Kaizen - your help is very much appreciated! I'm personally very excited to see this contribution that has come from the community, and you were able to pick up on some of the architectual patterns we use in Kaizen (such as the usage of tokens). 

I have some suggestions for changes - take a look and let me know if you have any questions.

pandurijal:
Hello @ActuallyACat, thank you for the review I totally love how you explain it to me. Besides that, I also amazed at how neat this project's code. I have already updated my PR based on your review, I hope I covered all your review here. Just let me know if there is anything I should change. Thanks!

ActuallyACat:
Branch preview: https://dev.cultureamp.design/ally/pandu/storybook/?path=/story/guidanceblock-react--prominent

ActuallyACat (approved):
LGTM - thank you Pandu :)

pandurijal:
Cool thank you for the amazing review!

ActuallyACat:
Just waiting for a designer to give this a quick once over before I merge this.

lindsey-collins:
Looks good, I'm happy to sign off on this one.

PR #807, opened 2020-10-07
fix: Navigation action buttons from being unnecessarily recreated

jeremy8883:
TitleBlockZen action buttons from being unnecessarily recreated. To do this, I've given proper keys to the primary and secondary buttons array.

The keys before were derived from the array index. This was resulting in a flaky cypress in perform-ui. We were trying to click the `primaryAction` button in the `TitleHeaderZen`. Every so often, the `click()` would fail, due to the element having disappeared from the dom. The cause being, in that short gap, the action item had been replaced with an exact replica.

I also noticed in the `ZenNavigationBar`, some menu items were getting recreated on every render, due to `key={uuid()}` on some of the components. I vaguely also remember having some issues trying to click on the top navigation bar in our cypress tests. I can't remember what the error was, but it sounds like this was likely the cause. Unfortunately I needed to use the array index, against react best practices, because the menu items have no unique identifier. To fix this, we'd need to create a breaking change, which will force all consumers to add a unique key for each menu item.


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MainActions.tsx
```diff
@@ -63,96 +63,123 @@ const MainActions = ({
...
+            align="right"
+            button={
+              <IconButton label="" reversed={reversed} icon={meatballsIcon} />
+            }
+          >
+            <MenuContent>
+              {overflowMenuItems.map((menuItem, idx) => (
+                <MenuItem
+                  {...menuItem}
+                  key={`main-action-overflow-item-menu-item-${idx}`}
```
sebpearce, line 176: Is there a way to generate a unique ID here? I've been assuming this should be changed to `uuid()`.

sebpearce:
> Unfortunately I needed to use the array index, against react best practices, because the menu items have no unique identifier. To fix this, we'd need to create a breaking change, which will force all consumers to add a unique key for each menu item.

If `uuid()` is out of the question, does this mean there is actually no way to generate robust unique IDs for sub-components within a component (not passed in)?

makenosound:
@sebpearce You could generate a hash from the input to ensure uniqueness. `menuItem` is a simple object is it?

makenosound:
Another option here could be to either generate hashed IDs (or use the index) only if the consumer hasn’t passed one in. Would mean that there’s no breaking change, but that you can give the consumer control if they need it.

jeremy8883:
> You could generate a hash from the input to ensure uniqueness

There are two downsides here:
* If two menu items are the same, there will be a key conflict. Sure, this wouldn't happen ever in production. But maybe a developer is in "WIP mode", and is playing around with the menu. Or they create a storybook story with repetitive menu items, thinking it's just for demo so it's fine. Dev time could be wasted trying to figure out the render bugs.
* If any of the properties were to change dynamically (ie. the disabled state changes from enabled to disabled), we will get the same recreation issue. ie. we end up with the same problem as using the array `index`, it can just happen in different scenarios.

> Another option here could be to either generate hashed IDs (or use the index) only if the consumer hasn’t passed one in.

I'm not sure anyone would know to use it, unless it's made as a required prop. But LMK if that's the preference, and I can add it in.

Something else to note, because the `ZenNavigationBar` is much less dynamic, these element replacement issues are much less likely to occur. It was mostly the `TitleHeaderZen` which was leading to the cypress test failures, and I was able to hard code key values thankfully. So maybe simply using the array `index` for now is fine, but we make sure we don't continue using this pattern in the future.

If I get time, I'll add [this lint rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) to perf-ui, because I've noticed this problem in our codebase as well.

jeremy8883:
> If uuid() is out of the question, does this mean there is actually no way to generate robust unique IDs for sub-components within a component (not passed in)?

If it's for a mapped array, then I think that's basically correct. At least there's nothing else I can think of.

sebpearce:
Sorry for not reviewing this yet @jeremy8883 – I'll go through this morning. I guess I was waiting to see how the conversation would turn out because the issue of how to approach dynamic keys seems like a Kaizen-wide issue rather than just this component.

I'll CC @ActuallyACat in here for visibility and ask advocates for opinions too.

ActuallyACat:
> Sorry for not reviewing this yet @jeremy8883 – I'll go through this morning. I guess I was waiting to see how the conversation would turn out because the issue of how to approach dynamic keys seems like a Kaizen-wide issue rather than just this component.

This feels like a titleblock specific issue to me... I can't think of any other component that we ask for an array of components to render, correct me if I'm wrong. If we're asking for consumers to give us a list of things to render (such as the case for titleblock), I don't think it's unreasonable to ask for an id/key for each item in that list. 

To avoid rolling out a breaking change this fix works well. This is a good opportunity to add an _optional_ `key` or `id` for each list item so we can make a future breaking change with less friction. My 2c.

sebpearce:
> This feels like a titleblock specific issue to me... I can't think of any other component that we ask for an array of components to render, correct me if I'm wrong.

Ah you may be right. I was thinking of things like `Menu`, but that actually relies on `children` rather than an abstracted array of items.

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/SecondaryActions.tsx
```diff
@@ -41,28 +41,32 @@ const SecondaryActions = ({
...
-            >
-              <MenuContent>
-                {a.menuItems.map(menuItem => (
-                  <MenuItem {...menuItem} />
-                ))}
-              </MenuContent>
-            </Menu>
-          )
+          return {
+            key: `${i}`, // We shouldn't use an index here
```
sebpearce, line 47: You left this comment which cleared this up for me:

> Unfortunately I needed to use the array index, against react best practices, because the menu items have no unique identifier. To fix this, we'd need to create a breaking change, which will force all consumers to add a unique key for each menu item.

IMHO it'd be more understandable if something similar was in these code comments too. I'm just trying to imagine what someone who hasn't seen this PR would think if they came across this comment saying "we shouldn't do this" but we're doing it anyway – having a bit of context on why we're doing it and what we should be doing in the future instead would be ideal, so anybody who wants to address this debt knows what to do.

I realise there are a few of them so we could even do something like have an explanation at the top of the file in a comment and all the comments direct the reader up there or something 🤷  Or alternatively we could create an issue and then link to that in the comment (I'm more than happy to do that if you like that idea).

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MainActions.tsx
```diff
@@ -63,96 +63,123 @@ const MainActions = ({
...
-                />
-              }
-            >
-              {menuContent}
-            </Menu>,
+            {
+              key: "primaryAction",
+              node: (
+                <Menu
+                  key="primaryAction"
```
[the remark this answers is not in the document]
jeremy8883: Sorry, I had forgotten to remove these.

sebpearce (approved):
LGTM 👍

PR #808, opened 2020-10-08
fix: Ensure delta-flat icon has path fill set to black

makenosound:
The svgo-loader expects this value to ensure automatically convert the fill to `currentColor`


ActuallyACat (approved):
LGTM :)

PR #810, opened 2020-10-09
feat: Allow TitleBlockZen to collapse when no navigation

makenosound:
The current implementation of `TitleBlockZen` intentionally includes a large space where the navigation tabs would appear, even when there are no tabs that would fill that space. This makes it rather awkward to implement designs that butt up more tightly to fill that area with content.

To counter that, this PR adds a `collapseNavigationAreaWhenPossible` prop to the component that decreases that space to a more "natural" feeling padding.

Example stories:

* [Admin variant](https://dev.cultureamp.design/allow-title-block-to-collapse-when-no-navigation/storybook/?path=/story/titleblockzen-react--admin-variant-navigation)
* [Default with collapsed nav](https://dev.cultureamp.design/allow-title-block-to-collapse-when-no-navigation/storybook/?path=/story/titleblockzen-react--default-collapsed-navigation)
* [Default with collapsed nav and card](https://dev.cultureamp.design/allow-title-block-to-collapse-when-no-navigation/storybook/?path=/story/titleblockzen-react--default-collapsed-navigation-card)

### DIFF
--- a/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
+++ b/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
@@ -283,9 +283,13 @@ const renderNavigationTabs = (
         [styles.navigationTabsContainerCollapsed]: collapse,
       })}
     >
-      <span className={styles.navigationTabEdgeShadowLeft} />
-      {navigationTabs}
-      <span className={styles.navigationTabEdgeShadowRight} />
+      {!collapse && (
+        <>
+          <span className={styles.navigationTabEdgeShadowLeft} />
+          {navigationTabs}
+          <span className={styles.navigationTabEdgeShadowRight} />
+        </>
+      )}
     </div>
   </div>
 )
--- a/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
+++ b/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
@@ -43,6 +43,7 @@ export interface TitleBlockProps {
   secondaryActions?: SecondaryActionsProps
   secondaryOverflowMenuItems?: TitleBlockMenuItemProps[]
   navigationTabs?: NavigationTabs
+  collapseNavigationAreaWhenPossible?: boolean
   textDirection?: TextDirection
   surveyStatus?: SurveyStatus
   titleAutomationId?: string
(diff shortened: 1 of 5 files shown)


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -425,6 +434,11 @@ const TitleBlockZen = ({
   const [isSmallOrMediumViewport, setSmallOrMediumViewport] = React.useState(
     false
   )
+  const hasNavigationTabs = navigationTabs && navigationTabs.length > 0
+  const collapseNavigationArea =
+    collapseNavigationAreaWhenPossible &&
+    (navigationTabs === undefined || navigationTabs) &&
```
sebpearce, line 440: Having a bit of trouble understanding the intention of the right side of the `||` here.

"If `navigationTabs` is undefined or `navigationTabs` is truthy"?

makenosound: Oh, that is because I am dumb and didn’t rewrite it when I extracted the intermediate var.

makenosound: Should be fixed now, @sebpearce

sebpearce:
> Totally open to a better name for this prop 😬

I'd sign off on the current one. Another option is `collapseBottomRowWhenPossible` (since the bottom row includes both the nav tabs and the secondary actions, and both need to be missing).

> Not sure how to usefully test this — any thoughts?

I'd personally be fine with using Storybook for this. I think the main thing we'd need to check is that each configuration looks right at mobile widths as well.

makenosound:
> I'd sign off on the current one.

Heh, cool. Can’t say it isn’t descriptive I suppose! I think your suggestion would work too, but feels like much of a muchness between them to me.

> I'd personally be fine with using Storybook for this. I think the main thing we'd need to check is that each configuration looks right at mobile widths as well.

Yeah, cool. I’ll add some easier-to-manually-check examples of things that should work at varying screen sizes.

makenosound:
@sebpearce I’ve fleshed out the example stories here to give a more clear indication of how things would collapse. This revealed an issue with the way we were hiding `navigationTabScrollerContainer`, and so I’ve rejigged it a little to handle that differently.

sebpearce:
@makenosound Code looks great and I was going to approve after checking Storybook, but realised we've still got this error: 
image

If you've smoke-checked the existing stories on your end then I'm happy to ✅  this.

sebpearce (approved):
Approved pending a check of the existing stories.

makenosound:
@sebpearce Annoyingly I’m seeing this locally now too. The two in the TitleBlockZen are:

```
// 1
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `ButtonAllowingAdditionalContent`.

// 2
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `MainActions`.
```

I also seem to have a Storybook full of:

```
export 'SomeExport was not found in 'some-package'
```

You said you’d seen this before? I had a look through Slack but couldn’t find any discussion that related to a cause that seemed likely for this.

sebpearce:
(☝️  Resolved on Slack – PR has just merged which should fix this.)

sebpearce:
Stories look good to me! 🚢

PR #815, opened 2020-10-12
feat: Add working variant of button

ellisjas:
# Objective

To add a "working" variant of button to be displayed when a particular process is loading.

# Motivation and Context 

This change adds a universal loading option. Often button makes a request that does not get an immediate response.

# Checklist

# Additional Considerations


ellisjas:
[Add working state to Kaizen button](https://trello.com/c/XUfaChmJ/210-add-working-state-to-kaizen-button)

ActuallyACat:
Language-wise, my gut-feel is to call this state `loading` rather than `working`. I think we both agree `disabled` isn't a great name, semantically. Curious if there are any particular reasons why you chose `working`?

As an aside, we may need the designers to update the UI Kit with whatever is chosen.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -52,6 +53,9 @@ export type LabelProps = {
   primary?: boolean
   secondary?: boolean
   reverseColor?: "cluny" | "peach" | "seedling" | "wisteria" | "yuzu"
+  working?: boolean | null
```
ActuallyACat, line 56: Any particular reason for using `null` over allowing the prop to be undefined when not specified? 

```
props.working -> undefined or false -> don't show working state
props.working -> true -> show working state
```

ellisjas: Yes good point

sebpearce:
Just wanted to add here for future reviewers that when I ported the CSS over to this component from the Button in `component-library` and then made changes to it, it was tricky to avoid unintentional weirdness on some variants due to the amount of complexity. I would recommend going through each individual Button story in Storybook and checking its hover, focused (via tabbing) and hovering-when-focused states.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -82,6 +89,17 @@ const GenericButton = forwardRef(
       },
     }))
 
+    if (
+      !props.iconButton &&
+      props.working !== undefined &&
+      !getWorkingLabel(props)
+    ) {
+      throw new Error(
```
ActuallyACat, line 97: I definitely agree we need to warn the user somehow that they need to use the correct aria attributes when using `working`, but I am not sure that throwing an error is the way to go 🤔 If there's no error boundary (or something else to catch this error) then the page could crash in unexpected ways. I think `console.warn` or `console.error` might be more appropriate here. Thoughts?

ActuallyACat:
Sorry for my slow review. Apart from my recent comment, everything else LGTM.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
ellisjas: Yeah that's fair. I'll replace with console.error.

ellisjas:
## Testing notes
For very zen button variant:

- [ ] Hover state fits design
- [ ] Focus state fits design
- [ ] Active state fits design
- [ ] Working state cross browser check

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
dougmacknz: Just want to throw one other idea in the mix - we might be able to enforce this with TypeScript.

If `working` is set to true, then you must specify a `workingLabel` prop. Then there can be an optional `workingLabelHidden` prop to decide whether it will be a visible label, or an `aria-label`

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -28,6 +28,14 @@ $caButton-verticalPaddingForm: calc(
   transition: none; // override Murmur global styles :(
 }
 
+@mixin working-state {
+  pointer-events: none;
```
dougmacknz, line 32: Given you have this, perhaps you can actually get rid of the `onClick` override in `GenericButton`?

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -28,6 +28,14 @@ $caButton-verticalPaddingForm: calc(
...
+
+  &:hover {
+    cursor: default;
+  }
```
dougmacknz, line 36: Looks like you don't actually need this, as long as you have the `pointer-events: none` above.

dougmacknz: This might actually go for all of the hover overrides that you have added below. I tried getting rid of them and didn't see any issues on hover.

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -149,23 +168,30 @@ const renderButton = (props: Props, ref: Ref<HTMLButtonElement>) => {
...
       onFocus={onFocus}
       onBlur={onBlur}
       onMouseDown={(e: any) => onMouseDown && onMouseDown(e)}
       type={type}
       title={label}
-      aria-label={label}
+      aria-label={working && btnWorkingLabel ? btnWorkingLabel : label}
+      aria-disabled={!!working}
       tabIndex={
         disableTabFocusAndIUnderstandTheAccessibilityImplications
```
dougmacknz, line 196: For documentation sake - I was going to say:

> I believe we'll also want to disable keyboard focus in the case that the button is `working` with `tabIndex={-1}`. You can just replace this line with a new var which is `const disableTabFocus =
    disableTabFocusAndIUnderstandTheAccessibilityImplications || working`.

On second thought, it's probably safer to not, even though it's not really interactable anymore.

dougmacknz (commented):
Looking good, I left a few comments. I'm excited about this addition!

draft-packages/stories/Button.stories.tsx
```diff
@@ -17,6 +17,8 @@ export default {
   },
 }
 
+const clickAction = () => alert("This shouldn't fire")
```
dougmacknz, line 20: Surprised that we don't have any tests for Button. This would be a good unit test to have.

dougmacknz (approved):

PR #816, opened 2020-10-13
fix: Add style disabled options in <Select>

makenosound:
# Objective

This adds a specific style to the `Option` values with the `<Select>` component when the `isDisabled` attribute for that option is set to `true.

image

image

You can view this live in the branch preview: https://dev.cultureamp.design/select/add-disabled-style/storybook/?path=/story/select-react--single

# Motivation and Context 

Currently the `isDisabled` attribute functions correctly — the option cannot be chosen — but it does not look any different from the enabled options and so could case confusion.

The disabled class sets the opacity to 30%, which is the same as the overall disabled style for the `<Select>`. It also adds a `not-allowed` cursor on hover to make it clear to users that the action is disabled.

# Checklist

# Additional Considerations
~- Does there need to be right-to-left (RTL) options for localization and internationalization?~
~- Has the test suite been updated?~
~- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?~
~- Have you updated any relevant documentation or left useful comments in the code?~
~- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?~
~- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?~
~- Have Storybook stories been updated?~

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.tsx
+++ b/draft-packages/select/KaizenDraft/Select/Select.tsx
@@ -150,6 +150,7 @@ const Option: typeof components.Option = props => (
       className={classNames("needsclick", styles.option, {
         [styles.focusedOption]: props.isFocused,
         [styles.selectedOption]: props.isSelected,
+        [styles.disabledOption]: props.isDisabled,
       })}
     />
   </div>
--- a/draft-packages/select/KaizenDraft/Select/styles.react.scss
+++ b/draft-packages/select/KaizenDraft/Select/styles.react.scss
@@ -74,6 +74,11 @@
     background-color: $kz-color-cluny-100;
   }
 
+  .disabledOption {
+    cursor: not-allowed;
+    opacity: 0.3;
+  }
+
   .noOptionsMessage {
     @include base-font-style();
   }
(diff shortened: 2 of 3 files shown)


ActuallyACat:
react-select

The disabled items still receive focus when navigating on keyboard.

makenosound:
@ActuallyACat That’s a feature of the underlying library, which I‘m not sure we’ll be able to easily undo. I would be reluctant to remove the focus style entirely (given it give feedback to the user about what is selected by the keyboard even if it’s not actionable). We could change the style slightly to make it look _more_ disabled when focused than it currently does?

ActuallyACat:
@makenosound Ah gotcha. Looks plenty disabled to me when it's in "faux-active" state.

ActuallyACat (approved):

makenosound:
Suhweet. Thanks for the review!

PR #819, opened 2020-10-15
fix: Move Title Block subcomponents out of peer deps

sebpearce:
# Objective

Moves the Kaizen components used by `TitleBlockZen` out of `peerDependencies` and into `dependencies` in the Title Block's `package.json`.

# Motivation and Context 

I discovered that when trying to work on the Stencil Title Block experiment in `node-packages`, the peer dependency configuration meant that i was using an out-of-date version of components like `Menu`, which had their version set at the root of `node-packages`. Updating this would mean a breaking change to the whole `node-packages` repo which I definitely did not want.

Having these components as regular dependencies gives us more peace of mind that changes to a component used by the Title Block will not get out of hand across multiple repos – there may be cases where we want e.g. the new version of `Menu` to be used in `TitleBlockZen`, but an older version of `Menu` in the rest of the consuming repo.

# Checklist


mbylstra:
My experience has been this: although the peer dependencies promise more consistency inside an app (same version of components so they have the same look and feel) and a way to gradually roll out customer facing changes, in practice they have created more confusion and hours of debugging pain, and I've yet to experience any benefits.

I think there was a time when we wanted to avoid having both blue and green primary buttons at the same time which might have motivated putting the draft-button in peerDeps, but I believe we're OK with "eventual consistency" (pardon the pun), as in having a few inconsistencies a long the way is find if it gets us to eventual consistency quicker.

mbylstra:
There's also an argument that it can reduce bundle sizes, but I hope we can avoid that by just having a better NPM refresh process.

jasononeil:
I think using direct dependencies, and reducing duplication (using `yarn deduplicate` to convince yarn to use the minimal number of package versions required) is the best bet at reducing bundle sizes and keeping things consistent without getting into impossible-to-manage peer dependency trees

sebpearce:
@mbylstra @jasononeil Would either of you gentlemen be so kind as to give this a review when you get a chance?

mbylstra (approved):
I think we're all in agreeance this is better 👍

mbylstra:
If you're wondering how I replied to quickly to this, it's because today I discovered Github's "real-time updates" Slack feature in Personal Settings -> Scheduled Reminders. Otherwise it would have just been sitting in my email inbox for days 😆

PR #821, opened 2020-10-15
feat: Change tooltip tag to span and add inline option

makenosound:
[👀📐 Storybook preview](https://dev.cultureamp.design/change-tooltip-tag/storybook/?path=/story/tooltip-react--inline)

# Objective

Allow tooltips to appear as inline elements.

# Motivation and Context 

There are situations where we’d like to use the tooltip as part of a block of text — usually to indicate that a link in that text opens in a new tab. The current implementation makes this awkward:

* The `display` value means if it’s a lengthy link it appears as a separate block
* The `<div>` wrapper means that we get annoying warning about invalid DOM nesting:

```
Warning: validateDOMNesting(...): <div> cannot appear as a child of <p>
```

The changes are:

* Change any `<div>` tags used in the tooltip to `<span>. They all have display properties/position set in such a way that this doesn’t affect their position
* Add an `inline` prop to the `<Tooltip>` component that sets the display to inline.

I’ve also fixed an issue I noticed when testing in IE11 where the content of the tooltip was misaligned. It is now correctly aligned in all the examples in IE11 _and_ modern browsers.

# Checklist

~[ ] I have considered likely risks of these changes and got someone else to QA as appropriate~


mbylstra:
The below variant isn't sitting below the tag, but that may well be an existing issue. Could be worth running a Chromatic build. Let me know if you haven't done that before and are interested in trying - I can show you what do (it's a pretty manual process at the moment).

image

makenosound:
@mbylstra Thanks for checking that. I swear I had done that, but perhaps I was looking at cached version. Have pushed up a few tweaks that fix that issue and make things a little more consistent overall.

I’ll run a Chromatic build now, which I am guess is pressing the big old button in Buildkite?

mbylstra:
No probs. Some of our stories (I forgot which ones) can be a bit broken. So it's never clear whether it's a problem in the component or just the story itself.

Yeah, it's just pressing the button, but to actually get to Chromatic, you have look at the bottom of the Buildkite output, and there's a link in plain text you have to copy and paste into the browser. I might need to give you access, let me know if you don't get let it. Another issue is we have to manually reset the baseline from the master build every now and then (this is not done often, so you'll likely get lots of irrelevant and out of date diffs). This is all because the Chromatic github integration asks for more permissions than we're allowed to give. It's all a bit of an ordeal, but definitely worth it if it's a sweeping change of the entire design system. In this case, you're already aware of what the issue is so it's not that necessary, but feel free to do it if you're interested!

makenosound:
Yeah, cool. I would definitely like to know how to do it, so I’ll have a squiz and see if I can get through the process. Thanks!

mbylstra:
Another thing I forgot to mention, is that even after you reset the baseline (you go to the latest kaizen buildkite build and run the chromatic build from there first) and the build has finished, it seems you some times have to wait 5-10 minutes for Chromatic to compare against the latest baseline. That can be pretty confusing.

makenosound:
Have run that build and amusingly it is useless here because it doesn’t show the `:hover` interaction state. Thanks for the pointers on how it works though, I will definitely find it useful in the future.

Do the updates here look OK to you now?

mbylstra:
Looks good! Sorry for not getting back to this sooner. Your replies were sitting in my email inbox and I'm terrible at checking that. Feel free to PM me in Slack in future to let me know as soon as you've addressed a PR review in future for a quicker turnaround.

Ah yeah, I hadn't thought that the story requires interaction. It'd be great if we had stories that demonstrated all of the states without requiring user interaction (for Chromatic, and just to quickly see states when viewing stories), but that's probably not our highest priority at this point in time.

mbylstra (approved):

makenosound:
@mbylstra No worries at all. Thanks for checking it over again!

> It'd be great if we had stories that demonstrated all of the states without requiring user interaction (for Chromatic, and just to quickly see states when viewing stories), but that's probably not our highest priority at this point in time.

Yeah, I think for that we’d probably need to make it part of our core workflow for Kaizen to have it work well, since it requires us thinking about how we structure our components so that things like tooltip visibility can be controlled by props.

mbylstra:
I was wondering if it's even possible to capture hover states in Chromatic, and it seems not (https://www.chromatic.com/docs/hoverfocus). You'd have to rewrite the component to be controlled with props as you say.

I did find this storybook addon: https://github.com/Ergosign/storybook-addon-pseudo-states, but I haven't looked into how it works (or if it works without having rewrite the components).

PR #826, opened 2020-10-16
feat: TitleBlockZen. Add a  variant for the badge

vicchirino:
# Objective
Include a `default` variant for the badge that the component display for survey status.

# Motivation and Context 
In #team_apollo we are currently working on Evaluation cycles admin dashboards. In one of the pages that display feedback requests information we want to display the `due date` of the current Evaluation Cycle.
Since we only allowed bags for `live` and `draft` status, thought about adding a status that will use the `default` variant of the `Tag` component. 
In our case will be used, as mentioned above, for `due date` .

image

image

📹 : http://g.recordit.co/BKZ7DMGU9W.gif

# Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,7 +9,7 @@
 # Exclude modifications to existing packages by assigning it to no one
 /packages/hosted-assets @ghost
 
-# Pipelines are maintained by Delivery Engineering:
-/.buildkite/ @cultureamp/delivery-engineering
-/docker-compose.yml @cultureamp/delivery-engineering
-/lerna.json @cultureamp/delivery-engineering
+# Pipelines are maintained by Foundation team FE Engineers
+/.buildkite/ @cultureamp/frontend-ops-approvers
+/docker-compose.yml @cultureamp/frontend-ops-approvers
+/lerna.json @cultureamp/frontend-ops-approvers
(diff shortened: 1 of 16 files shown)


jeremy8883 (approved):
LGTM!

sebpearce (changes requested):
Not sure if my understanding is off here, but I think there may be some confusion around "Badge" vs. "Tag". This is what we call Badges and Tags in Kaizen:

image

[(Link to Badge sticker sheet in Figma)](https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A14397)

The update to `TitleBlockZen` looks exactly right to me 👌 , but the stories need some changes to the wording to make them fit.

I apologise for the lack of clearer documentation on the structure of the Title Block – here is a quick summary of what each bit is called:

image

So for this PR, I would suggest getting rid of all references to the Badge from these stories, including the hook that holds state for them, since this change is not actually related to Badges.

The stories ideally highlight just one specific feature, so you might opt to have just a `title` and `primaryAction` or something (with no badge) as well as the `surveyStatus` that you want to draw attention to. You could even omit the `primaryAction` entirely if you want.

Also please let me know if there was any documentation in particular that confuses the meaning of Badges and Tags – I'll update it ASAP.

vicchirino:
Thanks for the review folks. @sebpearce make sense your comments, i think it was more my confusion than a documentation issue. Let me know if the changes that i did make sense. Thanks again

sebpearce (approved):
Looks great 👌  Thanks for that.

PR #828, opened 2020-10-19
feat: point $ca-default-font-xxx to $ca-inter-font-xxx

mbylstra:
This is a **canary release**

This removes all traces of Open Sans from Kaizen, and makes Inter the default for `$ca-default-font-xxx`

The `$ca-default-font-xxx` vars are legacy (in both files) and should be removed eventually, but this is far easier (not a breaking change, codewise). The nice thing is that the name `$ca-default-font-xxx` is still applicable, so there's no need to change the name like we needed to do for `$ca-font-ideal-sans` (which was a huge change❗). 

The consuming repo (guess which one!) will just need need to bump Kaizen, and hopefully most of the Open Sans usages will dissappear.

This has not been tested (by for example yarn linking locally), so there may be mistakes. Please let me know if find any in code review! I made it a canary release to be able to use our screenshots / visual regression tool against the consuming repo. This test can be used to decide if we want to make this an offical release (and bump the consuming repo)

### DIFF
--- a/packages/component-library/styles/type.scss
+++ b/packages/component-library/styles/type.scss
@@ -11,9 +11,14 @@ $ca-weight-medium: 500;
 $ca-weight-semibold: 500; // Note: in Sketch, semibold is 600. But murmur has an existing value of semibold=500 that is heavily used.
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
-$ca-default-font-family: "Open Sans", Helvetica, Arial, sans-serif;
-$ca-default-font-base-size: 0.875rem; /* 14px */
-$ca-default-font-descender-height: 0.115;
+$ca-inter-font-family: "Inter";
+$ca-inter-font-base-size: 1rem; /* 16px */
+$ca-inter-font-descender-height: 0.14;
+
+// Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
+$ca-default-font-family: $ca-inter-font-family;
+$ca-default-font-base-size: $ca-inter-font-base-size;
+$ca-default-font-descender-height: $ca-inter-font-descender-height;
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
 $ca-ideal-sans-font-family: "Ideal Sans A", "Ideal Sans B",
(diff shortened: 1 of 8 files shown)


packages/component-library/styles/type.scss
```diff
@@ -11,30 +11,30 @@ $ca-weight-medium: 500;
 $ca-weight-semibold: 500; // Note: in Sketch, semibold is 600. But murmur has an existing value of semibold=500 that is heavily used.
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
-$ca-default-font-family: "Open Sans", Helvetica, Arial, sans-serif;
-$ca-default-font-base-size: 0.875rem; /* 14px */
-$ca-default-font-descender-height: 0.115;
+$ca-inter-font-family: "Inter";
```
ActuallyACat, line 14: This is the variable that's used globally in other repos, right? Should It have some fallback?

mbylstra: What I realised is that none of the other fonts (including Inter) have fallbacks 😬 , so I kind of wanted to defer that. I mean it seems pretty harmless to add fallbacks, but I'm not 100% sure what the current fallback best practice is. I'll add a github issue about this. Now that I look at it, they kz-xxxx mixins have the fallbacks, just not the legacy `$ca-inter-font-family`. I'm just not sure of the ramifactions of changing that. I guess now is a good time though 🤷‍♂️

mbylstra: I also wonder if Helvetica is a better fallback than just `sans-serif`, because the system font in Apple products is no longer Helvetica. I think it's "SF Pro"?

ActuallyACat: That's not ideal 😬 At the very least it should be `Inter, sans-serif` (with the view to come back to this later). I'm not sure what a better fallback is at this point in time.

packages/component-library/styles/type.scss
```diff
@@ -11,30 +11,30 @@ $ca-weight-medium: 500;
 $ca-weight-semibold: 500; // Note: in Sketch, semibold is 600. But murmur has an existing value of semibold=500 that is heavily used.
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
-$ca-default-font-family: "Open Sans", Helvetica, Arial, sans-serif;
-$ca-default-font-base-size: 0.875rem; /* 14px */
-$ca-default-font-descender-height: 0.115;
+$ca-inter-font-family: "Inter", sans-serif;
```
mbylstra, line 14: Previously this was falling back to `$ca-default-font-family `which used to be Open Sans. I don't know how intentional that was. `sans-serif` is a temporary fallback until we figure out the best practice for font fallbacks (https://github.com/cultureamp/kaizen-design-system/issues/829).

mbylstra:
BTW, I also rebased this on master since the build bugfix. I thought I should also rebase the `canary` branch so it gets the fix, but I wasn't able to push to that branch directly. I'm not sure if that's necessary.

ActuallyACat (approved):

PR #833, opened 2020-10-21
feat: Remove outside modal click close from input modals

dougmacknz:
# Objective
Disables the ability to close input edit modals via clicking on the overlay.

# Motivation and Context 
We are getting a lot of reports from frustrated customers on Capture where they have lost their comment because they accidentally clicked outside of the modal. This is not a complete fix for that issue, but more of a quick fix to stop the bleeding on that issue.

I brought up this change in the advocados channel to check if we're happy to make this change across all input edit modals:
https://cultureamp.slack.com/archives/GLZGHEKG9/p1603240449048600

We may want to add some flexibility on this in the future to allow some modals to close via overlay click, but for now we have decided that removing this is a better default.

# Additional Considerations


sebpearce (approved):
🔪

sebpearce:
Happy to see this binned after losing many a comment.

PR #841, opened 2020-10-28
chore: Expose NavigationTabProps for external consumption

dougmacknz:
# Objective
Exports ﻿`NavigationTabProps` for external consumption

# Motivation and Context 
I'm doing a bit of work on the title block in Murmur and could use these props to avoid rewriting them.


sebpearce (approved):
Good call, I should've exposed these from the start!

PR #843, opened 2020-10-28
docs: new drawer guidelines

zakwash01:
docs: created new drawer guidelines

Preview: https://dev.cultureamp.design/zak/new-drawer-guidelines/components/drawer/

# Objective
Created new documentation for the 'Drawer' component

# Motivation and Context 
We've recently added a new slide-out drawer that is used for adding and editing individual user details. 
This documentation is intended to bring this new drawer into part of a family that includes the help centre, and the dossiers that are used in Performance. 

# Checklist

[x] Visuals have been reviewed by Marc 
[x] Michael Bylstra has reviewed for FE
[x] Design Advocates group has contributed to guidelines


ActuallyACat:
Recently we added the dossier to the titleblock. Are drawers a subset of dossiers? Or should we instead refer to dossiers as drawers? What is the difference between the two (if any)?

site/docs/components/drawer.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+*   See [animation guidelines](/guidelines/animation/)
+*   Provide a [skeleton loading pattern](/components/loading-skeleton/) when possible.
+
+**Loading patterns**
+*   Slide in from the right 
+*   Timing: 300ms 
+*   Ease-in-out easing
+
+### Accessibility
```
ActuallyACat, line 92: This is fantastic, I'm glad this guide includes accessibility ⭐️

My only thought is whether we should also trap focus when we're using parallel drawers - it seems a little inconsistent to allow clickable actions outside the drawer for mouse users, but lock to the drawer for keyboard users.

site/docs/components/drawer.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+### Medium vs Small
+
+**Use Medium Drawers (600 px):**
+*   When you need to display additional or supplemental page content
+*   When you want to provide additional navigation within the Drawer such as [Tabs](/components/tabs/).
+
+**Use Small Drawers (375 px):**
+*   When it’s important to stay in context and when the action applies changes to the parent page.
+*   When you expect this functionality to be required on mobile. Because a small drawer does not need to rearrange or re-flow like a page or modal does, it is often a friendlier experience on mobile.
```
mbylstra, line 37: It's not a big deal, but I'm not sure "re-flow" is accurate here. This usually refers to when browsers need to re-draw the page due to the user changing the window size (if the content changes this can also trigger a reflow/redraw). But you wouldn't ever change the window size on mobile so it wouldn't need to reflow technically speaking (it would just render once). I guess what you're saying is that a small drawer will look the same on mobile as it will be the same width, whereas a medium one would get squashed?

site/docs/components/drawer.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+Users can dismiss a Drawer by:
+*   Clicking an X close button
+*   Pressing the Esc key
+*   Clicking the primary action (e.g. Save or Submit) 
+*   Upon dismissal, the drawer focus should be restored to the triggering element
+
+### Mobile/Responsiveness
+
+*   A Small Drawer maintains the same behavior all the way down to viewports of 300px, at which point it begins taking up 100% of the viewport width.
```
mbylstra, line 79: I'm confused here. Wouldn't it start taking up 100% of the viewport width at 375px if the drawer is 375px wide? I'd suggest something more like "all the way down to viewports of 405px" (375px + 20px). Because if you have a screen that's 380px and the drawer is 375px wide, you'll get a very narrow (and ugly?) 5px gap to the left, so it might be better to fill the whole screen at screen widths less than about ~400px.

mbylstra (approved):
From a technical perspective this gets a thumbs up from me 👍 
There's a few details about responsiveness that are a bit unclear, but I think details on responsiveness are hard to describe in words or without getting overly technical, so it might be better for a developer to just pair with a designer to nut out the details when it comes time to implement it. The general gist of responsiveness makes sense.

zakwash01:
> Recently we added the dossier to the titleblock. Are drawers a subset of dossiers? Or should we instead refer to dossiers as drawers? What is the difference between the two (if any)?

The intention is that Dossiers use the Parallel Drawer component. Dossier is just a customer-facing name. 

I think it's fine for 'Drawer' to be an internal Culture Amp term and then the actual user-facing functionality can be given a fit-for-purpose name. Another example would be that the Help Centre can just be the Help Centre, rather than the Help Drawer.

PR #846, opened 2020-10-29
feat: Added y box offset option to Popover component

stephwestgarth:
# Objective
Due to the fact we don't have a way to set yPosition relative to target in popover, this is an intermediary solution to fix that problem. Ideally, we'd actually do this as part of the popover class, but this method allows us to set this in the meantime prior to the popover component re-engineering process.

# Motivation and Context 
In the 1-on-1s tool, we want to set positioning of the popover to be above an icon, but due to the way we set the x offset, it makes the y offset positioning unnecessarily difficult.

A conscious decision that was made was that we will not be adding the padding for the little notch as part of this y offset so all of those calculations are up to the user; in the eventual ideal solution, this would be taken into consideration.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/popover/KaizenDraft/Popover/Popover.tsx
+++ b/draft-packages/popover/KaizenDraft/Popover/Popover.tsx
@@ -22,7 +22,7 @@ export interface Props {
   readonly dismissible?: boolean
   readonly singleLine?: boolean
   readonly children: React.ReactNode
-  readonly boxOffset?: number
+  readonly boxOffset?: BoxOffset
 }
 
 type Variant =
@@ -102,12 +102,22 @@ const Popover: Popover = React.forwardRef<HTMLDivElement, Props>(
   )
 )
 
-const getRootStyle = (boxOffset: number | undefined) => ({
-  transform:
-    boxOffset == null
-      ? "translateX(-50%)"
-      : `translateX(calc(-50% + ${boxOffset}px)`,
-})
+type BoxOffset = number | undefined | {
+  xOffset: string,
+  yOffset: string
+}
+
+const getRootStyle = (boxOffset: BoxOffset) => {
+  if (boxOffset == null) {
+    return { transform: "translateX(-50%)"}
+  }
+
+  const translate = typeof boxOffset === "number"
+    ? `translateX(calc(-50% + ${boxOffset}px))`
+    : `translate(${boxOffset.xOffset}, ${boxOffset.yOffset})`
+
+  return { transform: translate }
+}
 
 const mapVariantToBoxClass = (variant: Variant): string => {
   switch (variant) {
--- a/draft-packages/popover/KaizenDraft/Popover/Popover.tsx
+++ b/draft-packages/popover/KaizenDraft/Popover/Popover.tsx
@@ -38,6 +38,14 @@ type Position = "start" | "center" | "end"
 
 type Size = "small" | "large"
 
+type BoxOffset =
+  | number
+  | undefined
+  | {
+      xOffset: number // to ensure a non-breaking change, xOffset can only be a number
+      yOffset: string
+    }
+
 type Popover = React.FunctionComponent<Props>
 
 const Popover: Popover = React.forwardRef<HTMLDivElement, Props>(
@@ -102,14 +110,6 @@ const Popover: Popover = React.forwardRef<HTMLDivElement, Props>(
   )
 )
 
-type BoxOffset =
-  | number
-  | undefined
-  | {
-      xOffset: string
-      yOffset: string
-    }
-
 const getRootStyle = (boxOffset: BoxOffset) => {
   if (boxOffset == null) {
     return { transform: "translateX(-50%)" }
@@ -118,7 +118,7 @@ const getRootStyle = (boxOffset: BoxOffset) => {
   const translate =
     typeof boxOffset === "number"
       ? `translateX(calc(-50% + ${boxOffset}px))`
-      : `translate(${boxOffset.xOffset}, ${boxOffset.yOffset})`
+      : `translate(${boxOffset.xOffset}px, ${boxOffset.yOffset})`
 
   return { transform: translate }
 }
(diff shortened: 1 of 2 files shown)


draft-packages/popover/KaizenDraft/Popover/Popover.tsx
```diff
@@ -38,6 +38,14 @@ type Position = "start" | "center" | "end"
 
 type Size = "small" | "large"
 
+type BoxOffset =
+  | number
+  | undefined
+  | {
+      xOffset?: number // to ensure a non-breaking change, xOffset can only be a number
+      yOffset?: string
+    }
```
thrgamon, line 47: Neat! The only thing I wonder if it is worth the more complicated interface of allowing a string as `yOffset` for the extra flexibility that provides?

It sounded like this component might be getting re-designed - what is the timescale for that change?

stephwestgarth: So I've allowed string as it means that people can set calculations etc to make sure it's set just right. Ideally, x would also be string but that would potentially be a breaking change

thrgamon: Yep, so you think the extra flexibility justifies the slightly more complex interface (accepting two different types of values for similar fields)?

stephwestgarth: I do, as I think not allowing this flexibility for the y offset would actually significantly decrease its usefulness.

thrgamon (approved):
Looks great!

draft-packages/popover/KaizenDraft/Popover/Popover.tsx
sebpearce: @stephwestgarth Could you not enjoy the same flexibility for `xOffset` without a breaking change by making it a union type of `number | string`?

stephwestgarth: Do you have a suggestion of what that y offset number would represent? Would it be `px` or `%`? I think that is one of the reasons why I also left it to the user to specify

sebpearce: I think I'd lean towards a `px` value if it's just a number. For a percentage I think I would want to see a name like `yOffsetAsPercentage` or a decimal value like `0.5`.

stephwestgarth: Sorry, I forgot to tag you @sebpearce! 😅

stephwestgarth: Hm, I wonder if that is starting to over-engineer what is meant to be an unblocking change. Applying the [yagni principle](https://martinfowler.com/bliki/Yagni.html), especially as we want to re-engineer how this works entirely so it's not as dumb of a component, is that adding a feature in that won't be used? I like the idea but get concerned about adding in complexity like that for no reason

sebpearce: I think we're seeing this from different angles – I thought from the comment below, you meant that you would like to have `xOffset` as a string, but couldn't because it would create a breaking change:
image
I didn't consider that this would be adding a feature.

My main concern with any temporary/unblocking change is: how hard is this going to be to climb out of when we do actually want to do that eventual "proper" change? If you feel that it'll be fairly easy, then I say go for it.

stephwestgarth: Ah, I'm on the same page now. To clarify, x offset is what the old version of boxOffset represented. Y offset is entirely new, and adding in extra complexity for something that _hopefully_ won't exist long is a bit of overkill imo.

So my impression of how we'd resolve this in future is it wouldn't be css driven at all (or css would be for minute changes, in which case it would likely be px like x offset); the ideal solution would actually be driven by js (intelligent target positioning to prevent off-screen popovers etc), so how this works would be changed entirely.

sebpearce: Sounds good to me!

ActuallyACat: Heh yeah, this was a suggestion from me. I think the `number | undefined | { xOffset: string, yOffset }` is a good middle ground for now. The tradeoff that @thrgamon mentioned is spot on - a slightly more complicated API now to avoid a breaking change for this small fix. My rationale is that popovers need a little more love in the future and I'd rather set aside a breaking change to rethink a few things about this component (like better OOTB position that @stephwestgarth mentioned - probably using something like popperjs).

What I think may have happened is this component was initially made to support one orientation (hence only needing to support xoffset).

jeremy8883: Hey! So I would have added this property originally, so I can provide some context.

The `boxOffset` property was designed as a way to offset the main popover rectangle from its arrow. eg.

image

I'm not sure if this is still used anymore in perform-ui, now that we have the new zen headers. So it is possible that we could remove it.

With that said, my understanding with this update is that it offsets the entire popover, not just the rectangle. This is different functionality, therefore if this feature ends up staying in, it should be added as a separate property. Maybe `popoverOffset` or `rootOffset`?

> What I think may have happened is this component was initially made to support one orientation (hence only needing to support xoffset).

Almost yeah. Since `boxOffset` only moves the rectangle and content, there'd never be a reason to move the box away from its pointer arrow. If we were ever to support "left" | "right" popover positioning, `boxOffset` would act on the y axis.

I don't have full context of the problem to solve, but as an alternative, maybe a wrapper around the Popover could work? That way you wouldn't need to update kaizen eg.

```
<div style={{ transform: "translateY(20px)" }}><Popover /></div>
```

Or a negative margin could also work - nested transforms can sometimes cause weird aliasing effects.

jeremy8883: And apologies for the lack of documentation here.

stephwestgarth: Hey @jeremy8883 , so I can see that we also use it in the getRootStyle function, which implies this is a style of the root. I guess that is why we've kept it here, as it is used in both the root offset and this arrow offset.

jeremy8883: Yeah sorry for the confusion, the reason why it does an offset there, is because it needs to compensate for the change made to the rectangle position. eg. before the compensation transform:

image

after the compensation transform:

image

stephwestgarth: So from what you're saying, I think it is in the right place for now given how it currently works, but we need to rethink how we do this in future entirely (which is already on the todo list!)

PR #848, opened 2020-10-29
docs: Remove CheckboxGroup/RadioGroup story styles

### LINKED ISSUE #770: Global styles for stories affects all stories
Any global styles that have been defined in storybook apply to all stories. As an example, checkboxGroup and radioGroup stories include global CSS. This was the cause of a visual regression that required [this patch](https://github.com/cultureamp/kaizen-design-system/pull/768/files).
https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/stories/CheckboxGroup.stories.scss#L3-L9 
Screen Shot 2020-09-08 at 4 53 39 pm

A suitable fix would include: 
1. Scope the styles for these two stories, so they don't effect other stories. 
2. Use/apply a stylelint rule that bans global CSS.

sebpearce:
# Objective

Prevent global type selectors in some stories from affecting the look of other stories. Closes #770.

# Motivation and Context 

We have occasionally run into cases where an `a` type selector in two particular stories files has affected another story, due to the fact that they end up in the global namespace. This is dangerous because it could mislead engineers and potentially cause visual bugs in production.

# Checklist

### DIFF
--- a/draft-packages/stories/CheckboxGroup.stories.scss
+++ b/draft-packages/stories/CheckboxGroup.stories.scss
@@ -1,9 +1,0 @@
-@import "~@kaizen/component-library/styles/color";
-
-a {
-  color: $ca-link-text-color;
-  text-decoration: none;
-  &:hover {
-    text-decoration: underline;
-  }
-}
--- a/draft-packages/stories/CheckboxGroup.stories.tsx
+++ b/draft-packages/stories/CheckboxGroup.stories.tsx
@@ -1,7 +1,6 @@
 import { CheckboxGroup } from "@kaizen/draft-checkbox-group"
 import { CheckboxField, Label } from "@kaizen/draft-form"
 import * as React from "react"
-import styles from "./CheckboxGroup.stories.scss"
 
 interface RenderProps {
   checkedStatus: string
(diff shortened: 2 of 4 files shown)


sebpearce:
@ActuallyACat

One of the suggested fixes here was "Use/apply a stylelint rule that bans global CSS", which I agree with.

But it looks like stylelint removed its rule to prevent type selectors in 8.0.0 (screenshot). From what I can see, the only way to do it now is to write a custom rule 🙃  

image

sebpearce:
Just revisited this today – I can't actually see a way to specify a disallowed selector! You can have disallowed pseudoclasses and disallowed combinators and disallowed attribute selectors, but you can't ban things like `a {}`.

Courses of action I can see from here:

1. Merge this PR and close the issue it references (#770).
2. Merge this PR, but instead of closing the issue, edit it so that the issue just says "find a way to ban type selectors unaccompanied by classes" and include the context of this PR in the issue description, knowing that it may not be dealt with any time soon.
3. Actually write a custom rule in JS for stylelint that does what we want.

I'm leaning towards (1), but open to other opinions!

PR #850, opened 2020-10-29
fix: Clean up eslint warnings

### LINKED ISSUE #849: There are a lot of eslint warnings for Kaizen
Engineers will run into the following when pushing to Kaizen, since a push will trigger eslint to run:

image

These should be cleaned up so they don't clutter people's terminals.

sebpearce:

Consumers that were passing in a now-restricted type to a prop that accepted e.g. "any" as a type may now break. Fortunately there are not many of these in Kaizen components' APIs.

# Objective

Remove the 23 eslint warnings that were cluttering up the terminal every time someone pushes (since a git hook runs eslint). Closes #849.

# Motivation and Context 

Having lots of warnings is annoying, distracting, and trains people to stop paying attention to them.

# Checklist


packages/component-library/components/Button/components/GenericButton.tsx
```diff
@@ -36,7 +36,7 @@ type LabelProps = {
 
 type Analytics = {
   eventName: string
-  properties: object
+  properties: Record<string, unknown>
```
sebpearce, line 39: I'm unsure whether this needs to remain `object` for the purposes of consistency. Can anyone more familiar weigh in on this?

makenosound:
Changes look fine to me, @sebpearce. I’m unsure about the `Analytics` type, but I _think_ that it’s likely safe too.

lijuenc:
Hmm, looks like a whole bunch of TS compilation failures because of this change: https://github.com/cultureamp/kaizen-design-system/pull/850/files#diff-0619a5bc06936366438463e9672232b7e31d1e517088d30f431b2905c3c24eb0

lijuenc:
Also, any thoughts about making these report as errors instead of warnings since they're things that we want to enforce anyway?

sebpearce:
> Hmm, looks like a whole bunch of TS compilation failures because of this change: https://github.com/cultureamp/kaizen-design-system/pull/850/files#diff-0619a5bc06936366438463e9672232b7e31d1e517088d30f431b2905c3c24eb0

Thanks for catching that @lijuenc – I might just keep it as `object` and make the linter ignore it.

> Also, any thoughts about making these report as errors instead of warnings since they're things that we want to enforce anyway?

Depending on the error/warning, I can see arguments both ways. The main one in favour of keeping them as warnings would be that it's fewer non-dealbreaker things to slow you down when scrambling to get something out. But the advantages of making them errors are pretty clear too. And then there are some warnings I don't even know if we should keep at all, like line length. Were there any particular warnings you thought should be errors?

lijuenc:
>Were there any particular warnings you thought should be errors?

Oh just in generall. But in this PR, for example, there are a couple of changes from plain functions to arrow functions. If the warnings just keep building up (since they're not enforced, just warnings) until they get to a point where they become  annoying and someone has to go back and change them anyway, wouldn't it make sense to make them errors earlier on?

ActuallyACat:
Some of the rules were intended to be errors ([like this one](https://github.com/cultureamp/kaizen-design-system/blob/master/.eslintrc.js#L35-L36)) but weren't at the time as it _could_ (although unlikely) break something for our consumers. Anything with a TODO on it should be flipped to error now.

lijuenc:
Nice, nearly there I think. There's still something incompatible with going from `object` to `Record<string, unknown>`. I think what you might want is `Record<string, any>` instead. `unknown` is probably more technically correct than `any` but since this is just an internal function not exposed anywhere I think it's fine. Type guarding against all the possible types to satisfy `tsc` may be too much trouble.

> Hmm, looks like a whole bunch of TS compilation failures because of this change: https://github.com/cultureamp/kaizen-design-system/pull/850/files#diff-0619a5bc06936366438463e9672232b7e31d1e517088d30f431b2905c3c24eb0

sebpearce:
Thanks @lijuenc – I had actually made the change to `any` last Friday but forgot to push it! Will do now.

lijuenc (approved):

PR #851, opened 2020-10-29
feat: Add table header wrapping and alignment support

jeremy8883:
# Objective
* Add table header wrapping and alignment support. The props to supply are `wrap` and `align`

# Motivation and Context 
* In perf-ui, we have some table headers with long headings, and not much real estate. Our only option is to display them on multiple lines on small and medium screen sizes.

# Screenshots

image

# Checklist

* [ ] If this contains visual changes, has it been reviewed by a designer? 
  (I've considered the risks, haven't gotten someone to QA it yet)


draft-packages/table/KaizenDraft/Table/styles.scss
```diff
@@ -35,10 +35,25 @@ $row-height: 60px;
...
+  white-space: normal;
+}
+
+.headerRowCellAlignCenter {
+  text-align: center;
+  justify-content: center;
+}
+
+.headerRowCellAlignEnd {
+  text-align: right;
```
jeremy8883, line 55: Just confirming, does this work as expected for RTL languages?

draft-packages/table/KaizenDraft/Table/styles.scss
```diff
@@ -35,10 +35,25 @@ $row-height: 60px;
   text-overflow: ellipsis;
   overflow-x: hidden;
   white-space: nowrap;
-  padding: 0 $ca-grid;
-  height: $ca-grid * 1.5;
+  padding: 8px $ca-grid;
```
jeremy8883, line 38: This removal of the fixed height was required to allow for text wrapping.

The 8px is to get the height of the cell roughly the same size as it was before. I was going to do `$ca-grid *0.27`, but I wasn't sure if that defeats the point of the ca-grid constant.

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -91,6 +93,14 @@ export const TableHeaderRowCell: TableHeaderRowCell = ({
   onCheck,
   active,
   automationId,
+  // I can't say for cetin why "nowrap" was the default value. Normally you wouldn't
+  // want to clip off information because it doesn't fit on one line.
+  // My assumption is that because since the cell width rows are decoupled, a heading
+  // cell with a word longer than the column width would push the columns out of
+  // alignment? I'm not sure.
+  // Anyway, we can override this default behaviour by setting wrapping to "wrap".
+  wrapping = "nowrap",
```
jeremy8883, line 102: This is to preserve existing behaviour. Would it make sense to either turn wrapping on by default, or turn on an ellipsis instead of using `nowrap`?

draft-packages/table/KaizenDraft/Table/styles.scss
ActuallyACat: How I usually think about this is any [measured alignments](https://liferay.design/articles/best-practices/measured-vs-optical-alignment/) should always use `$ca-grid`, and everything else (optical adjustments, things like this) is okay to use other values as long as it's clear why. In this case, I would drop a comment for the next person.

draft-packages/table/KaizenDraft/Table/styles.scss
ActuallyACat: Depends what you're expecting. The easiest way to check (or at least what I do) is to add `#root { direction: rtl; }` in the dev console. There look to be some alignment issues: 
Screen Shot 2020-11-04 at 10 21 18 am

ActuallyACat: Use `text-align: right / left` when you want to ignore the user's direction preferences or in situations where alignment is independant or direction, and `text-align: start / end` when you want to change alignment based on direction

draft-packages/table/KaizenDraft/Table/Table.tsx
ActuallyACat: > I can't say for cetin why "nowrap" was the default value.
I'd hazard a guess it's because of this:

```
Responsive:

    On small screens, the table is sideways scrollable.
    On touch devices (small or large), scroll is available on touch.
```
https://cultureamp.design/components/table/

But I could be wrong ¯\_(ツ)_/¯

ActuallyACat (changes requested):

draft-packages/table/KaizenDraft/Table/styles.scss
jeremy8883: Yeah I never used `start / end` that because of IE11, but now that you mention it, the IE11 styling doesn't need to be perfect. I've fixed that up now.

draft-packages/table/KaizenDraft/Table/Table.tsx
jeremy8883: If that's the case, would you still think it make sense to turn wrapping on by default? You can still have wrapped header cells along with a table that requires horizontal scrolling.

ActuallyACat: It's probably safer to leave as is.

ActuallyACat (approved):

PR #853, opened 2020-11-05
feat: Add 'guidance' as a variant for popovers

thrgamon:
# Objective
Add the 'guidance' variant for popovers.

# Motivation and Context 
We need to add this variant in as the lightbulb icon is going to be introduced to represent recommendations or tips, and we do not yet have a popover variant that allows for that. It is also important so we can achieve consistency as demonstrated in the design.

## Design
Screen Shot 2020-11-05 at 11 33 34

## Storybook component
Screen Shot 2020-11-05 at 11 22 59

# Checklist

### DIFF
--- a/draft-packages/popover/KaizenDraft/Popover/styles.scss
+++ b/draft-packages/popover/KaizenDraft/Popover/styles.scss
@@ -80,6 +80,16 @@ $large: 450px;
   @include arrow($kz-color-cluny-100, $kz-color-cluny-200);
 }
 
+.guidanceBox {
+  @extend %box;
+  background: $kz-color-cluny-100;
+  border-color: $kz-color-cluny-200;
+}
+
+.guidanceArrow {
+  @include arrow($kz-color-cluny-100, $kz-color-cluny-200);
+}
+
 .positiveBox {
   @extend %box;
   background: $kz-color-seedling-100;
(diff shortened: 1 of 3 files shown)


draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -80,6 +80,16 @@ $large: 450px;
...
+.guidanceBox {
+  @extend %box;
+  background: $kz-color-cluny-100;
+  border-color: $kz-color-cluny-200;
+}
+
+.guidanceArrow {
+  @include arrow($kz-color-cluny-100, $kz-color-cluny-200);
+}
+
```
stephwestgarth, line 92: If these are identical to the informativeBox/Arrow, could we just use `.informativeBox, .guidanceBox` instead so they get updated at the same time (or at least it's more obvious two places should be affected by those changes?)

draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -136,6 +146,10 @@ $large: 450px;
   color: $kz-color-cluny-500;
 }
 
+.guidanceIcon {
+  color: $kz-color-cluny-500;
+}
+
```
stephwestgarth, line 152: Same as above for Icon. Others may disagree with this thinking but assuming they should be the same, I think we should make that as obvious as possible in the code.

stephwestgarth (commented):
LGTM, just a few comments around how you've structured the CSS.

PR #859, opened 2020-11-11
feat: Add tooltip to the Table header

jeremy8883:
# Objective
Adds a `tooltipInfo` property to the `Table` component. Using this prop will wrap the cell in a tooltip, and add a blue info icon to the left hand side.

The problem with simply wrapping the heading cell with a `Tooltip`, is that it ruined a lot of the flex styling. To overcome this, I needed to make some minor refactors to `TableHeaderRowCell`.

# Motivation and Context 

On the new Evaluation Cycle dashboards work, we needed to include a tooltip, to provide more context on the table header column name. We've had some confusion from customers in regards to when a request is "from" someone, vs "for" someone. We have very little horizontal realestate on this table, so a Tooltip was our option of last resort.

# Screenshots

image

# Checklist

* [x] If this contains visual changes, has it been reviewed by a designer?  - Approved by Stuart

### DIFF
--- a/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
+++ b/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
@@ -9,14 +9,18 @@ type Props = {
   position?: Position
   text: string
   children?: React.ReactNode
-  className?: string
+  classNameAndIHaveSpokenToDST?: string
 }
 
 const Tooltip = (props: Props) => (
   <span
-    className={classnames(styles.tooltipWrap, props.className, {
-      [styles.inline]: props.inline === true,
-    })}
+    className={classnames(
+      styles.tooltipWrap,
+      props.classNameAndIHaveSpokenToDST,
+      {
+        [styles.inline]: props.inline === true,
+      }
+    )}
   >
     {props.children}
     <span
--- a/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
+++ b/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
@@ -9,11 +9,12 @@ type Props = {
   position?: Position
   text: string
   children?: React.ReactNode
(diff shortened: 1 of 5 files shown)


draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -9,11 +9,12 @@ type Props = {
   position?: Position
   text: string
   children?: React.ReactNode
+  className?: string
 }
 
 const Tooltip = (props: Props) => (
   <span
-    className={classnames(styles.tooltipWrap, {
+    className={classnames(styles.tooltipWrap, props.className, {
```
jeremy8883, line 17: I know this one breaks the rules, but I'm not really sure how else to fix the issue I'm having with the tooltip.

Basically I want the width and height to be 100%, as so I can appropriately align the content inside. See the `.headerRowCellTooltip` class definition above to see what I mean.

jeremy8883: Alternatively, I could pass down 4 new properties to tooltip: flex, width, display, align-items

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -9,11 +9,12 @@ type Props = {
   position?: Position
   text: string
   children?: React.ReactNode
+  className?: string
```
ActuallyACat, line 12: `classNameAndIHaveSpokenToDST`

ActuallyACat: Just to discourage future users of tooltip

ActuallyACat (approved):
LGTM - could you get a designer to give it a once over before merging?

jeremy8883:
Thanks! All approved now by Stuart

PR #860, opened 2020-11-12
chore: Remove unused styles from TextField

### LINKED ISSUE #592: Styles missing definitions
Was poking around `TextField` just now and noticed that the implementation references a bunch of styles that aren’t defined in the importing `.scss` file: https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx#L76-L80

Looks like maybe they were copied over from the `<Input>` implementation or something? They’re rendering some `undefined`s in the output at the moment.

sebpearce:
# Objective

When TextField was created, it was copied over from a previous component along with some styles that it doesn't use. This PR removes those redundant styles. Closes #592.

# Checklist


ActuallyACat (approved):
Noice

PR #864, opened 2020-11-16
feat: Add ability to change the table sorting arrow from descending to ascending

jeremy8883:
* Add ability to change the table sorting arrow from descending to ascending. Before, we only had an `active` prop, which accepted true or false. Now there's a `sorting` prop, with `ascending` | `descending` | undefined.

image

* Side update: add `href` property to the table header. In perform, our sort values are stored in the url. Whenever we use routing, we should (pretty much) always use `a`nchor tags, this allows the user to do things like open the link in a new tab etc.
image

# Checklist

* [ ] If this contains visual changes, has it been reviewed by a designer? 

### DIFF
--- a/draft-packages/table/KaizenDraft/Table/Table.tsx
+++ b/draft-packages/table/KaizenDraft/Table/Table.tsx
@@ -88,6 +88,9 @@ type TableHeaderRowCell = React.FunctionComponent<{
    * @deprecated
    */
   active?: boolean
+  /**
+   * Shows an up or down arrow, to show that the column is sorted.
+   */
   sorting?: "ascending" | "descending"
   wrapping?: "nowrap" | "wrap"
   align?: "start" | "center" | "end"
--- a/draft-packages/table/KaizenDraft/Table/Table.tsx
+++ b/draft-packages/table/KaizenDraft/Table/Table.tsx
@@ -74,7 +74,10 @@ const ratioToPercent = (width?: number) =>
 type TableHeaderRowCell = React.FunctionComponent<{
   labelText: string
   automationId?: string
-  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => any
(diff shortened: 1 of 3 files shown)


jeremy8883:
In case this is being reviewed at the moment, I've snuck in a side update, which adds a new `href` property to the table header cell.

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -73,14 +74,27 @@ const ratioToPercent = (width?: number) =>
...
   flex?: string
   icon?: React.SVGAttributes<SVGSymbolElement>
   checkable?: boolean
   checkedStatus?: CheckedStatus
   onCheck?: (event: React.ChangeEvent<HTMLInputElement>) => any
+  /**
+   * This boolean would show a "sort by" icon in the table cell header.
+   * The problem was that the arrow was pointing in the descending direction only.
+   * Please use `sorting` prop instead.
+   * @deprecated
```
lijuenc, line 91: `@deprecated` is so useful :+1:

draft-packages/table/KaizenDraft/Table/styles.scss
```diff
@@ -77,6 +77,8 @@ $row-height: 60px;
 
 .headerRowCellButton {
   @include button-reset;
+  text-decoration: none;
+  color: $kz-color-wisteria-800;
```
lijuenc, line 81: How about using `.anchorReset` from this same file?

lijuenc (approved):

draft-packages/table/KaizenDraft/Table/styles.scss
jeremy8883: Ah thanks! I didn't realise we had that. Updated now

PR #865, opened 2020-11-16
feat: add automation id to heading

Confidenceman02:
# Objective
There are cucumber tests that rely on finding a heading with a certain `data-automation-id`. These headings were not using the Kaizen component. Adding the automation id to the heading will ensure these cucumber tests don't break.

# Motivation and Context 
Ensures cucumber tests looking for a specific data-automation-id on a heading don't break when updating to kaizen heading 

# Checklist
[x ] I have or will communicate these changes to the front end practice

### DIFF
--- a/packages/component-library/components/Heading/Heading.tsx
+++ b/packages/component-library/components/Heading/Heading.tsx
@@ -41,6 +41,7 @@ export interface HeadingProps {
    */
   classNameAndIHaveSpokenToDST?: string
   children: React.ReactNode
+  automationId?: string
   /**
    * HTML elements that are allowed on Headings. When not supplied, the tag is inferred from
    * the variant. E.g. display-0 will infer h1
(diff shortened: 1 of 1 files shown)


packages/component-library/components/Heading/Heading.tsx
```diff
@@ -59,8 +60,12 @@ export const Heading = ({
   tag,
   variant,
   color = "dark",
+  automationId,
   ...otherProps
 }: HeadingProps) => {
+  const dataAutomationId = automationId
+    ? { "data-automation-id": automationId }
+    : { "data-automation-id": "" }
```
makenosound, line 68: @Confidenceman02 I think you could set this to `undefined` rather than an empty string in the case that the `automationId` isn’t passed — should mean that the attribute doesn’t render at all, rather than rendering with no value.

Confidenceman02: looks like we don't even need this! data-automation-id can be passed without it existing on the props.

makenosound: Ah yeah, of course.

PR #866, opened 2020-11-18
feat: Add autogrow feature to TextArea/TextAreaField

dougmacknz:
# Objective
Adds the ability to specify an `autogrow` prop on our React `TextArea` and `TextAreaField` components, which will grow and shrink the textarea with content as seen below. 

# Motivation and Context
The primary motivation for me adding this is so that I can replace out the free text question textareas in Capture with the Kaizen textarea. The current textareas on Capture have autogrow by way of jQuery.

This also seems to be a fairly common functionality added to text areas, so I would imagine it will have use in other places.

## Considerations
* I haven't added an ability to specify max rows (a height to cap out at), which is a bit more complex to implement. This can always be added later if we want, and I didn't want to invest too much time in this in case we will be rewriting this component in Stencil at some point anyway.

# Screenshots
textarea-autogrow

# Checklist

# Additional Considerations


sebpearce:
Code looks great. No libraries needed 😎 

I noticed that on `master` the textarea isn't resizable, but this one is, which at the moment behaves like this when focused:



Is the resizableness something we have to take on if we use autogrow? (Is it an accessibility problem to disable it?) I've never thought about how we might deal with the focus ring changing on resize 🙃

makenosound:
Perhaps we limit the `resize: vertical` property? Would mean we avoid the awkward display but allow users to open up the space if they want to.

dougmacknz:
> I noticed that on master the textarea isn't resizable, but this one is

@sebpearce That's... weird. I have just checked master via https://cultureamp.design/storybook/?path=/story/textareafield-react--default-kaizen-site-demo on Chrome and FF and they are resizable. I haven't made any changes to that so I would be confused if it were different.

sebpearce:
@dougmacknz Welp 😂  I almost didn't write my comment originally because I thought "if it's resizable on master then this PR didn't affect it at all". I must've absent-mindedly checked the dev branch twice or something 🤦  Sorry for the confusion.

I'll file an issue for it so we don't lose track.

sebpearce (approved):

dougmacknz:
> Perhaps we limit the resize: vertical property? Would mean we avoid the awkward display but allow users to open up the space if they want to.

This seems reasonable. We just have to hope that no one uses a text area with little width, in which case a horizontal resize would useful. But at that point, we shouldn't be relying on resize to make the experience better anyway.

PR #868, opened 2020-11-25
ci: Update docker images to use Node v14

ActuallyACat:
# Objective
Updating Node to v14

# Motivation and Context 
v14.15.1 is the version Node recommends for most users. 

# Checklist

~~[ ] If this contains visual changes, has it been reviewed by a designer?~~ 


.nvmrc
```diff
@@ -1 +1 @@
-v12.16.3
+v14.5.0
```
makenosound, line 1: Should this match the specific node version you’re pinning on the Docker images?

mbylstra (approved):
I'm guessing this is for Stencil :)

package.json
```diff
@@ -7,7 +7,7 @@
     "site"
   ],
   "engines": {
-    "node": "^12.16.3"
+    "node": "^14.5.0"
```
makenosound, line 10: Feels like we should perhaps bump this too (even though it’s not strictly necessary). Having it consistent with the rest of the values will mean it’s better kept in sync.

.nvmrc
ActuallyACat: I _don't think_ a patch will cause us grief 🤔 I will update just to be sure though. Thanks :)

package.json
ActuallyACat: I'm not sure why, but it appears I can't download v14.5.1 from Node (https://nodejs.org/dist/) using asdf :/

makenosound: Oh, `14.5.1` doesn’t exist I think? You want `14.15.1` (which because strings shows up further down the list).

ActuallyACat: 😬 I can't read. 14.1.x =/= 14.15.x

PR #871, opened 2020-11-27
feat: Allow generic textarea props on TextArea and TextAreaField

dougmacknz:
# Objective
Allows consumers of `TextAreaField` and `TextArea` to send in any generic `textarea` props.

# Motivation and Context 
If you need to add a generic prop to our text area component (such as `maxLength` or any `aria` property`) - we need to go through the process of adding it to the component in Kaizen, then update the package on the consuming repo, which is a time consuming process, and causes code bloat. 

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.tsx
@@ -3,7 +3,7 @@ import classnames from "classnames"
 import React, { useState, useEffect, useRef } from "react"
 import styles from "./styles.scss"
 
-type Props = {
+interface Props extends React.HTMLAttributes<HTMLTextAreaElement> {
   id: string
   automationId?: string
   rows?: number
@@ -35,6 +35,10 @@ const TextArea = (props: Props) => {
     onBlur,
     onFocus,
     automationId,
+    value,
+    onChange: propsOnChange,
+    textAreaRef: propsTextAreaRef,
+    ...genericTextAreaProps
   } = props
 
   const [textAreaHeight, setTextAreaHeight] = useState("auto")
@@ -43,7 +47,7 @@ const TextArea = (props: Props) => {
     autogrow ? defaultValue : undefined
   )
   // ^ holds an internal state of the value so that autogrow can still work with uncontrolled textareas
-  const textAreaRef = props.textAreaRef || useRef(null)
+  const textAreaRef = propsTextAreaRef || useRef(null)
 
   useEffect(() => {
     if (!autogrow) return
@@ -63,8 +67,8 @@ const TextArea = (props: Props) => {
         // see https://medium.com/@lucasalgus/creating-a-custom-auto-resize-textarea-component-for-your-react-web-application-6959c0ad68bc#2dee
 
         setInternalValue(event.target.value)
-        if (props.onChange) {
-          props.onChange(event)
+        if (propsOnChange) {
+          propsOnChange(event)
         }
       }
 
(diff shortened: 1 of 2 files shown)


lijuenc:
Just copying my comment from Slack in case people want to discuss it here:

> I like the how it extends the native element’s props:
`interface Props extends React.HTMLAttributes<HTMLTextAreaElement>`
To me it makes sense to allow this for those components which are essentially wrapping some native component and making it look/behave nicer

makenosound:
I think this broadly makes sense, and I’m in favour of us adopting this approach. For ease of reference, here’s the interface we’d be allowing:

```
interface TextareaHTMLAttributes<T> extends HTMLAttributes<T> {
    autoComplete?: string;
    autoFocus?: boolean;
    cols?: number;
    dirName?: string;
    disabled?: boolean;
    form?: string;
    maxLength?: number;
    minLength?: number;
    name?: string;
    placeholder?: string;
    readOnly?: boolean;
    required?: boolean;
    rows?: number;
    value?: string | ReadonlyArray<string> | number;
    wrap?: string;

    onChange?: ChangeEventHandler<T>;
}
```

Only thing I’m wondering if if we want to create a new interface that _omits_ specific attributes that don’t make sense for us. We probably don’t want to allow people to use `col` or `row` to size textareas for example, since we’d prefer to control those attributes with CSS.

*****

Also TIL `<textarea form="id_of_form_somewhere_on_page">` 🙃

dougmacknz:
> Only thing I’m wondering if if we want to create a new interface that omits specific attributes that don’t make sense for us. We probably don’t want to allow people to use col or row to size textareas for example, since we’d prefer to control those attributes with CSS.

This component currently encourages the use of `rows`. I recommend that over controlling height with CSS for `textarea`s to avoid having to factor in line height else potentially end up with a height that ends up cutting off text vertically, especially considering inconsistencies across devices.

`cols` on the other hand I wouldn't use, but I wouldn't go out of my way to block its usage.

ActuallyACat:
I'm mostly cool with this. Speaking generally, the issues I see with this approach in the wider system:
* Naming collisions, or naming duplication. E.g. if we called `disabled` attribute something like `isDisabled` we would then have two ways to do the same thing. 
* If it's in the interface there's an implicit expectation for consumers that attribute would work the same way they would expect it to work on a native element. Is this the case? Do we test all these attributes?

dougmacknz:
> Naming collisions, or naming duplication. E.g. if we called disabled attribute something like isDisabled we would then have two ways to do the same thing.

Yeah, we'll want to be weary of not duplicating native behaviour when using this. The consequences shouldn't be too bad, other than needing to do a big sweep of fixing usages if we find a duplicate that's been heavily used.

> If it's in the interface there's an implicit expectation for consumers that attribute would work the same way they would expect it to work on a native element. Is this the case? Do we test all these attributes?

I would imagine we won't be able to test every possible thing and combination of different attributes, so this approach definitely will open us up to potential issues like "I can send `disabled` to this component but it doesn't look right", and I suppose there is a possibility of some native functionality not working due to a conflict with some functionality we have introduced to the component. I'd say this would be a very rare case though. The only case I see where this happens is if you have heavily adjusted functionality on a native element, in which case this pattern probably shouldn't be used.

Another thing to be weary of, in this instance I have allowed `textarea` attributes to be sent to `TextAreaField` which is a wrapper on `TextArea`. So there is an assumption that the props you send with go through to `TextArea` (rather than be applied to the container, or the label, etc). In this case I'd say it still makes sense, but something to keep in mind.

makenosound:
> Yeah, we'll want to be weary of not duplicating native behaviour when using this.

I think we’d want consistency of API as much as possible, but it seems legit to me to overload the native duplicate in some ways potentially for a behaviour like `disabled`. As long as we’re translating the _spirit_ of the native API through to the native element then I think that sort of overloading is fine.

***

Feels like we perhaps first need to decide conceptually what we consider _our_ `<Textarea>` component to be: is it an lightweight wrapper for the native `<textarea>` or is it a component we want to build tighter control aroun?

lijuenc:
>As long as we’re translating the spirit of the native API through to the native element then I think that sort of overloading is fine

wonder if TypeScript's `extends` (or something else) can make it obvious enough that a property on the native element is being overloaded, that way the person implementing it is at least made aware that they are getting close to native behaviour and can avoid drastic changes.

ActuallyACat (approved):

PR #876, opened 2020-12-01
fix: allow title-block zen to use v2 of design tokens

jasononeil:
# Objective

One of our packages still had a requirement on the `@kaizen/design-tokens` v1 package, even though it was perfectly compatible with v2. This produced a warning when installing dependencies with Yarn or NPM.


ActuallyACat (approved):
LGTM

PR #877, opened 2020-12-01
fix: Extend tsconfig from the distribution config

ActuallyACat:
# Objective
Use the correct tsconfig for the new @kaizen/react-deprecate-warning package.

# Motivation and Context 
The react-deprecate-component package was extending from the old distribution config (tsconfig.json), which includes stories that we'd rather not publish with the package. 

# Checklist

~[ ] I have considered likely risks of these changes and got someone else to QA as appropriate~


okeeffed (approved):
LGTM ✅

PR #882, opened 2020-12-03
feat: Allow for the modal confirmation button to be optional

jeremy8883:
# Objective
Allow for the modal confirmation button to be optional

# Motivation and Context 
I was requested to show a cautionary modal, to let admins know that they can't view their own reviews. The designs only have one button.

image

# Side updates

Export props for the Popover and SplitButton components. Sometimes in perform, we add wrappers around the kaizen components to enhance the functionality. The common use case being linking them to react-router, or automatically assigning the translation strings. Exporting the props in kaizen, then allows for the wrapper component to take in the same props as its child.

# Checklist

[X] If this contains visual changes, has it been reviewed by a designer? 
[X] I have considered likely risks of these changes and got someone else to QA as appropriate
[X] I have or will communicate these changes to the front end practice

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.tsx
@@ -24,7 +24,7 @@ interface Props {
   readonly isOpen: boolean
   readonly type: ModalType
   readonly title: string
-  readonly onConfirm: () => void
+  readonly onConfirm?: () => void
   readonly onDismiss: () => void
   readonly confirmLabel?: string
   readonly dismissLabel?: string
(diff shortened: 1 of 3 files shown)


draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.tsx
```diff
@@ -58,48 +58,53 @@ const ConfirmationModal = ({
...
-          })}
-        >
-          <div className={styles.iconContainer}>
-            <div className={styles.spotIcon}>{getIcon(type)}</div>
+}: Props) => {
+  const footerActions: Array<{ label: string; action: () => void }> = []
+  if (onConfirm) {
+    footerActions.push({ label: confirmLabel, action: onConfirm })
+  }
+  footerActions.push({ label: dismissLabel, action: onDismiss })
```
jeremy8883, line 66: The only update in this block of code is this `footerActions` constant, which then gets passed down to the `ModalFooter` component

sebpearce (approved):

PR #884, opened 2020-12-04
docs: Update story background colour settings

ActuallyACat:
# Objective
Fix our storybook stories that are using incorrect background settings.


sebpearce (approved):
👌

PR #893, opened 2020-12-09
feat: new FilterMenuButton draft component

martindavid:
# Objective
Add new `FilterMenuButton` component

# Motivation and Context 
This is an iteration from the previous [FilterDrawer](https://cultureamp.design/components/filter-drawer/) component but with an emphasise of the new design that will support the new [Global Filter Menu](https://www.figma.com/file/kz2iLEjs3HIOMfF6k5X4l1/Global-Filter-Menu?node-id=54%3A0) component.

This component will support the behaviour of `Simple` and `Advanced` form of the Global filter menu by the `onClearFilter` props. It's the consumer responsibility to determine whether the filter dropdown is in `Simple` mode or `Advanced` mode.

### Notes
There will be a follow up [PR](https://github.com/cultureamp/kaizen-design-system/pull/899) that will remove the existing `FilterDrawer` component from Kaizen.

# Screenshots (if appropriate)
CleanShot 2020-12-10 at 15 03 53@2x

# Checklist

[x] If this contains visual changes, has it been reviewed by a designer? 
[x] I have considered likely risks of these changes and got someone else to QA as appropriate
[x] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline? 
- Have Storybook stories been updated? 

**If you're new to Kaizen, please ask #prod_design_systems to set up an onboarding session to get you up to speed.** If you have an urgent PR to merge before that happens, it is safest to ask Design Systems team to review it to catch any issues.

PR #899, opened 2020-12-14
chore: remove filter drawer component

martindavid:
# Objective
Remove the `FilterDrawer` component from kaizen library

# Motivation and Context 
As we introduce a new component called `FilterMenuButton`, this component becomes irrelevant anymore and need to be removed from Kaizen to not make it confusing with `FilterMenuButton`

# Checklist

# Additional Considerations


sebpearce:
The `FilterMenuButton` (#893) is not actually just a button; it's the whole filter component, right?

Given that the designers refer to this component as the "Filter Drawer" (as far as I know), wouldn't it make more sense to keep this package (`FilterDrawer`), copy the code from the `FilterMenuButton` PR to this one, and delete `FilterMenuButton`? 

That assumes this one isn't currently used of course. If it is used, maybe we need a name like `FilterMenu` to match the name in Figma... I don't know what the designers' intention is with regard to the naming, but it would clear up a bit of future confusion if we could find out.

martindavid:
@sebpearce It's just the button part. It looks similar to the existing `FilterDrawer` (a component that acts as a container for the dropdown component). This is the break down on the whole `FilterMenu` itself https://www.figma.com/file/kz2iLEjs3HIOMfF6k5X4l1/Global-Filter-Menu?node-id=743%3A0 
CleanShot 2020-12-14 at 12 22 55@2x


My intention to create a new component instead of modifying the `FilterDrawer` is to make it clear that the `FilterMenuButton` is a new component and also to learn how to build new draft-component from scratch using the generator.

sebpearce:
@martindavid Ah thanks, I thought I might be misunderstanding something. So to clear this up:

• `FilterMenuButton` is just the button
• `FilterMenuDrawer` is getting deleted
• ____________________ is the component that does the filtering behaviour

☝️  What would go in the space here?

martindavid:
It should be called `FilterMenuDropdown`. But it won't be built in Kaizen for now as we've built it in parallel in perf-ui. The reason for that is because there are some variances of it and we're still not sure what's the best way to construct those variants if we put it in Kaizen
CleanShot 2020-12-14 at 12 40 37@2x

sebpearce:
Ahhh OK, thanks, I see now. I didn't realise there was an empty component spot that would be filled by `FilterMenuDropdown` eventually 👍

sebpearce:
@ActuallyACat Just wanted to CC you in here so you're aware this component is about to be deleted.

(The replacement is Filter Menu, which doesn't fully exist in Kaizen yet, but there is FilterMenuButton in #893.)

sebpearce:
> It should be called `FilterMenuDropdown`.

Side note – I wanted to make a suggestion here for the future that we don't called it `FilterMenuDropdown`, but instead `FilterMenuContent`.

I'm assuming that we're going to use the Kaizen `Menu` component for that, and the language we use there refers to the _container_ (i.e. the bit that Kaizen gives you) as the dropdown, while the actual stuff inside the container that the consumer provides is called the _content_.

image

mbylstra:
> I don't know if this component is being consumed anywhere or not – was there a quick way to find out?

I'm very confident that it's not being used, but we can't know 100% for sure without scanning every codebase in the github murmur org. I think it's worth the risk because in the unlikely event it is being used it's not even so bad:
- team sees a major release that deletes the filter-drawer in @kaizen/draft-filter-drawer
- (hopefully team doesn't just blindly merge the breaking change), sends a WTF message to #prod_design_systems
- we can git revert that release

site/docs/guidelines/header.mdx
```diff
@@ -20,7 +20,7 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
 - Anatomy:
     - **Navigation bar:** Contains product switcher (between Engagement and Performance) or Culture Amp logo for customers with only 1 product.
     - **Title Block:** Contains the page title, breadcrumb, primary action, default action, secondary actions, and secondary navigation.
-    - **Skirt:** Draws the user's eye into the content area. Contains the [Filter Drawer](/components/filter-drawer/), viewed or sorted by functionality ([Select](/components/select/)), and page [Search Field](/components/search-field).
+    - **Skirt:** Draws the user's eye into the content area. Viewed or sorted by functionality ([Select](/components/select/)), and page [Search Field](/components/search-field).
```
sebpearce, line 23:
```suggestion
    - **Skirt:** Draws the user's eye into the content area.
```
The rest of the sentence that begins, "Viewed or sorted..." is referring to the Filter Drawer.

martindavid: @sebpearce I updated this.

sebpearce (approved):

PR #901, opened 2020-12-14
fix: Restore defaultValue on textarea

dougmacknz:
When autogrow was added to this component, I created something called `internalValue` for when the textarea is uncontrolled by the consumer, but autogrow still needs to know the value. React throws an error if you specify both a `value` and `defaultValue` on a textarea, so I took away this defaultValue assignment, which was no longer required at the time with `internalValue` now taking over. I then decided that I wanted to avoid holding the internal state if I could (when autogrow is false), but didn't add this back in. This meant that uncontrolled textareas no longer had their default value going through.


ActuallyACat (approved):
This LGTM - something we'll have to be careful about if/when we move other form inputs to extend from the native interface.

stephwestgarth (approved):
Thanks for fixing this!

PR #907, opened 2020-12-17
fix: Add explicit visibility style rule

sebpearce:
# Objective

Add an explicit `visibility: visible` rule to the `hr` element in the Divider component.

# Motivation and Context 

Murmur, our biggest consuming repo, has [global styles](https://github.com/cultureamp/murmur/blob/master/app/assets/stylesheets/legacy/modules/_body.scss#L95) that include
```
hr {
  visibility: hidden;
}
```
This makes for a poor user experience when someone uses the Divider component in Murmur, because they can't see it and may waste time figuring out why.

Removing this global rule from Murmur is risky, and adding an explicit rule to Kaizen seems like a compromise that is fairly harmless and does not couple the component to Murmur in any way.

### DIFF
--- a/draft-packages/divider/KaizenDraft/Divider/styles.module.scss
+++ b/draft-packages/divider/KaizenDraft/Divider/styles.module.scss
@@ -7,7 +7,7 @@
   margin: 0;
   border-radius: $kz-border-solid-border-radius;
   // This is here to protect against a global style in a consumer.
-  // See app/assets/stylesheets/legacy/modules/_body.scss in Murmur.
+  // https://github.com/cultureamp/murmur/blob/master/app/assets/stylesheets/legacy/modules/_body.scss
   visibility: visible;
 }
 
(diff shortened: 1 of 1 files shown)


makenosound:
Seems fair to me. Worth linking to the offending code in Murmur with a comment though so we know why it’s necessary?

makenosound (approved):

lijuenc (approved):
Looks reasonable to me too. We probably will keep seeing these types of workarounds :)

sentience:
👍 with the comment. I kind of wish we had a convention for `// hack for Murmur` that would enable us to find-replace in future, but this seems well enough documented that we'll know when we can remove it in the future.

sentience:
I know you've just merged, but I might have suggested linking to a specific SHA in the comment, so that if the code is refactored in the future it doesn't break the link.

sebpearce:
Ah good call. I'll do that next time for sure!

PR #910, opened 2020-12-18
feat: Add draft-page-layout package (Content, Container, Skirt, SkirtCard components)

makenosound:
This PR includes implementation of an `@kaizen/draft-page-layout` package that includes four related components.

The first two are true layout-related components:

* `Content` — a max-width wrapper that matches the `TitleBlockZen` component
* `Container` — a wrapper for `Content` to ensure it is positioned correctly

The intention with these we can use them to easily and consistently match the size of our page content to the sizes outlined by the `TitleBlockZen` component, while retaining the flexibility to do more interesting layout things if we need to (like the Skirt).

The package also includes two more specific layout-related components:

* `Skirt` — an implementation of the skirt from the Zen UI kit that works in concert with the `TitleBlockZen` component.
* `SkirtCard` — a wrapper for `<Card>` that implements the most common use case of content alongside the Skirt.

There are a couple of intesting things to note about the `Skirt` and `SkirtCard`

* The design of the `Skirt` dictates some behaviour that is impossible to achieve in CSS only, and so I’ve added that functionality using JavaScript. The behaviour is that as the skirt grows with its content, it needs to be limited in height to a specific value, but it also needs to be able to sit effectively against content that is _smaller_ than that maximum value. `max-height` percentages alone aren’t enough to cover the cases required for that.

  The measurements from this behaviour are implemented using a custom `useResizeObserver` hook that ideally is abstracted into an open-source package (it needs to be open-source as otherwise we’d need registry access for Kaizen itself). @sampnorris has some WIP of that and when it’s ready we can move this across there.
* The `SkirtCard` is _mostly_ identical to `Card` but contains some positional changes at smaller screen sizes (that are part of the Zen UI designs) that are subtle and would be awkward to implement for each consumer.

Demos in Storybook:

* https://dev.cultureamp.design/add-layout-and-skirt/storybook/?path=/story/pagelayout-react--default-story
* https://dev.cultureamp.design/add-layout-and-skirt/storybook/?path=/story/titleblockzen-react--default-with-content


draft-packages/stories/Skirt.stories.tsx
```diff
@@ -0,0 +1,223 @@
+import * as React from "react"
```
ActuallyACat, line 1: These stories are great - it gave me a good idea of anticipated usage.

draft-packages/skirt/README.mdx
```diff
@@ -0,0 +1,39 @@
+---
```
ActuallyACat, line 1: Would be great to grab a designers 2c on this documentation. _Usually_ (and I don't agree that this is the correct thing to do) website documentation = primarily designers audience and storybook = primarily engineer audience

draft-packages/skirt/KaizenDraft/Skirt/SkirtCard.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react"
+import { Card, CardProps } from "@kaizen/draft-card"
+import classNames from "classnames"
+import styles from "./SkirtCard.scss"
+
+export const SkirtCard = (props: CardProps) => {
```
ActuallyACat, line 6: Would love to hear your thoughts on a thin wrapper of card to create SkirtCard. Is it to encourage usage? Reinforce pattern?

ActuallyACat (commented):
I like the overall approach. I would consider whether these should all be separate packages or whether they could be consolidated into something generic like "PageLayout". The Titleblock will soon become its own service, potentially being deprecated from Kaizen, so given there will just be Skirt and ContentBlock it might make sense to keep them coupled together. WDYT?

draft-packages/skirt/KaizenDraft/Skirt/SkirtCard.tsx
makenosound: It’s a little of pattern-reinforcement/easy-usage, yep, since the `SkirtCard` is something that’s tied fairly closely to the `Skirt` in design-terms.

The primary reason though is that there’s some design subtlety for the `SkirtCard` at smaller sizes that’s applied here and if we were leaving it up to consumers to implement we’d need to make sure they implement the same details.

You can see in `SkirtCard.scss` that as the screen collapses in size the card gets negative margins so it touches the full width of the screen, and it has slightly border-radius variations.

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
```diff
@@ -0,0 +1,66 @@
...
+   * Not recommended. A short-circuit for overriding styles in a pinch
+   * @default ""
+   */
+  classNameAndIHaveSpokenToDST?: string
+
+  /**
+   * Not recommended. A short-circuit for overriding styles in a pinch
+   * @default ""
+   */
+  style?: React.CSSProperties
```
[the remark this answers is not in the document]
makenosound: I had forgotten I wrote that and meant to call it out as semi-controversial. My thought here is that there are going to be times when we want to apply specific style adjustments to a core building block in a way that _might_ not be possible with pure CSS — for example some dynamic calculation based on height, etc — and that having that as an out is a useful escape mechanism in those situations.

makenosound:
> I would consider whether these should all be separate packages or whether they could be consolidated into something generic like "PageLayout".

I think that’d be pretty reasonable change (naming is always the hardest). Would you encapsulate that in the name of all components? `PageLayoutContentArea` and `PageLayoutContentBlock` or would you prefer they were more simply named but part of a combined package?

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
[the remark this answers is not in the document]
makenosound: @ActuallyACat — Did you have further feels about this?

makenosound:
@ActuallyACat — Picking this up again. How do you feel about the naming within a consolidated package?

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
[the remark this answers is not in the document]
ActuallyACat: Got it - so `classNameAndIHaveSpokenToDST` = static style overrides, `style` = dynamic style overrides. I think that should be specifically called out in the docs. 

I am wondering if this is too "catch-all" and if it could be a little more specific. What kind of dynamic overrides can we reasonably anticipate? If that's too much of an unknown this approach seems reasonable to me.

ActuallyACat:
> I think that’d be pretty reasonable change (naming is always the hardest). Would you encapsulate that in the name of all components? PageLayoutContentArea and PageLayoutContentBlock or would you prefer they were more simply named but part of a combined package?

I'd prefer simply named since they'll be in the same package, the `PageLayout` prefix would be implicit.

makenosound:
> I'd prefer simply named since they'll be in the same package, the PageLayout prefix would be implicit.

Cool, I was hoping you’d say that. I think this could work then:

```
import { ContentContainer, ContentBlock } from "@kaizen/draft-page-layout";

function Page() {
  return (
    <ContentContainer>
      <ContentBlock>
        <p>I am content</p>
      </ContentBlock>
    </ContentContainer>
  )
}
```

Or if we wanted to be as simple as possible:

```
import { Container, Content } from "@kaizen/draft-page-layout";

function Page() {
  return (
    <Container>
      <Content>
        <p>I am content</p>
      </Content>
    </Container>
  )
}
```

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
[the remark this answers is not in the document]
makenosound: > I am wondering if this is too "catch-all" and if it could be a little more specific. What kind of dynamic overrides can we reasonably anticipate? If that's too much of an unknown this approach seems reasonable to me.

It is a _little_ tricky to anticipate I think. What we’d probably want to limit it to is any positional properties: margin, padding, position, top, left, etc.

ActuallyACat:
No preference here - contributors choice :D

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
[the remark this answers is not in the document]
ActuallyACat: Something like this? 
`Pick<React.CSSProperties, "padding", "margin", "position", "top", "right", "bottom", "left">`

If you think other use cases might arise that are outside of these properties then what you already have is 💯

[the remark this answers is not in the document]
makenosound: The other applications I can think of here are transformations, so if we picked:

* padding (and all variants)
* margin (and all variants)
* position
* top
* right
* bottom
* left
* transform (and all variants)

Then that’d _probably_ account for the kinds of layout changes I can imagine wanting to apply dynamically.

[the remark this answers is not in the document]
makenosound: Have added these variants in now!

makenosound:
> No preference here - contributors choice :D

Heh, I’m running with the simple version then: `Container` and `Content` it is. I think this is probably good to go bar some documentation tweaks. Will write up the PR description a little more clearly and see if @sebpearce can cast an eye across it too.

draft-packages/stories/Skirt.stories.tsx
ActuallyACat: Now that these packages are consolidated, can you combine `Skirt.stories.tsx` and `PageLayout.stories.tsx` ?

draft-packages/content-block/KaizenDraft/ContentBlock/ContentBlock.tsx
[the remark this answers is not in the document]
ActuallyACat: Looking good!

draft-packages/skirt/README.mdx
ActuallyACat: I wonder if we could fold this into the page layout documentation. It would be good to get @dlimiter or @katsambarcus' thoughts here too :)

draft-packages/stories/Skirt.stories.tsx
makenosound: Yeah, actually they’re not _all_ consolidated yet (the skirt is a separate package). Can move those bits into `PageLayout` too and consolidate the stories if you think that’s the right move.

ActuallyACat: My thoughts are that someone who is using this Skirt component would probably not use it in isolation, and instead would use it in conjunction with the page layout components. They seem related to me.

makenosound: Yep, fair. Have done that now.

draft-packages/skirt/README.mdx
makenosound: I realise most of the skirt stuff is covered in the header docs already in terms of design, so I’ve removed this when consolidating the packages together. I reckon it’s worth thinking about whether we should have a place for some code-usage documentation that sits outside the other documentation, or whether they should all be together.

sebpearce (approved):
Very nice work. I had a look at this on Friday and just had another one now – I think @ActuallyACat said everything I would have thought of along the way. Will be great to have this available to use 👌

makenosound:
Suhweet. Thanks for checking it over, @sebpearce!

@ActuallyACat — any final thoughts?

ActuallyACat (approved):
LGTM 🚢

makenosound:
⛵ 🛥️ 🚢  @ActuallyACat — Ta, mind ticking again? Had to rebase with your form merge.

ActuallyACat (approved):

PR #913, opened 2020-12-23
fix: console error, where boolean was passed into `className`

jeremy8883:
In an effort to clean up the console errors in perf-ui, here's one of them. It appears we were passing a boolean into `className`

image


ActuallyACat (approved):
LGTM

PR #915, opened 2021-01-05
feat: Use project references 

ActuallyACat:
# Objective
Use project references. I will be testing these changes in frontend template and some other repos to ensure the publishing works as expected. Tested locally using `npm pack`. 

Related PRs: https://github.com/cultureamp/kaizen-design-system/pull/919


.buildkite/scripts/release.sh
```diff
@@ -44,6 +44,7 @@ release() {
   git checkout master && git pull
 
   yarn install --frozen-lockfile
+  yarn build
```
ActuallyACat, line 47: I'm not sure if `build` is required here, or whether `lerna publish` will take care of all the type generation for us.

mbylstra (approved):
I have to admit I didn't really review it, but I figure you just want to be able to experiment with the Canary build!

PR #919, opened 2021-01-07
ci: Setup for project references 

ActuallyACat:
# Objective

This PR moves all stories to their corresponding package and ensures the stories are not bundled when publishing. 

# Motivation and Context 
This is a task in my housekeeping list and is something (comparatively) small that I could slice out of the project references work: https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/1924630820/Kaizen+Housekeeping

# Checklist


draft-packages/hierarchical-select/docs/hierarchicalStoriesHelpers.ts
```diff
@@ -0,0 +1,134 @@
+/* eslint import/no-extraneous-dependencies: 0 */
```
ActuallyACat, line 1: Technically, since this is a story and not something that is to be distributed, I could have linked to this file that is also present in `draft-packages/hierarchical-menu/docs/hierarchicalStoriesHelpers.ts`. It felt like code smell / something that would encourage bad practice so I have duplicated this file instead.

.gitignore
```diff
@@ -9,3 +9,4 @@ lerna-*.log
 /packages/**/*.js.map
 /packages/**/*.d.ts
 .DS_Store
+dist
```
ActuallyACat, line 12: In the future, all distributed assets will belong in a `dist` directory instead of being compiled in place.

draft-packages/hierarchical-select/docs/hierarchicalStoriesHelpers.ts
lijuenc: Ah ok, so with this new structure, every package's stories should be self-contained?

In this case, `hierarchical-menu` uses `hierarchical-select` internally. So we're saying it's ok to references other packages for implementation code, but not for the stories.

lijuenc (approved):
>moves all stories to their corresponding package

Makes a lot of sense.

draft-packages/hierarchical-select/docs/hierarchicalStoriesHelpers.ts
ActuallyACat: Yeah, pretty much. The stories are not within the distributed package so one would have to import using a relative link.

Issue #911, opened 2020-12-21
TextArea and ToggleSwitch have no unit tests


dougmacknz:
All of our other form primitives have Jest tests except for these two. A recent refactor of TextArea resulted in an incident, which would have been easily caught by unit tests.

ActuallyACat:
@dougmacknz Do you recall the incident that prompted this ticket? From memory, it was something about a defaultValue not being reflected in the component - does that sound correct?

dougmacknz:
This was the fix for it https://github.com/cultureamp/kaizen-design-system/pull/901, basically I broke `defaultValue`

PR #922, opened 2021-01-13
ci: Added test for text area and test stubs for other test cases

aaronknol:
# Objective

Adds tests for the TextArea component

# Motivation and Context 
A recent refactor of the TextArea component resulted in an incident, which test cases would have easily caught. This PR adds tests so this doesn't happen again in the future. This has been raised as an issue 
https://github.com/cultureamp/kaizen-design-system/issues/911


aaronknol:
This isn't ready to be merged yet, more test cases required

PR #933, opened 2021-01-14
feat: Allow greater customisation of action buttons on all modals

dougmacknz:
BREAKING CHANGE: `submitDisabled` prop on `InputEditModal` has been removed. Use `submitButtonProps={{ working: true }}` instead.

# Objective
Allows consumers of all modal types to specify any `ButtonProps` for its action buttons.

eg.
```
<ConfirmationModal
  (...usual modal props)
  confirmButtonProps={{
    label: "Confirm",
    working: true,
    workingLabel: "Working…",
  }}
/>
```

# Motivation and Context 
The reason this has come up is because we have a need to use the `working` prop on the confirm button for `ConfirmationModal`. This is something that has come up in the past with the `InputEditModal`, which caused a `submitDisabled` prop to be added (the `working` variant of `Button` didn't exist at the time).

# Alternate solution / Pros and Cons

There is an alternative solution, where we only expose the functionality we need, ie. we create a new prop on `ConfirmationModal` for `confirmButtonWorking`.

Pros of the approach in this PR:
* We only need to do this once. If there's ever a need for consumers to customise these buttons further, or if the Button API changes, we don't need to then adjust the modal API as well each time.

Cons of the approach in this PR:
* Bit less intuitive for consumers and opens up possibilities for consumers to use modals in some weird ways.
* Creates a bit of awkwardness with the existing `confirmLabel` and `dismissLabel` props. Do we remove them and make those customisations go through this new prop?

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.tsx
@@ -9,6 +9,7 @@ import {
   PositiveFemale,
 } from "@kaizen/draft-illustration"
 
+import { ButtonProps } from "@kaizen/draft-button"
 import {
   GenericModal,
   ModalAccessibleDescription,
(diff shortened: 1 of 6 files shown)


dougmacknz:
Closing in favour of https://github.com/cultureamp/kaizen-design-system/pull/941

PR #967, opened 2021-01-21
feat: Token switching using CSS variables and theme objects

jasononeil:
# List of problems

- [ ] We have compile-time functions for `add-alpha` and `add-shade`/`add-tint`
- [ ] We need a way to generate CSS custom properties from token files
- [ ] We need the sass/less to reference custom properties
- [ ] We need a way to toggle a `brand2021` class
- [ ] Any breaking changes will need fixing in other repos (codemod?)

- For add-alpha, we could do this: Codepen: https://codepen.io/jasononeil/pen/wvovWBP?editors=1100
- For add-tint and add-shade, we could just find the nearest color value. Instead of `add-tint(wisteria, 20)` `wisteria-200`

## Changes

- Define our current tokens as CSS custom properties on `:root`
- Define our rebrand tokens as CSS custom properties on `.brand2021`
- Redefine the existing SASS variables to be `$kz-color-cluny-500: var(--kz-color-cluny-500);` etc
- Create new SASS variables for `$kz-color-cluny-500-default` which has the hex value for the token's default (not the rebrand)

**NOTE: if we want to actually release this, we should be generating a token file like this using the generator rather than writing it by hand**

## Why?

This allows us to use a `.brand2021` class on the body to toggle all tokens for the page, allowing us to more easily QA and release with care.

What about IE11? IE11 users will be stuck on the default token set (so even with the feature flag won't be updated).

## What else

- [x] We'll need to update places where `add-tint` `add-shade` and `add-alpha` are used to now reference the `$kz-{token_name}-default` variables. These mixed colours won't be toggleable using the class.
- [x] We'll need to fix a pinned dependency in title-block-zen

# Checklist

# Additional Considerations


jasononeil:
Abandoning in favour of newer stuff already merged

PR #976, opened 2021-01-26
feat: restructure the SearchBox and move it to draft-form package

martindavid:
# Objective

Rename the `SearchBox` component to `SearchField` and move it to `@kaizen/draft-form` component so it aligns with the design guideline.

# Motivation and Context 

# Checklist

# Additional Considerations


ElliottWanless (approved):

ActuallyACat:
Can you hold of merging this until we have an advocate look over the changes? 🙏

ElliottWanless:
@ActuallyACat we can but currently we are the only ones using this and we don't have one in our team.

ActuallyACat:
The good thing is you don't need one in your team - if you ask in #prod_design_systems someone will be able to assist. Our PR template suggests having a designer review any visual changes, but it's also a good idea to have an engineering advocate review new components too. If you're new to Kaizen we can also run an onboarding session to get you up to speed.

The main concern with not following this process is there's coordination needed with deprecating components and breaking API changes. Just because your team is the only one needing this right now doesn't mean another team won't see this component and begin using it. Having an advocate (in this case, both a design and engineering advocate) review your changes will decrease the liklihood of needing to introduce many small breaking changes.

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -4,10 +4,10 @@ import searchIcon from "@kaizen/component-library/icons/search.icon.svg"
...
-import { FieldMessage, Label } from "@kaizen/draft-form"
-import styles from "./styles.module.scss"
+import { FieldMessage, Label } from ".."
+import styles from "./SearchField.module.scss"
 
-export interface SearchBoxProps {
+export interface SearchFieldProps {
   id: string
   placeholder?: string
   labelText?: string | React.ReactNode
```
sebpearce, line 13: `labelText` is an optional prop here, but should it be? Is there a scenario in which there would be no label?

And if there is a default label as a fallback, shouldn't that be specified on [line 35](https://github.com/cultureamp/kaizen-design-system/pull/976/files#diff-32ba8cb1fb45d7fad8385031e190a9802d23fc9581d5c8da35f2a85cb4bb6f7bR35)?

(I'm not saying there _is_ a default label – that's a design question.)

sebpearce:
This is the current proposed API for this component:

image

I am on the fence as to whether the props that have been taken from Button should be made explicit, e.g. `Pick`ed from Button so that they are actually coupled.

@ActuallyACat Did you have any other thoughts about this API?

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
dougmacknz: The [documentation](https://cultureamp.design/components/search-field/) for this component has some stuff about labels. I would make `labelText` required, and we can create a `labelVisible` prop
image

dougmacknz:
One thing I'd really like to see here is a change from building out most of the component's elements explicitly, to making this a lightweight shim over the `TextField` component. This would remove a whole lot of duplicate code and make this much easier for us to maintain in the future.

You will need to do a little bit of work on `TextField`/`Input` in order to make this happen, but nothing too difficult:
* Create a new `variant` prop for `Input` and `TextField` so that there's a variant with the search field's border radius.
* Create a new `labelHiddenAndIUnderstandTheAccessibilityImplications` prop.

dougmacknz:
An easier way to review the API is to compare it to our `TextField` props, I've just done a comparison:

SearchField specific props:
```
className?: string
working?: boolean
onClearText: (event: React.MouseEvent<HTMLButtonElement>) => void
onPressEnter?: (event: React.KeyboardEvent<HTMLInputElement>) => any
```

Props that exist on TextField but not SearchField:
```
validationMessage?: string
inputType?: InputType
status?: InputStatus
required?: boolean
```

Plus there's the difference in `labelText` being optional (see thread above).

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -48,7 +48,7 @@ export const SearchBox = ({
   className,
```
ActuallyACat, line 48: Use `classNameAndIHaveSpokenToDST` here. The component should accomodate all _known_ scenarios, but we typically add the "escape hatch" of `classNameAndIHaveSpokenToDST` for folks to use in a pinch. We don't want to encourage it's usage without understanding what new scenarios that we haven't considered..

ActuallyACat:
> I am on the fence as to whether the props that have been taken from Button should be made explicit, e.g. Picked from Button so that they are actually coupled.

`Pick`ed from Button or TextField? I really like the direction of SearchField being a thin wrapper over the existing TextField so `pick`ing the props from TextField makes that relationship really explicit. :chef-kiss: Things like states, container height, and other styles are _very_ similar between the two.

sebpearce:
> > I am on the fence as to whether the props that have been taken from Button should be made explicit, e.g. Picked from Button so that they are actually coupled.
> 
> `Pick`ed from Button or TextField? I really like the direction of SearchField being a thin wrapper over the existing TextField so `pick`ing the props from TextField makes that relationship really explicit. :chef-kiss: Things like states, container height, and other styles are _very_ similar between the two.

Ah you're right... I glanced at the Button being rendered and saw `secondary`, `fullWidth` and `label` and assumed those were in the main props 🤦

ashleecharetteCA:
Sorry for the late response to this, I've taken a look at the component in storybook and from a design perspective this looks good to me.

PR #995, opened 2021-02-02
fix: Remove hardcoded color from security-tip icon

lijuenc:
# Objective
Remove hardcoded color from security-tip icon.

# Motivation and Context
This allows the icon to inherit color from the text or content surrounding it. This behaviour is consistent with the other icons in the library and is necessary for components such as https://cultureamp.design/storybook/?path=/story/inlinenotification-react--dismissible-cautionary, which have colored icons.

# Checklist
- [x] If this contains visual changes, has it been reviewed by a designer? N/A - this just makes it consistent with the other icons


ActuallyACat (approved):
This seems like something we'd want to be consistent across all icons in kaizen, not just this icon. Approving because it's low risk (going from #35374a to black if no color can be inherited feels low risk anyway).

PR #996, opened 2021-02-03
feat: Add active state prop to menu items

sebpearce:
# Objective

This adds the ability to set an active state for a MenuItem:

image

# Motivation and Context

Menus don't have active items, because they are just a collection of links or actions.

However, the new Unified Navigation UI that Unify is building uses the Menu component for its dropdowns. These menu items represent pages in the platform that the user can visit, and if the user can see which page they are currently on via an active state, it helps them orient themselves.

I have included some in-code documentation to tell engineers that this is not for standard use and that it is discouraged.

# Checklist


dougmacknz (approved):
LGTM

Issue #1011, opened 2021-02-08
It's not clear how consumers should implement bold text


sebpearce:
- [x] If using a Paragraph component, there are no stories that deal with bold text.
- [x] Implementation of bold text is not explained on the Typography guidelines page on the Kaizen Site.
- [x] There is a style for bold text (`$kz-typography-paragraph-bold-font-weight`) in our typography tokens, but it only exists here and is not part of the Paragraph component.
- [x] The Paragraph component includes no rules for nested selectors like child `<strong>` elements to apply the correct token, so we may not be getting the correct font weights if people use a `<strong>` inside a Paragraph component.

PR #1013, opened 2021-02-08
feat: Add strong selector for bold paragraph text

sebpearce:
# Objective

Adds a rule for nested `strong` elements inside Paragraph components to give them the correct font weight.

# Motivation and Context

Addresses part of #1011.

Using `<strong>` is a standard way of implementing bold text, and is more flexible than creating a bold variant for the Paragraph component.

⚠️  This will potentially cause visual changes if people have already used `<strong>` tags inside a Paragraph component. If they have done that, they will currently be seeing the incorrect font weight (700) instead of the correct one (600) unless they have written a similar nested rule in their own code (it's a no-no for consumers to reach into Kaizen components this way, but possible).

# Checklist


mbylstra (approved):
Interesting one whether this is feat or breaking! My gut feeling was to make it a breaking change (erring more on the side of breaking than feat when in doubt), but thinking it through some more it's really hard to imagine a scenario where this could make things any worse or be contraversional, so it doesn't seem worth "freaking people out" by a breaking change.

(even though my interpretation of "breaking change" is more like "user facing change" rather than breaking in the sense of something being _broken_ or the potential for something breaking)

Long term I'm not sure whether we should _encourage_ people to use html tags directly rather than use a component. My thought process is that it makes it more possible to target completely different platforms (like native mobile) in the future with the same components, but that is very hypothetical and probably not worth overthinking (and we don't even have a way to do strong/bold with components at the moment anyway)

sebpearce:
Cheers 👍 

Re: encouraging tags, it has come up a few times that using tags for both 1) links and 2) strong/emphasis styles allows consumers to have more flexibility in terms of dynamically rendering content inside a component. I guess we can start a thread or issue on this to flesh it out if we need though.

jasononeil (approved):
I don't view this as a breaking change. I'm personally in favour of using `<strong>` as is. Would defer to @allybeaton for a tech lead decision though :)

sebpearce:
Going to merge this as I think it's pretty safe.

Issue #829, opened 2020-10-19
modernise and refactor font fallbacks


mbylstra:
There is currently some inconsistency in our font fallbacks.

`kz-typography-xxxx` mixins  have  `"#{$kz-typography-xxx}, Helvetica, Arial, sans-serif"`, but `ca-type-inter` (although this is a legacy var) doesn't have any.

It feels like we could standardise and have tokens for font fallbacks.`Helvetica, Arial, sans-serif` seems outdated. The Apple system font is now "SF Pro" (I believe), so Apple products will default to Helvetica which could look quite ugly in 2020 depending on how sensitive you are to typography 😆 .  I think the current best practise is to just use `"Font Name", sans-serif`. In the past I think you need to add Arial and Helvetica because some browsers would default to a serif font even if you specified `sans-serif`. 

I think there should be a design token for fallback fonts, but I don't think there just just be _one_. Perhaps it should be heading style dependent. Maybe even it should be baked into every design token... wait... it is! (https://github.com/cultureamp/kaizen-design-system/blob/master/packages/design-tokens/tokens/typography.json#L5). I think that was added later. Still, my point on Helvetica stands!

Maybe we could get fancy and find the closest Apple system font to Greycliff for example (Avenir perhaps?), and then whatever the closest Microsoft font is (?) These are just some quick thoughts -  I think this could do with some research.

mbylstra:
According to this site, http://www.identifont.com/find?similar=greycliff&q=Go, the closet Apple system font to Greycliff is San Francisco Display, which is better than Helvetica, and better again than "SF Pro Text" (the default body font I believe). It could be worth doing this, as I think it would make PDF renders potentially look way better (assuming we still have a PDF service that's unable to load the external fonts)

ActuallyACat:
Just adding historical context / conversations to this issue. 

This has been discussed as the fallback in the internal DST chats and a few other places. 

--------------

"_We need the intended font family (Inter/Greycliff), Noto Sans as our fallback for any non-Latin characters e.g. CJK languages, then any installed or system fonts that would look sensible if the fonts are blocked or don't load for some reason. As a first cut:_"
```
"Greycliff", "Noto Sans", Helvetica, Arial, sans-serif;
"Inter", "Noto Sans", Helvetica, Arial, sans-serif;
```

--------------

Since it's relevant to this discussion, here's the Noto Sans exploration that @katsambarcus looked into:
[...]

PR #1021, opened 2021-02-09
feat: point legacy Open Sans vars at Inter (release canary PR to master)

mbylstra:
# Objective

Release this Canary PR (https://github.com/cultureamp/kaizen-design-system/pull/828) to master (more details are in that PR).

Copy of the commit message regarding the breaking change:

```
BREAKING CHANGE

The breaking change is in only in the most legacy parts of Kaizen (`ca-` scss mixins and variables)
that are only used in the Big M repo.

For all other repos (non Big M), this won't affect them, so this release can be blindy
upgraded without testing. However, if you're unsure you can just do a search
for `ca-` in scss files in your repo, and if there are no results, no need to
worry about the major update

If you are bumping the Big M repo with this release, please PM michael.bylstra@cultureamp.com
for more details.

Description of changes:

- This removes all traces of Open Sans from Kaizen, and
makes Inter the default for $ca-default-font-xxx
- the $ca-default-font-xxx vars are legacy (in both files) and should be
removed eventually, but this is far easier (not a breaking change, codewise)
The nice thing is that the name $ca-default-font-xxx is still applicable, so
there's no need to change the name like we needed to do for $ca-font-ideal-sans!
```

# Motivation and Context
Get rid of Open Sans which is now 3 uplifts old!

# Checklist

### DIFF
--- a/packages/component-library/styles/type.scss
+++ b/packages/component-library/styles/type.scss
@@ -11,9 +11,14 @@ $ca-weight-medium: 500;
 $ca-weight-semibold: 500; // Note: in Sketch, semibold is 600. But murmur has an existing value of semibold=500 that is heavily used.
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
-$ca-default-font-family: "Open Sans", Helvetica, Arial, sans-serif;
-$ca-default-font-base-size: 0.875rem; /* 14px */
-$ca-default-font-descender-height: 0.115;
+$ca-inter-font-family: "Inter";
+$ca-inter-font-base-size: 1rem; /* 16px */
+$ca-inter-font-descender-height: 0.14;
+
+// Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
+$ca-default-font-family: $ca-inter-font-family;
+$ca-default-font-base-size: $ca-inter-font-base-size;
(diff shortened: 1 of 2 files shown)


packages/component-library/styles/type.scss
```diff
@@ -11,30 +11,30 @@ $ca-weight-medium: 500;
 $ca-weight-semibold: 500; // Note: in Sketch, semibold is 600. But murmur has an existing value of semibold=500 that is heavily used.
 
 // Combine these into a Sass map ($ca-default-font) once node-sass includes libsass 3.5.0.beta.3 with this bug fix: https://github.com/sass/libsass/issues/2309
-$ca-default-font-family: "Open Sans", Helvetica, Arial, sans-serif;
-$ca-default-font-base-size: 0.875rem; /* 14px */
-$ca-default-font-descender-height: 0.115;
+$ca-inter-font-family: "Inter", sans-serif;
```
dougmacknz, line 14: Do we definitely not want any fall back fonts (Helvetica, Arial) here? Genuinely asking, I'm not sure whether just `sans-serif` is good enough nowadays?

mbylstra: I'm not sure to be honest! I created this Kaizen issue a little while back to do more research: https://github.com/cultureamp/kaizen-design-system/issues/829. This seemed "good enough" for now.

dougmacknz (approved):

packages/component-library/styles/type.scss
mbylstra: Adding "Noto Sans" does seem important though. I will make this change to and at least keep it the same as the tokens.
 
I realise I can just use on tokens, but I didn't want to have the dependency (and get into potential dependency issues). The real goal is to get rid of the `ca-` stuff entirely.

PR #1025, opened 2021-02-10
feat: Add aria attrs to menu button

sebpearce:
# Objective

Adds `aria-haspopup="true"` and `aria-expanded="true|false"` depending on state.

This change only applies to the "stateful" menu (the one that manages its own state), since for the stateless menu it would have to be managed by the consumer.

# Motivation and Context

Unify needs this to make our new navigation bar accessible, but we realised that all instances of the Menu component should have this by default, since they're menus.

# Checklist


dougmacknz (approved):
Nice hacks

PR #1027, opened 2021-02-10
feat: Add ids for menu items and headers

sebpearce:
# Objective

- Adds the ability to pass an ID to a MenuHeader
- Adds the ability to pass a `headerId` to a MenuItem, which is applied as an `aria-labelledby` attribute

# Motivation and Context

Headers provide a visual clue as to the context of menu items, but without a visual, it isn't as easy for users to figure out which menu item is associated with which header.

image

This gives us a way of handling that.

# Checklist

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/components/MenuHeader.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/components/MenuHeader.tsx
@@ -2,8 +2,8 @@ import * as React from "react"
 
 import styles from "./MenuContent.module.scss"
 
-const MenuHeader = (props: { title: string }) => (
-  <div className={styles.header}>
+const MenuHeader = (props: { title: string; id?: string }) => (
+  <div className={styles.header} id={props.id}>
     <span className={styles.header__title}>{props.title}</span>
   </div>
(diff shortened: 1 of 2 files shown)


dougmacknz:
I don't think `aria-labelledby` is actually what you're after here. What labelledby will do is **replace** the label on the element. Eg.
`<div id="another-element">Some description</div>`
`<button aria-labelledby="another-element">The actual label<button>`

When a screen reader navigates to the button - it will read "Some description". "The actual label" won't be read at all. So in the same way as `aria-label` - you only want to use `aria-labelledby` if the element doesn't have a real label.

What you might be after is `aria-describedby`, which would read the label, then add the contents afterwards. So in the example above, it would read:
"The actual label, Some description"

sebpearce:
@dougmacknz Ahhhh. Yes. That's what I want... 

I think I need some different docs on this :/ Will update, cheers 🙏

sebpearce:
After chatting about this in more detail, I don't think this PR is necessary. Screenreaders should be able to handle the menu situation without issues so this is likely an unnecessary detail. Closing.

PR #1028, opened 2021-02-10
fix: Fix spacing issues on Menu dropdown

sebpearce:
# ⚠️ (POTENTIAL) VISUAL BREAKING CHANGE

# Objective

- Replaces container padding with margin on content elements
- This allows the divider to bleed to the edges (and makes other custom full-bleed sections easier)
- Also uses flexbox on the main container (with `flex-direction: column`) instead of relying on menu items to make themselves 100% width

⚠️  (This message is also in a commit body) This includes a potential visual breaking change, since it restructures the way margins and padding work. If any consumers are using the Menu in non-standard ways, such as creating a section at the top of the dropdown that bleeds all the way to the edges by using a negative margin to override the dropdown container's padding, those could break.

Under the new model, instead of the container using padding to add space, the *content* is responsible for setting its own margins. (This is handled within the component, so consumers don't have to do anything - but as mentioned above, they could run into problems if they had been overriding either the margin or padding.)

# Motivation and Context

A recent design review of the new navigation bar in Team Unify included a request for the dividers on all Menus to bleed to the edges.

# Screenshot
image

# Checklist

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/components/MenuContent.module.scss
+++ b/draft-packages/menu/KaizenDraft/Menu/components/MenuContent.module.scss
@@ -16,7 +16,7 @@ $side-padding: 1/2 * $ca-grid;
   color: $kz-color-wisteria-800;
   border-radius: $kz-border-solid-border-radius;
   box-shadow: $kz-shadow-large-box-shadow;
-  padding: 6px 4px;
+  padding: 6px 0;
   text-align: left;
   width: 100%;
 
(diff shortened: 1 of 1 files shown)


mbylstra:
Regarding whether this a breaking change or not, I tend to veer on the side of it being a breaking change when unsure. As long as there's good commit message explaining the circumstances I don't think it matters if it's a major release. The flipside is that people might put off upgrading because they are scared of the major update and never update, but I don't think they should be! That's my 2c.

lijuenc (approved):
Makes sense to me 👍 . Pretty unlikely for it to be a real breaking change but no harm in being safe and marking it as so like Michael said.

Issue #989, opened 2021-02-01
Rebranding and token switching project


dmisdm:
# Context
We, the Foundations team, are in the process of coordinating a rebrand of all Kaizen components, design tokens, and **primary consumers** of them (Murmur and Performance UI).
Because we would like to gradually roll out changes to design tokens, we cannot simply change the values of them. We must design a solution that allows one to change these values at runtime.
After lengthy discussions and prior work, [CSS Variables](https://www.w3schools.com/css/css3_variables.asp) have been proposed as the mechanism that can most cleanly and maintainably support this requirement: runtime theme switching.

The changes to the design system are outlined in [this google doc](https://docs.google.com/document/d/1G5iN3zZ0XHc0K5FKIFpa263nGW3l8d-jV2nWf6Gcucs/edit#heading=h.9j0pf0mabvt3) and should be the guiding light for what changes need to be made to, not just token values, but also the keys/names of tokens (e.g. Wisteria -> Purple).

## Rebrand process outline

Another document has been formulated for this project: https://cultureamp.atlassian.net/wiki/spaces/CPlatform/pages/2182218300/Migrating+CultureAmp+s+repositories+to+CSS+variable+Kaizen+design-tokens.

Outlined below is a rough execution plan.

- [x] Add new sibling CSS variable tokens and implement a method of switching them at runtime (ThemeManager). #1030 

- [x] Attempt to replace usages of `add-tint` and `add-shade` with tokens. And `add-alpha` with `rgba`
- [x] Add Tiempos font to heart theme; ensure it can be downloaded by consumers.
- [x] React wrapper that allows components to subscribe to a theme context, and use the theme using a hook.
- [x] A guide for how to use the new design tokens and themes; Recommended approach to theming e.g. import `default-theme.css` at the top level of your app if you need it (e.g. for haml), or just use the ThemeProvider react wrapper and useTheme hook.
- [x] Migrate kaizen components to use the new tokens
- [x] Migrate the rest of kaizen design systems styles to use new tokens and remove the old
- [x] Migrate [rebrand-token-remapper](https://github.com/cultureamp/node-packages/tree/master/packages/rebrand-token-remapper) to use new design tokens, and it's method (ThemeManager) for theme switching.
- [x] Upgrade to the new kaizen components that support CSS variable tokens in primary consumers (murmur and perf-ui)
- [x] Solution/program for automatically migrating old tokens to new, as much as possible.
[...]

PR #1030, opened 2021-02-11
feat: New sibling CSS variable tokens, heart theme, and theme switching

dmisdm:
# Objective
This PR is the first in the series of the token switching and rebrand project: https://github.com/cultureamp/kaizen-design-system/issues/989.
It aims to release a new version of design tokens that contains:
- New tokens that hold CSS variable functions, which reference design tokens (variables expected to be defined by the consumer, through importing or configuring a theme).
- A ThemeManager, responsible for containing the source of truth of the current theme, and allowing consumers to change it. The current theme can also be subscribed to.
- A React wrapper around ThemeManager, for accessing theme in a React-idiomatic way using context (`ThemeProvider`) and hooks (`useTheme`)

# Motivation and Context
For more context as to why we are making these changes you can read the overarching project issue https://github.com/cultureamp/kaizen-design-system/issues/989

# Checklist

# Additional Considerations


packages/design-tokens/bin/buildCssVarTokens.ts
```diff
@@ -0,0 +1,205 @@
...
+}`,
+    { parser: "css" }
+  )
+
+/**
+ * WIP: Need a better name and articulation of this.
+ * See {@link augmentCssVariable} to understand what happens to each leaf variable in the theme.
+ * Writing this as a solution to the add-alpha and add-tint/shade problem, and to spit out sass variables with `-default` and `-rgb-params` suffixes (where applicable).
+ * We need to have additional tokens that reference variables, which contain a triple (R, G, B).
+ * This triple can then be used within the CSS [runtime] function as a CSS variable, e.g. `rgba(var(--kz-color-wisteria-800-rgb))`.
```
ActuallyACat, line 47: tuple?

packages/design-tokens/src/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export const makeCSSVariableTheme = (theme: Theme) =>
+  mapLeafsOfObject({ [cssVariableThemeNamespace]: theme }, (path, value) =>
+    path[path.length - 1].endsWith("-default")
+      ? value
+      : objectPathToCssVarReference(path)
+  )
+
+/**
+ * Use this to generate an object containing `${key}: value`, `${key}-default: value`, and `${key}-rgb-params: rgb(r, g, b)` if the value is a color.
```
ActuallyACat, line 94: I wonder if the presence of "default" suffix might confuse people. This will break lots of things, but maybe the trade-off is worth it?

packages/design-tokens/tokens/border-vars.json
```diff
@@ -0,0 +1,44 @@
+{
```
ActuallyACat, line 1: Now that we're moving the source of truth from json to ts, we could probably say goodbye to these tokens

packages/design-tokens/bin/buildCssVarTokens.ts
dmisdm: Uhummm actually
image

dmisdm: Yeah, i'm smart

ActuallyACat: Hahaha, I stand corrected.

ActuallyACat:
Looks like you broke storybook: https://dev.cultureamp.design/danno/css-variable-tokens-and-theming/storybook/

packages/design-tokens/package.json
```diff
@@ -17,14 +17,30 @@
...
-    "prepublish": "yarn build:less && yarn build:sass"
+    "clean": "rimraf 'dist/**' './(!node_modules)/**/*.js' 'tokens/**/*.d.ts' 'tokens/**/*.map' ",
+    "prepublish": "yarn build:json && yarn build:less && yarn build:sass"
   },
   "devDependencies": {
-    "json-to-flat-sass": "^1.0.0"
+    "json-to-flat-sass": "^1.0.0",
+    "yargs": "16.2.0",
+    "@types/yargs": "16.0.0",
+    "prettier": "2.2.1",
```
ActuallyACat, line 34: Do you need all of these dependencies? I would expect prettier/ts-node to use the workspace root packages.

packages/design-tokens/bin/buildCssVarTokens.ts
```diff
@@ -0,0 +1,206 @@
...
+      JSON.stringify({
+        [cssVariableThemeNamespace]: {
+          typography:
+            customPropertiesTheme[cssVariableThemeNamespace].typography,
+        },
+      })
+    )
+  )
+
+  /* Write CSS variable theme files */
```
ActuallyACat, line 191: This is great - I like that you're writing out plain ol' css files.

packages/design-tokens/package.json
dmisdm: @ActuallyACat   i added them because eslint was complaining about extraneous dependencies...should i ignore that?

ActuallyACat: Oh, I see - my mistake. For packages that are _only_ used in `package.scripts` I just let those default to whatever is version is defined [here](https://github.com/cultureamp/kaizen-design-system/blob/master/package.json). You're not just using prettier in `package.scripts`, you're also using it [here](https://github.com/cultureamp/kaizen-design-system/pull/1030/files#diff-053c9716520eec2ec0f883918db9c2acab7d818654cc4209189f0c9b457253d1R3). So what you have here is valid.

packages/design-tokens/src/types.ts
```diff
@@ -0,0 +1,224 @@
+import type { CSSProperties as ReactCSSProperties } from "react"
```
ActuallyACat, line 1: This is really elegant :chef-kiss:

dmisdm:
@ActuallyACat couple of extra fixes, along with this:
https://dev.cultureamp.design/danno/css-variable-tokens-and-theming/storybook/?path=/story/design-tokens-overview--color-demos

ActuallyACat (approved):
This looks good to me.

ActuallyACat (changes requested):
Actually.... I forgot about variable fallbacks. If someone funky happens on the consumer's end we should fallback to hex values. Could you add something like `$kz-color-wisteria-100: var(--kz-color-wisteria-100, #F0F1F4)`

ActuallyACat:
> Actually.... I forgot about variable fallbacks. If someone funky happens on the consumer's end we should fallback to hex values. Could you add something like `$kz-color-wisteria-100: var(--kz-color-wisteria-100, #F0F1F4)`

Ignore the suggestion, but I'd like to discuss this further before merging.

ActuallyACat (approved):

PR #1032, opened 2021-02-11
fix: implicit any amendments

dmisdm:
## Objective, Motivation, and Context.
This PR improves the type safety of the codebase by fixing implicit any instances.
Set "noImplicitAny" to true, observe the errors, fix each one by one.
This commit got from 105-ish -> 73 errors.

"noImplicitAny" is set back to false, as this should be a gradual process. Otherwise the PR would be massive.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
+++ b/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
@@ -47,7 +47,7 @@ class NavigationBar extends React.Component<Props, State> {
     onNavigationChange: () => null,
   }
 
-  constructor(props) {
+  constructor(props: Props) {
     super(props)
     this.state = { mobileKey: 1 }
   }
@@ -119,8 +119,8 @@ class NavigationBar extends React.Component<Props, State> {
 
     return (
       <nav className={styles.links}>
-        {Object.keys(children).map(key =>
-          this.renderNavSection(key, children[key])
+        {(Object.keys(children) as Array<keyof typeof children>).map(key =>
+          this.renderNavSection(key, children[key] || [])
         )}
       </nav>
     )
@@ -141,7 +141,7 @@ class NavigationBar extends React.Component<Props, State> {
     ) : null
   }
 
-  renderNavItem(link: NavigationItem, section) {
+  renderNavItem(link: NavigationItem, section: string) {
     const { props: linkProps } = link
     const isFinal = section === "final"
     const linkWithProps = {
--- a/legacy-packages/title-block/KaizenDraft/TitleBlock/TitleBlock.tsx
+++ b/legacy-packages/title-block/KaizenDraft/TitleBlock/TitleBlock.tsx
@@ -95,7 +95,7 @@ class TitleBlock extends React.Component<Props, State> {
     const { surveyStatus } = this.props
     if (surveyStatus == undefined) return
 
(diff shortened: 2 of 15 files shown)


packages/component-library/components/NavigationBar/NavigationBar.tsx
```diff
@@ -69,8 +69,8 @@ class NavigationBar extends React.Component<Props, unknown> {
 
     return (
       <nav className={styles.links}>
-        {Object.keys(children).map(key =>
-          this.renderNavSection(key, children[key])
+        {(Object.keys(children) as Array<keyof typeof children>).map(key =>
+          this.renderNavSection(key, children[key] || [])
```
dmisdm, line 73: This is a logical and not purely type base change because there is an actual issue. All children props could be undefined, but the renderNavSection expects it to be defined.

draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
```diff
@@ -119,8 +119,8 @@ class NavigationBar extends React.Component<Props, State> {
 
     return (
       <nav className={styles.links}>
-        {Object.keys(children).map(key =>
-          this.renderNavSection(key, children[key])
+        {(Object.keys(children) as Array<keyof typeof children>).map(key =>
+          this.renderNavSection(key, children[key] || [])
```
dmisdm, line 123: See comment below for the same change

legacy-packages/title-block/KaizenDraft/TitleBlock/TitleBlock.tsx
```diff
@@ -95,7 +95,7 @@ class TitleBlock extends React.Component<Props, State> {
     const { surveyStatus } = this.props
     if (surveyStatus == undefined) return
 
-    let variant
+    let variant: React.ComponentPropsWithoutRef<typeof Tag>["variant"]
```
lijuenc, line 98: Interesting. So it's possible to reach into some other component's props typing without explicitly exporting it from that component?

packages/component-library/package.json
```diff
@@ -35,7 +35,7 @@
     "@kaizen/hosted-assets": "^1.1.0",
     "@types/classnames": "^2.2.6",
     "@types/lodash": "^4.14.132",
-    "@types/uuid": "^8.3.0",
+    "@types/uuid": "^3.3.2",
```
lijuenc, line 38: Is it intentional to go back major versions?

lijuenc (approved):
Always nice to have more accurate typing. Did have some questions though.

packages/component-library/package.json
jasononeil: It seems this now matches the actual library version we're using (the types package had run ahead of the actual package)

legacy-packages/title-block/KaizenDraft/TitleBlock/TitleBlock.tsx
dmisdm: Yeah i prefer to the alternative, I think this is more descriptive as to what the type means, i.e. we don't care what it is, just that it is the same as the one the component accepts.

packages/component-library/package.json
dmisdm: Yeah exactly I think it's meant to be consistent. I had to do this because there was a type error when importing from it.

PR #1038, opened 2021-02-14
fix: upgrade to Elm 0.19.1

mbylstra:
This PR upgrades from 0.19.0 to 0.19.1. The main change here is that the Elm compiler is much stricter with how the elm module names like `module ElmStories.TagStories exposing (main)` match the directory structure. We can no longer use the existing `docs/Thing.stories.elm` structure, so each package has a new directory named `ElmStories` at the root of the package (it can't be put inside `docs` any more). Also this uses an upgraded version of `cultureamp/elm-storybook` that's compatible with Elm 0.19.1 (and requires a bit more boilerplate to use)

Here is the Chromatic build: https://www.chromatic.com/build?appId=5e12c297b9d9020020789ca3&number=234. I'm pretty sure the changes there are just the baseline snapshots are a little behind master.


draft-packages/illustration/package.json
```diff
@@ -35,7 +35,7 @@
     "@kaizen/hosted-assets": "^1.1.0"
   },
   "devDependencies": {
-    "@cultureamp/elm-storybook": "cultureamp/elm-storybook#0.1.0",
+    "elm-storybook": "cultureamp/elm-storybook#0.1.0",
```
jasononeil, line 38: Should this be 0.3.0?

It seems this was the only package still on 0.1.0 so may have been missed in a find/replace

jasononeil: It wasn't actually used in this package (no Elm version, I'm guessing it just came from the boilerplate). I'm going to bump it for consistency.

jasononeil (approved):
How I am reviewing:

- [x] Checked changes to package.json
- [x] Checked all other changes are to Elm storybook files
- [x] Visually checked Elm storybook
- [ ] Get Chromatic to check differences

I'm still learning more about how to do that last step properly, but I'm confident this is safe to merge.

jasononeil (dismissed):
Reapprove after rebase

.tool-versions
```diff
@@ -1,3 +1,3 @@
 nodejs 14.15.1
 yarn 1.17.2
-elm 0.19.0
+elm 0.19.0
```
mbylstra, line 3:
```suggestion
elm 0.19.1
```

jasononeil (approved):

PR #1041, opened 2021-02-15
feat: Tooltip positioning improvements + allow it to work inside an `overflow: hidden` element

jeremy8883:
# Objective
Integrate the `popper` library into our `Tooltip` component. In doing so, it has the following benefits:
* The tooltips no longer get cropped, if found inside a `overflow: hidden` or `overflow: auto` element.
* If the tooltip is too close to the edge of the screen, it will automatically change its positioning to best fit.

Additional stories have been added to storybook.

The tooltip will now show upon focus, or if any of its children are focused.

There is also now support for setting the `display` property via props. Before, we only had the `inline` property, but the boolean value didn't allow us to pass in `flex`, `inline-flex`, or `inline-block`.

# Motivation and Context
The main motivation for us was that we have a table in perform. That table has horizontal scrolling, as well as tooltips. The `overflow: auto` was clipping the tooltips.

# Screenshots

image

image

# Known issues
* There used to a be a subtle positioning animation for the old tooltip. I was unable to get this working correctly. It's difficult to explain, but the cursor would flash between `default` and `pointer` during the animation, and sometimes the `isHover` state wouldn't get updated appropriately. It's strange, because I have set `pointer-events: none`.

Before:
Feb-15-2021 13-04-54

After:
Feb-15-2021 12-58-53

* In the "inline" story, for some reason, there is a 1px movement inside the text when hovering in and out. Notice what happens next to the period, after the hyperlink.

https://user-images.githubusercontent.com/4495057/107897916-8279ed80-6f8e-11eb-82ce-d63e4409a3d4.mp4

* I can't seem to get the width correct. Ideally we want the width of the box to hug the width of the text content, until it hits a max width of 400px. This sometimes works, but sometimes does not. I added a `min-width: 200px`, so we don't end up with tooltips that are too tall and narrow.

Without the 200px min-width:

image

After the 200px min-width:

image

* I did not apply this update to the elm version of the component. My assumption is that since we are no longer properly supporting it, then it wouldn't be worth the effort. But let me know otherwise.

* The `inline` property of the `Tooltip` component is now deprecated, please use `display="inline"` instead. If you're pressed for time, the old property will still work for now.
* `popper` has been integrated into the `Tooltip` component. Although I've tried to avoid adding breaking changes, I can't provide guarantees. Please test your tooltips after upgrading this package.

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer? - NA
- [X] I have considered likely risks of these changes and got someone else to QA as appropriate


draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
```diff
@@ -18,124 +22,89 @@
...
-.contentWrap.below {
-  bottom: 0;
-}
-
-.tooltipContent {
+  max-width: 400px;
+  // I'm having a problem getting the tooltip width to match the size of the text content.
+  // As a workaround, I'm just going to set a min width of 200px. It should be enough
+  // in most cases.
+  min-width: 200px;
```
ActuallyACat, line 29: This feels like a reasonable compromise to me.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -1,64 +1,154 @@
...
+  display?: "block" | "inline" | "inline-block" | "flex" | "inline-flex"
+  /**
+   * This is more a "desired position". The tooltip will automatically change
+   * its position, if there's not enough room to show it in the one specified.
+   */
   position?: Position
   text: React.ReactNode
   children?: React.ReactNode
   classNameAndIHaveSpokenToDST?: string
+  portalSelector?: string
```
ActuallyACat, line 31: Is `portalSelector` used anywhere?

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -1,64 +1,154 @@
...
 import styles from "./Tooltip.scss"
+import AppearanceAnim from "./AppearanceAnim"
+import { useUuid } from "./useUuid"
 
 type Position = "above" | "below"
 
 export type TooltipProps = {
+  /**
+   * Use `display="inline"` instead
+   * @deprecated
```
ActuallyACat, line 13: Nice 💯

draft-packages/tooltip/KaizenDraft/Tooltip/TooltipElm.scss
```diff
@@ -0,0 +1,147 @@
...
+@import "~@kaizen/design-tokens/sass/shadow";
+@import "~@kaizen/component-library/styles/type";
+@import "~@kaizen/component-library/styles/animation";
+@import "~@kaizen/deprecated-component-library-helpers/styles/type";
+
+/**
+ * The elm tooltip uses the legacy css code. Since we are soon to be dropping support
+ * for elm, I didn't think it was worth me spending the time to update the elm component
+ * to use popper.
+ */
```
ActuallyACat, line 12: Agreed here.

ActuallyACat (approved):
LGTM - just a question about `portalSelector`

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
jeremy8883: Sorry, no it isn't. I forgot to remove this

Issue #971, opened 2021-01-22
Modal content should be centered


jasononeil:
Our modal designs have centered content in the modal body, but the component does not do this.

Amusingly, all of the storybook stories added a `<div style={{align: center}}>` to the modal body rather than fixing the component :joy:

sebpearce:
We can make this change for some variants, like the Confirmation Modal, but for others, the body could contain things that would almost certainly break if we added `text-align: center`:

![2021-02-15 16 13 12](https://user-images.githubusercontent.com/6406263/107907914-f7f2b780-6fa8-11eb-90a6-0333549101ef.gif)

In the Input-Edit Modal above, the form elements are passed in as children so the component is agnostic to what goes in its body.

We could
1. still proceed with this and make it a breaking change, or
- try to avoid breakages by adding our own nested selectors for things like `label` and fixing the alignment.

mbylstra:
I think option 3 is the best (for variants other than Confirmation/Roadblock ). Even if most of the elements will want to be centered, this should be opt-in, not opt-out (which the consumer would have to do for input labels). I don't like the idea of having to opt out of magically inherited stuff! (I can imagine a future where we have very sophisticated constraints at a code level on how things are centered but I think we're a long way off from that, and I'm not a fan of doing this with CSS selectors which is a just a form of inheritance/magic to me)

If people are needing to wrap text in an inline style like `<div style={text-align: center}>...</div>`, that is pretty gross. I wonder if there's a way we can make centering simpler and more consistent? Either a `centered` prop to our Paragraph component ) or a tailwind style CSS class? 🤔

sebpearce:
Yeah, I'm leaning towards your view too. If the body was a consistent type of thing rather than a misc bucket, I'd feel better about it being opt-out with centering.

The prop is a good idea! It has the downside of not automatically fixing anyone's centering, but the upside of being a cleaner way to opt in.

jasononeil:
My thoughts:

- I think the `Confirmation` variant is pretty safe to center, I'd be up for making that change. I might explore just _how_ breaking it would be (how many places are using anything other than a paragraph)
- The `Information` variant is left aligned in UI Kit, and has the image off to the side, so that should stay as is.
- The `InputEdit` variant has the problem you mentioned. I'd be open to either suggest (2) (Fix the labels in Kaizen to be left aligned) or (3) Let consumers worry about it.

I created the issue because of this:

[...]

PR #1048, opened 2021-02-15
fix: Center text on Confirmation/ Roadblock Modals

sebpearce:
# Objective

This centres text on the body of the Confirmation and Roadblock Modals.

# Motivation and Context

Other modals such as the Input-Edit Modal often have form elements in them, and if we use `text-align: center` on the body of those, it will introduce visual bugs by centering form labels. We may want to leave these cases up to the consumer to handle any text centering (see #971 for more details).

# Checklist

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.scss
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/ConfirmationModal.scss
@@ -46,6 +46,7 @@
 
 .body {
   padding: $kz-spacing-md $kz-spacing-xxxxl $kz-spacing-sm;
+  text-align: center;
 }
 
 .iconContainer {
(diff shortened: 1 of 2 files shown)


mbylstra:
Even though Roadblock and Confirmation modals are only meant to have text in the body, the type of children is `React.Node` so we can't know for sure that people aren't putting whatever they like inside (tables, images, forms, who knows!).

So, I think this should be a breaking change. Even if the odds are super low that anything would "break", I still think it's good to let upgraders know that they should ideally look at the code and remove any redundant `text-align: center` css that they must have added (it will be messy having the duplicates, and potentially harder to change later if we _didn't_ want text to be centered for some reason). This is my opinion on what makes a "breaking changes" - it's the question of whether the upgrader can blindly upgrade or not. We can leave it in their court to decide if they want to take the risk or if they could be bothered cleaning up the duplicate css.

sebpearce:
Agree. I mentioned that in [the issue](https://github.com/cultureamp/kaizen-design-system/issues/971) by the way but didn't add the breaking change text to this PR yet, FWIW.

PR #1052, opened 2021-02-15
chore: Allow running storybook for one package only.

jasononeil:
# Objective

Give the option of starting storybook for only one package at a time. This results in significantly better start-up times, and slightly better rebuild times.

This reduces startup time from 1m18s, to 0m27s for the "modal" package as an example.

# Checklist

# Additional Considerations
- [x] Have you updated any relevant documentation or left useful comments in the code?


ActuallyACat (approved):
>30s time savings is great! LGTM

PR #1058, opened 2021-02-16
feat: Add portal support to Tooltip

jeremy8883:
# Objective
Add react portal support to Tooltip.

# Motivation and Context
When I first tested the popper library with a component with `overflow: auto`, in storybook, everything seemed to work without portals. I was confused by this. I didn't know it was possible. It was magic! When I tested this in perform, the tooltip was getting clipped. So it looks like we need to use react portals after all.

# Screenshots

image

# Checklist


draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -95,8 +97,9 @@ const Tooltip = ({
   text,
   inline,
   display = "block",
-  position,
+  position = "above",
```
jeremy8883, line 100: We were already defaulting to `above`, but I thought I'd make this more explicit.

draft-packages/tooltip/docs/Tooltip.stories.tsx
```diff
@@ -125,6 +125,9 @@ export const OverflowScroll = () => (
           position="below"
           display="inline-block"
           text="This should not get cropped"
+          // Normally, you'd specify a div by ID, but since this is only in storybook,
+          // using `body` is fine (I think).
+          portalSelector="body"
```
jeremy8883, line 130: I wasn't sure how to insert a `<div id="tooltips" />` into the dom, so I just used `body`

ActuallyACat (approved):
lgtm

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -126,6 +126,10 @@ const Tooltip = ({
     </AppearanceAnim>
   )
 
+  const portalSelectorElement: Element | null = portalSelector
+    ? document.querySelector(portalSelector)
+    : null
```
jeremy8883, line 131: Thanks for fixing this. Not a big deal, but I would probably check if this `querySelector` fails, to show a console warning as a result. Otherwise if we fail silently, a developer could be confused as to why their tooltip isn't showing in the correct location in the DOM tree.

PR #1065, opened 2021-02-17
fix: Remove CSS properties when theme is Zen

dmisdm:
## Context
We have the need to use different tokens depending on which theme is active. The most frictionless way we identified to do this was to allow for usages like: `var(--kz-color-stone-500, $kz-color-wisteria-800)`, which says: If there exists a CSS variable --kz-color-stone-500, use it, otherwise use wisteria 800. In effect, if we don't set any CSS variables when Zen is active, this means we will see stone 500 on Heart theme, and wisteria 800 on Zen theme.

Great Success.

The nice thing about this is that there is effectively no change to theme switching ability. Because we already expose CSS variables with defaults in SASS, like: `$kz-var-color-wisteria-800: var(--kz-var-color-wisteria-800, #35374a);` where the default hex value refers to wisteria-800 on Zen theme, the non-existence of CSS variables is the same as activating the Zen theme.
There is of course the edge case of consumers using CSS variables without using our SASS variables, and without defaults, but this is extremely unlikely as the release of CSS variable tokens has not been out for long.

## Changes

- In ThemeManager > Add an if statement to detect when theme is Zen, and remove rather than add CSS properties
- Expose variable identifiers, `$kz-var-id-*`, in json/sass/less, so that consumers have a safer way to either declare their own `var()` usages in SASS, e.g: `color: var($kz-var-id-color-stone-500, $kz-color-wisteria-800)`, or perhaps access variables in JS e.g. `style.getPropertyValue(identifiers.themeKey)`.

# Checklist

# Additional Considerations


ActuallyACat (approved):

ActuallyACat (approved):
LGTM - we'll loop back to the docs after customer research

PR #1074, opened 2021-02-18
feat: align breakpoints to mobile design token to make rollout of new title block and navigation easier

sampnorris:
# Objective, Motivation and Context

One legacy, and one soon-to-be legacy component, have different breakpoints for mobile. Specifically the Legacy Title Block and the Current Navigation.

We will be rolling out a new navigation with a mobile breakpoint defined design tokens. This will align the legacy components with their replacements. 

The problem exists because we will be rolling the new navigation out behind a feature flag, but we won't be upgrading the legacy title block as there is no new version in Elm. The title block needs to make space for the hamburger menu. And frankly it will just be 100x easier to work with.

This issue actually persists before this change so I guess I'm fixing a bug? IDK

Screen Shot 2021-02-18 at 12 53 25 pm


sebpearce (approved):
👌  Love it

PR #1110, opened 2021-02-24
feat: Add automatic positioning for the Menu component, using popper

jeremy8883:
Add automatic positioning for the Menu component, using popper. This is similar to what was recently implemented for `Tooltip` ([PR 1](https://github.com/cultureamp/kaizen-design-system/pull/1041), [PR 2](https://github.com/cultureamp/kaizen-design-system/pull/1058)). It's basically so we can display a dropdown menu, even when inside a scrollable element. It will also automatically detect when a menu will go outside the viewport, and find the next best position for it. - 196973b, 6c9b00b

image

image

* `popper` has been integrated into the `Menu` components. Although I've tried to avoid adding breaking changes, I can't provide guarantees. Please test your dropdown menus after upgrading this package.

# Side tasks
* Convert `MenuItem` to a functional component - 1d58ba7
* Move code from the `Menu.render` function, into `StatelessMenu`. This was because I can't use hooks inside a regular function (well...you can, but the linter will complain). The other benefit is that it removes some of the duplicated code. - 229a711
* Fix to the tooltips, so they don't display flush up against the viewport - 196973b

Before:
image

After:
image

* Add a warning for tooltips if the portal selector does not find anything. This is to help developers, just in case there's a typo. Before it was failing silently. - ba5a136

image

# Checklist

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/Menu.tsx
@@ -1,7 +1,7 @@
 import { ButtonProps } from "@kaizen/draft-button"
-import React, { useRef, useState } from "react"
+import ReactDOM from "react-dom"
+import React, { useEffect, useRef, useState } from "react"
 import MenuDropdown from "./MenuDropdown"
-
 import styles from "./styles.scss"
 
 export type GenericMenuProps = {
@@ -38,6 +38,7 @@ export type GenericMenuProps = {
    * The content to appear inside the dropdown when it is open
    */
   children: React.ReactNode
+  portalSelector?: string
 }
 
 type ButtonPropsWithOptionalAria = ButtonProps & {
@@ -53,16 +54,17 @@ export type MenuProps = GenericMenuProps & StatefulMenuProps
 
 type Menu = React.FunctionComponent<MenuProps>
 
-const Menu: Menu = props => {
-  const {
-    align = "left",
-    dropdownWidth = "default",
-    autoHide = "on",
-    menuVisible = false,
-  } = props
-
-  const dropdownButtonContainer: React.RefObject<HTMLDivElement> = useRef(null)
-
+const Menu: Menu = ({
+  align = "left",
+  dropdownWidth = "default",
+  autoHide = "on",
+  automationId,
+  dropdownId,
+  children,
+  button,
+  portalSelector,
+  menuVisible = false,
+}) => {
   const [isMenuVisible, setIsMenuVisible] = useState<boolean>(menuVisible)
 
   const toggleMenuDropdown = () => {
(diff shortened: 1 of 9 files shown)


draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
```diff
@@ -1,31 +1,111 @@
...
+  /**
+   * Determines when the menu should automatically hide.
+   * @default: "on"
+   */
+  autoHide?: "on" | "outside-click-only" | "off"
+  /**
+   * The content to appear inside the dropdown when it is open
+   */
+  children: React.ReactNode
+  portalSelector?: string
```
sebpearce, line 34: Could we please add a JSDoc comment here explaining what a `portalSelector` is? It might be confusing to people unfamiliar with it otherwise.

sebpearce (commented):
This is some serious work 👍  I hadn't heard of `popper` before.

draft-packages/menu/KaizenDraft/Menu/Menu.tsx
```diff
@@ -73,64 +35,21 @@ const Menu: Menu = props => {
...
+      {...rest}
+      isMenuVisible={isMenuVisible}
+      toggleMenuDropdown={toggleMenuDropdown}
+      hideMenuDropdown={hideMenuDropdown}
+      // `StatelessMenu` is one of the newer components, so it uses a render function,
+      // as opposed to `React.cloneElement`.
+      // `cloneElement` has its problems, mainly because it's somewhat magical, can lead
+      // to unexpected behaviour, and it doesn't self document. Hence, the switch
+      // to the render function. It would be nice if the `Menu` component also
+      // used this pattern, but it's probably not worth the time and effort.
```
jeremy8883, line 49: I didn't do the original code, but I added this comment. LMK if I made incorrect assumptions here.

sebpearce: That was the general feeling about it so it's all good. From memory that `cloneElement` was itself ported from an even earlier version of the component 🙃

sebpearce:
I'm happy with this, but will wait for @allybeaton's take since she was going to have a look too.

draft-packages/menu/KaizenDraft/Menu/MenuDropdown.tsx
```diff
@@ -15,126 +15,102 @@ type MenuDropdownProps = {
...
+  )
+  const { styles: popperStyles, attributes } = usePopper(
+    referenceElement,
+    popperElement,
+    {
+      modifiers: [
+        {
+          name: "offset",
+          options: {
+            offset: [0, 6], // value used from the $kz-spacing-xs scss variable
```
ActuallyACat, line 42: /meganit and you don't need to make this change. You can still use the tokens if you need them in javascript, like so: 

```
import { spacing } from "@kaizen/design-tokens/tokens/spacing"

options: {
   offset: [0, parseInt(spacing.kz.spacing.xs, 10)], 
 },
```

ActuallyACat (approved):
LGTM - the stories do a really good of documenting the changes. 👍

sebpearce (approved):
Adding my ✅  here in case it was blocking the merge.

PR #1151, opened 2021-03-01
feat: Add popper layout engine to `Popover`

jeremy8883:
* Add the popper layout engine to `Popover`. This will allow the popover to automatically change 

image

For the sake of backwards compatibility, internally I created `PopoverLegacy`, which gets used if the `referenceElement` prop has not been included. Keeping this up to date is going to be a problem. I'm tempted to go "No, you must go through every `Popover` component in your codebase, and update the props accordingly". I'm happy to fix all the existing code in perf-ui. LMK if you'd like to go down this path.

Contrary to the Tooltip and Menu components, I did not add a `portalSelector` prop. It will be the consumer's responsibility to wrap the popover in a portal if needed. This is due to a difference in the component design. The Popover doesn't wrap around an element, you pass it the reference.

# BREAKING CHANGE:
* To quickly fix all breaking changes, simply import `PopoverLegacy` instead of `Popover` from `@kaizen/draft-popover`. Codemod:

```
codemod -m -d ./src --extensions tsx,jsx 'import \{([\,\w\s]*)Popover([\w\W]*)\} from "@kaizen\/draft-popover"' 'import {\1PopoverLegacy\2} from "@kaizen/draft-popover"'
codemod -m -d ./src --extensions tsx,jsx '<Popover([\w\W]*)<\/Popover>' '<PopoverLegacy\1</PopoverLegacy>'
``` 

With that said, it is recommended that you start using the `usePopover` hook (see the stories for more examples), or the new `Popover` component if you are using a class component (see the `usePopover` hook to see how this needs to be wired up).

* The `boxOffset` property has been removed. Now that the new `Popover` uses popper, it will automatically determine the offset. This property was also being overloaded with two types of offsets, depending on the value type, which was confusing.
* With the new version of `Popover`, the `side` and `position` props are no longer used. Use `placement` instead. Also note, the `placement` describes the position of the popover, relative to the referenceElement, which is unlike the legacy popover, where it describes the placement of the arrow.
* We now use popper as the positioning engine, which shouldn't cause any breaking change, but please test your popovers when upgrading this package.

# Side updates
* Fix the `automationId` prop not getting added to the root div element
* Remove the `id` prop, which wasn't getting used.

# Checklist


draft-packages/popover/KaizenDraft/Popover/PopoverLegacy.tsx
```diff
@@ -0,0 +1,108 @@
...
+  mapArrowSideToClassLegacy,
+  mapArrowVariantToClass,
+  mapLineVariant,
+  mapSizeToClass,
+  mapVariantToBoxClass,
+  mapVariantToIcon,
+  mapVariantToIconClass,
+} from "./classMappers"
+
+export type LegacyPopoverProps = {
```
jeremy8883, line 20: Nothing really new in this file. It's the old `Popover.tsx` code moved over. With the following exceptions:
* The class mappers have been moved to `classMappers.ts`.
* The `automationId` prop is fixed
* The `id` prop is removed, because it wasn't getting used.
* Minor changes to how the arrow works, similar to what was done in Popover.elm

draft-packages/popover/KaizenDraft/Popover/PopoverModern.tsx
```diff
@@ -0,0 +1,204 @@
...
+  /** For almost all intents and purposes, you should be using a pre-defined variant.
+   Please avoid using a custom icon unless you have a very good reason to do so. **/
+  readonly customIcon?: React.SVGAttributes<SVGSymbolElement>
+  readonly referenceElement: HTMLElement | null
+  /**
+   * Render the tooltip inside a react portal, given the ccs selector.
+   * This is typically used for instances where the menu is a descendant of an
+   * `overflow: scroll` or `overflow: hidden` element.
+   */
+  readonly portalSelector?: string
```
jeremy8883, line 47: To summarise the changes to these props compared to the legacy ones:
* new prop `portalSelector`
* new prop `referenceElement`
* `position` and `side` has been merged into the new prop `placement`

draft-packages/popover/KaizenDraft/Popover/classMappers.ts
```diff
@@ -0,0 +1,118 @@
+import * as React from "react"
+import informativeIcon from "@kaizen/component-library/icons/information.icon.svg"
+import positiveIcon from "@kaizen/component-library/icons/success.icon.svg"
+import negativeIcon from "@kaizen/component-library/icons/exclamation.icon.svg"
+import styles from "./styles.scss"
+import { Side, Size, Variant, Position } from "./types"
+
+export const mapVariantToBoxClass = (variant: Variant): string => {
```
jeremy8883, line 8: These are all pretty much moved from the `Popover.tsx` file. The only exception being that I added the `Legacy` suffix where appropriate.

draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -9,37 +9,34 @@
...
 
 // Suggestion: with this implementation, the anchor point of the popover is at the base of the triangle. But,
 // we would ideally want it at the tip of the triangle. Otherwise, when using the popover, the consumer needs to manually
 // add something like a `margin-top: 8px` to get the popover properly positioned.
 // I didn't update this, as I didn't want to introduce a breaking change.
-@mixin arrow($background-color, $border-color, $size: $default-size) {
-  position: absolute;
-  top: 100%;
-  left: 50%;
-
```
jeremy8883, line 24: Thankfully, I was able to pretty much use the same classes as before. What I did was separate anything that related to popover placement into separate classes. So for the modern popover, I simply didn't use those classes.

draft-packages/popover/docs/Popover.stories.tsx
```diff
@@ -26,212 +24,365 @@ const Container = ({ children }: { children: React.ReactNode }) => (
...
 )
 
-export const DefaultKaizenSiteDemo = () => (
-  <Container>
-    <Popover heading="Default">
-      Popover body that explains something useful, is optional, and not critical
-      to completing a task.
-    </Popover>
-  </Container>
+const TargetElement = ({
```
jeremy8883, line 27: The updates made in this file:
* Using the new `referenceElementRef` prop, which enables the modern popover
* Adding a "Targen element" div, as so the popover can point to something
* Adding a "Legacy Popover" story, since this still needs to function for backwards compatibility
* Add a "overflow: scroll" story

draft-packages/popover/KaizenDraft/Popover/Popover.tsx
```diff
@@ -1,255 +1,35 @@
...
-  readonly onClose?: (event: React.MouseEvent<HTMLButtonElement>) => any
-  readonly variant?: Variant
-  readonly side?: Side
-  readonly size?: Size
-  readonly position?: Position
-  readonly heading?: string
-  readonly dismissible?: boolean
-  readonly singleLine?: boolean
-  readonly children: React.ReactNode
-  readonly boxOffset?: BoxOffset
```
jeremy8883, line 25: cc @stephwestgarth, this property will be removed since popper now takes care of the offset for us. I will this up on perf-ui. LMK if any issues.

draft-packages/popover/KaizenDraft/Popover/PopoverModern.tsx
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * const [referenceElement, referenceElementRef] = usePopoverReferenceElementRef()
+ *
+ * return (<>
+ *   <button ref={referenceElementRef}>
+ *     Hello world
+ *   </button>
+ *   <Popover referenceElement={referenceElement} />
+ * </>)
+ *
```
makenosound, line 161: This API feels a bit awkward to me. It’s not super difficult to understand, but it feels like most of this work could be abstracted away from users in _most_ use cases (though you might want to leave the basic API of `<Popper>` intact.

I can imagine something like this:

```tsx
const [popoverReferenceRef, PopoverWithCurriedRef] = usePopover();

return (
  <>
    <button ref={popoverReferenceRef}>Hello, World!</button>
    <PopoverWithCurriedRef>
      <p>Content</p>
    <PopoverWithCurriedRef>
  </>
)
```

With `usePopover` being something vaguely like:

```tsx
function usePopover() {
  const [refElement, setRefElement] = useState<HTMLElement | null>(null)
  const [popoverComponent, setPopoverComponent] = useState<React.FunctionComponent>(() => undefined)
  const ref = useCallback(node: HTMLElement => {
    setRefElement(node)
  })

  useEffect(() => {
    if (refElement) {
      setPopoverComponent((props) => {
        return <Popover {...props} referenceElement={refElement}/>
      })
    }
  }, [refElement, setPopoverComponent])

  return [ref, popoverComponent]
}
```

That _feels_ like a more natural API to me, since consumers don’t have to worry about handling nearly as much passing variables around. WDYR?

jeremy8883: Yeah, that would be easier to understand as a consumer. I've updated this now, LMK if all good. Only difference, I used `useMemo` instead of `useEffect`/`useState`, to avoid a double render.

draft-packages/popover/KaizenDraft/Popover/PopoverModern.tsx
```diff
@@ -147,33 +147,46 @@ export const PopoverModern: PopoverModernType = ({
...
 ] => {
   const [referenceElement, setReferenceElement] = useState<HTMLElement | null>(
     null
   )
 
-  return [referenceElement, setReferenceElement]
+  // I guess the problem with this pattern, is that every time referenceElement
+  // changes, a brand new component is generated, which would be bad for memoization.
+  // In this situation however, the value is rarely going to change, and
+  // popovers aren't going to include content with expensive render times.
```
makenosound, line 182: This is a good comment, but I agree that it’s unlikely to be a problem in practice, so I think the implementation is sensible.

draft-packages/popover/KaizenDraft/Popover/PopoverModern.tsx
```diff
@@ -147,33 +147,46 @@ export const PopoverModern: PopoverModernType = ({
   )
 }
 
+type PopoverPropsWithoutRef = Omit<ModernPopoverProps, "referenceElement">
+
 /**
  * How to use:
  *
- * const [referenceElement, referenceElementRef] = usePopoverReferenceElementRef()
+ * const [referenceElementRef, Popover] = usePopoverReferenceElementRef()
```
makenosound, line 155: This should be `usePopover` now, yeah?

makenosound (commented):
> I'm tempted to go "No, you must go through every Popover component in your codebase, and update the props accordingly". I'm happy to fix all the existing code in perf-ui. LMK if you'd like to go down this path.

My feeling is that it’d be better to properly deprecate it and create some separation. Could we instead:

* Expose the legacy component separately
* Write a codemod for the change (probably not super necessary given it’d be a simple find-replace)

jeremy8883:
Sweet, I've split the two components up now

jeremy8883:
For the codemod, where should these be located? I've included some in the PR description, which I'll also paste into the squashed commit.

```
codemod -m -d ./src --extensions tsx,jsx '<Popover([\w\W]*)<\/Popover>' '<PopoverLegacy\1</PopoverLegacy>'

codemod -m -d ./src --extensions tsx,jsx 'import \{([\,\w\s]*)Popover([\w\W]*)\} from "@kaizen\/draft-popover"' 'import {\1PopoverLegacy\2} from "@kaizen/draft-popover"'
```

makenosound:
I imagine that’s probably enough in the codebase? You could promote them heavily in your post in #updates_design_systems though.

makenosound (approved):

PR #1166, opened 2021-03-02
feat: Include notification prop to display on Information Modal

vicchirino:
# Objective
Allow to display an inline notification on the Information Modal.

# Motivation and Context
On Performance we are updating modals when a manager is going to share a review. Currently we display an inline notification on the modal saying what is going to happen after the notification is shared. 

Include that notification as part of the `children` prop was generating that be stretched to the left. 

image

Added a `contentHeader ` prop into the modal to display the inline notification component between the title and modal content.

After a chat with Marc, we want the `InformationModal` when have an image on the right, flip the order of the actions buttons to have the primary action on the left side. For this include a `variant` prop on the `ModalFooter` component to avoid breaking changes.
Slack conversation: https://cultureamp.slack.com/archives/C0189KBPM4Y/p1614785352180800

image

# Checklist


draft-packages/modal/docs/Modal.stories.tsx
```diff
@@ -1,3 +1,4 @@
+import { title } from "process"
```
ActuallyACat, line 1: This import looks like an accident.

draft-packages/modal/docs/Modal.stories.tsx
```diff
@@ -705,6 +710,74 @@ InformationModalWithImage.story = {
...
+            </Heading>
+          </ModalAccessibleDescription>
+          <Padding />
+          <Paragraph variant="body">
+            • What did this person achieve over the last 6 months? What specific
+            impact did they have on the team’s goals?
+          </Paragraph>
+          <Padding />
+          <Paragraph variant="body">
+            • What progress have they made on their learning anddevelopment
```
ActuallyACat, line 758: /nit `anddevelopment` => `and development`

draft-packages/modal/KaizenDraft/Modal/Presets/InformationModal.tsx
```diff
@@ -25,6 +32,7 @@ export type InformationModalProps = Readonly<
     renderBackground?: () => React.ReactNode
     image?: React.ReactNode
     children: React.ReactNode
+    notification?: InlineNotificationProps
```
ActuallyACat, line 35: Something like `contentHeader?: children` would be preferred so it's a little more generic.

vicchirino: yes will do that! Thanks 🙏

ActuallyACat (approved):
LGTM - nice one @vicchirino

PR #1202, opened 2021-03-08
chore: Don't pin specific versions

jasononeil:
# Objective

I want to ensure dependencies are able to be updated using normal processes (manual npm-refresh or dependabot) and having versions specifically pinned makes that harder.


ActuallyACat (approved):
LGTM - Nothing I could see in git blame indicated these have been pinned for a particular reason.

ActuallyACat (approved):

mbylstra (approved):

PR #1224, opened 2021-03-10
fix: Remove max-width to prevent bugs outside grid

sebpearce:
# ⚠️ VISUAL BREAKING CHANGE: Because TileGrid no longer sets its own max-width, the surrounding container or page must take on this responsibility of constraining its width. If this is not already the case on a page that is affected by this update, the TileGrid will be wider than it should be.

# Motivation and Context

The `max-width` rule on TileGrid was the source of a confusing bug in a consuming repo owned by Team Unify. It was causing margins to not be respected on a containing element a few layers up.

Since
- turning it off solves the problem, and
- the TileGrid should not be responsible for constraining its own width anyway (that is the page's responsibility),

I decided to remove it.

# Checklist


ActuallyACat (approved):
lgtm

PR #1225, opened 2021-03-10
feat: Add Heart support to loading spinner

ActuallyACat:
# Objective
**There are no visual changes to this PR**

* Removed unused imports
* Removed import that was not a dependency in Kaizen. The alternative is to add component-library as a dependency in this package which feels unnecessary given the size of component-library. 
* Use the new animation variable tokens


dmisdm (approved):

ActuallyACat:
Can we remove the component-library package from loading spinner ?

PR #1226, opened 2021-03-10
feat: Persist storybook theme with query param support

dmisdm:
Allow people to set kaizen themes in storybook using &theme=zen|heart, and persist the current theme to localStorage.

# Motivation and Context
This is to support design reviews. Designers (Marc) are wanting a more frictionless CSS inspecting experience, and this PR will allow them to set the theme in the addon panel or URL, and subsequently **open the story canvas in a separate tab and see the same theme**.
The latter action is also a driver for this PR, because it will enable the usage of this addon (It doesn't allow the inspection of the inner story/canvas iframes): https://chrome.google.com/webstore/detail/css-peeper/mbnbehikldjhnfehhnaidhjhoofhpehk?hl=en

And, this addon might improve the design review process.

# Checklist

# Additional Considerations


storybook/theme-switcher-addon/themeManager.ts
```diff
@@ -0,0 +1,25 @@
...
+    case "heart":
+      return heartTheme
+    default:
+      return zenTheme
+  }
+}
+
+export const getInitialTheme = () =>
+  themeOfKey(
+    window.location.search.match(/(\?|\&)theme=(zen|heart)/)?.[2] ||
```
ActuallyACat, line 20: I would think that the logic would work something like this (feel free to challenge, of course)

```
If query param exists for theme then 
   set that theme in local storage (overriding existing value if it exists)
   set that theme's variables on root of document
else
   if theme is defined in local storage then
      use local storage theme
   else
      use default theme (zen)
```

I've opened a new tab with the theme, and tried to override the theme with a query parameter and it seems the only way to change the theme is to open another tab and change the theme there.

dmisdm: @ActuallyACat i did some digging because i could swear my experience of the logic was exactly as you described.
Turns out the useEffect hook that updates localStorage isn't run when viewing the canvas/viewOnly tab, because the addons aren't loaded!
Amending this now.

storybook/theme-switcher-addon/tokens.ts
```diff
@@ -1,197 +0,0 @@
-// @TODO - these should use tokens & values from @kaizen/design-tokens
```
ActuallyACat, line 1: Thanks for cleaning this up!

ActuallyACat (dismissed):

ActuallyACat (approved):
Looks good!

PR #1227, opened 2021-03-10
docs: Show RGB and HEX values for colors story

dmisdm:
Show RGB and HEX values on each color within the story board.

# Motivation and Context
Originally motivated by Marc's need to know RGB values for each token, but this should be useful for other designers and devs.

image

# Checklist

# Additional Considerations


packages/design-tokens/docs/StoryBoard.stories.tsx
```diff
@@ -140,9 +141,38 @@ const ColorDemo = (props: { color: string; name?: string }) => {
...
           ...theme.border.solid,
         }}
-      ></div>
+      >
+        {parsedColor && (
+          <Heading variant="heading-5">
+            <span
+              style={{
+                color: props.color,
+                filter: "invert(1) grayscale(1) contrast(100)",
```
jasononeil, line 156: This is clever. 

Back on the old Kaizen site I had [a whole contrast calculation thing](https://github.com/cultureamp/cultureamp-style-guide/blob/master/guide/src/pages/styles/colors/_ColorBlock.js#L18) to determine if text should be white or black. If you did something like that you could use the Heading's actual props to toggle colour, rather than this span with inline styles.

Not sure it's worth it though.

jasononeil (approved):
The overlap with this page is unfortunate: https://cultureamp.design/guidelines/color/

But, the Gatsby site doesn't have a theme switcher yet, so 🤷 

Again - the divide between storybook and the Gatsby site is going to be a concern.

Anyway, that's not something that should hold up this PR. ✅

packages/design-tokens/docs/StoryBoard.stories.tsx
dmisdm: Yeah this is just to potentially help Marc, so probably not worth worrying about too much, and as you say, the crossover between storybook and gatsby will need to be cleared up - we'll probably ditch this color demo at some point maybe.

dmisdm:
> The overlap with this page is unfortunate: https://cultureamp.design/guidelines/color/
> 
> But, the Gatsby site doesn't have a theme switcher yet, so 🤷
> 
> Again - the divide between storybook and the Gatsby site is going to be a concern.
> 
> Anyway, that's not something that should hold up this PR. ✅

Good pickup thank you, i did not know about that page!

PR #941, opened 2021-01-18
feat: Add ability to use 'working' confirm/submit buttons on input/edit and confirmation modals

dougmacknz:
**BREAKING CHANGE:** submitDisabled has been removed from InputEditModal, use submitWorking instead.

**BREAKING CHANGE:** ModalFooter's 'actions' prop now accepts ButtonProps[], which means the 'action' property was renamed to 'onClick'

# Objective
We recently added the `working` variant of `Button`, and the need has popped up for us to use this on Confirmation and Input/Edit Modals. The Input/Edit modal had a `submitDisabled` prop added to it a while ago for the same purpose, this replaces that out.

# Motivation and Context 
On Capture, we have a Confirmation Modal that shows before the survey is submitted. As part of the work to add a request queue to Capture, we would like to hold the user on the modal after they press confirm to check that all of the events have processed.

While I was making the change to the ConfirmationModal, I've also made the same change to the InputEditModal, replacing the `submitDisabled` prop with a `submitWorking` prop.

# Screenshots

image

image

# Checklist

[ ] If this contains visual changes, has it been reviewed by a designer? 
[ ] I have considered likely risks of these changes and got someone else to QA as appropriate
[ ] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline? 
- Have Storybook stories been updated? 

**If you're new to Kaizen, please ask #prod_design_systems to set up an onboarding session to get you up to speed.** If you have an urgent PR to merge before that happens, it is safest to ask Design Systems team to review it to catch any issues.

PR #1243, opened 2021-03-12
feat: InformationModal include working button

vicchirino:
# Objective
Include working button props on `InformationModal`

Add `WorkingActionProps` on `InformationModal` to allow primary button to display loading state when some action is happening. This is needed on `performance-ui` when we share performance reviews using this modal.

Add story to reflect this change
 
https://user-images.githubusercontent.com/3228237/110971716-f9aa6380-8339-11eb-8b6c-49add712efb1.mov

Include update on contentHeader layout to have same padding that the main content.

**Before:**
image

**After: **
image


dougmacknz:
Hi @vicchirino, we have a pattern for `working` primary buttons on modals that was created in https://github.com/cultureamp/kaizen-design-system/pull/941. (The only reason I didn't do it for `InformationModal` there is because I didn't think this modal was used for submitting/confirming things, I was wrong.)

The usage looks like this:
`<ConfirmationModal confirmWorking={{ label: "Working…" }} />`
`<InputEditModal submitWorking={{ label: "Submitting…" }} />`

They can also receive an optional `labelHidden` prop like so (hence the object):
`<ConfirmationModal confirmWorking={{ label: "Working…", labelHidden: true }} />`

This PR would introduce a new pattern looking like this, which is going to be inconsistent:
`<InformationModal isWorking={true} workingLabel="Sharing review" />`

Can I please get you to look over that PR I linked above and copy the way it was done in `ConfirmationModal`. You will want to use the wording `confirmWorking` because the `InformationModal` has a `confirm` button (rather than a `submit` button).

The other behaviour missing here is that when `working` is enabled, `onDismiss` should be hijacked to block users from closing the modal while it's working. Again, please just copy the way it was done in https://github.com/cultureamp/kaizen-design-system/pull/941.

keirah:
> Hi @vicchirino, we have a pattern for `working` primary buttons on modals that was created in #941. (The only reason I didn't do it for `InformationModal` there is because I didn't think this modal was used for submitting/confirming things, I was wrong.)
> 
> The usage looks like this:
> `<ConfirmationModal confirmWorking={{ label: "Working…" }} />`
> `<InputEditModal submitWorking={{ label: "Submitting…" }} />`
> 
> They can also receive an optional `labelHidden` prop like so (hence the object):
> `<ConfirmationModal confirmWorking={{ label: "Working…", labelHidden: true }} />`
> 
> This PR would introduce a new pattern looking like this, which is going to be inconsistent:
> `<InformationModal isWorking={true} workingLabel="Sharing review" />`
> 
> Can I please get you to look over that PR I linked above and copy the way it was done in `ConfirmationModal`. You will want to use the wording `confirmWorking` because the `InformationModal` has a `confirm` button (rather than a `submit` button).
> 
> The other behaviour missing here is that when `working` is enabled, `onDismiss` should be hijacked to block users from closing the modal while it's working. Again, please just copy the way it was done in #941.

Thanks for the detailed feedback @dougmacknz! @vicchirino is on a well deserved holiday right now so I'm going to try and move these changes forward.

keirah:
Have pushed some changes, but would like to get them working with the [corresponding PR](https://github.com/cultureamp/performance-ui/pull/2768) that Vic has in the Performance-UI repo (to make sure it still works for his use case), before requesting a re-review.

draft-packages/modal/KaizenDraft/Modal/Presets/InformationModal.tsx
```diff
@@ -34,75 +36,97 @@ type InformationModal = React.FunctionComponent<InformationModalProps>
...
+        workingLabelHidden: confirmWorking.labelHidden,
+      }
+    : {}
+
+  if (onConfirm) {
+    const confirmAction = { label: confirmLabel, onClick: onConfirm }
+
+    footerActions.push({ ...confirmAction, ...workingProps })
+  }
+
```
keirah, line 66: The `ConfirmationModal` has a separate footerActions included

`  footerActions.push({
    label: dismissLabel,
    onClick: onDismiss,
    disabled: !!confirmWorking,
  })
`

Unsure if this is something that is required on the InformationModal?

dougmacknz (approved):

PR #1279, opened 2021-03-18
docs: update header

zakwash01:
# Objective
Updating header documentation

# Motivation and Context
New IA has introduced flow-of-work admin pages, requiring clarification about when to use the Account admin header.

### DIFF
--- a/site/docs/guidelines/header.mdx
+++ b/site/docs/guidelines/header.mdx
@@ -24,7 +24,7 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
 - Variants:
     - **Default (Wisteria):** This is for most content and reporting pages. These pages let you consume data and take action.
     - **Education (Cluny):** This is for informative content, such as manager learning.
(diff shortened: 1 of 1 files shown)


katsambarcus (changes requested):
@zakwash01 Should we refer to Account admin as 'Account settings'? I think this is the language that @lucy-oldridge wants to try and instill.

zakwash01:
> @zakwash01 Should we refer to Account admin as 'Account settings'? I think this is the language that @lucy-oldridge wants to try and instill.

To avoid confusion, I've left it 'Account admin' for now as that's the current label. When we release the new nav I'll come back and update it!

katsambarcus (approved):

katsambarcus:
> > @zakwash01 Should we refer to Account admin as 'Account settings'? I think this is the language that @lucy-oldridge wants to try and instill.
> 
> To avoid confusion, I've left it 'Account admin' for now as that's the current label. When we release the new nav I'll come back and update it!

Approved on that basis.

Issue #1184, opened 2021-03-04
@packages/draft-modal incompatible with React 17


lijuenc:
When using the Kaizen modal with React 17 you might encounter an error that says `Unable to find node on an unmounted component`.

![Screen Shot 2021-03-04 at 12 51 26 pm](https://user-images.githubusercontent.com/4900094/109898435-5da0ac80-7ce8-11eb-8ec0-7a1a75a53be8.png)

This affects all React modals as the error appears to be originating from the use of a `CSSTransition` in our `GenericModal`. `CSSTransition` is a component from the `react-transition-group` package and is used to perform animations when opening and closing the modal.

https://github.com/cultureamp/kaizen-design-system/blob/c39749f6c2ffcebbdff44c5cb31fe4d9e1b51a02/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx#L170

Using something as simple as `<GenericModal isOpen={isOpen}>test</GenericModal>` and toggling the value of `isOpen` from true to false or false to true will cause this error.

I have a repository with a minimal reproducible scenario here. It just has the draft modal package, React 17, and a button that attempts to open a modal.

https://github.com/cultureamp/lijuenc-kaizen-modal

lijuenc:
Fixed by https://github.com/cultureamp/kaizen-design-system/pull/1301

PR #1301, opened 2021-03-22
fix: React 17 fix - replace react-transition-group in draft-modal

lijuenc:
# Objective
Replace CSSTransition from react-transition-group with Transition from headlessui/react. The API is very close making it a near drop-in replacement.

For this PR, the relevant change happens in `GenericModal.scss` and `GenericModal.tsx`. There is also a minor change to `Modal.elm` to make the CSS class names consistent.

The rest of the PR is adding tests to GenericModal and the presets we have (ConfirmationModal, InformationModal, InputEditModal, RoadblockModal) to hopefully catch issues like this in the future.

# Motivation and Context
react-transition-group is not actively maintained (https://github.com/reactjs/react-transition-group/issues/630#issuecomment-797733168) and is not fully compatible with React 17 (https://github.com/cultureamp/kaizen-design-system/issues/1184).

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer? N/A

# Additional Considerations
A basic test case has been added for GenericModal, however, we would like to have the test suite running in both React 16 and React 17 to catch issues such as this.

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Modal.elm
+++ b/draft-packages/modal/KaizenDraft/Modal/Modal.elm
@@ -241,7 +241,7 @@ viewContent (Config modalConfig) =
                     [ ( .animatingElmEnter, True ) ]
 
                 Closing_ ->
-                    [ ( .animatingElmExit, True ) ]
+                    [ ( .animatingElmLeave, True ) ]
 
                 _ ->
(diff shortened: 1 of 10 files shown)


ActuallyACat (approved):
LGTM 🎉

PR #1308, opened 2021-03-23
refactor: Migrate remaining usages of non-variable tokens

dmisdm:
# Migration to new design tokens

**Commit messages contain Step 1, Step 2 etc, but the numbering doesn't exactly correlate with this outline**

1. Replace $kz with $kz-var (unless already replaced) across all files, including some MDX/MD.

    

    

2. Rebuild the library (just in case the keys in token files (sass/typography.scss etc) get changed )
3. Commit
4.  Replace imports

    "~@kaizen/design-tokens/sass/(color|border|typography|spacing|layout|animation|shadow)"

    "~@kaizen/design-tokens/sass/$1-vars"

    

5. Manually amend one file

    

6. Commit
7. **Manually find and fix** all rgba functions that don't contain an rgb-params variable
regex query: `rgba\(.*(?<!(-rgb-params)),`

    

    Example: change $ca-base-text-color to $kz-var-color-wisteria-800-rgb-params in rgba functions

    

8. Commit
9. Manually find and fix usages of add-alpha, add-tint, add-shade, mix, darken, lighten

    Resulted in only 2 changes: 

    

    ^^ Outdated screenshot (i added mix, darken, lighten after, but it made no difference)

10. Commit

### Retrospective steps

These are some steps that we're completed in later commits, that should have been done in the above steps.
11. Find all negations of kaizen variable tokens, and convert to `calc(-1 * $kz-var-blah)`
12. Find all @media queries and ensure no CSS variable tokens are used.
13. Ensure that any typography mixins (which are deprecated) aren't using tokens. There is a function called `strip-unit` which is the root of all failures (can't remove the unit from a CSS variable and just return the number).
14. Make further changes after comparing the CSS snapshot against master. See https://github.com/cultureamp/kaizen-design-system/pull/1336 - This PR was rebased off this one, then changed PR base to master, then all traces of it was removed at the end so that the other PR can be the one to integrate CSS snapshots. 

### Verification
This is how we've verified that the migration has gone smoothly
- `yarn storybook` to find sass errors
-  Fix those errors as best as possible. The deprecated typography mixins couldn't be fixed due to the `strip-unit` function being used. Still thinking of a good solution, but for now i've just used the old concrete tokens.
- CI/CD builds the gatsby site too, which includes more potentially flawed SASS.
- Run a chromatic review. This finds any visual regressions, and has helped in identifying further required migration steps
- **Generate a whole CSS snapshot of master (see https://github.com/cultureamp/kaizen-design-system/pull/1336), then generate one after the changes, and compare them. You can review the snapshot changes in the late commits**

### Lessons learned
- Gatsby production build of SASS gives very ambiguous errors sometimes, causing a lot of time spent on debugging.
- Bulk converting $kz -> $kz-var causes silent failures in `calc()` functions that contain tokens.
- Negations of CSS variables silently fail in SASS (e.g. `-$kz-var` which equals `-var(--blah)`)
- Transitive usages of tokens (e.g. `$large: $kz-var-layout-breakpoints-large;` then `@media (min-width: $large + $sidebarWidthPadding) {`) silently fail sometimes, like in media queries. This kind of issue is most suited to being solved with a smart codemod/linter that can resolve/understand the value within a variable like `$large`.
- You can't use CSS variables in `@media` queries, which prevented the migration of some files. **Also they all silently fail gahhh damn it SASS**.  This in particular is also a dangerous one because Chromatic doesn't run across all screen sizes. 😅
- Just to reiterate: Be very aware of transitive usages of tokens. Don't find+replace using `$kz-blah` thinking those are the only usages you need to worry about.

# Checklist

# Additional Considerations


draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
```diff
@@ -1,6 +1,6 @@
-@import "~@kaizen/design-tokens/sass/color-vars";
-@import "~@kaizen/design-tokens/sass/border-vars";
-@import "~@kaizen/design-tokens/sass/shadow-vars";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/shadow";
```
jasononeil: What happened here? (It looks like this commit was about fixing rgba these lines revert imports to non-variable versions, and it's not clear why)

draft-packages/table/KaizenDraft/Table/styles.scss
```diff
@@ -176,7 +176,10 @@ $row-height: 60px;
 .expanded {
   // We want to have NO background on Heart theme (when CSS variables are defined), but keep stone backgrond on Zen. Need to do this using 0 opacity.
   // We also only want this switch for the expanded state. Non-expanded hovering will resort to stone background, as seen in .hasHoverState
-  --dt-hover-background-heart: rgba(var(#{$kz-var-id-color-white}), 0);
+  --dt-hover-background-heart: rgba(
+    var(#{$kz-var-id-color-white-rgb-params}),
```
jasononeil, line 180: Not suggesting a change, but I'm curiouso: is the `#{...}` interpolationo syntax needed here? We don't seem to be using it in other places in this commit.

draft-packages/button/KaizenDraft/Button/styles.scss
```diff
@@ -103,18 +102,20 @@ $caButton-verticalPaddingForm: calc(
 
       // show custom focus ring when :focus-visible
       &:global(.focus-visible)::after {
-        $focus-ring-offset: ($kz-var-border-focus-ring-border-width * 2) + 1px;
+        $focus-ring-offset: calc(
+          (#{$kz-var-border-focus-ring-border-width}#{" * 2"}) #{" + 1px"}
```
jasononeil, line 106: Sass 🤮 

According to sass-meister this should work the same... can we try it?

```suggestion
          (#{$kz-var-border-focus-ring-border-width} * 2) + 1px
```

This style of syntax is pretty gross and I'm worried it'll scare away contributors, so if there is a saner syntax that still compiles I'd like to find it.

(Same goes for a bunch of other changes in this specific commit)

jasononeil (dismissed):

draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
jasononeil: Looks like it got fixed by the end of the PR 🤷

dmisdm:
@jasononeil regarding all the string interpolations: the reason for them is to remove any possibility of the SASS compiler doing a calculation/transformation, e.g.
```
padding: calc(#{kz-var-spacing-md} + 1);
```

turns into
```
padding: calc(var(--kz-var-spacing-md)1);
```
because it thinks it's concatenating strings.

And it could just become more ambiguous:
```
padding: calc(#{kz-var-spacing-md} + 3 * 1px);
```
If it's BODMAS then it should be fine, but it's a bit of extra cognitive load for the developer.

The behaviour of the SASS compiler, i.e. having a sense of what the `calc(...)` will look like at the end, is a bit of a mystery, at least to me. I think it is to most people but i don't know.

Going through each calc + interpolation usage in SASS meister and double checking it transforms into something correct is a good idea, but i actually think the ugly string interpolation is more maintainable because it means any developer that comes to it in the future doesn't have to do a manual SASSmeister (or other) check, or keep track of that complexity in their head, and thus less prone to human error. 

It also expresses what we want exactly (in a very ugly way); it puts the operators, scalars and constants in exactly as they are without processing.

Tbh i've mentally debated this a lot and don't know what the best outcome is. Knowing all this, if you've still got the drive to move away from the interpolation chaos, lets do it.

draft-packages/button/KaizenDraft/Button/styles.scss
dmisdm: @jasononeil not sure if you saw my message but i'll fix these up and remove the extra literals

dmisdm:
There were flaws in what i said lol! I think what i tried was `#{$someVar + 1}` which was causing the string concat, while `#{$someVar} + 1` is fine.

snapshot.css
```diff
@@ -1340,11 +1453,11 @@ html:lang(ar)
   flex-direction: row;
   align-items: center;
   justify-content: flex-start;
-  margin: 0 -0.375rem;
+  margin: 0 -var(--kz-var-spacing-sm, 0.75rem) / 2;
```
dmisdm, line 1456: This is an error:
Should be fixed in `draft-packages-filter-menu-button-KaizenDraft-FilterMenuButton-styles-module__filterMenuButton--3RBsO`

snapshot.css
```diff
@@ -478,7 +531,9 @@
 }
 .js-focus-visible
   .draft-packages-button-KaizenDraft-Button-components-GenericButton-module__secondary--H8yYm.focus-visible
-  .draft-packages-button-KaizenDraft-Button-components-GenericButton-module__content--1fJr9,
+  .draft-packages-button-KaizenDraft-Button-components-GenericButton-module__content--1fJr9 {
+  padding: calc(0.75rem - 2px) calc(0.75rem - 2px);
```
dmisdm, line 535: This is new? Should probably investigate to make sure.

snapshot.css
```diff
@@ -5621,10 +5964,10 @@ html:lang(ar)
   border-width: 2px;
   border-style: solid;
   border-color: #0168b3;
-  top: -5px;
-  left: -5px;
-  right: -5px;
-  bottom: -5px;
+  top: calc(-1 * $focus-ring-offset);
```
dmisdm, line 5967: This should be fixed. The path to the source is in the class name:  `.draft-packages-likert-scale-legacy-KaizenDraft-LikertScaleLegacy-styles-module__container--1Iqv-.focus-visible`

The string literal `$focus-ring-offset` is being used within a calc.

snapshot.css
```diff
@@ -11669,9 +12185,9 @@ html:lang(ar) .draft-packages-tag-KaizenDraft-Tag-Tag__textContent--BwO_T {
   position: relative;
   display: flex;
   padding: 0.375rem;
-  margin-top: -var(--kz-var-spacing-xs, 0.375rem);
+  margin-top: calc(-1 * 0.375rem);
```
dmisdm, line 12188: Was the original/master source incorrect here? I think so.

snapshot.css
```diff
@@ -59421,7 +60095,8 @@ html:lang(ar)
 .site-src-docs-components-animation-AnimationSequencesExample__is-animating--3SbM9
   .site-src-docs-components-animation-AnimationSequencesExample__spin-2--dJhhr {
   animation-name: site-src-docs-components-animation-AnimationSequencesExample__pulsate--1HQh5;
-  animation-delay: 0.3s;
+  animation-delay: 0svar (--kz-var-animation-duration-deliberate, 700ms)
```
dmisdm, line 60098: Should be fixed.

snapshot.css
```diff
@@ -59448,7 +60123,8 @@ html:lang(ar)
 .site-src-docs-components-animation-AnimationSequencesExample__is-animating--3SbM9
   .site-src-docs-components-animation-AnimationSequencesExample__spin-3--3rVoJ {
   animation-name: site-src-docs-components-animation-AnimationSequencesExample__pulsate--1HQh5;
-  animation-delay: 0.6s;
+  animation-delay: 0svar (--kz-var-animation-duration-deliberate, 700ms)
```
dmisdm, line 60126: Again here.

snapshot.css
```diff
@@ -5288,377 +5316,6 @@ html:lang(ar)
...
-  0% {
-    opacity: 0.5;
-    -webkit-transform: scale(1);
-  }
-  100% {
-    opacity: 0;
-    -webkit-transform: scale(1.78);
-  }
-}
-
```
dmisdm, line 5301: All these getting removed makes me suspicious of the snapshot generation...but it's hard to say...

dmisdm: Pretty sure the postcss-calc plugin bombed out on this file, so didn't include it.

dmisdm: It came back after removing the plugin

ActuallyACat (approved):
This LGTM - I couldn't find anything awry here. 

After looking at the snapshot I can see there are some negative vars that affect other components (i.e. not introduced in this PR) so I'll get that triaged
```
.draft-packages-tag-KaizenDraft-Tag-Tag__dismissIcon--3iOQK {
  position: relative;
  display: flex;
  padding: 0.375rem;
  margin-top: -var(--kz-var-spacing-xs, 0.375rem);
```

PR #1312, opened 2021-03-24
fix: bumping tooltip z-index to much higher value, by using layers vars

nickrm97:
# Objective
Give tooltip a more reasonable, higher z-index. The structure of the tooltip has also been reworked so there are fewer stacking contexts created. 

Instead of the tooltip content being wrapped in a div that controls the animations, like this: 
```
<div class="animation__visibleState animation__defaultState"> // opacity: 1, will-change: opacity
   <div class="tooltip-content"> // position: absolute, transform: translate(x,x) 
      Tooltip content
   </div>
   <div class="tooltip-arrow" />  // position: absolute, transform: translate(x,x) 
</div>
```

It has become: 
```
<div class="animation__visibleState animation__defaultState">  // position: absolute, transform: translate(x,x), opacity: 1, will-change: opacity
   <div class="tooltip-content"> // z-index: 1070
      Tooltip content
   </div>
   <div class="tooltip-arrow" />  // position: absolute, transform: translate(x,x), z-index: 1070
</div>
```

# Motivation and Context

Before: 
Menu dropdowns (and other absolutely position elements) are stacked over the tooltip
Screen Shot 2021-04-01 at 4 59 05 pm

After
Tooltip is layered on top of other elements, as it should be
Screen Shot 2021-04-01 at 5 07 19 pm

# Checklist

# Additional Considerations


draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
```diff
@@ -17,13 +18,14 @@ $dt-tooltip-content-border: var(
...
   pointer-events: none;
+  z-index: $ca-z-index-tooltip;
 }
 
 .tooltipContent {
   border-radius: $kz-var-border-solid-border-radius;
   transition: opacity $ca-duration-fast, transform $ca-duration-fast;
   padding: $ca-grid / 2 $ca-grid / 2;
-  z-index: 2;
+  z-index: $ca-z-index-tooltip;
```
sentience, line 28: If we set `z-index: $ca-z-index-tooltip` for the `.tooltip` element above, this property may not be necessary at all. If it needs to be stacked on top of `.arrow` and its descendants, then `z-index: 1` should be all that is needed here, since we would be working inside the stacking context created by `.tooltip`. Using `$ca-z-index-tooltip` within a context that is already at `$ca-z-index-tooltip` relative to the rest of the page seems like overkill to me, and a little confusing.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
```diff
@@ -40,7 +42,7 @@ $dt-tooltip-content-border: var(
 }
 
 .arrow {
-  z-index: 2;
+  z-index: $ca-z-index-tooltip;
```
sentience, line 45: As for `.tooltipContent`, I don't believe the `z-index` here is required if we decide to have `.tooltip` create the stacking context for the entire tooltip. If `.arrow` should stack on top of `.tooltipContent`, then the automatic stacking order will do this without us meddling with `z-index` values. If `.tooltipContent` should stack on top of `.arrow`, then we only need to specify an elevated `z-index` property for `.tooltipContent`, and we can leave `.arrow` on the base layer.

draft-packages/tooltip/docs/Tooltip.stories.tsx
```diff
@@ -1,20 +1,28 @@
...
 import { withDesign } from "storybook-addon-designs"
 import { Button } from "@kaizen/draft-button"
 import { Tooltip } from "@kaizen/draft-tooltip"
+import isChromatic from "chromatic/isChromatic"
 import { figmaEmbed } from "../../../storybook/helpers"
 
+const excludeFromChromatic = Story => {
+  if (!isChromatic()) return <Story />
+  return null
+}
```
sentience, line 16: Why exclude the tooltip stories from Chromatic? If anything, I would expect it would be good to have screenshot coverage of this stacking issue, so we avoid a regression.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -150,37 +154,39 @@ const Tooltip = ({
...
-          setIsFocus(true)
-        }}
-        onBlurCapture={() => {
-          setIsFocus(false)
-        }}
-        aria-describedby={tooltipId}
-      >
-        {children}
-      </div>
+    <AnimationProvider isVisible={isHover || isFocus}>
```
sentience, line 157: I don't understand what `AnimationProvider` is doing for us, or what the changes to `AnimationProvider` above are for. I think I'll need someone to explain this to me. 🙏

sentience (changes requested):
I think there's still some “z-index overkill” going on here, where we're setting high z-indexes inside of high z-indexes unnecessarily. Suggest we jump on a Zoom to figure it out together.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
```diff
@@ -17,6 +17,7 @@ $dt-tooltip-content-border: var(
 );
 
 .tooltip {
+  position: relative; // Set an initial position for the tooltip, which will then be overridden by Popper
```
sentience, line 20:
```suggestion
  position: relative; // Make this a positioned element so z-index works. (Note: Popper overrides this to absolute.)
```

sentience (dismissed):
One small suggestion above to make a comment clearer, but happy to approve this as-is if you prefer

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
ActuallyACat: Much clearer - committed

jasononeil (approved):
Approving based on Kevin's prior comment that was approving other than updating a single comment, which is now updated.

PR #1316, opened 2021-03-25
fix: Select component style regression on Zen

dmisdm:
## Objective
Fix the border color on the Select component which was picked up by chromatic, and occurred when updating to Heart.


ActuallyACat (approved):
LGTM

PR #1318, opened 2021-03-25
fix: InlineNotification issue from Heart update

dmisdm:
# Objective
Fixes the negative margin on InlineNotification's cancel button (needed a `calc()` function to negate a CSS variable; silently failed)


jasononeil (approved):
Good pickup!

PR #1320, opened 2021-03-26
feat: add additional background colors to HeroCard component

talbet:
# Objective

Adding the ability to change the background color of the left hand side of the HeroCard component. Now several options are available including the default background colour, currently set to wisteria700.

I thought about adding a transparent background, and then controlling the background color in a child element, but there was too much absolute positioning and/or negative margins involved for it to feel like a clean solution.

This is still not a very extensible way to define variations, but I think a larger conversation needs to happen before dynamic styling is included in any components.

# Motivation and Context

Previously the HeroCard component used wisteria700 as a hardcoded background color, this change allows variations to the background color, allowing the new survey summary page to use it and match the current designs.

Mar 26 11-35-11

### DIFF
--- a/draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
+++ b/draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
@@ -26,7 +26,6 @@ $left-content__width: 221px;
   display: flex;
   border-radius: $kz-border-solid-border-radius 0 0
     $kz-border-solid-border-radius;
-  background-color: $kz-color-wisteria-700;
   color: $kz-color-white;
   padding: ($ca-grid / 2) ($ca-grid * 2);
   flex-direction: column;
@@ -38,6 +37,22 @@ $left-content__width: 221px;
   .fullWidth & {
     width: 33%;
     max-width: 450px;
+  }
+
+  &.wisteria700 {
+    background-color: $kz-color-wisteria-700;
+  }
+
+  &.wisteria200 {
+    background-color: $kz-color-wisteria-200;
+  }
+
+  &.cluny200 {
+    background-color: $kz-color-cluny-200;
+  }
+
+  &.seedling200 {
+    background-color: $kz-color-seedling-200;
   }
 }
 
--- a/draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.tsx
+++ b/draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.tsx
@@ -3,6 +3,12 @@ import * as React from "react"
 
 import styles from "./HeroCard.scss"
 
+type BackgroundColors =
+  | "wisteria700"
+  | "wisteria200"
+  | "cluny200"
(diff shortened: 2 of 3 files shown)


draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
```diff
@@ -39,6 +38,26 @@ $left-content__width: 221px;
...
+    background-color: $kz-color-cluny-200;
+  }
+
+  &.seedling200 {
+    background-color: $kz-color-seedling-200;
+  }
+
+  &.transparent {
+    background-color: transparent;
+  }
```
makenosound, line 60: This one seems to not be connected?

makenosound (commented):
Asked one question inline, @talbet.

The rest of the code looks fine, and I think we can merge this. There’s something that feels slightly weird to me about allow-listing a set of colours here when the set is semi-random (aligned with a specific design but not necessarily a considered set of principles at the product level). I don’t have a better suggestion for the moment though, and I think anything more flexible is likely to make it harder to keep track of which colours are being used and handle things like the current rebrand.

lijuenc:
> weird to me about allow-listing a set of colours here when the set is semi-random

Yeah I thought the same. Typically we'd have a set of "variants" that have some semantic meaning defined by the product design but I suppose we don't have that yet for a HeroCard.

Examples:

https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/badge/KaizenDraft/Badge/Badge.tsx#L10 

https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/tag/KaizenDraft/Tag/Tag.tsx#L10

draft-packages/hero-card/KaizenDraft/HeroCard/HeroCard.scss
talbet: good catch, this is a hold over from my trial with transparency, I'll remove.

talbet:
@lijuenc @makenosound I hear you both, not convinced this is a good path forward myself. "variants" doesn't make much sense here, as there is no semantic meaning to the colour changes (green is not positive in this case). Adding every colour is not a great option, since they won't be used, and the foundations team would be in a better to create an abstraction if one were required.

I could duplicate this component in the micro-frontend I am implementing and extend it there if we think that would be a better, but still not perfect approach?

makenosound:
> I could duplicate this component in the micro-frontend I am implementing and extend it there if we think that would be a better, but still not perfect approach?

Heh, six of one, half a dozen of the other really. I feel like it’s probably better to have the imperfect API within the existing component since the blast radius is pretty small. Will allow us to change it centrally and force us to consider the references that use the specified colours and correct them.

lijuenc:
Yeah agree ☝️  
Also I believe this should be a `feat` rather than `fix` as it's a feature addition and would trigger a minor version release (0.1.0) rather a patch version release (0.0.1).

lijuenc (approved):

PR #1345, opened 2021-03-30
feat: Add a bootstrap component for Kaizen-specific styles

ActuallyACat:
# Objective
This package solves a few problems: 
* We need a place for Kaizen-specific styles to reduce duplication between apps.
* The font-face definitions that all our components depend on is hidden away in component-library. If an app were to consume Kaizen components, but not consume the component-library package, their app will _not_ be loading the fonts (Inter/Greycliff) it needs to render correctly. 
* Maybe this could also be the place for reasonable CSS fallback, e.g ( `body { font-family: "Greycliff CF", sans-serif; }` 🤷‍♀️
* Reduce magic / improve clarity. Everything Kaizen needs to consistently render across apps is in this package.

Why is all of this in one package? 
* Because it _just works_. IMO this is a better experience than authoring multiple packages that an app needs to consumer to ensure their apps render Kaizen components correctly. With the addition of this component, that means there are two wrapping components users will need to consider for their app: Bootstrap and ThemeProvider.

## Questions
* Should this also include JS polyfills?
* Should component-level overrides live in this package, or elsewhere?

## TODO
* ~~Abstract ThemeManagement into Container~~
* ~~Fix the README~~

# Checklist

# Additional Considerations


packages/bootstrap/README.md
```diff
@@ -0,0 +1,25 @@
...
+  <Bootstrap>
+    /* your React app goes here */
+  </Bootstrap>
+);
+```
+
+If for whatever reason this is not possible, you can import the Sass file which will give you all the necessary @font-face definitions: 
+```
+// index.module.sass
+@import "~@kaizen/boostrap/fonts.module.css"
```
ActuallyACat, line 24: If it's not possible for a user to use the React wrapper (for whatever reason), should there be an alternate file that adds these definitions to the `body` so that a consumer just needs to import one Sass file? Something like...

```
// index.module.scss
@import "~@kaizen/boostrap/fonts.module.css"

body {
  display: flex;
  flex-direction: column;
  min-height: 100vh;
  background-color: $kz-color-stone;
}
```

makenosound:
Strong endorse for something along these lines. I think it’s be useful to have a discussion about what the boundaries of this should be and what problems we’re trying to solve. Feels to me like there are two-ish questions to ask:

1. What CSS/layout defaults to we need to provide to create a Kaizen-supported page within the product?
2. What other things should we provide alongside that to create a Kaizen-supported page within the product?

Feels like the latter should eventually encompass the former in some structure to make it easy to do the right thing. Can imagine something like this for example:

```ts
import { KaizenApp } from "@kaizen/react-app";

function MyCoolApp() {
  return (
    <KaizenApp>
       ...
    </KaizenApp>
  )
}
```

Where `KaizenApp` is some  mix like this:

```ts
import { ThemeProvider } from "@kaizen/theme";
import { Page } from "@kaizen/page-layout";

function KaizenApp({ children }) {
  return (
    <ThemeProvider>
      <Page>
        {children}
      </Page>
    </ThemeProvider>
  )
}
```

> The hardest part of this is to clearly differentiate responsibilities between this boilerplate and these polyfils https://github.com/cultureamp/frontend-ops/blob/master/packages/frontend-runtime/lib/index.ts

Feels to me like we could replace the CSS concerns in the `frontend-runtime` with a Kaizen-owned thing and that that would create a _clearer_ separation of concerns. Sucking in `normalize` to all our front-ends through the runtime feels smelly to me.

packages/bootstrap/README.md
```diff
@@ -0,0 +1,25 @@
+# @kaizen/bootstrap
+
+Bootstrap is a place for Kaizen-specific styles that cannot be contained within a component package (such as Button).
+
+Note that this package is _not_ for browser normalisations or polyfills. Instead refer to @cultureamp/frontend-runtime:
+https://github.com/cultureamp/frontend-ops/blob/master/packages/frontend-runtime/lib/index.ts
+
+This package exports two assets; a Sass file and a React component. For most use cases it is easiest to wrap your App in the React component:
```
ActuallyACat, line 8: Now that Elm is deprecated I'm going to drop the a Sass file recommendation, and steer people to use the React wrapper.

packages/bootstrap/README.md
ActuallyACat: Nope, bad idea. This flexibility is not required.

ActuallyACat:
Those are good questions. Here's me having a stab at them: 

> What CSS/layout defaults to we need to provide to create a Kaizen-supported page within the product?
I'm thinking there are two things here... Component level defaults (margin/padding resets, and box-sizing), and whole page defaults. I can take a stab at some sensible component defaults tomorrow.

> What other things should we provide alongside that to create a Kaizen-supported page within the product?
I like this `KaizenApp` approach - It feels right to abstract the theme management, and CSS defaults into this shell. I'd potentially expose the theme manager so folks can change themes if they need to. 

>  Sucking in normalize to all our front-ends through the runtime 
I didn't agree with this at first, but thinking about it further it seems like a good idea to bring it into Kaizen. We should be able to drop Kaizen into any web app (not just the new repos!) and it brings everything it needs to ensure it renders as consistently as possible. Unless there's a specific reason why this belongs in frontend-runtime I'm all for moving it in here.

makenosound:
> Sucking in normalize to all our front-ends through the runtime

Yeah, I feel like normalize is much more strongly aligned with Kaizen than the frontend-runtime. In my mind, the runtime should take care of any common runtime tooling that doesn’t impose side-effects. I know it’s all side-effects really, but things like normalize are side-effects that change _everything_ CSS related down the tree, whereas polyfilling `fetch` is essentially seamless.

packages/bootstrap/lib/Wrapper.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+// Polyfill for :focus-visible pseudo-selector
+// See: https://github.com/WICG/focus-visible
+import "focus-visible"
+
+// Standard base stylesheet used across Culture Amp products
+// See: https://github.com/necolas/normalize.css/
+import "normalize.css"
+
+export default ({ children }: { children: React.ReactNode }) => (
```
ActuallyACat, line 13: This probably seems redundant, but it leaves the door open in the future if we need to wrap components in some kind of DOM

packages/bootstrap/lib/globals.module.scss
```diff
@@ -0,0 +1,13 @@
...
+ * IMPORTANT
+ * Before adding styles to this class, think _very carefully_ about
+ * possible side effects this may have on apps. These styles will
+ * be applied to all apps that consume Kaizen
+ */
+
+/** 
+ * Component-level overrides 
+ */
+[data-kaizen-component] {
```
ActuallyACat, line 11: This feels very smelly to me, and I'm not sure where I am landing w.r.t convenience and flexibility. Another thing to note is this has the same level of specificity as class selectors, so to consistently override these defaults the components wrapper override will have to bump specificity.

The alternative here is to expose a sass mixin in either this package, or another new one, and have all components depend on and consume that mixin. Components will have more control if they need to opt out. 

WDYT?

makenosound: Yeah, I think that’s OK though. I wonder if we could choose a name that’s more indicative of its purpose so that components can _feel_ like they can reasonably opt-out. Maybe something like:

```
data-kaizen-component-defaults
```

packages/bootstrap/lib/globals.module.scss
```diff
@@ -0,0 +1,13 @@
...
+  box-sizing: border-box;
```
makenosound, line 12: For this specific case, do we need to opt all descendants into this rule too?

packages/bootstrap/lib/Wrapper.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+// Global styles
+import "./globals.module.scss"
+
+// Supported / active face declarations
+import "./fonts.module.scss"
+
+// Initialise the ThemeManager, this is a singleton so this should not be redeclared
+export const themeManager = new ThemeManager(getCurrentTheme())
+themeSwitcher({ themeManager })
```
jasononeil, line 21: I'd still prefer to not include this here. Nothing else in Kaizen expects to be able to call a Culture Amp Platform API, and I don't really want to introduce that.

It'll be weird in places like Storybook (or our Gatsby site for that matter)

packages/bootstrap/lib/globals.module.scss
ActuallyACat: Yeah, 🤔 I would hope that if an engineer is opting out of these component defaults then they should set reasonable defaults that work for them to stop this cascade. That's not really clear from this though. For clarity this feels like the expected behaviour to me:
https://codepen.io/actuallyacat/pen/abJwGZR?editors=1100

packages/bootstrap/lib/Wrapper.tsx
ActuallyACat: Time for `frontend-services` - let's pair on this? If I don't have a better place to put the theme management my preference is to leave it here to avoid double handling of work and comms when this component rolls out.

jasononeil: I'm happy to pair on it asap, yeah.

packages/bootstrap/lib/globals.module.scss
jasononeil: Hey @makenosound, 

If we're adding an attribute to tag components like this, I'd love to keep the door open for using defo to collect info on components as they're added/removed from the page so we can start to get some analytics on component usage.

Would a goal like that affect the attribute name we pick?

makenosound: A tiny bit. It would need to be a `data-` attribute but apart from that we’re all good. Probably makes sense to do something like `data-kz-*` to establish a Kaizen namespace essentially.

.buildkite/pipeline.yaml
```diff
@@ -2,6 +2,7 @@ env:
   KAIZEN_ROLE_ARN: ${BRANCH_ROLE_ARN}
   KAIZEN_DOMAIN_NAME: ${BRANCH_DOMAIN_NAME}
   KAIZEN_DISTRIBUTION_ID: ${BRANCH_DISTRIBUTION_ID}
+  GITHUB_REGISTRY_TOKEN: ${GITHUB_REGISTRY_TOKEN}
```
jasononeil, line 5: TODO: If we're not using theme-switcher, we might not need these anymore (though we might want them in future?)

packages/bootstrap/lib/Wrapper.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default ({ children }: { children?: React.ReactNode }) => <>{children}</>
```
jasononeil, line 23: Should we put a `ThemeProvider` here?

packages/bootstrap/README.md
```diff
@@ -0,0 +1,20 @@
+# @kaizen/bootstrap
```
jasononeil, line 1: Naming thoughts

* Clash with another famous "bootstrap"
* `KaizenContainer`?
* `Kontainer`?
* `KaizenContext`?
    * We could use React Context for interesting things, like a single resize observer at the root of the page
    * And a ThemeProvider?
* `KaizenProvider`?

packages/bootstrap/docs/Bootstrap.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+  title: "KaizenApp Container (React)",
+  parameters: {},
+  component: Container,
+}
+
+export const DefaultContainer = () => (
+  <Container>
+    <div>Wrap your app in container to ensure it is consistent</div>
```
jasononeil, line 12: Might be nice to use a `Header` component so we can demonstrate fonts loading?

jasononeil: It might be nice to have stories that don't include this as a default decorator, and one story with it explicitly added, and one without, to demonstrate font loading, CSS normalizing, etc

packages/bootstrap/docs/Bootstrap.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react"
+import Container from "../index"
+
+export default {
+  title: "KaizenApp Container (React)",
+  parameters: {},
+  component: Container,
```
jasononeil, line 7: - [ ] Update the naming with whatever we decide

storybook/preview.tsx
```diff
@@ -1,26 +1,14 @@
...
-  defaultTheme,
-  heartTheme,
-  ThemeProvider,
-  zenTheme,
-} from "@kaizen/design-tokens"
 import { addParameters } from "@storybook/react"
 import { addons } from "@storybook/addons"
+import KaizenContainer from "@kaizen/bootstrap"
 import { backgrounds } from "./backgrounds"
 import { themeManager, themeOfKey } from "./theme-switcher-addon/themeManager"
```
jasononeil, line 7: How are we going to share a `themeManager` between our KaizenProvider and the theme-switcher-addon? We could have the addon use the exported instance in `Wrapper.tsx`, but it does get messy

packages/bootstrap/lib/Wrapper.tsx
```diff
@@ -0,0 +1,23 @@
...
+import "normalize.css"
+
+// Global styles
+import "./globals.module.scss"
+
+// Supported / active face declarations
+import "./fonts.module.scss"
+
+// Initialise the ThemeManager, this is a singleton so this should not be redeclared
+export const themeManager = new ThemeManager(getCurrentTheme())
```
jasononeil, line 20: Should we split out the polyfills and global instances etc, from the React Wrapper?

If there's a haml page, or an Elm only app, without React, it'd be nice to have a simple file to import still.

(If you use the React wrapper, it could import that file for you)

jasononeil (dismissed):
I've left comments, once they're addressed I'm keen to see this merged.

We'll have to create an issue for various places to clean up other things that aren't needed once this exists (like removing font declarations from other packages)

.buildkite/pipeline.yaml
ActuallyACat: Agree(ish). The way we're heading we may include private packages in this repo in the future. I'll rip this out of this PR as we don't need it here anymore.

packages/bootstrap/README.md
ActuallyACat: I've changed to `Container` (for now). Naming things is hard.

package.json
```diff
@@ -87,15 +87,13 @@
...
     "eslint-plugin-sort-imports-es6-autofix": "^0.5.0",
-    "focus-visible": "^5.2.0",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.6.3",
     "jest-circus": "^24.9.0",
     "jest-static-stubs": "^0.0.1",
     "lerna": "^3.20.2",
     "lodash": "^4.17.21",
     "node-sass": "^4.12.0",
-    "normalize.css": "^8.0.1",
```
ActuallyACat, line 98: These aren't needed in the workspace anymore, since the Container package encapsulates these.

storybook/preview.tsx
ActuallyACat: I agree that it is getting messy. 

We might want to rethink the responsibilities of the design-tokens package. Having theme management intertwined with the tokens means consumers need to use the design tokens directly to bootstrap their app. This isn't really compatible with the wrapper - we'd be exporting the ThemeManager and telling consumers _not_ to use it, and instead use the ThemeManager from Container only to avoid multiple declarations.

My current thoughts are the design tokens _should_ be for token definitions and compilations into different formats only (such as json to less/sass) and to provide these in their raw format (i.e no sass or React helpers) With that in mind, it makes sense to move the theme management responsibilities to the Container package. 

So we'd end up with something like this....

```
import KaizenContainer, { themeManager } from "@kaizen/container"
import App from "./App"

export const MyCoolApp = {
   themeSwitcher({ themeManager })

   return (
      <KaizenContainer>
        <App />
      </KaizenContainer>
   )
}
```

Then the implementation of ThemeSwitcher can be updated to consume from kaizen/container. 

WDYT?

mbylstra:
FWIW I just ran into a box-sizing expectation issue. I ran into this with Empty State which would create a horizontal scrollbar if you don't have `* { box-sizing: border-box }`, but we don't have that in our micro frontend (it's not super clear if we should or not). It's the opposite issue to the one Oli had in the Slack post, but it's the same issue in that Kaizen either expects you to have (or in some cases expects you not to have) a global box sizing override. I'm scared to add it just to fix the Empty State issue, because I don't know it it will break other things.

TLDR - I highly support moving this along if there's anything I can help with! (perhaps next Kaizen health day?)

ActuallyACat:
> TLDR - I highly support moving this along if there's anything I can help with! (perhaps next Kaizen health day?)

:D That would be great - I had to deprioritise this to get the animation work finished. I feel like it's _pretty close_ and something we could smash out next health day.

mbylstra:
This PR is looking great. I've been thinking about global styles recently (as part of https://github.com/cultureamp/kaizen-design-system/issues/1723, not that I think we should have a global `a` style, but that it's not even an option at the moment but that it's not even an option if we wanted to), and just discovered this PR!

ActuallyACat:
This is clearly still needed - reopened and added it to triage.

packages/container/README.md
```diff
@@ -0,0 +1,20 @@
+# @kaizen/container
+
+Container combines all the boilerplate and setup to ensure Kaizen components are consistent across different codebases. This includes:
+* Theme management
+* Browser normalisation 
+* Fonts
+* CSS Polyfills
```
jasononeil, line 7: Crazy idea: We could use this component to cancel out global styles.

Say performance-ui has this annoying code (it does):

```
figure {
  margin: 1em 40px;
}
```

We could have our container component include a stylesheet called:

```
.overrideGlobalStyles {
    figure {
        margin: initial;
    }
}
```

These rules would have higher specificity than our annoying global styles, and would "reset" them for any child components. Assuming all Kaizen stuff renders inside this container in future, we now have a single place to cancel our global styles.

We could then go through murmur and perf-ui, build a comprehensive list of global styles, and reset them all here, and be done with that class of bug.

packages/container/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+To use, wrap App in the React component: 
+
+```
+// App.tsx
+import KaizenContainer from "@kaizen/container";
```
jasononeil, line 13: We seem to often use named exports rather than default exports. Did we want to standardize on this?

packages/container/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "!**/*.ts",
+    "!**/*.tsx",
+    "**/*.d.ts",
+    "!**/*.spec.*",
+    "!**/*.snap",
+    "!docs",
+    "!tsconfig.dist.json"
+  ],
+  "peerDependencies": {
+    "@kaizen/design-tokens": "^2.5.0",
```
jasononeil, line 32: Let's bump this to v3 (or at least, `^2.x | ^3.x` or whatever it is)

packages/container/README.md
jasononeil: Should we also document the `themeManager` export?

packages/container/lib/fonts.module.scss
```diff
@@ -0,0 +1,102 @@
...
+
+// =============== TIEMPOS ===============
+@font-face {
+  font-family: "Tiempos Headline";
+  src: url(asset-url("fonts/tiempos/tiempos-headline-bold.woff2")),
+    url(asset-url("fonts/tiempos/tiempos-headline-bold.woff"));
+  font-weight: 800;
+}
+
+// =============== GREYCLIFF ===============
```
jasononeil, line 11: Is it safe to remove Greycliff fonts now?

Or should we wait until it's entirely purged from our codebase

packages/container/lib/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react"
+import { ThemeManager, ThemeProvider, heartTheme } from "@kaizen/design-tokens"
+import "./polyfills"
+
+// Initialise the ThemeManager, this is a singleton so this should not be redeclared
```
jasononeil, line 5: It would be nice to add a JSDoc comment here for editor niceness

storybook/preview.tsx
```diff
@@ -1,28 +1,14 @@
...
 } from "./theme-switcher-addon/constants"
-// Polyfill for :focus-visible pseudo-selector
-// See: https://github.com/WICG/focus-visible
-require("focus-visible")
-
-// Standard base stylesheet used across Culture Amp products
-// See: https://github.com/necolas/normalize.css/
-require("normalize.css")
-
-require("@kaizen/component-library/styles/fonts.scss")
```
jasononeil, line 25: Gosh the diff in this file makes me happy.

yarn.lock
```diff
@@ -1727,6 +1727,24 @@
     "@types/yargs" "^15.0.0"
     chalk "^4.0.0"
 
+"@kaizen/component-library@^9.6.1":
```
jasononeil, line 1730: Any idea why this appeared?

We probably don't want an outdated component-library version being used inside the monorepo

packages/container/lib/globals.module.scss
```diff
@@ -0,0 +1,13 @@
+/**
+ * IMPORTANT
```
jasononeil, line 2: This warning is good, but I'm not super clear on what the `data-kz-component-defaults` is or where it should be used.

Should we be adding it to all our components? (And if so, should we do that in this PR?)

Should it be on the KaizenContainer only?

Is it strictly for internal use only?

Some guidance on when you should (and shouldn't) use this attribute would be nice.

packages/container/lib/polyfills.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Standard base stylesheet used across Culture Amp products
+// See: https://github.com/necolas/normalize.css/
+import "normalize.css"
+
+// Global styles
+import { } from "./globals.module.scss"
+
+// Supported / active face declarations
+import { } from "./fonts.module.scss"
```
jasononeil, line 13: Nitpick: I feel like these two should maybe be imported from the main `index.tsx` file, since these aren't really polyfills.

storybook/theme-switcher-addon/themeManager.ts
```diff
@@ -13,10 +18,11 @@ export const themeOfKey = (themeKey: string) => {
 export const getInitialTheme = () =>
   themeOfKey(
     window.location.search.match(/(\?|\&)theme=(zen|heart)/)?.[2] ||
-      localStorage.getItem(THEME_KEY_STORE_KEY) ||
-      "heart"
+    localStorage.getItem(THEME_KEY_STORE_KEY) ||
+    "heart"
     // ^This should ideally be `defaultTheme.themeKey` (defaultTheme imported from kaizen/design-tokens)
     // but this has been manually overridden to "heart" to avoid having to wait for that change (which has other factors blocking it)
```
jasononeil, line 24:
```suggestion
    defaultTheme.themeKey
```

storybook/preview.tsx
```diff
@@ -1,28 +1,14 @@
...
 import React from "react"
-import {
-  defaultTheme,
-  heartTheme,
-  ThemeProvider,
-  zenTheme,
-} from "@kaizen/design-tokens"
 import { addParameters } from "@storybook/react"
 import { addons } from "@storybook/addons"
+import KaizenContainer, { themeManager } from "@kaizen/container"
```
jasononeil, line 5: I'm not sure where the bug is, but in the storybook branch preview I'm getting:

```
Uncaught TypeError: can't access property "themeKey", _this.theme is undefined
```

packages/container/lib/globals.module.scss
jasononeil: This probably ties into @makenosound's comment about being clear on what the boundaries of this component are.

What should the KaizenContainer do? What should be included here, and what needs to find a different home? Putting that info up front in the README would be good

jasononeil (commented):
I've done a fresh review of this.

It's so good! There's a few things to address (sorry...) but we should get this over the line asap :)

packages/container/lib/fonts.module.scss
ActuallyACat: Yeah, given everyone is on Zen this should be removed.

packages/container/README.md
ActuallyACat: Yeah, good call. The defacto standard in Kaizen is named exports, I can't recall why I made this a default export at the time.

packages/container/lib/globals.module.scss
ActuallyACat: To be honest, I don't think I settled on that myself. We clearly need both global and component-level overrides. I think this PR solves global normalization and overrides well. 

I'd also like to solve component level overrides, but maybe this is not the right place for that. Ideally, there should be some kind of style utility for component-level defaults that would give the components the ability to opt-in, modify, extend the default component override as needed. However, we actively discourage consumers from extending Sass so I think a Sass class that consumers extend is a tacit endorsement of that approach. 

Admittedly, I think this approach is also kind of sucks.

ActuallyACat: Decided to remove this completely. Component-level overrides should be done at the component-level.

packages/container/README.md
ActuallyACat: This would definitely resolve this scope of issue for us in the long term, however I'm quite nervous about rolling this out in the first iteration as it will likely cause some unintended regressions.

jasononeil: I'd be keen to just get this merged.

If we did it we could also scope it with some props:

```
<Container globalStyleResets="murmur" />
```

... which makes me wonder if we can't just let the consumers (perf-ui and murmur are the only culprits) handle this in a custom component in their own codebase.

```
<Container>
   <MurmurGlobalStyleReset>
        {...}
    </MurmurGlobalStyleReset>
</Container>
```

jasononeil: Then the problems become entirely theirs to handle

packages/container/index.module.scss
```diff
@@ -0,0 +1 @@
+@import "./lib/fonts.module.scss";
```
jasononeil, line 1: Is this file being used? (It doesn't look like it to me?)

packages/container/README.md
```diff
@@ -0,0 +1,40 @@
...
+// App.tsx
+import { Container } from "@kaizen/container"; 
+
+const YourApp = () => (
+  <Container>
+    /* your React app goes here */
+  </Container>
+);
+```
+
```
jasononeil, line 28: I had a thought that we should document usage on non-react pages:

````suggestion
### Usage
To use, wrap App in the React component: 

```
// App.tsx
import { Container } from "@kaizen/container"; 

const YourApp = () => (
  <Container>
    /* your React app goes here */
  </Container>
);
```

You can also use this on non-react pages to initialise the theme manager and get the required font definitions, polyfills etc:

```
import "@kaizen/container"; 
```
````

jasononeil (dismissed):

jasononeil (approved):

PR #1398, opened 2021-04-09
fix: export elm select field clearable

mbylstra:
Export the `clearable` config option of the Elm SelectField so it can actually be used!


cset00 (approved):
:D

PR #1426, opened 2021-04-13
feat: adjust Collapsible component styles and add prop to disable background in header

talbet:
# Objective

This brings the Collapsible component closer in line with the branding guidelines and uses styles closer to those used in the Card component.

I have also added a new prop to control the background colour of the collapsible header

# Motivation and Context

This may not have been done previously because curved corners using border-radius and sticky headers tend not to play well. There is a solution for this in the new `overflow: clip` property that is being implemented in modern browsers.

As a fallback, we can still use `overflow: visible` for older browsers, this has the side effect of children without border radius overlapping their parents, but this is a minor visual glitch that many people won't notice.

Talking to Lindsey, we agreed that this can be considered progressive enhancement. We will prioritize correct behavior (sticky headers) over correct appearance (border radius).

Re: header background

While we can provide custom header JSX to the Collapsible component, we cannot determine if the header should be highlighted when open. One of the designs we are implementing for the survey summary page requires that the highlighting be disabled, so I have added that option.

The highlighting on hover is always enabled as this signals that the header is interactive.Ï

### Browser with `overflow: clip` support

Screenshot at Apr 13 13-40-26

### Browser __without__ `overflow: clip` support

Screenshot at Apr 13 13-40-06

# Checklist


dougmacknz (approved):
Awesome stuff, thanks for being so thorough with the sticky header things!

PR #1450, opened 2021-04-18
feat: avatar won't render initials if prop is passed

sampnorris:
# Objective

I have added an optional prop to disable the initials generation for use cases where this doesn't make sense.

This also fixes a bug where if the avatarSrc changes from nothing to something, then the component was not trying to fetch the new src.

# Motivation and Context

Sometimes, a users name is numeric, so we want an option to disable initials when implemented for cases such as this..

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
+++ b/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
@@ -31,8 +31,9 @@ export interface AvatarProps {
   /**
    * Default behaviour when an avatarSrc is not provided is to generate initials from the username.
    * This disables this feature and shows the generic avatar.
+   * Enable this prop when there is no specific individual or group identified.
    */
-  disableInitials?: string
+  disableInitials?: boolean
   /**
    * Src for the avatar image to load, if not passed we will derive initials from the full name.
    */
--- a/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
+++ b/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
@@ -30,7 +30,7 @@ export interface AvatarProps {
   fullName: string
   /**
    * Default behaviour when an avatarSrc is not provided is to generate initials from the username.
-   * This disables this feature and shows the generica avatar.
+   * This disables this feature and shows the generic avatar.
    */
   disableInitials?: string
   /**
--- a/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
+++ b/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
@@ -29,6 +29,11 @@ export interface AvatarProps {
    */
   fullName: string
   /**
+   * Default behaviour when an avatarSrc is not provided is to generate initials from the username.
+   * This disables this feature and shows the generica avatar.
+   */
+  disableInitials?: string
+  /**
    * Src for the avatar image to load, if not passed we will derive initials from the full name.
    */
   avatarSrc?: string
@@ -47,6 +52,7 @@ export interface AvatarProps {
 export const Avatar = ({
   fullName,
   avatarSrc,
+  disableInitials,
   size = "medium",
   isCurrentUser = true,
 }: AvatarProps) => {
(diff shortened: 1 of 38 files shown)


draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -54,10 +54,21 @@ export const Avatar = ({
     "none" | "error" | "loading" | "success"
   >(avatarSrc ? "loading" : "none")
 
+  useEffect(() => {
+    setAvatarState(avatarSrc ? "loading" : "none")
+  }, [avatarSrc])
+
+  const nameHasAlpha = fullName && /[a-zA-Z]+$/.test(fullName)
```
sebpearce, line 61: I think we need to be really careful here. This pattern will fail on names like `Étienne` or `Nguyễn` which contain diacritic characters.

sebpearce:
I'm curious about the need for this extra complexity. What prompted the PR? I see "This is required as we are rolling out the unified navigation bar across the whole platform" but not sure why we would be passing in numeric "names" into avatars.

> This also fixes a bug where if the avatarSrc changes from nothing to something, then the component was not trying to fetch the new src.

As a separate issue this looks like an important fix, though.

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
sebpearce: OK wait, I was too hasty 😬  I assumed every character had to match, but you're only looking for the presence of at least one alphabetic char. Oops

sampnorris: Yea, had a think about the logic,
It does add some complexity, but yes regarding your comment below. The prompt for this was, sometimes in our platform, users don't have names, they have numbers, but they live in the name field. So we either, do a check for a numeric name _outside_ this component, and just not pass in anything, which adds complexity to the implementation, or we just makes sure there's at least an alpha character, which would make implementation easier.

Open to your Kaizen-minded suggestion. This really only covers us for fully numeric strings (which is the case I am fixing for), and not a combination of the 2, but I'm not sure that's a case we need to cover for now.

That said, an approach that isn't tied to 'Culture Amp' as a platform would make sense. 
Maybe just a prop called `disableInitials`, and maybe just throw in the check upon implementation? Keen to discuss.

Prompt for this work is here: https://cultureamp.atlassian.net/browse/TU-342

sampnorris: @sebpearce  - I have had a think, and reworked this to just take a disableInitials prop, and will do the check for the numeric name upon implementation.

This makes more sense from a design system point of view, and will give us more control / less magic.

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -28,6 +28,11 @@ export interface AvatarProps {
    * We use this for the alt text of the avatar, and to derive intials when user has no avatar image.
    */
   fullName: string
+  /**
+   * Default behaviour when an avatarSrc is not provided is to generate initials from the username.
+   * This disables this feature and shows the generica avatar.
```
sebpearce, line 33:
```suggestion
   * This disables this feature and shows the generics avatar.
```

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
sebpearce: Ah nice one, I didn't think of that and it feels a lot more Kaizen-y. 👌

@ActuallyACat CCing you here since this is an API change you may want to be across.

sebpearce: @ActuallyACat Could we confirm that you are OK with this API change?

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -28,6 +28,11 @@ export interface AvatarProps {
    * We use this for the alt text of the avatar, and to derive intials when user has no avatar image.
    */
   fullName: string
+  /**
+   * Default behaviour when an avatarSrc is not provided is to generate initials from the username.
+   * This disables this feature and shows the generic avatar.
+   */
+  disableInitials?: string
```
ActuallyACat, line 35: /nitpick I think we can do a little more to inform when the user should use this option. Here's an excerpt from the docs - "When there is no specific individual or group identified, use the default user avatar."

sampnorris: I've updated the description to include this

ActuallyACat (approved):

PR #1473, opened 2021-04-22
feat: Added Heart spot illustrations

ActuallyACat:
# Objective
Add the Heart spot illustrations, and conditionally render them based on a theme.

# Other notes
Few things to call out: 
* I have used snapshot tests. The removal of any of these illustrations will be a breaking change and we should guard against that in our tests.
* When there is a Heart illustration without a corresponding Zen illustration, a warning will be logged when using these under a Zen theme. I don't feel that we should stop someone from using a Heart illustration in Zen since Zen will eventually be deprecated.
* I've also stopped rendering each of these illustrations as a separate story and have instead made a mega-story that matches the UI kit sticker sheet. I hope that this makes it a little easier to find the illustration needed.
* You're probably best to review this in Chromatic. 

Story is here: https://dev.cultureamp.design/ally/heart-spot-illustrations/storybook/?path=/story/illustration-spot-react--all-spot-illustrations

Screen Shot 2021-04-23 at 11 17 39 am

# Checklist


dougmacknz (approved):
Awesome, looking good to me

PR #1506, opened 2021-04-27
feat: Deprecated tokens stylelint plugin and codemod

dmisdm:
# Objective
We want people to be able to migrate to the new Kaizen tokens without much pain.
Blindly renaming the old -> new unfortunately doesn't work, because of the change from plain old values `#ff0022` to CSS variables `var(--kz-var-color-blah, #ff0022)`.
Some of the reasons why:
- You can't run SASS/LESS color manipulations on CSS variables
- You can't use them in `rgba|rgb` functions, or in `add-alpha`
- You can't use them in equations e.g. adding 5px to `$kz-var-spacing-md`
- You can't negate them e.g. `-$kz-var-spacing-lg`

Further background into the project: https://cultureamp.atlassian.net/wiki/spaces/CPlatform/pages/2182218300/Migrating+CultureAmp+s+repositories+to+CSS+variable+Kaizen+design-tokens.

# Proposed solution
A stylelint plugin, postcss plugin, and standalone CLI, to parse SASS and LESS and report on any instances of old kaizen variables that need to be migrated, and any instances where new variables are used incorrectly that might result in a silent failure or a SASS/LESS compilation error. Also, be able autofix many of these reported issues.

The plugin is split into multiple rules:
- `kaizen/no-deprecated-tokens` - Detects usages of deprecated tokens, with the ability to automatically migrate many instances.
- `kaizen/no-invalid-functions` - Detects invalid usages of Kaizen tokens within some functions that are known to be problematic such as `rgba`, `darken` or `mix`, and automatically fixes `rgba`, `rgb`, `add-alpha` instances by ensuring the correct `-rgb-params` tokens are used, OR by replacing unfixable color manipulation functions into their computed values (see [Sass Manipulate Color Functions] (https://www.w3schools.com/sass/sass_functions_color.asp))
- `kaizen/no-invalid-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a `calc()` function, which is going to be unsupported in the future due to CSS variables.
- `kaizen/no-transitive-tokens` - Prevent and fix usages of local non-Kaizen variables (only ones that are defined on the current sheet, ignoring any that are imported) that contain Kaizen tokens in their values. For example, `$card-bg: $kz-color-wisteria-800; .card { background-color: $card-bg; }` would be disallowed, and would be autofixed to `background-color: $kz-color-wisteria-800;`
- `kaizen/imports-no-unused` - Reports and fixes unused imports that relate to Kaizen tokens
- `kaizen/imports-no-extraneous` - Reports and fixes any imports that should exist but don't, based on the tokens that are used in the current stylesheet.

# Checklist

# Additional Considerations


packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
+# Kaizen Stylelint Plugin
+
+## Tl;dr
+
+`yarn add @kaizen/stylelint-plugin`
+
+Then, add the plugin and the rules you want to your stylelint configuration:
```
mbylstra, line 7: Hi, I left a comment in #team_foundations (https://cultureamp.slack.com/archives/C016N8Y1E10/p1619501865363200), but I was wondering if you'd consider exporting a base config, so that consumers just need to extend a base config, rather than opt in to each rule individually.

dmisdm: replying in slack :)

ActuallyACat (approved):

PR #1520, opened 2021-04-30
feat: Stylelint plugin and codemod (for updating deprecated tokens etc)

dmisdm:
This PR Introduce a Kaizen Stylelint plugin that we can use to perform linting, and codemodding through the linter's "autofix" feature.

It includes an initial set of rules to help with the token changes that are part of our change to the new Heart theme.

---

# Objective
We want people to be able to migrate to the new Kaizen tokens without much pain.
Blindly renaming the old -> new unfortunately doesn't work, because of the change from plain old values `#ff0022` to CSS variables `var(--kz-var-color-blah, #ff0022)`.
Some of the reasons why:
- You can't run SASS/LESS color manipulations on CSS variables
- You can't use them in `rgba|rgb` functions, or in `add-alpha`
- You can't use them in equations e.g. adding 5px to `$kz-var-spacing-md`
- You can't negate them e.g. `-$kz-var-spacing-lg`

Further background into the project: https://cultureamp.atlassian.net/wiki/spaces/CPlatform/pages/2182218300/Migrating+CultureAmp+s+repositories+to+CSS+variable+Kaizen+design-tokens.

# Proposed solution
A stylelint plugin, postcss plugin, and standalone CLI, to parse SASS and LESS and report on any instances of old kaizen variables that need to be migrated, and any instances where new variables are used incorrectly that might result in a silent failure or a SASS/LESS compilation error. Also, be able autofix many of these reported issues.

The plugin is split into multiple rules:
- `kaizen/no-deprecated-tokens` - Detects usages of deprecated tokens, with the ability to automatically migrate many instances.
- `kaizen/no-invalid-functions` - Detects invalid usages of Kaizen tokens within some functions that are known to be problematic such as `rgba`, `darken` or `mix`, and automatically fixes `rgba`, `rgb`, `add-alpha` instances by ensuring the correct `-rgb-params` tokens are used, OR by replacing unfixable color manipulation functions into their computed values (see [Sass Manipulate Color Functions] (https://www.w3schools.com/sass/sass_functions_color.asp))
- `kaizen/no-invalid-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a `calc()` function, which is going to be unsupported in the future due to CSS variables.
- `kaizen/no-transitive-tokens` - Prevent and fix usages of local non-Kaizen variables (only ones that are defined on the current sheet, ignoring any that are imported) that contain Kaizen tokens in their values. For example, `$card-bg: $kz-color-wisteria-800; .card { background-color: $card-bg; }` would be disallowed, and would be autofixed to `background-color: $kz-color-wisteria-800;`
- `kaizen/imports-no-unused` - Reports and fixes unused imports that relate to Kaizen tokens
- `kaizen/imports-no-extraneous` - Reports and fixes any imports that should exist but don't, based on the tokens that are used in the current stylesheet.

# Checklist

# Additional Considerations


dmisdm:
One comment left in previous PR
https://github.com/cultureamp/kaizen-design-system/pull/1506#discussion_r620884987

jest.config.js
```diff
@@ -4,7 +4,8 @@ module.exports = {
   testMatch: ["**/*.spec.ts?(x)"],
   moduleNameMapper: {
     "\\.(jpe?g|png)$": "jest-static-stubs/$1",
-    "\\.s?css$": "identity-obj-proxy",
+    "\\.s?css$(?<!((packages|@kaizen)/design-tokens/(sass|less|css)/.*))":
+      "identity-obj-proxy",
```
jasononeil, line 8: I've got to admit I don't understand what this is doing. Can this be split out into multiple simpler rules, or can you add a comment explaining it?

packages/stylelint-plugin/src/cssStandardFunctions.ts
```diff
@@ -0,0 +1,111 @@
+/*
+    A set of all function names that are part of the CSS specification.
+    Because they arent (shouldn't be) evaluated at compile-time (apart from some SASS overloaded ones like rgb, rgba), they should support CSS variables within their parameters.
+    Source: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions
+    To generate this list again, go to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions and run this in your dev console:
+    `Array.from(document.querySelectorAll("#content > article > div:nth-child(22) > div li a")).map(n => n.textContent.replace("()", ""))`
+*/
+export const cssStandardFunctions = new Set([
```
ActuallyACat, line 8: Can we extend this from types that already exist?

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `kaizen/no-invalid-functions` - Detects invalid usages of Kaizen tokens within some functions that are known to be problematic such as `rgba`, `darken` or `mix`, and automatically fixes `rgba`, `rgb`, `add-alpha` instances by ensuring the correct `-rgb-params` tokens are used, OR by replacing unfixable color manipulation functions into their computed values (see [Sass Manipulate Color Functions] (https://www.w3schools.com/sass/sass_functions_color.asp))
+- `kaizen/no-invalid-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a `calc()` function, which is going to be unsupported in the future due to CSS variables.
+- `kaizen/no-transitive-tokens` - Prevent and fix usages of local non-Kaizen variables (only ones that are defined on the current sheet, ignoring any that are imported) that contain Kaizen tokens in their values. For example, `$card-bg: $kz-color-wisteria-800; .card { background-color: $card-bg; }` would be disallowed, and would be autofixed to `background-color: $kz-color-wisteria-800;`
```
jasononeil, line 46: How desirable is this behaviour? I can see people using local variables as "decision tokens".

In the example you have setting `$card-bg: $kz-color-wisteria-800` makes it really clear that's a design decision, and if the `$card-bg` decision token is used in multiple places it makes it easy to update to the decision on what card background colours are (compared to a find/replace of the wisteria-800 option token, which might be used for things other than card backgrounds too)

jasononeil: (Apologies if this was previously discussed)

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+In the example above, `severity` is used, however it is inbuilt into stylelint. You can set this to "warning" or "error" to treat reports of anything as ok or fatal.
+
+## Features
+- Detect and report on usages of old/deprecated kaizen tokens
+- Doesn't auto-migrate anything that can't be safely migrated, e.g. variable declarations, color manipulation functions, or equations.
+  - However, it WILL migrate any variables that are used incorrectly within a `calc()` function
+- Migrates `rgba|rgb|add-alpha` usages.
+- Warns you when you use a new kaizen token incorrectly (most of the time)
+- Coming soon: any usages of darken, add-shade, add-tint, mix etc will be computed within the codemod and replaced with a HEX/RGB/HSL value.
```
jasononeil, line 77: Is this actually coming soon? :P

dmisdm: Lol should take this line out - it's been implemented in no-invalid-functions :)

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
+# Kaizen Stylelint Plugin
+
+## Tl;dr
```
jasononeil, line 3: Nitpick: I know _most_ people know what TL;DR is by now, but I feel "getting started" is just as good:

```suggestion
## Getting Started
```

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `--fix` tells it to fix anything it can, and thus not report on anything it has fixed.
+
+
+
+## Options:
+- `allowFixing: boolean` - Allows you to run stylelint fixes using the codemod, even though there may be incosistent results. Defaults to false.
+
+In the example above, `severity` is used, however it is inbuilt into stylelint. You can set this to "warning" or "error" to treat reports of anything as ok or fatal.
+
+## Features
```
jasononeil, line 71: I feel like this is burying the lede... We could move this up next to "Getting Started" to highlight the features earlier

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,77 @@
+# Kaizen Stylelint Plugin
```
jasononeil, line 1: Great README! I love this effort you've put into writing docs 😍

packages/stylelint-plugin/src/kaizenTokens.ts
```diff
@@ -0,0 +1,160 @@
...
+import { getParser } from "./utils"
+import { isVariable } from "./variableUtils"
+const scssParser = getParser("scss")
+
+/* Pass in just the name of a module which is used to import variable.
+  E.g. "color" or "color-vars", NOT "@kaizen/design-tokens/sass/color".
+  Assumes that the SASS and LESS modules contain the same variables
+*/
+const getVarsFromKaizenModule = (moduleName: string) => {
+  const sassModulePath = `@kaizen/design-tokens/sass/${moduleName}`
```
ActuallyACat, line 14: Potentially import the TS files from design-tokens since that is treated as the source of truth for token generation.

packages/stylelint-plugin/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "types": "dist/**/*.d.ts",
+  "main": "dist/src/index.js",
+  "license": "MIT",
+  "files": [
+    "src",
+    "dist",
+    "types"
+  ],
+  "bin": {
+    "kaizen-token-codemod": "./dist/src/cli.js"
```
jasononeil, line 13: Do you think we'll ever want to distinguish stylelint codemods from eslint (or similar) codemods? Do you think this name is still best in a world where other types of codemods exist?

(I think it's probably fine, but wanted to ask the question and see if you think of anything)

packages/stylelint-plugin/README.md
dmisdm: It's absolutely a tricky one.
It massively helped in the migration of murmur though, because doing those replacements allowed the other rules to run.
Apart from murmur, i haven't been using it. It might not be that useful outside of migrations.

packages/stylelint-plugin/types/stylelint.d.ts
```diff
@@ -0,0 +1,187 @@
+// Type definitions for stylelint 9.10
+// Project: https://github.com/stylelint/stylelint, https://stylelint.io
+// Definitions by: Alan Agius <https://github.com/alan-agius4>
+//                 Filips Alpe <https://github.com/filipsalpe>
+//                 James Garbutt <https://github.com/43081j>
+//                 Bob Matcuk <https://github.com/bmatcuk>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
```
jasononeil, line 8: Is there a reason you're not using the package from definitely-typed?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/stylelint/index.d.ts

jasononeil: (If there is a reason, it would be good to include a comment explaining why and any local changes you've made)

jest.config.js
dmisdm: This was originally saying: "ignore `@kaizen/design-tokens/(sass|less|css)` when mocking SCSS imports in jest"
Originally, if you tried to `require.resolve("@kaizen/design-tokens/sass/color.scss")` you would get `identity-obj-proxy` when executing in Jest, but we wanted the actual SCSS.

I just refactored the code that required this to not require the SCSS files and use the JSON tokens instead, so i reverted it.

packages/stylelint-plugin/package.json
dmisdm: In case you see this, I ended up un-exposing the CLI (keeping the program just in case we change our mind).
It was originally intended to be used because of some quirks of stylelint, which we're resolved (LESS and SCSS variables are represented differently in the AST when run with stylelint).

I think having both the CLI and stylelint plugin is unnecessary, for now at least.

packages/stylelint-plugin/src/cssStandardFunctions.ts
dmisdm: @ActuallyACat do you mean an external library?
I haven't been able to find an alternative yet, anything in mind?

packages/stylelint-plugin/src/rules/no-invalid-functions.ts
```diff
@@ -0,0 +1,360 @@
...
+        autofixAvailable: !isVariable(decl),
+      })
+    }
+  }
+}
+
+/**
+ * This function will compile a value, for example `mix($kz-color-wisteria-800, $white, 80%)` in place (which would result in #5d5f6e).
+ * It also uses the variables that are immediately available within the same stylesheet, so nothing too fancy, and will bail out with a lint error if it can't compile.
+ * ALSO, it will work with $kz-var-* variables too.
```
jasononeil, line 144: I think I made this comment elsewhere, but I'm uncomfortable with this behaviour. I think leaving the static sass variables (that contain actual values, not references to css variables) is safer than compiling hard coded values.

packages/stylelint-plugin/src/rules/no-invalid-functions.ts
```diff
@@ -0,0 +1,360 @@
...
+ */
+export const noInvalidFunctionsRule = (
+  stylesheetNode: Root,
+  options: Options
+) => {
+  walkDeclsWithKaizenTokens(stylesheetNode, ({ postcssNode }) => {
+    noInvalidFunctionsOnDeclaration(postcssNode, options)
+  })
+}
+export const declContainsInvalidFunctions = (
```
jasononeil, line 347: Having this export here used in a separate rule makes me think it's in the wrong place and should be in a separate utility module. (Not a deal-breaker if you want to leave it for now in the interest of merging)

packages/stylelint-plugin/src/rules/no-invalid-equations.ts
```diff
@@ -0,0 +1,165 @@
...
+  negatedKaizenVariableMessage,
+  tokenNotInterpolatedInCalcMessage,
+} from "../messages"
+import { isOperator } from "../patterns"
+import { Options } from "../types"
+import { isVariable, parseVariable, stringifyVariable } from "../variableUtils"
+import { walkDeclsWithKaizenTokens, walkWithParent } from "../walkers"
+
+export const noInvalidEquationsRuleName = "no-invalid-equations"
+export const noInvalidEquationsOnDeclaraion = (
```
jasononeil, line 14: This doesn't appear to be used in other modules, so we probably shouldn't export it

packages/stylelint-plugin/src/rules/no-invalid-equations.ts
```diff
@@ -0,0 +1,165 @@
...
+  walkWithParent(parsed, (node, parent) => {
+    if (node.type === "function" || node.type === "div") {
+      // Reset the word sequence if there is something meaningful separating them, like a function or a divider (comma)
+      lastTwoWordNodesNextToEachOther = []
+      return
+    }
+
+    const isWithinCalcFunction = parent?.value === "calc"
+    if (node.type === "word") {
+      // Detect the cose of the node being a negated variable (because postcss-value-parser doesn't detect the operator and variable as seprate nodes)
```
jasononeil, line 43:
```suggestion
      // Detect the case of the node being a negated variable (because postcss-value-parser doesn't detect the operator and variable as seprate nodes)
```

packages/stylelint-plugin/src/rules/no-invalid-equations.ts
```diff
@@ -0,0 +1,165 @@
...
+export const noInvalidEquationsRule = (
+  stylesheetNode: Root,
+  options: Options
+) => {
+  walkDeclsWithKaizenTokens(stylesheetNode, ({ postcssNode, parsedValue }) => {
+    noInvalidEquationsOnDeclaraion(postcssNode, parsedValue, options)
+  })
+}
+
+export const declContainsInvalidEquations = (
```
jasononeil, line 151: Again, I feel this export from one rule, being used in the `no-deprecated-tokens` rule, is a bit of a code-smell. I don't think rules should be importing utility functions from each other

packages/stylelint-plugin/src/rules/no-deprecated-tokens.ts
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      const decl = postcssNode
+
+      /*
+        These next two blocks bail out if a token is used in an equation or in an unsupported function.
+        Admittedly it's a bit weird because there is a double up on reporting if you include the rules that relate to these predicate functions.
+        However, this is what determines if a deprecated token can't automatically be migrated.
+        We want to be able to use the same logic as the other rules, but also keep them as separate rules...
+      */
```
jasononeil, line 74: This is a useful comment 💯

jasononeil (changes requested):
This is a huge piece of work - well done!

I'm pretty happy with the overall structure, and the test suite on codemod.ts gives me a lot of confidence.

Other than a few nitpicks, my one big ask is to not compile down tokens to hex values... I'd rather deprecated tokens (without the CSS variables) than a raw hex value. And if we publish this and a bunch of tokens are converted to hex values, that's going to be hard to undo in future.

packages/stylelint-plugin/src/codemod.spec.ts
```diff
@@ -0,0 +1,523 @@
...
+    expectedOutput:
+      '@import "~@kaizen/design-tokens/sass/spacing-vars"; .foo { padding: rgba(+, $kz-var-spacing-md) }',
+    expectedUnfixables: 0,
+  },
+  {
+    language: "scss",
+    testName: "compiles mix() function in place",
+    input:
+      '@import "~@kaizen/design-tokens/sass/color"; $customVar: white; .foo { color: mix($kz-color-wisteria-800, $customVar, 80%) }',
+    expectedOutput: "$customVar: white; .foo { color: #5d5f6e }",
```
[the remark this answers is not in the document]
dmisdm: Hey this is a really good point.

Because we already ran this on murmur and perf-ui it is unfortunately a little too late.
We do have git history/annotations though that i assumed would be of help when going back to fix them.

The other thing is, when going through a bunch of these deprecated colors in Murmur with Marc, the only thing that he was really interested in is the context around the color and the semantics (e.g. "this looks like the background of a primary button, it should be X, and it's so old anyway so let's not stress too much about precision").

packages/stylelint-plugin/src/rules/no-invalid-functions.ts
dmisdm: Agree this is strange, i'm not sure what the best solution is... e.g.
- move just this to a separate file but still import from the no-invalid-functions rule module
- completely refactor the rule so that any "checking" logic is separate from the actual rule, i.e. the rule is a composition of a set of primitives/utilities (i like this the best)

Same goes for no-invalid-equations.

dmisdm: The 2nd point might be the way to do. In any case i think this is a job for a subsequent PR.

jasononeil: FWIW, I'm happy to merge without this refactor in the interest of getting it done before you go on leave

.eslintignore
```diff
@@ -1 +1,2 @@
-/packages/design-tokens/dist
\ No newline at end of file
+/packages/design-tokens/dist
+/packages/stylelint-plugin/dist
```
ActuallyACat, line 2: nit/ might be time to glob these:
```
packages/**/dist
```

dmisdm:
@jasononeil I've made a subsequent PR for the color compilation changes:
https://github.com/cultureamp/kaizen-design-system/pull/1624

packages/stylelint-plugin/src/codemod.spec.ts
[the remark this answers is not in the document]
dmisdm: Following up in this PR https://github.com/cultureamp/kaizen-design-system/pull/1624

packages/stylelint-plugin/src/rules/no-invalid-equations.ts
dmisdm: Following up in https://github.com/cultureamp/kaizen-design-system/pull/1624

packages/stylelint-plugin/src/cssStandardFunctions.ts
ActuallyACat: scratch that - I couldn't find anything either .

packages/stylelint-plugin/test.less
```diff
@@ -0,0 +1,18 @@
...
+
+@import "~@kaizen/design-tokens/less/color-vars";
+
+@import "~@kaizen/design-tokens/less/spacing-vars";
+
+.test {
+  color: @kz-var-color-wisteria-800;
+}
+
+.testt {
```
ActuallyACat, line 11: This looks like a typo, but I'm guessing it's to differentiate from the other class?

dmisdm: yep just a play file, cleaned it up

packages/stylelint-plugin/dev-config.js
```diff
@@ -0,0 +1,13 @@
+// This is an entrypoint file that you can use in Kaizen's stylelint.config.js.
```
ActuallyACat, line 1: This is :chef-kiss:

packages/stylelint-plugin/README.md
```diff
@@ -0,0 +1,45 @@
...
+- `kaizen/all-token-imports-must-be-used` - Reports and fixes unused Kaizen token imports.
+- `kaizen/tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
+
+Also available, but not enabled by default:
+
+- `kaizen/no-tokens-in-variables` - Prevent and fix usages of local non-Kaizen variables (only ones that are defined on the current sheet, ignoring any that are imported) that contain Kaizen tokens in their values. For example, `$card-bg: $kz-color-wisteria-800; .card { background-color: $card-bg; }` would be disallowed, and would be autofixed to `background-color: $kz-color-wisteria-800;`. The main reason you might want to enable this rule is if you have a large amount of proxy variables like these, because the other rules' autofixes can't be safely applied to variable values, e.g. we can't transform `$card-bg: $kz-color-wisteria-800;` -> `$card-bg: $kz-var-color-wisteria-800;` because we don't know how `$card-bg` is being used (it could be imported in other files, and used within a calc function which may silently fail).
```
ActuallyACat, line 39: kaizen/no-tokens-in-sass-variables, or kaizen/no-tokens-in-preprocessor-variables, or something else?

packages/stylelint-plugin/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "dist",
+    "types"
+  ],
+  "dependencies": {
+    "@kaizen/design-tokens": "^2.8.1",
+    "@types/lodash.kebabcase": "^4.1.6",
+    "@types/node": "^14.14.37",
+    "@types/postcss-less": "^4.0.0",
+    "@types/sass": "^1.16.0",
+    "@types/stylelint": "^9.10.1",
```
ActuallyACat, line 17: Are these supposed to be devDeps?

packages/stylelint-plugin/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "files": [
+    "dist",
+    "types"
+  ],
+  "dependencies": {
+    "@kaizen/design-tokens": "^2.8.1",
+    "@types/lodash.kebabcase": "^4.1.6",
+    "@types/node": "^14.14.37",
+    "@types/postcss-less": "^4.0.0",
+    "@types/sass": "^1.16.0",
```
ActuallyACat, line 16: sass is removed, but types/sass is here

packages/stylelint-plugin/src/index.ts
```diff
@@ -0,0 +1,31 @@
+import { Configuration } from "stylelint"
```
ActuallyACat, line 1: 💯

packages/stylelint-plugin/src/messages.ts
```diff
@@ -0,0 +1,41 @@
+export const missingRequiredKaizeImport = (path: string) =>
```
ActuallyACat, line 1: typo `missingRequiredKaizenImport`

packages/stylelint-plugin/src/rules/all-used-tokens-must-be-imported.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    // For each required import:
+    // If the import doesn't exist, and we are in `fix: true` mode, add it.
+    requiredImports.forEach(path => {
+      if (!currentImports.has(path)) {
+        if (options.fix) {
+          addImport(stylesheetNode, path)
+        } else {
+          options.reporter({
+            message: missingRequiredKaizeImport(path),
```
ActuallyACat, line 37: I think this might be the same typo from before `missingRequiredKaizeImport`

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { getParser } from "../util/utils"
+import { noInvalidUseOfVarTokensInEquations } from "./no-invalid-use-of-var-tokens-in-equations"
+
+describe("no-invalid-use-of-var-tokens-in-equations rule", () => {
+  test.each([
```
ActuallyACat, line 5: TIL `test.each` - I've been wrapping test cases in `foreach` like a sucker.

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    ["$kz-something-user-defined + 2", false],
+
+    ["solid 1px + $kz-spacing-md black", true],
+
+    ["calc($kz-spacing-md + 5px)", true],
+
+    ["calc($kz-var-spacing-md + 5px)", true],
+
+    ["calc(#{$kz-var-spacing-md} + 5rem)", false],
+    ["-$kz-var-spacing-md", true],
```
ActuallyACat, line 19: It might be handly to add a test case for adjacent operators, something like these:
```
["calc(#{$kz-var-spacing-md} * -5rem)", false],
["calc($kz-spacing-md * -5px)", true],
```

packages/stylelint-plugin/src/util/cssStandardFunctions.ts
```diff
@@ -0,0 +1,111 @@
+/*
+    A set of all function names that are part of the CSS specification.
+    Because they arent (shouldn't be) evaluated at compile-time (apart from some SASS overloaded ones like rgb, rgba), they should support CSS variables within their parameters.
+    Source: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions
+    To generate this list again, go to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions and run this in your dev console:
+    `Array.from(document.querySelectorAll("#content > article > div:nth-child(22) > div li a")).map(n => n.textContent.replace("()", ""))`
```
ActuallyACat, line 6: /nit This feels like it's very prone to break / become out of date with the next update to this MDN doc. IMO the link to the document it enough.

packages/stylelint-plugin/README.md
ActuallyACat: I think this paragraph could be a little clearer, maybe something like this:
```
`kaizen/no-tokens-in-variables` - Prevent and fix the usage of Kaizen design tokens in Sass or Less variables. This does not include design tokens in variables that have been imported into the current stylesheet. For example, `$card-bg: $kz-color-wisteria-800; .card { background-color: $card-bg; }` would be disallowed and would be autofixed to `background-color: $kz-color-wisteria-800;`. 

This rule is not recommended as a default as transitive variables are hard to accurately detect. You may enable this rule if you have a large number of proxy variables. The other rules' autofixes can't be applied to variable values, e.g. we can't transform `$card-bg: $kz-color-wisteria-800;` -> `$card-bg: $kz-var-color-wisteria-800;` because we don't know how `$card-bg` is being used (it could be imported in other files, and used within a calc function which may silently fail).
```

An important bit of information that was missing is why this rule isn't enabled by default, so I've added some additional context.

ActuallyACat (changes requested):
It have a small nits. Otherwise, I think this is good - splitting out the rules definitely made this more approachable.

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
jasononeil: Yeah this was new for me too!

packages/stylelint-plugin/src/util/cssStandardFunctions.ts
jasononeil: Great call out. I thought the same at one point but then forgot.

packages/stylelint-plugin/README.md
jasononeil: Great call that we should explain "why" better. I've adapted your text and tweaked it even further

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
jasononeil: I've added these test cases, and refactored to a more traditional `test()` structure, rather than `test.each()` so I can clarify what each one is testing, because it wasn't obvious to me

ActuallyACat (approved):
LGTM 💪

PR #1531, opened 2021-05-03
fix: Use 100% height for dismissIcon on Tag to avoid global box-sizing css conflicts

sugendran:
# Objective
Replaces out top and bottom margins on our Tag dismissIcon with 100% height.

# Motivation and Context
`performance-ui` has the following global css coming from Materialize:
```
*, *:before, *:after {
    box-sizing: inherit;
}
```

This is causing the following issue with our Tag:
image

# Checklist

### DIFF
--- a/draft-packages/tag/KaizenDraft/Tag/Tag.scss
+++ b/draft-packages/tag/KaizenDraft/Tag/Tag.scss
@@ -48,14 +48,11 @@ $small: $kz-var-spacing-md;
 .dismissIcon {
   position: relative;
   display: flex;
-  padding: $kz-var-spacing-xs;
-  margin-top: calc(-1 * #{$kz-var-spacing-xs});
+  height: 100%;
+  align-items: center;
+  padding: 0 $kz-var-spacing-xs;
   margin-right: -0.6625rem;
-  margin-bottom: calc(-1 * #{$kz-var-spacing-xs});
   margin-left: -0.225rem;
-
-  // Override Materialize global style for performance-ui
-  box-sizing: content-box;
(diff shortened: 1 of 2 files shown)


sugendran:
An alternate to this is that we rip out the top and bottom margin calcs and use flex to align center

sebpearce:
@sugendran If you have the time I'd go for the flex idea... probably fewer headaches down the road 😬

sugendran:
@sebpearce I don't have time right now

dougmacknz:
@sebpearce Either way this is going to be an override, so it probably doesn't matter too much. If y'all don't mind, I'm going to add a comment to the added line and merge this (it's blocking our Kaizen refresh of performance-ui).

sebpearce:
@dougmacknz Sounds good. My thing about fewer headaches was related to the fact that Murmur applies something like `* { box-sizing: border-box }` which could conflict with a margin-based approach to positioning. Was thinking that the longer this component relies on margins to position itself, the more liable it is to being overridden by stuff like that.

sebpearce (dismissed):

dougmacknz:
Hmm, actually yeah that might be a problem. I'll have a quick look at taking out the margins.

dougmacknz:
I've removed the top and bottom margins and padding and used 100% height to get the same hitbox.
image

I've verified on storybook that this will fix the issue when this global style exists.
```
* {
    box-sizing: inherit;
}
```

@sebpearce can I please get another review?

sebpearce (approved):
Nice 👌 🚢

PR #1546, opened 2021-05-05
fix: Pass the missing onHide prop to the toast

### LINKED ISSUE #1545: ToastNotification drops `onHide` prop and never calls back
Hello, we're having an issue with the `ToastNotification` from `@kaizen/component-library` in performance-ui.
A recent upgrade has caused the notifications we save in state to now never disappear. The issue appears to be this commit: https://github.com/cultureamp/kaizen-design-system/commit/d84d8f4b5d1efd8340995bc6763dd57612f0ab37#diff-0c710c1332f581db3b3d144d2e6ba7c64240091f6526f198bd1d09bfca01e307
where `...otherProps` are being dropped and not passed through to the underlying components. The missing prop is the optional `onHide` callback, which is necessary for us to clear up old state.

We have a fix in our userland that removes our reliance on state cleanup and imports the newer API of this release, so we don't have to rely on this functionality anymore, but it may still be important for others that I am not aware of (I dunno who has bound their important business logic to a notification onHide callback but it's possible!). Even though this component has been deprecated, can we add the onHide callback back?

OliBridgman:
Fixes #1545.

# Objective
The `onHide` prop was being dropped from the ToastNotification. This re-adds it.

# Motivation and Context
[Issue 1545](https://github.com/cultureamp/kaizen-design-system/issues/1545)

# Checklist

# Additional Considerations

### DIFF
--- a/packages/component-library/components/Notification/ToastNotification.tsx
+++ b/packages/component-library/components/Notification/ToastNotification.tsx
@@ -28,6 +28,7 @@ const ToastNotification = ({
     autohideDelay,
     message: otherProps.children,
     persistent,
+    onHide: otherProps.onHide
   })
 
   return null
--- a/packages/component-library/components/Notification/ToastNotification.tsx
+++ b/packages/component-library/components/Notification/ToastNotification.tsx
@@ -28,7 +28,7 @@ const ToastNotification = ({
     autohideDelay,
     message: otherProps.children,
     persistent,
-    onHide: otherProps.onHide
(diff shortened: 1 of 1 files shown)


makenosound (dismissed):

makenosound:
Ah, good pick up, @OliBridgman.

FYI you’ll need to set the commit message to the same as the PR title here to get a 💚 from Moai (because you have a single commit).

makenosound (approved):

PR #1561, opened 2021-05-10
fix: Allow react nodes for heading text

sugendran:
# Objective
In order to support translations that happen in some legacy class style components we need to be able to provide a react node to the heading.

# Checklist


draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -30,7 +30,7 @@ type LayoutContextType = "sidebarAndContent" | "contentOnly"
 export type EmptyStateProps = {
   id?: string
   automationId?: string
-  headingText: string
```
dougmacknz, line 33: ReactNode includes string, so you could just do ReactNode here.

dougmacknz (approved):

PR #1571, opened 2021-05-11
feat: guidance block accepts spot component

OliBridgman:
# Objective

Add ability to pass a `Spot` component from  `@kaizen/draft-illustration` to the img prop of the `GuidanceBlock`.

# Motivation and Context

We need to render some guidance blocks and they have images. I assume those images can or should change depending on a theme, but at the moment I can only pass a src path to the img prop of the GuidanceBlock.

# Concerns

* Is the type guard good enough?
* Is the wrapper css stupid
* Should I use a different prop?
* Is this wrong all together, should these be defined as... like a reference string or somethin e.g:

```
<GuidanceBlock spot={'Affirmative'} />
```

Screen Shot 2021-05-11 at 5 24 09 PM

Screen Shot 2021-05-11 at 5 25 46 PM

# Checklist

# Additional Considerations


OliBridgman:
@dougmacknz I had a go at making some dynamic loading based on a reference string.

* We render an error fallback if we can't find the module (this is typed with a typing export from spot/illustrations but you can still cast it to any or be rendering from jsx so thought this was appropriate).
* We render an error boundary that renders an error falling incase lazy loading fails (It will!)
* We render a placeholder with suspense for when in flight.

Issues:

* `React.lazy` only works for default exports, and all the spots are named exports, we end up lazily loading the whole module anyway so I'm not too sure on the benefit of doing this.
* I had to modify spot illustration, i'm pretty new to monorepo/package management in general, but i feel like if we wanna do this, this is gonna require multiple PRs with version changes. 
* I haven't currently removed img, `img` and  `illustration` prop are both optional, so there's a placeholder rendered if none are passed. This shouldn't be released this way but I'm not sure if API changes like this also wanna go through a few versions!

ActuallyACat (changes requested):
I am going to preface this by saying that there's lots of good stuff in this PR and it's a quality contribution. 

I think the optimisation is perhaps too premature in this scenario. If those illustrations were moved to default exports and the illustration SVGs were inlined we would see much more of a perf boost with this change. I am thinking that the second commit in this PR is a good "now" solution but I would also like to see us work towards this approach between now and then.

OliBridgman:
Closing in favour (hopefully) of:

v1: https://github.com/cultureamp/kaizen-design-system/pull/1595
OR
v2: https://github.com/cultureamp/kaizen-design-system/pull/1594

PR #1576, opened 2021-05-13
feat: Create BrandMoment component

dougmacknz:
# Objective
Creates a new component for 'Brand Moments' (see designs here  https://www.figma.com/file/OgCmQ6H2uygDqhd2c86oNt/Brand-Moments).

# Motivation and Context
This will allow camps to easily implement these new screens.

# Notes
There's a few things that will be incomplete here:
* Hasn't been RTL tested.
* Doesn't respond well to smaller viewports.
* Stories aren't using the correct illustrations.

These will be addressed in follow up PRs.

# Screenshots
https://dev.cultureamp.design/create-brand-moment/storybook/?path=/story/brand-moment-react--capture-intro

# Checklist

# Additional Considerations

### DIFF
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -1,0 +1,48 @@
+{
+  "name": "@kaizen/brand-moment",
+  "version": "1.0.0",
+  "description": "The brand moment component",
+  "scripts": {
+    "prepublish": "tsc --project tsconfig.dist.json",
+    "build": "yarn clean && yarn prepublish",
+    "build:watch": "yarn clean && yarn prepublish --watch",
+    "clean": "rimraf '**/*.d.ts' '**/*.js' '**/*.map'"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cultureamp/kaizen-design-system.git",
+    "directory": "packages/brand-moment"
+  },
+  "bugs": {
+    "url": "https://github.com/cultureamp/kaizen-design-system/issues"
+  },
+  "files": [
+    "**/*",
+    "!**/*.ts",
+    "!**/*.tsx",
+    "**/*.d.ts",
+    "!**/*.spec.*",
+    "!**/*.snap",
+    "!draft-templates",
+    "!stories",
+    "!docs",
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/component-library": "^9.6.1",
+    "@kaizen/draft-button": "3.3.4",
+    "@kaizen/draft-illustration": "^1.8.0"
+  },
+  "devDependencies": {
+    "@kaizen/design-tokens": "^2.8.1",
+    "elm-storybook": "cultureamp/elm-storybook#0.3.0",
+    "rimraf": "^3.0.2"
+  },
+  "peerDependencies": {
+    "@kaizen/design-tokens": "^2.8.2",
+    "react": "^16.9.0"
+  }
+}
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -33,7 +33,8 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-library": "^9.6.1",
-    "@kaizen/draft-button": "3.3.4",
+    "@kaizen/draft-button": "^3.3.4",
+    "@kaizen/hosted-assets": "^1.1.1",
     "@kaizen/draft-illustration": "^1.8.0",
     "@types/classnames": "^2.3.1",
     "classnames": "^2.3.1"
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -41,7 +41,6 @@
   },
   "devDependencies": {
     "@kaizen/design-tokens": "^2.8.1",
-    "elm-storybook": "cultureamp/elm-storybook#0.3.0",
     "rimraf": "^3.0.2"
   },
   "peerDependencies": {
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -34,7 +34,9 @@
   "dependencies": {
     "@kaizen/component-library": "^9.6.1",
     "@kaizen/draft-button": "3.3.4",
-    "@kaizen/draft-illustration": "^1.8.0"
+    "@kaizen/draft-illustration": "^1.8.0",
+    "@types/classnames": "^2.3.1",
+    "classnames": "^2.3.1"
   },
   "devDependencies": {
     "@kaizen/design-tokens": "^2.8.1",
--- a/packages/brand-moment/src/BrandMoment.tsx
+++ b/packages/brand-moment/src/BrandMoment.tsx
@@ -1,10 +1,11 @@
-import React, { ReactNode } from "react"
+import React, { ReactNode, ReactElement } from "react"
 import { Heading, Box, Paragraph } from "@kaizen/component-library"
-import { IntroductionsCaptureIntro } from "@kaizen/draft-illustration"
 import { Button, ButtonProps } from "@kaizen/draft-button"
+import { SceneProps } from "@kaizen/draft-illustration"
 import styles from "./BrandMoment.scss"
 
 type Props = {
+  illustration: ReactElement<SceneProps>
   heading: ReactNode
   subheading?: ReactNode
   bodyText?: ReactNode
@@ -15,9 +16,7 @@ type Props = {
 export const BrandMoment = (props: Props) => (
   <div className={styles.container}>
     <div className={styles.left}>
-      <div className={styles.leftInner}>
-        <IntroductionsCaptureIntro alt="" />
-      </div>
+      <div className={styles.leftInner}>{props.illustration}</div>
     </div>
     <div className={styles.right}>
       <div className={styles.rightInner}>
--- a/packages/brand-moment/src/BrandMoment.tsx
+++ b/packages/brand-moment/src/BrandMoment.tsx
@@ -15,7 +15,7 @@ type Props = {
   body?: ReactNode
   primaryAction: ButtonProps
   secondaryAction?: ButtonProps
-  footer?: ReactNode
+  footerContent?: ReactNode
 }
 
 export const BrandMoment = (props: Props) => (
@@ -62,7 +62,7 @@ export const BrandMoment = (props: Props) => (
         </div>
       </div>
     </main>
-    {props.footer && (
+    {props.footerContent && (
       <footer className={styles.footer}>
         <Box mr={2}>
           <div className={styles.poweredByContainer}>
@@ -81,7 +81,7 @@ export const BrandMoment = (props: Props) => (
           </div>
         </Box>
         <Paragraph variant="extra-small" color="dark-reduced-opacity">
-          {props.footer}
+          {props.footerContent}
         </Paragraph>
       </footer>
     )}
--- a/packages/brand-moment/src/BrandMoment.tsx
+++ b/packages/brand-moment/src/BrandMoment.tsx
@@ -12,7 +12,7 @@ type Props = {
   illustration: ReactElement<SceneProps>
   heading: ReactNode
   subheading?: ReactNode
-  bodyText?: ReactNode
+  body?: ReactNode
   primaryAction: ButtonProps
   secondaryAction?: ButtonProps
   footer?: ReactNode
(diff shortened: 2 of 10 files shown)


packages/brand-moment/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@kaizen/component-library": "^9.6.1",
+    "@kaizen/draft-button": "^3.3.4",
+    "@kaizen/hosted-assets": "^1.1.1",
+    "@kaizen/draft-illustration": "^1.8.0",
+    "@types/classnames": "^2.3.1",
+    "classnames": "^2.3.1"
+  },
+  "devDependencies": {
+    "@kaizen/design-tokens": "^2.8.1",
+    "elm-storybook": "cultureamp/elm-storybook#0.3.0",
```
jasononeil, line 44: Maybe not needed

packages/brand-moment/src/stories/BrandMoment.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+  decorators: [story => <div style={{ margin: "-1rem" }}>{story()}</div>],
+}
+
+export const DemoIntro = () => (
+  <BrandMoment
+    backgroundColor="blue"
+    header={<MinimalBasic />}
+    illustration={<EmptyStatesAction alt="" />}
+    subheading="Welcome to Culture Amp"
+    heading="Let's dive in and see how it works"
```
makenosound, line 39: This is extreme nit-pick territory but since I can see these being copied around the place a bit, we should ideally use proper apostrophes in all the copy for these examples:

```diff
- Let's dive in and see how it works
+ Let’s dive in and see how it works
```

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { ReactNode, ReactElement } from "react"
+import { Heading, Box, Paragraph } from "@kaizen/component-library"
+import { Button, ButtonProps } from "@kaizen/draft-button"
+import { SceneProps } from "@kaizen/draft-illustration"
+import { assetUrl } from "@kaizen/hosted-assets"
+import classnames from "classnames"
+import styles from "./BrandMoment.scss"
+
+type Props = {
+  backgroundColor: "blue" | "purple" | "green" | "red"
```
jasononeil, line 10: Is there semantic meaning to these colours? Should `"green"` be `"success"` etc?

packages/brand-moment/src/stories/BrandMoment.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+export const Error = () => (
+  <BrandMoment
+    backgroundColor="red"
+    header={<FakeNavBar />}
+    illustration={<EmptyStatesNegative alt="" />}
+    heading="Missing pages are one of life's mysteries"
+    body={
+      <>
+        <Box mb={1.75}>
+          Sorry but we can't find the page you're looking for. Go back and try
```
makenosound, line 110: Same with the language here. I reckon we could change this example to be more direct (and in line with our guidelines):

```diff
- Sorry but we can't find the page you're looking for.
+ We can’t find the page you’re looking for.
```

packages/brand-moment/src/BrandMoment.tsx
jasononeil: From @dougmacknz: We can't do a straight `variant` prop because the colour and illustration need to be controlled independently.

We think we can trust the designer to pick the right colour and have control here.

We should cover this in documentation (which I guess we need to write)

packages/brand-moment/src/stories/ExampleHeaders.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styles from "./ExampleHeaders.scss"
+
+export const MinimalBasic = () => (
+  <div className={styles.header}>
+    <div className={styles.headerLeft}>
+      <img
+        src={assetUrl("brand/logo-horizontal-default.svg")}
+        alt="Culture Amp"
+        width={126}
+      />
```
makenosound, line 12: Should we default the logo to be a link to `/`? Or make it easy for it to people to set a link?

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
...
+  header: ReactNode
+  illustration: ReactElement<SceneProps>
+  heading: ReactNode
+  subheading?: ReactNode
+  body?: ReactNode
+  primaryAction: ButtonProps
+  secondaryAction?: ButtonProps
+  footer?: ReactNode
```
jasononeil, line 18: We discussed if the footer should contain the full footer as used in capture, or if it should be passed in, similar to the nav header.

For now, the only use case for the footer is capture.

We're intending to leave it (have the component define the footer layout, and the props pass in the footer content). This makes it easier to consume, and doesn't give flexibility we don't need.

One suggestion is to rename the prop to `footerContent` so that if we change our minds in future, we can enforce the change with the type system by creating a new prop (and perhaps removing the old prop if desired):

```suggestion
  footerContent?: ReactNode
```

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const BrandMoment = (props: Props) => (
+  <div
+    className={classnames(styles.body, {
+      [styles.blue]: props.backgroundColor === "blue",
+      [styles.green]: props.backgroundColor === "green",
+      [styles.red]: props.backgroundColor === "red",
+    })}
+  >
+    <header className={styles.header}>{props.header}</header>
+    <main className={styles.main}>
```
jasononeil, line 30: Nice use of semantic HTML 💯

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
...
+            </Box>
+          )}
+          <Box mb={1.5}>
+            <Heading variant="display-0" tag={props.subheading ? "h2" : "h1"}>
+              {props.heading}
+            </Heading>
+          </Box>
+          {props.body && (
+            <Box mb={1.5}>
+              <Paragraph variant="intro-lede">{props.body}</Paragraph>
```
jasononeil, line 50: Need to check: if there are any implications (html validity, accessibility, cascade styling) of embedding these

jasononeil: I initially liked the idea of wrapping in a paragraph _only_ when it's a literal string, but on second thoughts that'd break if you have text with some inline markup.

So we either always wrap or never wrap. Keen to hear @ActuallyACat's take

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
...
+      <div className={styles.right}>
+        <div className={styles.rightInner}>
+          {props.subheading && (
+            <Box mb={0.5}>
+              <Heading variant="heading-3" tag="h1">
+                {props.subheading}
+              </Heading>
+            </Box>
+          )}
+          <Box mb={1.5}>
```
jasononeil, line 43: I'm interested in a new utility component for having a flex column / stack here

packages/brand-moment/src/BrandMoment.tsx
```diff
@@ -0,0 +1,89 @@
...
+              {props.heading}
+            </Heading>
+          </Box>
+          {props.body && (
+            <Box mb={1.5}>
+              <Paragraph variant="intro-lede">{props.body}</Paragraph>
+            </Box>
+          )}
+
+          <div className={styles.actions}>
```
jasononeil, line 54: Also interested in a utility component for button rows, at some point :)

packages/brand-moment/src/stories/BrandMoment.stories.tsx
dougmacknz: Good call

packages/brand-moment/src/stories/BrandMoment.stories.tsx
dougmacknz: I'm cool with this but CC @katsambarcus because I've just taken this from the Brand Moments Figma, which should be updated as well if we're changing this.

packages/brand-moment/src/BrandMoment.tsx
dougmacknz: I've asked Marc and it does makes sense for this to be referred to by mood (positive, negative, informative) instead of colour, which also means this can be consistent with other Kaizen components like Modal, Popover, EmptyState. What's not so consistent is the name we have for this prop across other components. I've called it `variant` for now because that is the slightly more consistent one.

packages/brand-moment/src/BrandMoment.tsx
dougmacknz: Yeah.. I'm guessing you saw the work I ended up abandoning in the commit history. It also means if you pass in a `<Trans>` component from Lingui you will no longer get the Paragraph. I concluded it will just end up confusing consumers, and probably end up with some incorrect paragraph variants being used. 

There are other Kaizen components doing this already as well:
* GuidanceBlock
* HeroPanel
* GenericTile

packages/brand-moment/src/BrandMoment.tsx
dougmacknz: Here's to hoping consumers don't put this component inside a `<main>` element 😅

packages/brand-moment/src/stories/BrandMoment.stories.tsx
katsambarcus: I'd prefer if @lucy-oldridge made the call on these language changes.

dougmacknz: Lucy doesn’t have Github access (no more licenses) so responding on her behalf:
> Would prefer to keep it as is. I think it’s a lightweight way to create an apologetic tone (considering the user could well be frustrated or stressed), especially when the H1 is more abstract and brand-y.

packages/brand-moment/src/stories/ExampleHeaders.tsx
dougmacknz: 🤔 these `ExampleHeaders` are just for Kaizen storybook purposes only, the real thing will actually be created on the consuming side, or at least passed in from the consuming side. I'm not sure if we'll always want this to be a link or not, but that can be worked out during implementation time.

makenosound: It’d be useful to show an example with a link at least I reckon.

dougmacknz: Done. FYI I am going to look at creating a helper/util component for adding focus indicators, which would be useful for links like this.

packages/brand-moment/src/BrandMoment.tsx
jasononeil: The big difference there is that each of those components have a prop with a name that suggests it should only contain text (`information.text`, `text.description` etc) whereas this explicitly has some non-text, and in the stories we've already got an example of wanting to pass in a block level element, which is semantically incorrect.

But on a practical level, it seems fine. I'm not willing to die on the hill :) I would personally change it but I'll approve the PR anyway, I'm happy for you to do this if you're confident the API niceness is worth the weirdness of the html semantics

jasononeil (dismissed):
Approved (with a disagreement about one thing, but I don't think it needs to block you). 

Nice work! Keen to see this component evolve from here

packages/brand-moment/src/BrandMoment.tsx
dougmacknz: You've convinced me.

ActuallyACat (dismissed):

packages/brand-moment/src/BrandMoment.tsx
dougmacknz: I took another look at this with fresh eyes and have come out with a solution that I'm happy with:

* All of the props that will be wrapped in some sort of Typography component are now part of a `text` object prop. This pattern is consistent the components I listed above as you pointed out.
* Props that are intended that have no or very little wrapping remain - `header`, `body`.
* There's now two ways to pass in body content. If you would like it to be wrapped in Typography styling, you  into the `text` object, otherwise it just goes into the `body` prop.
* This also provides a good solution for the comment about the footer here as well. As it stands, there is a `text.footer` prop which will be styled, and we have room for adding a base `footer` prop in the future if we need a way to avoid the CA logo. 

Side note: I renamed `heading` and `subheading` to `title` and `subtitle` to avoid confusion with `header`, and I noticed `title` has been used elsewhere in Kaizen as well.

jasononeil: I'm a fan. Nice work!

jasononeil (approved):
Looks good to me. I don't have a strong opinion on making `text.body` and `body` (etc) mutually exclusive. I'm happy to approve either way.

Looking forward to the coming improvements (RTL, responsive designs, etc)

PR #1581, opened 2021-05-14
feat: add Heart theme option to Elm Modal and EmptyState (for Heart illustrations)

mbylstra:
# Summary
- Adds Heart theme option to the Elm Modal and Empty State  illustrations (cautionary, success, informative, negative, action). Defaults to Zen if not hooked up to a "prop drilling" solution of passing the react `themeKey` to the component.
- Adds elm code to @kaizen/design-tokens for themes. Includes `Theme` type (`Zen` or `Heart`) that needs to be passed down to the components. Also a `themeDecoder` for passing the theme to the Elm app via React.
- Exposes an `image` helper function/component in `Kaizen.HostedAssets.Image` to create images that use a path from our hosted assets S3 bucket (we are currently not using hosted assets anywhere in Kaizen for Elm yet, so this was required to use the new illustrations)

#### other things:
- I upgraded the usages of zen illustrations in Elm modal to use the ones in hosted-assets rather than the ones in this repo. I think there is now an opportunity to delete them.
- The stories always show the Heart theme, because the Elm storybook doesn't support theme switching (it's possible, but out of scope for this PR)
- The only other component I could think of with illustrations was Guideance block, but there is no Elm GuideanceBlock component.
- Supporting illustration animation is out of scope.

# Examples:

image

image

# Story links: 
https://dev.cultureamp.design/elm-modal-with-heart-illustrations/storybook/?path=/story/emptystate-elm--action
https://dev.cultureamp.design/elm-modal-with-heart-illustrations/storybook/?path=/story/modal-elm--confirmation-cautionary-shown-by-default

# Checklist

# Additional Considerations

### DIFF
--- a/packages/design-tokens/elm/Kaizen/Theme.elm
+++ b/packages/design-tokens/elm/Kaizen/Theme.elm
@@ -1,6 +1,29 @@
-module Kaizen.Theme exposing (Theme(..))
+module Kaizen.Theme exposing (Theme(..), defaultTheme, themeDecoder)
+
+import Json.Decode as Decode exposing (Decoder)
 
 
 type Theme
     = Heart
     | Zen
+
+
+defaultTheme =
+    Zen
+
+
+themeDecoder : Decoder Theme
+themeDecoder =
+    Decode.string
+        |> Decode.map
+            (\themeString ->
+                case themeString of
+                    "zen" ->
+                        Zen
+
+                    "heart" ->
+                        Heart
+
+                    _ ->
+                        defaultTheme
+            )
--- a/packages/design-tokens/elm/Kaizen/Theme.elm
+++ b/packages/design-tokens/elm/Kaizen/Theme.elm
@@ -1,0 +1,6 @@
+module Kaizen.Theme exposing (Theme(..))
+
+
+type Theme
+    = Heart
+    | Zen
--- a/packages/hosted-assets/Kaizen/HostedAssets.elm
+++ b/packages/hosted-assets/Kaizen/HostedAssets.elm
@@ -1,4 +1,8 @@
-module Kaizen.HostedAssets exposing (assetUrl)
+module Kaizen.HostedAssets exposing (Role(..), assetUrl, image)
+
+import Html exposing (Html, img)
+import Html.Attributes exposing (src)
+import Html.Attributes.Aria as Aria exposing (ariaHidden, ariaLabelledby)
 
 
 originBaseUrl =
@@ -17,3 +21,23 @@ assetUrl "some/blob.png" -> "<https://<origin>/some/blob.png">
 assetUrl : String -> String
 assetUrl path =
     [ originBaseUrl, path ] |> String.join "/"
+
+
+type Role
+    = Presentation
+
+
+image : String -> Role -> Html msg
+image assetPath role =
+    img
+        ((src <| assetUrl assetPath) :: a11yAttributes role)
+        []
+
+
+a11yAttributes : Role -> List (Html.Attribute msg)
+a11yAttributes role =
+    case role of
+        Presentation ->
+            [ Aria.role "presentation"
+            , ariaHidden True
+            ]
--- a/packages/hosted-assets/Kaizen/HostedAssets.elm
+++ b/packages/hosted-assets/Kaizen/HostedAssets.elm
@@ -1,8 +1,4 @@
-module Kaizen.HostedAssets exposing (Role(..), assetUrl, image)
-
-import Html exposing (Html, img)
-import Html.Attributes exposing (src)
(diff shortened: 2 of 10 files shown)


mbylstra:
I will wait for approval from a designer, to make sure this compromise is acceptable.

packages/hosted-assets/Kaizen/HostedAssets.elm
```diff
@@ -17,3 +21,23 @@ assetUrl "some/blob.png" -> "<https://<origin>/some/blob.png">
 assetUrl : String -> String
 assetUrl path =
     [ originBaseUrl, path ] |> String.join "/"
+
+
+type Role
+    = Presentation
```
mbylstra, line 27: The code that determines how the `img` is rendered is based on https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/components/Icon/Icon.elm. Only the `Presentation` variant has been implemented. The `Img` variant can be added if a need arises.

packages/hosted-assets/Kaizen/HostedAssets/Image.elm
```diff
@@ -0,0 +1,70 @@
+module Kaizen.HostedAssets.Image exposing (Config, Role(..), attributes, default, image)
+
+import Html exposing (Html, img)
+import Html.Attributes exposing (src)
+import Html.Attributes.Aria as Aria exposing (ariaHidden, ariaLabelledby)
+import Kaizen.HostedAssets exposing (assetUrl)
+
+
+type Role
+    = Presentation
```
mbylstra, line 10: This just supports presentation for now (completely hidden from screen readers), but other types where you want a description to be read to a screen reader could be added.

packages/hosted-assets/Kaizen/HostedAssets/Image.elm
```diff
@@ -0,0 +1,70 @@
...
+    img attributes_ []
+
+
+a11yAttributes : ConfigValue msg -> List (Html.Attribute msg)
+a11yAttributes config =
+    case config.role of
+        Presentation ->
+            [ Aria.role "presentation"
+            , ariaHidden True
+            ]
```
mbylstra, line 70: This just supports presentation for now (completely hidden from screen readers), but other types where you want a description to be read to a screen reader could be added.

mbylstra:
[Kaizen: support Heart illustrations for EmptyState and Modal](https://trello.com/c/9m0ArRF9/1040-kaizen-support-heart-illustrations-for-emptystate-and-modal)

packages/design-tokens/elm/Kaizen/Theme.elm
```diff
@@ -0,0 +1,29 @@
...
+            (\themeString ->
+                case themeString of
+                    "zen" ->
+                        Zen
+
+                    "heart" ->
+                        Heart
+
+                    _ ->
+                        defaultTheme
```
sentience, line 28: Any reason for this fallback? I wonder if the decoder should `fail` when it receives an unexpected string value, and if we want to tolerate a missing/invalid value then we can use `Decode.maybe` for that:

```elm
decodeValue (maybe (field "theme" themeDecoder)) flags
```

sentience (approved):
Looks good! Left a comment with one idea for improvement, but I don't consider it a blocker.

packages/design-tokens/elm/Kaizen/Theme.elm
mbylstra: Thanks for the review Kevin!

I was a bit torn on this. I definitely didn't want it to fail, because I'm not very confident this would  be handled well by consumers. Falling back the last theme (in this case Zen) seems better than the page crashing or showing a generic error page. Also, I just wanted to make this as easy as possible for consumers (eg, not having to handle Maybes), as there will be a lot of friction in using the theme switcher in Elm so I'd be more worried people wouldn't find the time to implement it at all (which would guarantee no theme switching!). Even with these new features in Kaizen, there's still a lot of work to do to hook up the new illustrations - using the React `ThemeContext`, pass a flag to Elm and then prop drilling the theme all that way down to every Elm component. I happen to have a good amount of time this week so I can go deep with the prop drilling, etc, to get in our app but I'm not sure I can expect people in other teams with Elm codebases to do the same. 

Earlier as a much easier approach I suggested just immediately upgrading all the Elm illustrations to Heart, which would mean some customers not on the feature flag would see some components with Zen styles but Heart illustrations (because I thought implementing theme switching in Elm would be too much work for people). The Heart rollout team wasn't comfortable with this idea though.

There might be a much more magical solution for doing something like "context" in Elm to magically pass vars down the chain, but that would require a lot more investigation of feasibility.

PR #1582, opened 2021-05-14
feat: Migrate color rule from @cultureamp/stylelint-kaizen

dmisdm:
# Context, objective, and motivation

- `src/rules/colors-prefer-token.ts`
- `src/rules/colors-prefer-token.spec.ts`

- Better equality matching of color strings. Will parse colors on both ends (the parsed value and the kaizen token) and compare them using a common stringification function (to hex). This means it recognises that `#fff`, `#ffffff`, and `rgb(255,255,255)` are all equal, and similar with any other color including ones that manipulate alpha.
- Will now replace with/recommend new $kz-var variables rather than the deprecated ones.

Unfortunately a circular dependency occurred and a bunch of functions needed to be moved around.

# Checklist

# Additional Considerations


ActuallyACat:
Might be worthwhile to cc/ @sugendran in this (if he hasn't seen it already)

sugendran:
Sweet!

packages/stylelint-plugin/src/kaizenTokens.ts
```diff
@@ -1,9 +1,11 @@
 import fs from "fs"
```
ActuallyACat, line 1: The responsibilities of this file and `variableUtils` are becoming a little murky. I have a small idea about how it could be reworked (take it or leave it) which would be to combine all the `postcss-value-parser` helpers together. 🤷🏻‍♀️

ActuallyACat (approved):
Overall it LGTM - I can see where you've added the types for the existing stylelint plugin and that all read well.

dmisdm:
Because this was stale for a while and the stylelint plugin changed significantly, i'm moving this to a new PR.
https://github.com/cultureamp/kaizen-design-system/pull/1795

PR #1592, opened 2021-05-15
test: mock and or fix deprecation warnings in console when running tests

OliBridgman:
# Objective
The console is a little noisy, cleaning it up so we can more easily catch warnings and errors when we (never!) add them.

# Motivation and Context
Theres some 'real' warnings /errors that are harder to see/catch because we have added warnings for consumers. While the warnings are good, we don't want to pollute our console when running tests.

# Changes
* Most of these add a `mockImplementation` to the spy that is being created for console.warn. I've changed any tests that are explicitly looking for this to check the `mockImplementation` call.
* The modals needed a `ModalAccessibleLabel` so I added one instead of mocking the warn function. Modals also have a seperate issue that will be addressed in a follow-up PR.

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer?
- [ ] If this introduces either a new component or a breaking change to an existing component, has it been reviewed by an advocate? (Ask in #prod_design_systems)
- [ ] I have considered likely risks of these changes and got someone else to QA as appropriate
- [ ] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?
- Have Storybook stories been updated?

**If you're new to Kaizen, please ask #prod_design_systems to set up an onboarding session to get you up to speed.** If you have an urgent PR to merge before that happens, it is safest to ask in #prod_design_systems to have a design system advocate review the PR to catch any issues.

PR #1593, opened 2021-05-15
fix: ensure the accessibility check is being run at the right time

OliBridgman:
# Objective
The accessibility check was never running when consuming a modal. I think it's because of the open/closing animation/mounting. The ref passed down is always false when we first check and we never check again.

# Motivation and Context
I wanted to add a test for the accessibility warning in this PR: https://github.com/cultureamp/kaizen-design-system/pull/1592 but found this issue when trying to cause the check to happen in storybook. To replicate this, remove the AccessbilityLabel from a modal and watch it never warn.

# Concerns
setTimeout with 0 ain't the nicest looking solution. requestAnimationFrame wasn't slow enough?!? I dunno, anyone got a better idea?
Also this is branched off  https://github.com/cultureamp/kaizen-design-system/pull/1592 , but that doesn't go changing 'real' code and so doesn't need a version change. 

# Checklist

# Additional Considerations


OliBridgman:
This is the commit I _want_ this PR to be. https://github.com/cultureamp/kaizen-design-system/pull/1593/commits/13293bd5b2c5310c85dc8139041ac64e17a33d46

draft-packages/illustration/KaizenDraft/Illustration/illustration.spec.tsx
```diff
@@ -4,7 +4,9 @@ import * as SpotIllustrations from "./Spot"
 import * as SceneIllustrations from "./Scene"
 
 afterEach(cleanup)
-
+beforeEach(() => {
+  jest.spyOn(global.console, "warn").mockImplementation(jest.fn())
```
lijuenc, line 8: without any mock restoration, would this end up unintentionally silencing other warnings?

lijuenc (commented):
>setTimeout with 0 ain't the nicest looking solution. requestAnimationFrame wasn't slow enough?!? I dunno, anyone got a better idea?

https://headlessui.dev/react/transition#transition

I think we'll need to perform the check after the Transition component finishes doing its things. It does provide an `afterEnter` prop to which you can supply a callback that starts the check.

draft-packages/illustration/KaizenDraft/Illustration/illustration.spec.tsx
OliBridgman: I had lazily assumed cleanup would do that but that is apparently not the case. So probably, thanks.

OliBridgman: Fixed (I think!? spy confuses me) in the other PR and rebased.

OliBridgman:
@lijuenc `this.modalLayer` is still null when `afterEnter` enter fires. unfortunately

lijuenc (dismissed):
fair enough, this is probably the next best thing.

lijuenc (approved):
Looks like it was just rebasing?

ActuallyACat (approved):
Nice, thanks @OliBridgman

PR #1600, opened 2021-05-17
fix: Make box-sizing explicit

sebpearce:
# Objective

The GlobalNotification shows this bug when used in consumers, but it doesn't show up in Storybook:

image

The bug appears to show because the component sets `box-sizing: border-box` at the top level, and does not reset this to `content-box` for the icon, despite the icon depending on it to show correctly.

# Checklist


sebpearce:
Just testing this via yarn link, do not merge yet please.

ActuallyACat (approved):

PR #1601, opened 2021-05-17
ci: Add GH workflow for chromatic

ActuallyACat:
# Objective

- [x] create a Chromatic project for Kaizen in the Culture Amp Chromatic account
- [x] add a `chromatic.yaml` file to `.github/workflows` that runs the official Chromatic github action, using the latest Turbosnap feature so that snapshots are only built for components that are affected by the change
- [x] delete the old buildkite chromatic step that was linked to a chromatic project on Michael's chromatic account.

# Motivation and Context
Currently the buildkite workflow for Chromatic is very painful to use: you have to reset the master baseline manually every time, no one knows how to use it and there's no green/red status in the github commit. We are now permitted to use the github integration via the official github chromatic action. This replaces the buildkite Ci script with the 3rd party chromatic github action. It uses the latest Turbsnap feature (https://www.chromatic.com/docs/turbosnap), so that only snapshots are created for components are affected by changes, so that we don't burn too many of our allotted credits for the month.

This is evidence that the Turbosnap feature is working: https://github.com/cultureamp/kaizen-design-system/pull/1601/checks?check_run_id=3225537272#step:6:45

https://github.com/cultureamp/kaizen-design-system/issues/1768

image

image

# Checklist

# Additional Considerations


mbylstra:
Hi @allybeaton, I found there is a way to manually run an action from the github ui: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow-on-github (from the Actions tab). It might not be ideal - it seems to force to have to choose the branch, which would be super annoying and error prone, but it's an option! I added a commit with `on: workflow_dispatch` which sets the action to only run when triggered manually (https://github.com/cultureamp/kaizen-design-system/pull/1601/files#diff-83fb98eb20533196e6c42eab9d3f551c2a002d4330848227bcb4618cd9634901R7). 

I'm not seeing the option in the UI yet, but I think that's because "To trigger the workflow_dispatch event on GitHub, _your workflow must be in the default branch_. Follow these steps to manually trigger a workflow run." https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow-on-github. So, I think that would have to be merged into master before the button to trigger shows up.

mbylstra:
Rather than having to find the action in the Actions tab and then having to choose the right branch, there is a REST API that can be used to trigger the action: https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event. I think we could use that in another custom github action that listens to commands in comments like `@run chromatic`

mbylstra:
This is an action I made a while ago for updating the github status based on whether checklist items were checked (never quite made it into Kaizen): https://github.com/cultureamp/pr-checklist-reminders/blob/master/action.yml. It's a github action that runs some typescript and is able to read comments. I think we could do something similar - read the the last comment for a string like `@run chromatic` and then trigger _the main chromatic action_ using the REST API.

mbylstra:
I think there might be a simple way to do this with just the one `chromatic.yml` file (and no need to write any js/ts or have two github actions or use the rest api):

- add`on: issue_comment` instead of `on: push` (confusingly issue_comment is for both issue comments and PR comments 😡 )
- check if the "issue comment" is actually a pr comment and not an issue comment (😡 ). described in https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issue_comment.
- get that text of the last comment (this is the part I'm not sure if it's possible to do purely in yaml?). Might be ablel to use this: https://github.com/marketplace/actions/find-comment#find-the-last-comment-containing-the-specified-string, ordered by last, and just search for an empty string?
- if text contains "@run chomatic" I think this is quite possible/easy
- then run the chromatic build.
- else stop doing anything. The action is run, but it doesn't do anythig.

 https://github.community/t/on-issue-comment-events-are-not-triggering-workflows/16784

mbylstra:
I'm also trying out  https://github.com/nektos/act to test locally, but I haven't had much luck with it.

My first finding is that you have to set `on: push` for it to do anything. And then I get a `❗  ::error::Input required and not supplied: token` error. If I can figure that out then I think it's a much better workflow for writing custom actions.

mbylstra:
update: I could get "act" to work by doing this:

- going into https://github.com/settings/tokens and creating a new token and ticking the `repo` permissions.
- temporarily update the action to `on: push`
- run `act -s GITHUB_TOKEN`, and pasting in the token at the prompt

mbylstra:
The next thing I'm stuck on with "act" is `line 1: yarn: command not found`. I think it might because the docker image doesn't have yarn.

mbylstra:
I gotta say I'm not super comfortable passing my github token to this "act" thing that runs an image named `catthehacker/ubuntu:act-latest` (don't make your username `catthehacker` if you want to inspire confidence in people!), but having no local development environment to build the github action with really sucks.

mbylstra:
I temporarily added:

```
      - name: install yarn
        run: |
          npm install --global yarn
```
to the chromatic.yml to install yarn, but then I get `The engine "node" is incompatible with this module. Expected version "^14.15.1". Got "12.22.1"`. I think I will give up on getting "act" to work at this point.

.github/workflows/chromatic.yaml
```diff
@@ -3,67 +3,44 @@
 # Workflow name
```
mbylstra, line 3: These changes are pretty confusing, but this file is still here because of an earlier attempt to conserve snapshots before we knew about Turbosnap. Consider this file to be deleted (you can ignore the changes).

.github/workflows/chromatic.yaml
```diff
@@ -3,67 +3,44 @@
...
           node-version: "14.15.x"
-
       - name: Install package dependencies
-        if: ${{ env.runChromatic == 'true' }}
         run: |
-          yarn install --frozen-lockfile
-
-      # 👇 Adds Chromatic as a step in the workflow
+          yarn install --frozen-lockfile --ignore-scripts
+          # --ignore-scripts ignores the preinstall script that checks if the executor is logged into the github registry
```
mbylstra, line 35: We will need to remove `--ignore-scripts` and implement the private registry authentication when we start using frontend-ops.

package.json
```diff
@@ -12,6 +12,7 @@
   "scripts": {
     "storybook": "start-storybook -p 6006 -c storybook",
     "storybook:build": "build-storybook -c storybook -o storybook/public",
+    "storybook:build-chromatic": "yarn storybook:build --webpack-stats-json",
```
mbylstra, line 15: `--webpack-stats-json` is the secret sauce that chromatic requires to detect which files have code changes

.github/workflows/chromatic.yaml
```diff
@@ -3,67 +3,44 @@
...
       - name: Publish to Chromatic
-        if: ${{ env.runChromatic == 'true' }}
         uses: chromaui/action@v1
         # Chromatic GitHub Action options
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           # 👇 Chromatic projectToken, refer to the manage page to obtain it.
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           autoAcceptChanges: master # 👈 Auto accept master builds (why https://www.chromatic.com/docs/github-actions#github-squashrebase-merge-and-the-main-branch)
+          buildScriptName: "storybook:build-chromatic"
```
mbylstra, line 45: use the special build for chromatic

.github/workflows/chromatic.yaml
```diff
@@ -3,67 +3,44 @@
...
+          onlyChanged: true
```
mbylstra, line 46: this is the Turbosnap feature

mbylstra (approved):

.github/workflows/chromatic.yaml
ActuallyACat: Thanks for dropping in this comment. No doubt it will be a gotcha when we roll out FE-ops in this repo

ActuallyACat (commented):
LGTM

mbylstra:
I added a visual change to test Chromatic (https://github.com/cultureamp/kaizen-design-system/pull/1601/commits/dea6c342f843a8e7ca14ee0d670e02618396a921),change the button css from round to square, but it doesn't look like it was picked up by Chromatic: https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=32 🤔

mbylstra (changes requested):
We need confirm whether it's actually working for the button example I added before merging.

mbylstra:
I added a more obvious change (tsx inside a story itself, and it still didn't get picked up) 🤔

mbylstra:
It might be because I accepted all the changes in https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=29, earlier in this PR, but I was kind of expecting it to pick up further changes.

mbylstra:
As far as I can tell from the docs, it should be comparing against the last accepted changes, so it should have picked up the button changes since the build that was accepted: https://www.chromatic.com/docs/branching-and-baselines#branches-and-baselines.

mbylstra:
In the last build (where the text "Chromatic Testing" was added to button):
```
Retrieving story files affected by recent changes
    → Traversing dependencies for 1 files that changed since the last build
Retrieved story files affected by recent changes
    → Found 0 story files affected by recent changes
```
So it looks like turbosnap detected that 1 file changed, but I'm not sure why it doesn't consider the story to be affected.

mbylstra:
I was drafting up this comm and was hoping to grab screenshots of what changes in Chromatic looks like when I noticed it was behaving as expected. I'm just leaving this half finished comm here in case anyone else wanted to roll with it:



:rainbow: :rainbow:  Announcing the General release of Chromatic (trial) :rainbow: :rainbow:
**What is it? How does it work?**
Chromatic (https://www.chromatic.com/) is a tool for detecting and reviewing visual changes which has been integrated with the Kaizen build pipeline.
Chromatic has been available in Kaizen for quite a while now, but it had some major limitation in that we could not enable the Github integration or link it to the Culture Amp account, so it was linked to my personal account :shh:, and the workflow was extremely clunky, so it's something we would just only manually on request when there were any scary changes.
This update brings Github Integration (fully automatic and can be interacted with from the Github UI) and a trial of a new Chromatic feature named "Turbosnap". We have a limited number of Chromatic "snapshots" available per month on our plan. As we have a large number of stories and we are frequently pushing commits, we would quickly run out of credits if we ran visual tests for every story for every commit. The use of Turbosnap
. Chromatic has been been accepted  , so we never
**How does this affect me?**
You will now be able to merge PRs without any fear of unexpected visual changes :tada:
If a PR contains any visual changes, it will need to be reviewed in Chromatic before Merging.

mbylstra:
To try to narrow down the issue, I created this branch https://github.com/cultureamp/kaizen-design-system/pull/1827/commits, which is exactly the same, but it's squashed (so chromatic builds aren't reused) and it has turbosnap turned off. I wanted to see if it's turbosnap that's the issue. Annoyingly, the node-sass build error came back (we had this when we were working on this 2 or 3 weeks ago, which magically fixed itslef the next day). The really strange thing is that the node-sass issue only happens on that branch, not this branch. I added two more commits to the other branch and now it is literally the same as this one (other than a small change to the button component), but that branch has the node-sass errors and this one doesn't 😖 . My only theory is that there is some issue with the github-action agent, and that that PR is being used on a different agent... or some other voodoo is happening. So, I wasn't able to get to the bottom of the original problem because of the node-sass problem 😞

mbylstra:
I think this might be the cause of the scss error (https://github.com/sass/node-sass/issues/1579). It looks like a heisenbug that we somehow haven't run into before (and for somereason is only happening on [this branch](https://github.com/cultureamp/kaizen-design-system/pull/1827) , which is almost identical to the branch of this PR.

mbylstra:
I got to the bottom of the node-sass issue, it was because of the `--ignore-scripts` line in the yarn install (https://github.com/cultureamp/kaizen-design-system/pull/1601/files#diff-83fb98eb20533196e6c42eab9d3f551c2a002d4330848227bcb4618cd9634901R34 ). The problem manifest itself in a very heisenbug way making it really hard to determine that's what was causing it (it's not happening on this branch, but it's happening on the other branch for some reason, I think because this branch edits some scss and the other one doesn't). So we either need to:

- authenticate with the private culture amp package repository 
OR
- remove this line https://github.com/cultureamp/kaizen-design-system/blob/master/package.json#L26, which isn't actually needed at the moment because we aren't yet using anything from the private package repository

So I was able to continue debugging the turbosnap issue...

mbylstra:
Unfortunately it doesn't really seem to be working properly, even for very basic changes directly in a story itself, rather than an indirect change from a component or a scss variable.

This is a change i introduced (changing the label for a button):

image

And this is what chromatic reported:

image

So it picks up that there was the code change, but incorrectly determines that there aren't any stories affected. I highlight the part with `--webpack-stats-json` to show that this is actually being run (there isn't a typo at the call site or something)

mbylstra:
From here, I'm not really sure what to do. We might need to ask Chromatic. I'm not sure how to give them debug info though as the github action logs won't be available to them:

image

Another thing we could try is to try replicating the problem locally and having a look inside the webpack stats json file to see if it mentions the story (could be a `--webpack-stats-json` issue to do with our monorepo or something).

Issue #1602, opened 2021-05-17
Run Chromatic using the Chromatic github action (+ switch to Culture Amp account)


mbylstra:
## TODO:
- [x] create a Chromatic project for Kaizen in the Culture Amp Chromatic account
- [x] add a `chromatic.yaml` file to `.github/workflows` that runs the official Chromatic github action
- [x] make chromatic builds opt-in per commit (the default is to run the on every commit, which would quickly eat up all of our credits)
- [ ] delete the old buildkite chromatic step that was linked to a chromatic project on Michael's chromatic account.

mbylstra:
We have started working on this in https://github.com/cultureamp/kaizen-design-system/pull/1601

mbylstra:
I created a branched PR that adds the functionality where the action is only run if you add a comment "run chromatic": https://github.com/cultureamp/kaizen-design-system/pull/1603

PR #1603, opened 2021-05-17
ci: add chromatic github action that triggers on PR comment

mbylstra:
# Objective
- Add Chromatic github integration using the official Chromatic Github Action
- link it to the Culture Amp chromatic account (was previously linked to a personal account)
- The github action only publishes to Chromatic if the latest comment in the PR contains "run chromatic"
- The action should run whenever someone adds a comment(however, there is no way to test until we merge to master, as the `issue_comment` trigger is apparently only triggered for actions on the master branch)
- I tested this with the `push` trigger, which can be run on a dev branch (the final commit switches to `issue_comment`). This is an example of a succesful run that published to chromatic: https://github.com/cultureamp/kaizen-design-system/runs/2760935408?check_suite_focus=true

### TODO
- [x] create a Chromatic project for Kaizen in the Culture Amp Chromatic account
- [x] add a `chromatic.yaml` file to `.github/workflows` that runs the official Chromatic github action
- [x] make chromatic builds opt-in per commit (the default is to run the on every commit, which would quickly eat up all of our credits)
- [ ] delete the old buildkite chromatic step that was linked to a chromatic project on Michael's chromatic account.

### Notes:
-  the `issue_comment` is for both issue comments and PR comments (that's not a mistake, the Github Actions API is just really confusing!)
- I plan to the delete the old buildkite chromatic step in a follow up PR (just in case this one isn't working, there is still the ability to optionally use the old one)

# Motivation and Context
https://github.com/cultureamp/kaizen-design-system/issues/1602


mbylstra:
This comment should kick of the chromatic action.

mbylstra:
Maybe changing from `pull_requestion_review_comment` to `issue_comment` will work? Does this comment tigger a chromatic build?

mbylstra:
run chromatic

mbylstra:
this is a comment without the special command, so chromatic should not run on next push

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,79 @@
+# Chromatic action from https://www.chromatic.com/docs/github-actions
+
+#v2
+
+# Workflow name
+name: "chromatic"
+
+on: push
```
mbylstra, line 8: this would be changed to:

```
on:
  issue_comment:
    types: [created]
 ```
 
 before merging, but the action can only be tested with `on: push`

File: .github/workflows/chromatic.yaml [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+            ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-
+            ${{ runner.os }}-
+
+      - uses: actions/setup-node@v1
+        if: contains(steps.latestComment.outputs.comment-body, 'run chromatic')
+        with:
+          node-version: "14.15.x"
+
+      - name: Install package dependencies
+        if: contains(steps.latestComment.outputs.comment-body, 'run chromatic')
```
mbylstra, line 66: I'm sure there's a way to turn this into a variable, but I don't know how!

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-
+            ${{ runner.os }}-
+
+      - uses: actions/setup-node@v1
+        if: contains(steps.latestComment.outputs.comment-body, 'run chromatic')
```
mbylstra, line 61: I'm sure there's a way to turn this into a variable to avoid repetition, but I don't know how!

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,79 @@
...
+        env:
+          prNumber: ${{ steps.getPr.outputs.number }}
+
+      - name: Get most recent comment
+        if: success() && steps.getPr.outputs.number
+        uses: peter-evans/find-comment@v1
+        id: latestComment
+        with:
+          issue-number: ${{ steps.getPr.outputs.number }}
+          body-includes: ""
```
mbylstra, line 41: this is a hack, but it seems to work!

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,79 @@
...
+          direction: last
+
+      - run: |
+          echo ${{ steps.latestComment.outputs.comment-id }}
+          echo ${{ steps.latestComment.outputs.comment-body }}
```
mbylstra, line 46: this was just some debug code that should be removed

mbylstra:
run chromatic

ActuallyACat:
I feel like we should update the documentation with instructions on how to run Chromatic now (potentially use the PR template)? https://github.com/cultureamp/kaizen-design-system/blob/master/.github/pull_request_template.md)

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,80 @@
...
+        id: latestComment
+        with:
+          issue-number: ${{ steps.getPr.outputs.number }}
+          body-includes: ""
+          direction: last
+
+      - run: |
+          echo "runChromatic=${runChromatic}" >> $GITHUB_ENV
+        env:
+          runChromatic: ${{ contains(steps.latestComment.outputs.comment-body, 'run chromatic') }}
```
ActuallyACat, line 47: /nit Something like `@chromatic run` is a little more extensible and would allow us to write new command should we need them in the future. I like the way Dependabot has approached it: 
```
@dependabot rebase will rebase this PR
@dependabot recreate will recreate this PR, overwriting any edits that have been made to it
@dependabot merge will merge this PR after your CI passes on it
... etc
```

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+      - run: echo "found run chromatic text in latest comment"
+        if: ${{ env.runChromatic == 'true' }}
+
+      - uses: actions/cache@v2 # use of cache for node_modules saves usually > 1min of build time
+        if: ${{ env.runChromatic == 'true' }}
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-${{ hashFiles('**/package.json') }}
```
ActuallyACat, line 56: I might be misunderstanding something, but why do we need to store this information (`key`)?

ActuallyACat (changes requested):
Just have a few small suggestions

.github/workflows/chromatic.yaml
mbylstra: I'm not really sure, that was copy pasta from the unified-home config.

.github/workflows/chromatic.yaml
mbylstra: I wasn't sure about that, because `@dependabot` actually at-mentions the dependabot account/user. I think @chromatic would at-mention chromatic, and they'd get a bunch of notifications. We could do `@cultureamp`, but I wasn't sure if that would also generate a bunch of notifications we don't want.

mbylstra:
>I feel like we should update the documentation with instructions on how to run Chromatic now (potentially use the PR template)? https://github.com/cultureamp/kaizen-design-system/blob/master/.github/pull_request_template.md) 

Ah yeah, I was intending to do that. I can add that to the follow up PR where the old chromatic buildkite step gets removed. I'm not 100% sure this works yet, and it can only be properly tested once merged, so I didn't want to add the instructions until we knew it works properly.

.github/workflows/chromatic.yaml
chromatic: > I think @chromatic would at-mention chromatic (this person: https://github.com/chromatic), and they'd get a bunch of notifications.

Confirmed! No harm done though.

ActuallyACat: Yeah, good call (and thanks for chiming in @chromatic 😄). We also have a @cultureamp-ci account that we can use https://github.com/cultureamp-ci. @jasononeil or @zioroboco might have some thoughts on whether a PR bot for common actions is something we might want to invest time in and make available for other repos.

ActuallyACat: (That being said, I don't think this conversation should block this PR and would be happy for this to be merged with `chromatic run` as the trigger)

.github/workflows/chromatic.yaml
ActuallyACat: Ah! Okay looking at unified's config and these docs it looks like a pretty standard way to cache deps: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

ActuallyACat (approved):
LGTM thanks @mbylstra 👍🏻

mbylstra:
Thanks for the review @ActuallyACat. I might hold off on merging until next Monday, as I think I'll be too stretched to follow it up this week if there are issues. I'm totally happy for you to move it forward if you have the time though.

ActuallyACat:
Good call.

PR #1624, opened 2021-05-21
feat: Replace rather than compile in color fns

dmisdm:
This is a change to the `no-invalid-functions` rule.
Rather than turn a statement like: `mix($kz-color-wisteria-800, white, 80%)` into it's compiled hex `#5d5f6e`, just replace the tokens with their values.
e.g. `mix($kz-color-wisteria-800, white, 80%)` -> `mix(#35374A, white, 80%)`


jasononeil:
This still doesn't feel like the desired behaviour to me. I've pushed commits to the original PR

PR #1632, opened 2021-05-24
feat: Delete component-library/Button

dougmacknz:
# Objective
The big finale. With all usages and references of the old Button removed from `performance-ui`, `murmur` and `kaizen-design-system` itself, this PR deletes the `Button` and `IconButton` components from `@kaizen/component-library` completely. Any remaining usages or references in smaller repos should be minimal enough for the person doing the NPM refresh to take on.

**BREAKING CHANGE**

Any usages or references to Button or IconButton from `@kaizen/component-library` need to be updated to come from `@kaizen/draft-button` instead.

Eg.
`import { Button } from "@kaizen/component-library"`
to
`import { Button, IconButton } from "@kaizen/draft-button"`

It's also been common to import the Button stylesheet for the `button-reset` mixin:
`@import '~@kaizen/component-library/components/Button/styles'`
Ideally a copy of the mixin should be created on the consuming repo instead, but this can be updated like this as a quick fix:
`@import '~@kaizen/draft-button/KaizenDraft/Button/styles'`

Some other variations to look out for:

Importing directly from the GenericButton file:
`import Button from "@kaizen/component-library/components/Button/components/Button"`

Importing ButtonProps:
`import { ButtonProps } from '@kaizen/component-library`

Elm usages of the old Button:
`import Button.Button as Button`
changes to:
`import KaizenDraft.Button.Button as Button`

# Motivation and Context
https://github.com/cultureamp/kaizen-design-system/issues/1466

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer?
- [ ] If this introduces either a new component or a breaking change to an existing component, has it been reviewed by an advocate? (Ask in #prod_design_systems)
- [ ] I have considered likely risks of these changes and got someone else to QA as appropriate
- [ ] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?
[...]

PR #1641, opened 2021-05-25
chore: Remove (Zen) from Button stories

dougmacknz:
Quick follow up to https://github.com/cultureamp/kaizen-design-system/pull/1632 - remove `(Zen)` from the Button stories because it no longer makes sense.


ActuallyACat (approved):
So glad the legacy button is no more 🚮 Nice one!

PR #1642, opened 2021-05-25
fix: Fix issue with autohide for draft-menu

makenosound:
# Objective

This moves the event listener that checks for document-level clicks (and decides whether or not to hide the dropdown menu) into the capturing phase rather than the bubbling phase. In doing so we also need to check that the `referenceElement` isn’t the targeted event (or one of its children).

# Motivation and Context

We’re seeing issues in multiple places where multiple `<Menu>` instances are open as users move between menus. I.e., if a user:

1. Opens a menu by clicking on its button
2. Attempts to open a different menu by clicking on its button
3. Both menus would be open

This strangely doesn’t manifest within the Kaizen Storybook (even though it should), but we’ve seen it occur in multiple different consuming applications (comparisons-service, unified-home). The issue occurs because of the `e.stopPropagation()` call within `renderButton` which, like it says on the box, stops the propagation through the bubbling phase of events.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/menu/KaizenDraft/Menu/MenuDropdown.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/MenuDropdown.tsx
@@ -60,12 +60,14 @@ const MenuDropdown = ({
       if (
         popperElement &&
         e.target instanceof Node &&
-        !popperElement.contains(e.target)
+        !popperElement.contains(e.target) &&
+        referenceElement !== e.target &&
+        !referenceElement?.contains(e.target)
       ) {
         hideMenuDropdown()
       }
     },
-    [popperElement, hideMenuDropdown]
+    [popperElement, referenceElement, hideMenuDropdown]
   )
 
   const handleDocumentResize = useCallback(() => {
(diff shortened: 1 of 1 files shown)


draft-packages/menu/KaizenDraft/Menu/MenuDropdown.tsx
```diff
@@ -79,23 +81,29 @@ const MenuDropdown = ({
...
   }
 
+  useEffect(() => {
+    window.addEventListener("resize", handleDocumentResize, false)
+
+    return () => {
+      window.removeEventListener("resize", handleDocumentResize, false)
+    }
+  }, [handleDocumentResize])
+
```
makenosound, line 91: I’ve separated the two `useEffect` calls here so they can be reasoned/changed independently.

draft-packages/menu/KaizenDraft/Menu/MenuDropdown.tsx
```diff
@@ -60,12 +60,14 @@ const MenuDropdown = ({
       if (
         popperElement &&
         e.target instanceof Node &&
-        !popperElement.contains(e.target)
+        !popperElement.contains(e.target) &&
+        referenceElement !== e.target &&
+        !referenceElement?.contains(e.target)
```
makenosound, line 65: Because we’re calling this all the time now, we need to check that the `referenceElement` and its children aren’t being clicked so that we can leave the rest of the event handling/toggling behaviour alone.

sebpearce (approved):
Wow. 👏  🚢

PR #1671, opened 2021-06-03
feat: elm tag allow text wrapping option + dismiss button size fix

mbylstra:
**NOTE: Elm only, no code that is consumed by React has been touched**

### allow text wrapping option
This is a option to deal with tags with user generated content
that is so large that the text escapes the available horizontal
space (such as in a multi-select field). Currently the tag
completely breaks if this happens. There's an existing option
to add truncation (ellipsis) with a fixed max width, but the problem
with that is there is no tooltip, there is no way to distinguish
tags that start with the same text (this is a real problem in production)

Ideally, we'd implement the tooltip when the tag truncates (as is specified in the docs for Tag), but this
is unfortunately in the way too hard basket for Elm. You need to measure
the width of the text with an effect, and then the component requires state
to keep track of it - it would be a major change to the component. Not only
would it be difficult to implement but it would be a major headache for consumers. Is it worth it for a technology we have contained?

So, as a quick fix, this option allows the tag's content to grow vertically so
that it can be read, and is in response to an urgent required fix in production.

Before

image

after

image

(the lack of vertical margin is expected - you have to add it yourself, that's just how the component already was - this isn't a problem in multi-select widget which adds it, this is just to demonstrate the wrapping. Also it's Zen theme because I forgot to switch themes when getting the screenshots)

### fix the dismiss button size
In doing this PR, I discovered that the Elm dismiss button became way too large recently. This has been fixed (it must have been introduced when the React Tag was changed without the Elm tag being tested)

before

image

after

image

### DIFF
--- a/draft-packages/tag/KaizenDraft/Tag/TagElmExtra.scss
+++ b/draft-packages/tag/KaizenDraft/Tag/TagElmExtra.scss
@@ -1,0 +1,13 @@
+@import "./Tag.scss";
+
+.mediumAllowingTextWrapping {
+  min-height: $medium;
+  padding-top: 4px; // padding required because vertical centering doesn't work with min-height
+  padding-bottom: 4px;
+}
+
+.smallAllowingTextWrapping {
+  min-height: $small;
+  padding-top: 1px; // padding required because vertical centering doesn't work with min-height
+  padding-bottom: 1px;
+}
--- a/draft-packages/form/ElmStories/SelectFieldStories.elm
+++ b/draft-packages/form/ElmStories/SelectFieldStories.elm
@@ -141,6 +141,7 @@ main =
                     [ SelectField.view
                         (SelectField.multi
                             { truncationWidth = Just 300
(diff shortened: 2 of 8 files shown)


draft-packages/tag/KaizenDraft/Tag/TagElmExtra.scss
```diff
@@ -0,0 +1,13 @@
+@import "./Tag.scss";
```
mbylstra, line 1: I intentionally put this in a separate file, because this does not need to be shared with React. React should implement the tooltip on truncation (which is much easier in React) - issue here: https://github.com/cultureamp/kaizen-design-system/issues/1672.

mbylstra:
[Problem Ticket: **HRBP Permissions display issue - new user modal**](https://trello.com/c/4BS7rXX0/997-problem-ticket-hrbp-permissions-display-issue-new-user-modal)

caseysinclair (approved):

PR #1674, opened 2021-06-04
chore: Default Storybook to Heart

dougmacknz:
The Storybook theme switcher add on is defaulting to Zen. Now that we've GA'ed Heart, let's switch this to default to Heart instead.

### DIFF
--- a/storybook/theme-switcher-addon/themeManager.ts
+++ b/storybook/theme-switcher-addon/themeManager.ts
@@ -1,17 +1,12 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import {
-  defaultTheme,
-  heartTheme,
-  ThemeManager,
-  zenTheme,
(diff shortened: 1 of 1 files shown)


ActuallyACat:
Some thoughts... https://github.com/cultureamp/kaizen-design-system/blob/master/storybook/theme-switcher-addon/themeManager.ts#L22

When theme is not cached in localstorage we default to defaultTheme (which will continue to be Zen until all customers have moved to Heart). What do you think about overriding this to fallback to heart?

dougmacknz:
> When theme is not cached in localstorage we default to defaultTheme (which will continue to be Zen until all customers have moved to Heart). What do you think about overriding this to fallback to heart?

Oh right, oops. I've done that now.

ActuallyACat (approved):

PR #1680, opened 2021-06-07
chore: Add dependabot config

ActuallyACat:
# Objective
This config will supercede the org-level defaults for Dependabot. 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,4 @@
-version: 1
+version: 2
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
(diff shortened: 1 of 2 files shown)


mbylstra:
Ah, so this is going back to the old dependabot V1 where it should only get updates on the weekend? Was that the intention?

mbylstra:
I noticed that v1 is being shut down on Aug 1: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/upgrading-from-dependabotcom-to-github-native-dependabot

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 1
```
mbylstra, line 1: is version 1 intentional? Not version 2?

ActuallyACat: Oops - fixed

ActuallyACat:
> Ah, so this is going back to the old dependabot V1 where it should only get updates on the weekend? Was that the intention?

My bad! I meant to use v2 - all other parts in the config relate to v2 config. There are some things the v2 config can do that v1 can't - which is limit the number of total pull requests and ignore patches. I will probably need to tweak it some more depending on how noisy / quiet it is but I'm hoping this is a good start.

mbylstra (approved):

PR #1690, opened 2021-06-07
ci: Allow consumption of private @cultureamp packages

ActuallyACat:
# Objective
Setup build so that we have the _option_ of consuming private packages in this repo in the future. 

This will allow us to share webpack config so that Kaizen configs do not need to be maintained separately.

### DIFF
--- a/scripts/check-if-authenticated-to-github-registry.bash
+++ b/scripts/check-if-authenticated-to-github-registry.bash
@@ -9,7 +9,8 @@ else
   echo
   echo "You need to log into the GitHub package registry to install private @cultureamp packages."
   echo
-  echo "Follow: https://github.com/cultureamp/node-packages/blob/master/README.md"
+  echo "Read this to find out how:"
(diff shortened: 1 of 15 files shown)


scripts/check-if-authenticated-to-github-registry.bash
```diff
@@ -0,0 +1,15 @@
...
+
+registry_user=$(npm whoami --registry "https://npm.pkg.github.com/")
+
+if [ $? -eq 0 ] || [ ! -z "$registry_user" ]; then
+  echo "Logged in to GitHub package registry as: ${registry_user}"
+else
+  echo
+  echo "You need to log into the GitHub package registry to install private @cultureamp packages."
+  echo
+  echo "Follow: https://github.com/cultureamp/node-packages/blob/master/README.md"
```
damncabbage, line 12: Super minor / optional:

```suggestion
  echo "Read this to find out how:"
  echo "  https://github.com/cultureamp/node-packages/blob/master/how-to-use-private-cultureamp-packages-on-your-laptop.md"
```

damncabbage (approved):
:+1:, some minor suggestions.

damncabbage (approved):

PR #1695, opened 2021-06-09
fix: change elm select "faded" placeholder from wisteria-300 to 800

mbylstra:
wisteria-300 is that ugly pink colour. This now matches the
React component for the default styling of a placeholder for
a select component.

before

image

after

image

the React version (for reference)

image


mbylstra:
[Kaizen: fix Elm Select placeholder color (not pink!)](https://trello.com/c/UEVhI0bI/1037-kaizen-fix-elm-select-placeholder-color-not-pink)

caseysinclair (approved):

PR #1700, opened 2021-06-10
fix: Remove hasRendered state from ModalHeader

OliBridgman:
# Objective
In `performance-ui`, when I'm writing tests that render modals, in the happy, multiple-`it`s-passing case, I end up polluting the console with:

```
    Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
        in ModalHeader (created by InformationModal)
```

I think this is because the ModalHeader has a setState in a requestAnimationFrame and I guess our tests are so damn good that we're unmounting too quickly?

To add a mount check, I needed to make `GenericModalSection` accept a forwardRef. So I made this a feat?

# Motivation and Context
I tried avoiding this in user-land but it's not possible to mock internal references that these components have to themselves so e.g

```
jest.mock("@kaizen/draft-modal", () => ({ ...requireActual('@kaizen/draft-modal'), ModalHeader: 'div' }))
```
Doesn't fly.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/ModalHeader.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/ModalHeader.tsx
@@ -13,32 +13,9 @@ export interface ModalHeaderProps {
   readonly children: React.ReactNode
 }
 
-export interface ModalHeaderState {
-  hasRendered: boolean
-}
-
-class ModalHeader extends React.Component<ModalHeaderProps, ModalHeaderState> {
-  state = { hasRendered: false }
-
-  componentDidMount() {
-    requestAnimationFrame(() => this.setState({ hasRendered: true }))
-  }
-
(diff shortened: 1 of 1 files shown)


dougmacknz:
Firstly, thanks for looking into this!

I was about to approve your PR, but I decided to have a look at why `ModalHeader` keeps track of `hasRendered` anyway. It was done to avoid focus going straight to the close button because, at the time, the modal was focusing the first interactive element found. Since that was done, we have changed this behaviour to focus the modal header itself always, instead of the first interactive element found.

So we can actually just rip out the code that was removing the close button from being focusable, and the `hasRendered` state along with it.

dougmacknz (approved):
Awesome, thanks!

PR #1704, opened 2021-06-15
feat: Add responsiveness to BrandMoment

dougmacknz:
# Objective
Main objective here was to add responsiveness to the `BrandMoment` component, which was previously mainly built around larger viewports and didn't scale down well.

Before:
https://user-images.githubusercontent.com/1811583/122691403-d1080300-d272-11eb-8412-6afc9cb6cb24.mov

After:
https://user-images.githubusercontent.com/1811583/122691472-365bf400-d273-11eb-9d18-8ac44371b140.mov

Notes:
* I've switched from flex to grid for the main container so that I could utilise `gap`.
* I've used a mobile first approach to media query styling. Small viewport styles are applied by default, then larger viewport styles are added under `min-width` media queries.

----

I've also bundled in a bunch of small tweaks that came out of a review with @katsambarcus 

### Small component adjustments:
- [x] Footer logo should link to cultureamp.com
- [x] ‘Powered by CA’ can go below text on small viewport
- [x] Remove reduced opacity from footer text

### Story adjustments:
- [x] Rename stories to reflect types of brand moments
- [x] Add <a> tags to example footer links
- [x] Make customer focused logo larger
- [x] Add box shadow to logo container
- [x] Add a non customer focused outro example
- [x] Put locale selector into capture examples
- [x] Add exit button to a story as an example
- [x] Add a 4px border radius to the logo inside of the container for logos that aren't transparent or white

**Note:** this doesn't yet changing the typography or CTA button width based on viewport as the designs suggest. There is work happening at the moment to make this easy to do, so we can come through once that's done and add that here.

### DIFF
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -41,11 +41,11 @@
   },
   "devDependencies": {
     "@kaizen/design-tokens": "^2.9.0",
+    "@kaizen/draft-select": "^1.16.2",
     "rimraf": "^3.0.2"
   },
   "peerDependencies": {
     "@kaizen/design-tokens": "^2.8.2",
-    "@kaizen/draft-select": "^1.16.2",
     "react": "^16.9.0"
   }
 }
--- a/packages/brand-moment/package.json
+++ b/packages/brand-moment/package.json
@@ -45,6 +45,7 @@
   },
   "peerDependencies": {
     "@kaizen/design-tokens": "^2.8.2",
+    "@kaizen/draft-select": "^1.16.2",
     "react": "^16.9.0"
   }
 }
--- a/packages/brand-moment/src/BrandMoment.scss
+++ b/packages/brand-moment/src/BrandMoment.scss
@@ -73,6 +73,7 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
   flex: 1 0 auto;
   display: flex;
   align-items: center;
+  margin-bottom: $kz-var-spacing-xxl;
 }
 
 .mainInner {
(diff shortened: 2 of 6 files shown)


packages/brand-moment/src/BrandMoment.scss
```diff
@@ -1,9 +1,15 @@
 @import "~@kaizen/design-tokens/sass/color-vars";
 @import "~@kaizen/design-tokens/sass/spacing-vars";
+@import "~@kaizen/design-tokens/sass/layout";
+
+.container {
+  max-width: 1392px;
```
sebpearce, line 6: Sure you don't want to use: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/design-tokens/tokens/layout.json#L4 ?

dougmacknz: oooh, thanks!

ActuallyACat:
Screen Shot 2021-06-22 at 10 09 10 am

This one is probably for @katsambarcus - The CTA button is not in the viewport for smaller devices... We might want to consider knocking down the size of the typography or the illustration to bring the button into the view.

dougmacknz:
@ActuallyACat Yep there are plans to reduce the typography size on the smaller viewport, but that's going to be on the second pass once we have worked out a good way to do that (hence the discussion that's happening at the moment).

packages/brand-moment/package.json
```diff
@@ -45,6 +45,7 @@
   },
   "peerDependencies": {
     "@kaizen/design-tokens": "^2.8.2",
+    "@kaizen/draft-select": "^1.16.2",
```
ActuallyACat, line 48: Correct me if I'm wrong, but this looks like it's only used in stories. Should this instead be a `devDependency`?

dougmacknz: You're right

ActuallyACat (approved):

PR #1715, opened 2021-06-22
feat: controlHasUnconstrainedHeight option for Elm Select

mbylstra:
# Objective
This adds an optional `controlHasUnconstrainedHeight` option that defaults to false. The default behaviour (as is current) is that the control* is always only takes up one line, and that if the select item is longer than the available spac,e it gets truncated with an ellipsis. That is problematic if being able to read the full text of the selected item is very important for the use case. This gives an option to view the the full option, with text wrapping instead of truncation (with unconstrained height). 

*WTF is "a control", as in the noun, not the verb? You know, that thing you click on that's not the dropdown that opens up. It's just what it's called, don't blame me! ... also, WTF is "a select" - "select" is a verb, not a noun! There is no noun form of "select" in the dictionary. Yeah, I know... don't get me started, it's just what it's called 😞 

# Motivation and Context
https://trello.com/c/QigAAifj/978-should-have-need-a-way-to-check-long-values-on-user-modal

# Story

https://dev.cultureamp.design/ted/elm-select-grow-vertically-option/storybook/?path=/story/select-elm--single-clearable-controlhasunconstrainedheight

# Screenshots
long sentence with spaces

image

long word with no spaces (uses `word-break: break-word;`)

image

A usual short word

image

A short word, after clicking the control, when the `searchable` option is enabled and the cursor is showing:

image

A short word, after clicking the control, when the `searchable` option is enabled and the cursor is showing:

image
Why is the cursor in a weird spot? See Caviats.

#### Caviats

- The cursor is in a misaligned spot when you pick a long word, and then you search. Unfortunately I don't think there's a way to solve this without significantly rewriting the existing css (and risk breaking edge cases that are widely used in the product). I tried to implement this in a way that there would be no changes to the implementation of the existing CSS. Because this component has been implemented on the assumption that there would be only one line, everything is vertically centered. It would be extremely fiddly, messy  and risky to undo that assumption (either updating the existing implementation, or having a massive amount of overrides for the `controlHasUnconstrainedHeight`). Also, seeing as have contained Elm, it doesn't seem worth the risk or effort.
- I would have liked to have put more padding at the top and the bottom when there is a long word, but again, this is tricky because everything is vertically centered and doesn't have much padding.
- The is a bigger gap between the long text and the "clear" button that is necessary. This is really tricky to fix without also breaking the existing implementation (it's a really complex component!)

I'm hoping we can accept these compromises as the end result of being able to read the full text of long words is a huge UX improvement.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/styles.elm.scss
+++ b/draft-packages/select/KaizenDraft/Select/styles.elm.scss
@@ -12,6 +12,8 @@
   // box-shadow: 0 0 0 1px $color;
 }
 
+$control-height: $ca-grid * 2;
+
 .basePlaceholder {
   @include kz-typography-paragraph-body;
   margin-left: 2px;
@@ -20,6 +22,12 @@
   position: absolute;
   transform: translateY(-50%);
   box-sizing: border-box;
+
+  .controlHasUnconstrainedHeight & {
+    position: relative;
+    top: 0;
+    transform: none;
+  }
 }
 .container {
   position: relative;
@@ -37,7 +45,7 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
-  min-height: 48px;
+  min-height: $control-height;
   position: relative;
   box-sizing: border-box;
   border: $kz-var-border-solid-border-width $kz-var-border-solid-border-style
@@ -70,7 +78,13 @@
   overflow: hidden;
 
   .disabled & {
-    position: static;
+    position: static; // TODO calc based on border width
+  }
+
+  .controlHasUnconstrainedHeight & {
+    min-height: calc(
+      #{$control-height} - (#{$kz-var-border-solid-border-width} * 2)
+    );
   }
 }
 
@@ -80,6 +94,12 @@
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
+
+  .controlHasUnconstrainedHeight & {
+    white-space: normal;
+    overflow: visible;
+    word-break: break-word;
+  }
 }
 
 .placeholder {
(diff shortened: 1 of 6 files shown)


draft-packages/select/KaizenDraft/Select/styles.elm.scss
```diff
@@ -219,3 +239,17 @@
     background-color: transparent;
   }
 }
+
+.autosizeInputContainer {
+  @include kz-typography-paragraph-body;
+  @include ca-inherit-baseline;
```
mbylstra, line 245: These includes fix a long standing bug where the search text font would be the browser default (a sans serif font). I fixed it at the same time, because I needed to make sure the vertical alignment was correct.

draft-packages/select/KaizenDraft/Select/styles.elm.scss
```diff
@@ -37,7 +45,7 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
-  min-height: 48px;
+  min-height: $control-height;
```
mbylstra, line 48: I tried to avoid changing any of the existing CSS (because that would require a lot of QA, and there'd be a risk of ) and this is the only real change I made. It's just replacing `48px` with rem units of `2 * $ca-grid`, so that I could use that to calculate the height minus the borders.

mbylstra:
[(Should have) Need a way to check long values on User Modal](https://trello.com/c/QigAAifj/978-should-have-need-a-way-to-check-long-values-on-user-modal)

idamadam (approved):
This looks good to me Michael, you already covered the points I thought of in your documentation and I'm happy with the trade offs. Thanks for writing such a thorough PR 👏

mbylstra:
Thanks for the review @idamadam!

caseysinclair (approved):
LGTM!

PR #1716, opened 2021-06-22
feat: Create @kaizen/responsive package with useMediaQueries hook

dougmacknz:
Credit to @makenosound for this idea.

## Objective

Creates a `useMediaQueries` hook under a new `@kaizen/responsive` package. The `useMediaQueries` hook returns a set of `queries` (booleans), and `components` which can be used to determine whether to show/hide things, or used in any conditionals.

There are queries / components that are built in, based on Kaizen's breakpoints:
* `isSmall` / `<SmallOnly>`
* `isMedium` / `<MediumOnly>`
* `isLarge` / `<LargeOnly>`
* `isMediumOrSmaller` / `<MediumOrSmaller>`
* `isMediumOrLarger` / `<MediumOrLarger>`

responsive-booleans

And there is the ability to send in your own media queries to add to the returned set:

```jsx
const { queries, components } = useMediaQueries({
  prefersReducedMotion: "(prefers-reduced-motion: reduce)",
})
const { prefersReducedMotion } = queries
const { PrefersReducedMotion } = components

<>
  <div>Prefers reduced motion? {prefersReducedMotion ? "Yes" : "No"}</div>
  <PrefersReducedMotion>Prefers reduced motion</PrefersReducedMotion>
</>
```

reduced-motion


packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+type Props = { [key: string]: string }
+
+export const useMediaQueries = (propQueries: Props) => {
+  const theme = useTheme()
+
+  // TODO: max-width usage needs to be -1px
+  const kaizenQueries = {
+    isSmall: useMediaQuery({
+      query: `(max-width: ${theme.layout.breakpoints.medium})`,
```
sebpearce, line 13: We'd need a `- 1` here to make it consistent with the approach of "each breakpoint represents the minimum in its range, not the maximum of the range below it".

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    isMedium: useMediaQuery({
+      query: `(min-width: ${theme.layout.breakpoints.medium}) and (max-width: ${theme.layout.breakpoints.large})`,
```
sebpearce, line 16: Same here for the `max-width`, subtract 1px.

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    isLarge: useMediaQuery({
+      query: `(min-width: ${theme.layout.breakpoints.large})`,
```
sebpearce, line 19: 👌

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    isMediumDown: useMediaQuery({
+      query: `(max-width: ${theme.layout.breakpoints.large})`,
```
sebpearce, line 22: Subtract 1px here.

packages/responsive/src/useMediaQueries.tsx
sebpearce: Sorry, just noticed the bullet point addressing this in the todo stuff 🤦

sebpearce:
This is a boss-level PR 😎  Love the GIFs and detail.

What do you think of changing `isMediumDown` and `isMediumUp` to `isSmallOrMedium` and `isMediumOrLarge`? Could be personal taste but I find those easier to understand.

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react"
+import { useMediaQuery } from "react-responsive"
+import { useTheme } from "@kaizen/design-tokens"
+
+type Props = { [key: string]: string }
+
+export const useMediaQueries = (propQueries: Props) => {
```
OliBridgman, line 7: Want to default this to an empty object? Save the consumer 2 characters!! :P

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,168 @@
+import React, { FC, useEffect, useState } from "react"
+import { useTheme } from "@kaizen/design-tokens"
+
+type Props = { [key: string]: string }
+
+const minus1Px = (breakpoint: string) => `${parseInt(breakpoint, 10) - 1}px`
+
+export const useMediaQueries = (
+  propQueries: Props
```
makenosound, line 9: I reckon @OliBridgman’s previous note to default to an empty object is still a good one. No value in forcing consumers to pass through empty objects in the caller all the time.

dougmacknz: Yeah I want to do it but ran into problems with TS given the dynamic nature of props coming in. i.e. I wasn't able to make `[key: string]: string` optional, but I can have another go at it. Let me know if you know of a good way to do it.

makenosound: Ah, right, I’ll have a squiz and see if I can see the issue.

makenosound: You shouldn’t need to make the arguments optional, it should be enough to set a default:

```ts
export const useMediaQueries = (
  propQueries: Props = {}
): {
```

That’ll allow the caller to not pass the default `useMediaQueries()` without making anything "optional" in the arguments themselves.

dougmacknz: Nice, ty!

packages/responsive/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/design-tokens": "^2.9.0"
+  },
+  "devDependencies": {
+    "@kaizen/draft-button": "^3.3.15",
```
dmisdm, line 38: Looks like this package isn't being used at all?

packages/responsive/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "!draft-templates",
+    "!stories",
+    "!docs",
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/design-tokens": "^2.9.0"
```
dmisdm, line 35: Should this be a peer?

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,169 @@
...
+  // Create Kaizen helper components
+  // ---------------------------------------
+  const kaizenComponents = {
+    SmallOnly: (props: any) => <>{kaizenMatches.isSmall && props.children}</>,
+    MediumOnly: (props: any) => <>{kaizenMatches.isMedium && props.children}</>,
+    LargeOnly: (props: any) => <>{kaizenMatches.isLarge && props.children}</>,
+    MediumOrSmaller: (props: any) => (
+      <>{kaizenMatches.isMediumOrSmaller && props.children}</>
+    ),
+    MediumOrLarger: (props: any) => (
```
dmisdm, line 149: Look, maybe nitpicky so up to you, but the `any` here could be replaced with something like `{children: React.ReactNode}` - in which `children` is a required prop.

packages/responsive/src/useMediaQueries.tsx
```diff
@@ -0,0 +1,169 @@
...
+    `(max-width: ${subtractOnePixel(theme.layout.breakpoints.medium)})`
+  )
+  const mediumMatchMedia = window.matchMedia(
+    `(min-width: ${
+      theme.layout.breakpoints.medium
+    }) and (max-width: ${subtractOnePixel(theme.layout.breakpoints.large)})`
+  )
+  const largeMatchMedia = window.matchMedia(
+    `(min-width: ${theme.layout.breakpoints.large})`
+  )
```
dmisdm, line 44: Probably nothing to do, but it's hard to say how often these are going to get called (probably not too often but perhaps if someone used this hook in a table component with lots of cells/rows etc).
It's worth noting the performance difference when memoizing this: https://jsbench.me/sbkqzzsdpn/1
Probably not worth prematurely optimising, but if you wanted to now or in the future, you could wrap these with `React.useMemo`

dmisdm (approved):
I sound like a mega nerd when i say this but i can't wait to use this hook.
Awesome work :)

packages/responsive/src/useMediaQueries.tsx
makenosound: We’ll potentially end up with a ton of these in the future, so some optimisation is probably worthwhile on a proactive level. We’d need to trigger the `useMemo` based on a resize handler, yeah?

dmisdm: i think it's much simpler: 
```
const largeMatchMedia = React.useMemo(() => window.matchMedia(
    `(min-width: ${theme.layout.breakpoints.large})`
  ), [theme.layout.breakpoints.large])
```

In other words: `largeMatchMedia` is a computed value that depends on `theme.layout.breakpoints.large` - only re-calculate it when `theme.layout.breakpoints.large` changes (which is, like, never)

dmisdm: i think we should be cautious of bikeshedding: this hook could be optimised in a number of ways.
we don't really have any evidence to support decisions, only good guesses really.

now i'm thinking i might setup a story with heaps of instances of it to test it...

makenosound: > i think it's much simpler

Ah, you’re right. I wasn’t really sure about what `matchMedia` returns but it’s just an instance so we should really be fine.

sebpearce (approved):

ActuallyACat (approved):
Approving based on @dmisdm review :D

PR #1718, opened 2021-06-22
fix: elm select also stop propagation on keyboard events

mbylstra:
# Objective

When the Elm Select is used inside a form in modal, if you hit the escape key when the Select dropdown/menu is open, you expect it to just close the dropdown, but it also closes the dropdown (and you lose your work). So there are issues when the Select is used in a modal.

### Solution:
The component currently adds`preventDefault` to keyboard events that it listens to such as escape, space, arrow buttons. The solution is to also add `stopPropagation`. Although the problem found is specifically for the escape key event, I couldn't see a reason why you _wouldn't_ want to the stop the propagation of the events we are capturing, and it made the code simpler to treat all keyboard events the same. @caseysinclair : can you think of any reason why we'd not want to stop propagation?

# Motivation and Context
https://trello.com/c/apd6MKnq/1071-pressing-the-escape-key-when-the-select-widget-or-the-date-picker-is-open-should-not-close-the-modal

# Checklist

# Additional Considerations


mbylstra:
[Pressing the escape key when the select widget or the date picker is open should not close the modal.](https://trello.com/c/apd6MKnq/1071-pressing-the-escape-key-when-the-select-widget-or-the-date-picker-is-open-should-not-close-the-modal)

caseysinclair (approved):
Makes sense to add the stop propagation, can't think of any downsides!

PR #1720, opened 2021-06-23
fix: Add better RTL support to BrandMoment

dougmacknz:
# Objective
Adds better RTL support to the `BrandMoment` component, mainly by swapping out directional properties with the awesome new [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties). Browser support: https://caniuse.com/css-logical-props

Also adds an ltr/rtl switcher to the Kaizen storybook for easy RTL support checking on any component.

# Motivation and Context
I had descoped this from the first pass with the intention of coming back and tidying this up. It wasn't too bad because I used a lot of grid and flex, but there were still a few things to fix up.

# Notes
* Our `Box` component's RTL support isn't so great - it requires you to send in an `rtl` prop which I would like to deprecate in favour of just using logical properties as I have done here. As a result - the spacing between two action buttons isn't there on RTL, until we make that update. Not a blocker IMO.

# Screenshots

Before:
image

After:
image

ltr/rtl Storybook switcher:
rtl-storybook

### DIFF
--- a/packages/brand-moment/src/BrandMoment.scss
+++ b/packages/brand-moment/src/BrandMoment.scss
@@ -10,7 +10,7 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
 // --------------------------------
 
 .body {
-  width: 100%;
+  inline-size: 100%;
   display: flex;
   flex-direction: column;
   align-items: center;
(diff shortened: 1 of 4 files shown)


sebpearce (dismissed):
Nice! I didn't realise we could use these – for some reason thought Edge didn't support them, but it does.

dougmacknz:
Soz forgot to list this as a draft, still a few more changes to make

sebpearce (approved):
🚢

PR #1722, opened 2021-06-25
fix: Collapsible fixes: type=button and opaque sticky header

dougmacknz:
* Adds type=button to the Collapsible header to avoid forms submitting.
* Uses a hex instead of transparency in the Collapsible header to avoid this issue with sticky headers, a regression that I introduced in https://github.com/cultureamp/kaizen-design-system/pull/1694
collapsible-header


makenosound (dismissed):

lisatassone (approved):
Looks fine to me. I couldn't test the sticky behaviour on storybook, but the hex should do it

ActuallyACat (approved):

PR #1724, opened 2021-06-25
chore: Delete the @kaizen/draft-search-box package

dougmacknz:
# Objective
Deletes the @kaizen/draft-search-box package.

# Motivation and Context
This package was created [here](https://github.com/cultureamp/kaizen-design-system/pull/957) but it wasn't reviewed by advocates and has a bunch of changes that need to be made, including a move from its own package to the form package and a rename from `SearchBox` to `SearchField` at the least, with some other wholesale changes being recommended as well (see https://github.com/cultureamp/kaizen-design-system/pull/976).

This version needs to be removed to avoid people finding it, not realising all of ☝️ and using it in their apps.


ActuallyACat (approved):
LGTM.

This is _very unlikely_, but if someone creates a `@kaizen/draft-search-box` package in the future (unlikely since we're moving away from `drafts` and this component will move to form) we'd need to start it on a different version because we can't publish a package over a version that already exists. 

Make sure the advocates / @martindavid are across this change.

PR #1728, opened 2021-06-28
fix: use module.exports to fix stylelint plugin

jasononeil:
# Objective

Fix the stylelint plugin so it actually works in consuming repos 😓 

# Motivation and Context

For some reason, stylelint wants `module.exports = {}` rather than `export default = {}`

Typescript was compiling `export default {}` to `exports.default = {}`, but it leaves `module.exports = {}` as is.

Q: Why didn't I notice this earlier?
A: I only tested the plugin in Kaizen, and in Kaizen we use a local `dev-config.js` that indeed uses `module.exports` already.

Q: Isn't this already working in murmur?
A: No, it's in murmur but silently failing, so no rules are being applied.


ActuallyACat (approved):
> No, it's in murmur but silently failing,

Uhh, my bad 😅

Issue #1729, opened 2021-06-29
design-tokens and stylelint-plugin are not cleaning compiled files


ActuallyACat:
When compiling (or running `yarn install`) in a package, we typically have a clean script to remove compiled files. Design-tokens and stylelint-plugin are not cleaning up all compiled files and as such are throwing confusing false positive issues when linting: 

<img width="842" alt="Screen Shot 2021-06-29 at 9 59 24 am" src="https://user-images.githubusercontent.com/1621182/123718175-d3e1a400-d8c1-11eb-94fc-9a978cc5e8a6.png">

PR #1730, opened 2021-06-29
feat: add brand component to kaizen

lloydstubber:
# Objective

# Motivation and Context

# Checklist

# Additional Considerations

### DIFF
--- a/packages/brand/index.ts
+++ b/packages/brand/index.ts
@@ -1,0 +1,1 @@
+export * from "./src/Brand/Brand"
--- a/packages/brand/src/Brand/Brand.tsx
+++ b/packages/brand/src/Brand/Brand.tsx
@@ -1,0 +1,23 @@
+import React from "react"
+import { assetUrl } from "@kaizen/hosted-assets"
+
+export interface BrandProps {
+  variant:
+    | "logo-horizontal"
+    | "logo-vertical"
+    | "enso"
+    | "collective-intelligence"
+  isReversed: boolean
+  alt: string
+}
+
+export const Brand = (props: BrandProps) => {
+  const brandTheme = props.isReversed ? "-reversed" : "-default"
+
+  return (
+    <img
+      src={assetUrl(`brand/${props.variant}${brandTheme}.svg`)}
+      alt={props.alt}
+    />
+  )
+}
--- a/packages/brand/src/Brand/Brand.tsx
+++ b/packages/brand/src/Brand/Brand.tsx
@@ -7,12 +7,12 @@ export interface BrandProps {
     | "logo-vertical"
     | "enso"
     | "collective-intelligence"
-  isReversed: boolean
+  reversed?: boolean
   alt: string
 }
 
 export const Brand = (props: BrandProps) => {
-  const brandTheme = props.isReversed ? "-reversed" : "-default"
+  const brandTheme = props.reversed ? "-reversed" : "-default"
 
   return (
     <img
--- a/draft-packages/empty-state/CHANGELOG.md
+++ b/draft-packages/empty-state/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [2.3.3](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/draft-empty-state@2.3.2...@kaizen/draft-empty-state@2.3.3) (2021-06-30)
+
+**Note:** Version bump only for package @kaizen/draft-empty-state
+
+
+
+
+
 ## [2.3.2](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/draft-empty-state@2.3.1...@kaizen/draft-empty-state@2.3.2) (2021-06-25)
 
 **Note:** Version bump only for package @kaizen/draft-empty-state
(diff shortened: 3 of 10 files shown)


packages/brand/Kaizen/Brand/Brand.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import { assetUrl } from "../../../hosted-assets"
+
+interface BrandProps {
```
ActuallyACat, line 4: nit/ You'll probably want to export these props. There _could_ be situations where a consumer might want to know them.

packages/brand/Kaizen/Brand/Brand.tsx
```diff
@@ -0,0 +1,22 @@
...
+  isReversed: boolean
+}
+
+export const Brand = (props: BrandProps) => {
+  const brandTheme = props.isReversed ? "-reversed" : "-default"
+
+  return (
+    <img
+      src={assetUrl(`brand/${props.variant}${brandTheme}.svg`)}
+      alt="Culture Amp"
```
ActuallyACat, line 19: Also nitpicking, and @dougmacknz might be able to advise better, but are we _always_ going to want this `alt` to be `Culture Amp`? I'm a little unsure of where the collective intelligence logo is going to be used.

ActuallyACat (changes requested):
🤏🏻 Very small nitpicks. Looking good though!

packages/brand/Kaizen/Brand/Brand.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import { assetUrl } from "../../../hosted-assets"
```
dougmacknz, line 2: Think you'll want to import from `"@kaizen/hosted-assets"` instead and add that package as a dependency.

packages/brand/Kaizen/Brand/Brand.tsx
dougmacknz: 🤔 In general this is what we'll want, except for the collective intelligence one which will need something different. I see a couple of options with pros and cons:
* We have sensible defaults here, with an optional `alt` prop for consumers to override if they like. 
* We make a required `alt` prop and have no default.

The latter would force consumers to think about what they want, but the former would avoid potential incorrect usages.

packages/brand/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./Kaizen/Brand/Brand"
```
dougmacknz, line 1: I know some components have the component living in a `Kaizen` folder, but I don't really know why. `src` makes more sense and seems to be what we're moving towards. Mind changing this?

ActuallyACat: It's because we use the same package for React and Elm code. The consistent directory structure ensures (or ensure**d**) the way Elm modules are imported is consistent. E.g. `import KaizenDraft.ComponentName.ComponentName as ComponentName`

ActuallyACat: It's not needed since I am doubtful this component will have a corresponding Elm component before the Elm code is shuffled out of this repo. 
fyi @mbylstra

packages/brand/Kaizen/Brand/Brand.tsx
lloydstubber: Maybe the later? Just so it's a little more flexible? I guess that word is a double-edged sword though

lloydstubber:
Addressed the feedback! Let me know if I should also work with a designer to get the README.mdx sorted in this PR as well @ActuallyACat @dougmacknz

storybook/helpers/index.js
```diff
@@ -0,0 +1,27 @@
+"use strict"
```
ActuallyACat, line 1: There are some compiled js/declaration files that shouldn't be here.

lloydstubber: 😓 I'll understand this yarn clean stuff one day

ActuallyACat: This is our bad - they should be in .gitignore and be removed with `yarn clean` (which they're not). I've got an issue and PR to clean some of these up https://github.com/cultureamp/kaizen-design-system/issues/1729

ActuallyACat (approved):
Look good - thanks @lloydstubber !

packages/brand/src/Brand/Brand.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import { assetUrl } from "@kaizen/hosted-assets"
+
+export interface BrandProps {
+  variant:
+    | "logo-horizontal"
+    | "logo-vertical"
+    | "enso"
+    | "collective-intelligence"
+  isReversed: boolean
```
dougmacknz, line 10: Soz I missed this on the previous look overs. Can this prop name please be `reversed`? Looking over other components with a reversed prop, there are two components with `isReversed` but the vast majority have this prop named `reversed`. It can also be optional.

lloydstubber: @dougmacknz no worries. This has been updated now 👍

dougmacknz (approved):
Awesome, thanks for addressing everything! I'm not sure if we want a Components -> Brand documentation page for this, seeing as we already have guidelines for brand here: https://cultureamp.design/guidelines/brand/. We can address that later anyway.

🚢🚢🚢

PR #1731, opened 2021-06-29
docs: empty state updates

dlimiter:
# Objective
Updates to empty state guidance particularly around:

- responsive variants
- negative mood
- when to use/not use

Branch preview direct link: https://dev.cultureamp.design/dlimiter-empty-state-updates/components/empty-state/

### DIFF
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
@@ -39,22 +39,22 @@ import Dont from "docs-components/Dont"
 ## To keep in mind
 
 * There are different Empty States including:
-    * **Positive empty state**: nothing is here because everything is working/you’ve completed all tasks
-    * **Informative empty state**: things are happening and there are things to be done, but no direct action required
-    * **Action empty state**: there is an action the user can take right now
-    * **Neutral empty state**: there is nothing to show and no user action to take  
+    * **Positive empty state**: nothing is here because everything is working/you’ve completed all tasks.
+    * **Informative empty state**: things are happening and there are things to be done, but no direct action required.
+    * **Action empty state**: there is an action the user can take right now.
+    * **Neutral empty state**: there is nothing to show and no user action to take.
     * **Negative empty state**: 
-        * when we couldn’t find anything, such as a search (or filter) that returns no results
-        * To fill the space when a system error results in no page content and there is no other content on the page. If there is other content on the page consider using an [inline notification](/components/inline-notification) instead
+        * when we couldn’t find anything, such as a search (or filter) that returns no results.
+        * to fill the space when a system error results in no page content and there is no other content on the page. If there is other content on the page consider using an [inline notification](/components/inline-notification) instead.
 
 * They also come in three different sizes to accommodate responsive page layouts:
-    * Large: For viewports or containers larger than 1080px
-    * Medium: For viewports or containers between 375px and 767px
-    * Small: For viewports or containers up to 375px
+    * Large: For viewports or containers larger than 1080px.
+    * Medium: For viewports or containers between 375px and 767px.
+    * Small: For viewports or containers up to 375px.
 
 * Call To Actions (CTAs):
-    * We use Default Buttons unless it is the only action on the page, then we use Primary Action Buttons (label + icon)
-    * If no action is appropriate, don’t show a button. Instead, you might include hyperlinks in the body copy as needed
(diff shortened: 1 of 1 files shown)


draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
```diff
@@ -133,10 +135,10 @@ Your CTA button should compliment the title of the empty state. If a user scans
 Based on the overall user flow and the nature of the empty state, make sure you adjust your tone accordingly. A positive empty state should contain encouraging language, while an action empty state should get to the point quickly. Here are some helpful tips for each type of empty state:
 
 *   **Positive empty state** - Include words of affirmation here. If a user has nothing left to do after this point, let them know that. Be careful though, as this type of language can sometimes come off as condescending.
-*   **Informative empty state** - Reassure the user that work is being done in the background and that we’ll notify them when there’s data on that page for them to see or action. This is an opportunity to educate.
+*   **Informative empty state** - Reassure the user that work is being done in the background and that we’ll notify them when there’s data on that page for them to see or action. This is not used as a holding state to replace data that is in the process of being updated. This is an opportunity to inform.
```
lindsey-collins, line 138: This is a little confusing to me. What do we mean by "This is not used as a holding state..."? It contradicts the first sentence when we tell the user we'll notify them when there's something for them to do - isn't that kind of a holding state? Is letting the user know that there's data being updated not informing them?

draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
```diff
@@ -40,23 +40,25 @@ import Dont from "docs-components/Dont"
 
 * There are different Empty States including:
```
lindsey-collins, line 41: Moods above don't match the actual empty state moods listed below - should they?

lindsey-collins, line 41: We have options for border style listed at the top of the page but I've never seen us mix and match border style with the empty states and I don't think we should start doing that. Can we clear that information out?

lindsey-collins: This is also very picky but should the order of the list above (moods) and list below (empty state types) match?

lindsey-collins (commented):
I've made a couple of comments @dlimiter - just a few small bits that stood out to me, mostly some semantic stuff about this doc in general and one phrasing quibble.

draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
dlimiter: Ah the idea we're trying to convey is that is shouldn't be used to replace existing data while it's being updated or something. How about if I remove the reference to holding state like "This is not to be used to replace existing data that is in the process of being updated."?

draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
dlimiter: Haha good catch, I've updated the list

dlimiter: Yeah we're not even using them in the Heart states which have become a lot more concrete since then – I'll remove it.

draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
katsambarcus: That change helps, @dlimiter

draft-packages/empty-state/KaizenDraft/EmptyState/README.mdx
```diff
@@ -26,37 +26,35 @@ import Dont from "docs-components/Dont"
...
 *   Mood:
-    *   Default
     *   Positive
     *   Informative
-    *   Cautionary
+    *   Action
+    *   Neutral
     *   Negative
-*   Border style:
-    *   Solid
```
katsambarcus, line 35: None of our empty states have border styles anymore. Can you remove this reference altogether, 
@dlimiter

dlimiter: Ah yep that's the delta – it's removed 👍

katsambarcus (changes requested):
Have left one small comment to remove Border styles guidance.

katsambarcus (approved):

PR #1733, opened 2021-06-29
feat: Add Pagination component

JakePitman:
- Copies the Pagination component from Perf, and changes it to accept an onPageChange callback. This allows us to pass in the behaviour from outside the component (we will use this to update the current page in the parent state, and re-render the appropriate page)
- Adds a default Pagination story
- Also fixes a styling issue with the page indicators & nav arrows. This PR gets them in line with the design
focus-ring-fixed

*NOTE
This PR is only for getting the existing Pagination component into Kaizen. Behavioural or code-health improvements are out of scope


jakubriedl:
Is it expected to look like this after click?
Snímek obrazovky 2021-06-30 v 10 56 24 dop

draft-packages/pagination/docs/pagination.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const [currentPage, setCurrentPage] = React.useState(1)
+  return (
+    <Pagination
+      pageCount={50}
+      currentPage={currentPage}
+      onPageChange={(newPage: number) => {
+        setCurrentPage(newPage)
+        // fetch new page
+        // update currentPage value in state
+        // scroll to top of page
```
sampnorris, line 26: Remove these comments, not needed.

draft-packages/pagination/docs/pagination.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+      `,
+    },
+  },
+}
+
+export const basic = () => {
+  const [currentPage, setCurrentPage] = React.useState(1)
+  return (
+    <Pagination
+      pageCount={50}
```
sampnorris, line 20: Let's add a storybook control for this :)

sampnorris (changes requested):
@JakePitman - Let's pair on that bug we found where truncation was happening for when there were 3 pages.

dougmacknz:
@JakePitman @sampnorris @jakubriedl @allybeaton 

Disclaimer: I haven't reviewed this PR at all yet, only seen this part:

> This PR is only for getting the existing Pagination component into Kaizen. Behavioural or code-health improvements are out of scope

I'm going to be a bit of a party pooper here because we need to crack down on the tech debt we add to Kaizen. New component contributions to Kaizen are not something that should be done lightly. When taking on a new component contribution we need to focus on getting everything as "correct" as possible right out of the gate, having reviewed many different use cases around the platform, the documentation, designer input, etc. We absolutely want to avoid copy + pastes of a single implementation from a consuming repo, it ultimately creates a large mess for us down the line.

It might be the case that that comment is out of date because I see quite a few commits coming through here, let me know if that is the case. However, if you don't have capacity to create this component in a way that would be considered future-proofed for many different use cases, please just create this in your repo and we can come back to this at a time where we can invest in those things.

sampnorris:
@dougmacknz I completely understand the need,

We planned to move this component here then implement it in 2 places: 

1- We will replace the existing implementation where this was copied from
2- We will be using this in unified home

This _has_ been tweaked to consider multiple use cases, this should be usable anywhere - and this already is up to design standards, and seems to be pretty robust code.

Happy to pull this up to kaizen standards before merge, what are the requirements here for documentation, and accessibility, would you mind reviewing the current state and adding your input as to what we need to do to get this to standard?

sebpearce:
Just wanted to mention that there was a failure of communication here on our part ("our" meaning Kaizen advocates excluding @JakePitman).

For context, basically there was a group decision made a while back that new components should be built in consuming repos and polished until they meet certain standards required for Kaizen components around reusability, accessibility, etc. But we haven't done a good job of communicating that. I'm now thinking about how that info could be made obvious (including to us, since we forgot to mention it!).

sebpearce:
> Happy to pull this up to kaizen standards before merge, what are the requirements here for documentation, and accessibility, would you mind reviewing the current state and adding your input as to what we need to do to get this to standard?

@sampnorris The reason for the change in the way we do it (build it in consumer first) is that we don't actually know what a "ready for Kaizen" standard looks like until the thing has been used in prod for a while. The approach is basically "let's assume changes will need to be made, and make that easier on ourselves when it does happen".

draft-packages/pagination/KaizenDraft/PageIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+  selected,
+  ariaLabelPage,
+  onPageClick,
+}: PageIndicatorProps) => (
+  <button
+    key={`page-indicator-${page}`}
+    className={cx(styles.pageIndicator, {
+      [styles.pageIndicatorSelected]: selected,
+    })}
+    aria-label={`${ariaLabelPage} ${page}`}
```
dmisdm, line 23: I don't know how important this is, but `ariaLabelPage` is optional, and used here as if it isn't. It's possible that this template literal might spit out `undefined 1`, `null 2` etc.
```suggestion
    aria-label={`${ariaLabelPage || "Page"} ${page}`}
```

draft-packages/pagination/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/component-library": "^10.0.3",
+    "classnames": "^2.3.1",
+    "path-to-regexp": "^6.2.0"
+  },
+  "devDependencies": {},
+  "peerDependencies": {
+    "@kaizen/design-tokens": "^2.5.0",
```
dmisdm, line 41: I think the current version of design-tokens is 2.9.3. If theres no reason not to bump this, i think we should.
```suggestion
    "@kaizen/design-tokens": "^2.9.3",
```

dmisdm: I realised this was a peer sorry so probably doesn't matter!

draft-packages/pagination/KaizenDraft/Pagination.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const nextPageDisabled = currentPage >= pageCount
+
+  return (
+    <div className={styles.container}>
+      <button
+        className={cx(styles.arrowIconWrapper, {
+          [styles.arrowIconWrapperDisabled]: previousPageDisabled,
+        })}
+        aria-label={ariaLabelPreviousPage}
+        aria-disabled={previousPageDisabled}
```
dmisdm, line 143: I'm pretty sure `disabled` should be used in favor of `aria-disabled` as `<button>` natively supports the former.
Disagree with me though, i'm not 100% sure.

dmisdm (changes requested):
Left some comments but this is some awesome stuff 🤓😃

draft-packages/pagination/KaizenDraft/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.arrowIconWrapper {
+  height: 36px;
+  width: 36px;
+  border-radius: 18px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border: $kz-var-border-focus-ring-border-width solid $kz-var-color-cluny-200;
```
dmisdm, line 20: Your decision to make, but `$kz-var-border-focus-ring-border-style` can be used in place of `solid` (but maybe you want `solid` to be declared explicitly).
https://cultureamp.design/storybook/?path=/story/design-tokens-api--page

draft-packages/pagination/KaizenDraft/Pagination.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }
+  return range
+}
+
+type Props = {
+  currentPage: number
+  pageCount: number
+  ariaLabelNextPage?: string
+  ariaLabelPreviousPage?: string
+  ariaLabelPage?: string
```
dougmacknz, line 23: Similar to Dan's comment on `PageIndicator`, these aria label props should be required I'd imagine.

draft-packages/pagination/KaizenDraft/Pagination.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const pagination = () => {
+    const items: JSX.Element[] = []
+
+    // truncateSize is 1 now but could be 0 if we add the ability to hide it.
+    const truncateSize = 1
+
+    // Simplify generation of pages if number of available items is equal or greater than total pages to show
+    if (
+      1 + 2 * truncateSize + 2 * siblingPagesRange + 2 * boundaryPagesRange >=
```
dougmacknz, line 70: Feels like all the calculations happening here belong in a separate file with tests.

dougmacknz (commented):
Here's my thoughts so far. I haven't really reviewed the CSS because I would love to see us reuse the `IconButton` instead (more information in one of the review comments).

Let's have a chat about all this when you're free and we can talk about our options if you don't have capacity to take on addressing all of these.

Issue #1713, opened 2021-06-21
Toggle Switch guidance is missing


katsambarcus:
It appears the original guidance for **Toggle Switch** has gone AWOL. 

You used to be able to access it via two places: 

1. Guidelines > [Checkbox, Radio and Toggle Switch](https://cultureamp.design/guidelines/checkbox-radio-toggle-switch/) (click on Toggle Switch link takes you to a 'Not found' page.)
2. Components > Toggle Switch (which is missing completely from the component list)

PR #1737, opened 2021-07-01
docs: Move form readmes into component folders

dougmacknz:
Fixes https://github.com/cultureamp/kaizen-design-system/issues/1713

These readmes weren't displaying on the Kaizen site, something to do with the way they were named `README.CheckboxField.mdx`.

I've moved this into their component folder and just named them `README.mdx`. This is consistent with others and they are now appearing on the sidebar.

Side note: it would be great to have some categorising of our component documentation, something to keep in mind when we rethink our documentation.


ActuallyACat (approved):
:chef-kiss: Thanks @dougmacknz !

PR #1742, opened 2021-07-05
chore: Delete the hero-panel component and package

dougmacknz:
# Objective
Deletes the `hero-panel` component and package.

# Motivation and Context
This component doesn't exist in our UI Kit or our documentation. It was built for a particular usage in `performance-ui`, but that has since been deleted.

It's possible we would like a component like this in the future (Marc has added it to the designer board for triaging), but even still it makes more sense to delete what's here and come at that with a fresh take, rather than having to work with what was built here.

FWIW, here are a few issues that would need to be worked through if we kept this component, which has led to me saying we should just delete this:
* Package is called `draft` (this alone means we will need to create a new package eventually).
* Hard coded English
* API for the buttons only allows `onClick`, no `href` compatibility.
* Usage of the deprecated `Text` component.


ActuallyACat (approved):
Love seeing code being deleted <3

PR #1741, opened 2021-07-05
feat: Add new generation of token names

dmisdm:
This PR:
- Adds new colors that are in sync with Heart/the rebrand (e.g. Wisteria -> Purple, Ash -> Gray-300)
- Adds new tokens which don't contain the `kz-var` prefix - (we are now just using tokens without it e.g. `$color-purple-100`)
- Adds new tokens for RGB, with just a `-rgb` suffix (we're renaming `-rgb-params` to `-rgb`)
- Adds new tokens for the CSS variable ID using a `-id` suffix. Likewise to the previous point, we'll be deprecating `variable-identifiers.scss` in the future.
- Expose dataViz tokens in `color.(scss|less|json)`, as they weren't before.

This is part of the bigger plan to completely remove the old tokens, which we'll do in a later commit.
The rationale behind having both old and new in the same repo temporarily is that the alternative (e.g. pinning a certain version of design tokens in every package which depends on it and doing some release trickery) is a lot more awkward.

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer?
- [ ] If this introduces either a new component or a breaking change to an existing component, has it been reviewed by an advocate? (Ask in #prod_design_systems)
- [ ] I have considered likely risks of these changes and got someone else to QA as appropriate
- [ ] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?
- Have Storybook stories been updated?

**If you're new to Kaizen, please ask #prod_design_systems to set up an onboarding session to get you up to speed.** If you have an urgent PR to merge before that happens, it is safest to ask in #prod_design_systems to have a design system advocate review the PR to catch any issues.

PR #1743, opened 2021-07-05
feat: Deprecated and removed tokens rule uplift

dmisdm:
BREAKING CHANGE: Replaced two rules, `no-invalid-use-of-var-tokens-in-functions` and `prefer-var-tokens`, with a single one: `no-invalid-use-of-tokens`. Unnecessary exports from library files we're unexposed, and some exports we're renamed.

This PR allows the stylelint plugin to support the new generation of design-tokens, for example `kz-var-color-wisteria-100` -> `color-purple-100` etc (for more  details see https://github.com/cultureamp/kaizen-design-system/pull/1741).

One important bit of this is that it will now be able to know about tokens that don't exist within the current workspace version of `@kaizen/design-tokens`, by downloading tokens from an older version at build time. This is crucial because once we release the new version of design tokens that won't contain older ones, we still want this plugin to know how to recommend and replace old ones to new. If we didn't do this, the plugin would just ignore variables starting with `kz` or `kz-var` within stylesheets.

## Rationale behind the solution for knowing about deprecated tokens
The main problem at hand is that the stylelint plugin must know about tokens that might not exist within `@kaizen/design-tokens` at a later version. This is a real use case; it is practical to think about this problem because we are in fact going to release such a change, that is, remove a large portion of deprecated tokens from `@kaizen/design-tokens` and release version 3.
In ideation and experimentation I discovered the following solutions:
1. Depend on an aliased older version of design-tokens in `package.json`, e.g. something like: `"@kaizen/design-tokens-deprecated": "@kaizen/design-tokens@2.9.3"`.
  - Good: we can determine what all the deprecated tokens are in a concrete rather than heuristic way.
  - Good: we know about what all the tokens contain and where they are from.
  - Good: the plugin doesn't need to change too much to accommodate for it.
  - Bad: although the two effectively installed instances of design-tokens won't conflict, their transitive dependencies might.
  - Bad: depending on two versions of design-tokens (but this might not be impractical)
  - Bad: we only need the files in `tokens/*.json`, but we have to depend on the whole package.

2. Generate the names of the removed tokens using a grammar or chain of regular expression replacements.
  - Good: dependency-less. We don't need to depend on a previous version or file.
  - Bad: we won't know anything about the tokens, e.g. if they we're a CSS variable, or which module they came from.
  - Bad: code is a bit harder to comprehend because of the task's complexity.
  - Bad: the jest spec that tests whether this works will fail once we upgrade to a version of design-tokens that doesn't contain the deprecated tokens (unless we fetch them from NPM or a CDN during the test)

  2b. (An idea discovered late) Generate all bits of information about deprecated variables by doing almost the same thing as `buildCssVarTokens.ts` in `@kaizen/design-tokens`: the step that builds tokens from the theme declared in TypeScript. This means that we would be maintaining the functions that generate the old version of design tokens, and calling them upon initialisation of the stylelint plugin.

3. Fetch a certain deprecated version of all required design-token files from an NPM CDN  like unpkg at runtime, and cache them
  - Good: only depends on the relevant `tokens/*.json` files
  - Good: won't be subject to NPM dependency issues
  - Bad: non-negligible refactor of the plugin to support the async initialisation of the database of deprecated tokens, and comes at the cost of readability.
 
4. Do the same thing as the previous, but at build time.
  - Good: all the benefits of the previous step but without the refactor to support async.
  - Good: as opposed to any method which generates the tokens at runtime, these methods are very exact, i.e we can fetch design tokens from an exact version in time, and claim that the stylelint plugin "only knows about tokens from version X.Y.Z onwards".
  - Bad: another build step - but this isn't practically bad.

5. Generate a list of old to new token mappings at build time, and use only this to know how to make replacements.

~~At the moment, the solution implemented is 2, but this is subject to change in favor of 4 (which was discovered after implementing 2).~~

~~After experimenting with 2, 2b, and 4, 4 turned out to be the cleanest and easiest to understand. Adding an extra build step is almost negligible, there's no need to maintain code that we don't want sitting around anymore, and it's a much simpler solution than any of the others (we just download all deprecated token json files at build time so that they are require-able during runtime).~~

After a few iterations, we settled on this: Generate a map of `old token name` -> `new token name` and use this information to know how to make replacements. We generate this map using `./buildDeprecatedTokens.ts`, which downloads a previous version of design-tokens to generate the output file `deprecatedTokens.ts`. This also has the potential for custom token replacements if needed.

# Checklist

# Additional Considerations


packages/stylelint-plugin/src/rules/rename-colors.ts
```diff
@@ -0,0 +1,75 @@
...
+            options.reporter({
+              message: usageOfOldColorName(oldName, newName),
+              node: declOrAtRule,
+              autofixAvailable: true,
+            })
+          }
+        })
+      })
+
+      // Here is where we replace the postcss node with a new modified value, using the modified postcssValueParser nodes.
```
dmisdm, line 53: #
```suggestion
      // Here is where we replace the postcss node value with something new: a stringified postcssValueParser AST, which will have been mutated in the previous block that walked through each node.
```

packages/stylelint-plugin/README.md
```diff
@@ -32,7 +31,11 @@ These rules are included by default:
 - `kaizen/no-invalid-use-of-var-tokens-in-functions` - Detects invalid usages of Kaizen variable tokens within some functions that are known to be problematic such as `rgba`, `darken` or `mix`, and automatically fixes `rgba`, `rgb`, `add-alpha`, `transparentize` instances by ensuring the correct `-rgb-params` tokens are used.
 - `kaizen/no-invalid-use-of-var-tokens-in-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a runtime `calc()` function.
 - `kaizen/all-token-imports-must-be-used` - Reports and fixes unused Kaizen token imports.
-- `kaizen/tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
+- `kaizen/all-used-tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
```
ActuallyACat, line 34: Renaming a rule is a breaking change - someone may be overriding this rule in particular (either ignoring it completely or changing the default log to error) and this would break their pipeline.

dmisdm: Hey sorry i wasn't trying to rename it, i was trying to fix the README because currently the name of the rule is actually `kaizen/all-used-tokens-must-be-imported` as per the file `packages/stylelint-plugin/src/rules/all-used-tokens-must-be-imported.ts`

ActuallyACat: Oh - my mistake. Carry on then 😬

packages/stylelint-plugin/README.md
```diff
@@ -32,7 +31,13 @@ These rules are included by default:
...
+- `kaizen/rename-tokens` - Reports and automatically renames any ex-generation or deprecated color names, removes the `kz-var` prefix as it is not used anymore, renames `-rgb-params` to `-rgb`, and renames `kz-var-id-*` to `*-id`. E.g:
```
jasononeil, line 35: I think let's keep this sentence simpler and rely on the examples. We might add new use cases to this rule in future.

```suggestion
- `kaizen/rename-tokens` - Reports and automatically renames any deprecated color names to the newer name. E.g:
```

packages/stylelint-plugin/src/messages.ts
```diff
@@ -39,3 +39,6 @@ export const transitiveKaizenTokenUsage = (
   nameOfVariableThatContainsKaizen: string
 ) =>
   `Variable ${nameOfVariableThatContainsKaizen} contains a Kaizen token, and is not allowed because it prevents other rules from detecting invalid kaizen tokens`
+
+export const usageOfDeprecatedName = (oldName: string, newName: string) =>
+  `Usage of "${oldName}" is deprecated and should be renamed to "${newName}"`
```
jasononeil, line 44:
```suggestion
  `The token "${oldName}" is now called "${newName}" and should be renamed`
```

packages/stylelint-plugin/test.less
```diff
@@ -1,14 +1,12 @@
 // This is a file you can use to play with the stylelint plugin
-
 @import "~@kaizen/design-tokens/less/color-vars";
-
 @import "~@kaizen/design-tokens/less/spacing-vars";
```
jasononeil, line 3: This is interesting... are these imports going to change when we no longer use the `var` prefix everywhere?

packages/stylelint-plugin/test.less
```diff
@@ -1,14 +1,12 @@
...
 
 .test {
-  color: rgba(@kz-var-color-cluny-600-rgb-params, 80%);
-  background-color: rgb(@kz-var-color-cluny-600-rgb-params);
+  color: rgba(@kz-var-color-blue-600-rgb-params, 80%);
+  background-color: rgb(@kz-var-color-blue-600-rgb-params);
 
-  border-color: rgba(@kz-var-color-wisteria-800-rgb-params, 70%);
+  border-color: rgba(@kz-var-color-purple-800-rgb-params, 70%);
```
jasononeil, line 9: This whole file need a cleanup as they'll be stuck in an unfixable state (neither the old token names or the new ones)

packages/stylelint-plugin/test.scss
```diff
@@ -1,22 +1,17 @@
...
 @import "~@kaizen/component-library/styles/layout";
 
-$test-color: $kz-var-color-wisteria-800;
+$test-color: $kz-var-color-purple-800;
 
-/* $test-offset: -1 * ($kz-spacing-md + 1px);
- */
 .test {
-  background-color: rgb($kz-var-color-cluny-600-rgb-params);
+  background-color: rgb($kz-var-color-blue-600-rgb-params);
```
jasononeil, line 10: This whole file need a cleanup as they'll be stuck in an unfixable state (neither the old token names or the new ones)

packages/stylelint-plugin/src/rules/rename-tokens.ts
```diff
@@ -0,0 +1,98 @@
...
+} from "../util/walkers"
+
+const renameRules: Array<[RegExp, string]> = [
+  [/color-wisteria/, "color-purple"],
+  [/color-cluny/, "color-blue"],
+  [/color-yuzu/, "color-yellow"],
+  [/color-coral/, "color-red"],
+  [/color-seedling/, "color-green"],
+  [/color-peach/, "color-orange"],
+  [/color-white/, "color-neutral-100"],
```
jasononeil, line 18: Where did we land on this discussion?

packages/stylelint-plugin/src/rules/rename-tokens.ts
```diff
@@ -0,0 +1,98 @@
...
+      // Walk through each node within the declaration value. e.g. If the declaration was "prop: $kz-color-wisteria-800 $kz-color-cluny-500", two word nodes would be visited.
+      parsedValue.walk(node => {
+        renameRules.forEach(renameRule => {
+          const [oldPattern, replacement] = renameRule
+
+          // We only care about word nodes becasue only they can be variables.
+          if (node.type !== "word") return
+
+          // We don't care about nodes that don't contain old names.
+          if (!oldPattern.test(node.value)) return
```
jasononeil, line 49: I'm a fan of these inline comments on the guard lines 👏

packages/stylelint-plugin/src/rules/rename-tokens.ts
```diff
@@ -0,0 +1,98 @@
...
+
+          const replacementValue = node.value.replace(oldPattern, replacement)
+
+          const variable = parseVariable({ ...node, value: replacementValue })
+
+          // After we've made a replacement, DOUBLE CHECK whether the replaced variable is an existing kaizen token. If it's not, just return and don't report.
+          // We do this because the rename rules are not declared in a way where there is 100% certainty that the replacement will be correct 100% of the time.
+          // Rabbit hole: using the Theme type, and TypeScript's template literal power, we could probably strongly type those rename rules.
+          if (!variable?.kaizenToken) return
```
jasononeil, line 58: This made me go down a different rabbit hole... will the token changes here require changes to our existing stylelint rules?

For example:
- using `-rgb` instead of `-rgb-params`
- no longer needing a rule to `prefer variable tokens` etc

packages/stylelint-plugin/src/rules/rename-tokens.ts
```diff
@@ -0,0 +1,98 @@
...
+  [/color-yuzu/, "color-yellow"],
+  [/color-coral/, "color-red"],
+  [/color-seedling/, "color-green"],
+  [/color-peach/, "color-orange"],
+  [/color-white/, "color-neutral-100"],
+  [/color-stone/, "color-neutral-200"],
+  [/color-ash/, "color-neutral-300"],
+  [/color-iron/, "color-neutral-500"],
+  [/color-slate/, "color-neutral-600"],
+  [/kz-var-/, ""],
```
jasononeil, line 23: What are we going to do about the legacy `kz-` tokens (eg `kz-color-coral-400`). 

These were hardcoded values (hex for colours, etc). Are these going to exist in some form? (eg `color-coral-400-hex`)

And if they do, are we going to handle the replacements here?

jasononeil (changes requested):
I'm happy with the overall direction here.

I've got some questions to answer before approving (and a few tiny suggestions)

packages/stylelint-plugin/src/rules/rename-tokens.ts
jasononeil: This comment probably needs to go on #1741

packages/stylelint-plugin/test.scss
dmisdm: @jasononeil i think i might just delete these test files...

packages/stylelint-plugin/src/rules/rename-tokens.ts
dmisdm: Thanks for picking this up, I missed the need to change from -rgb-params to -rgb within the other rule.
Although, if you enable both rules (no invalid functions + rename tokens) you'll get the same effect.

dmisdm: For the most part, however, the rules work without any changes because of how we are preprocessing the variables in `utils/kaizenTokens.ts`. A token is declared deprecated if it is part of an old module like "color.scss" and not "color-vars.scss", and that's how the `prefer-variable-tokens` determines whether something should be reported/fixed (i think it used to be called `no-deprecated-tokens` or something).

The `-rgb-params` suffix was hard coded in the no invalid functions rule, along with a few other hard codings that i'm working now

packages/stylelint-plugin/README.md
```diff
@@ -21,18 +21,25 @@ Using this preset allows you to adopt the Kaizen rules easily, and allows the pa
 - It can autofix any variables that are used incorrectly within a `calc()` function, such as ones that aren't interpolated when they should be, or negated variables that need to be wrapped in `calc()` to support CSS variables (e.g. `-$kz-spacing-md` -> `calc(-1 * #{$kz-var-spacing-md}))`
 - Fixes alpha manipulation functions (`add-alpha`, `rgba`, `rgb`, `transparentize`) by using `-rgb-params` variable replacements with `rgba()`.
 - Warns you when you use a new CSS variable Kaizen token incorrectly (most of the time), e.g. when you use it in a function that doesn't support CSS variables such as `darken`, `add-tint`, `mix`, or a custom mixin/function.
-- Any usages of color manipulation functions such as `darken`, `add-shade`, `add-tint`, `mix` etc will be computed within the codemod and replaced with a HEX/RGB/HSL value, using the `sass` compiler.
```
dmisdm, line 24: Removing this was just part of cleanup - it doesn't happen any more

packages/stylelint-plugin/README.md
```diff
@@ -21,18 +21,25 @@ Using this preset allows you to adopt the Kaizen rules easily, and allows the pa
...
 
 These rules are included by default:
 
-- `kaizen/prefer-var-tokens` - Detects usages of non-themable tokens, and attempts to migrate them to our tokens that support theming with CSS variables.
+- `kaizen/no-deprecated-or-removed-tokens` - Reports and autofixes (if replacements exist) deprecated or completely removed tokens, for example: 
+  - `kz-color-cluny-100` -> `color-blue-100`
+  - `kz-var-color-wisteria-100 -> color-purple-100`
+  - `kz-var-spacing-md` -> `spacing-md`
+  - `kz-var-id-color-peach-100` -> `color-orange-100-id`
+  - `kz-var-color-seedling-500-rgb-params` -> `color-green-500-rgb`
```
dmisdm, line 35:
```suggestion
  - `kz-var-color-seedling-500-rgb-params` -> `color-green-500-rgb`
  - `kz-deprecated-color-lapis` -> no automatic replacement
```

packages/stylelint-plugin/README.md
```diff
@@ -21,18 +21,26 @@ Using this preset allows you to adopt the Kaizen rules easily, and allows the pa
...
+  - `kz-deprecated-color-lapis` -> no automatic replacement
 - `kaizen/no-invalid-use-of-var-tokens-in-functions` - Detects invalid usages of Kaizen variable tokens within some functions that are known to be problematic such as `rgba`, `darken` or `mix`, and automatically fixes `rgba`, `rgb`, `add-alpha`, `transparentize` instances by ensuring the correct `-rgb-params` tokens are used.
 - `kaizen/no-invalid-use-of-var-tokens-in-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a runtime `calc()` function.
 - `kaizen/all-token-imports-must-be-used` - Reports and fixes unused Kaizen token imports.
-- `kaizen/tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
+- `kaizen/all-used-tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
+
+
+
```
dmisdm, line 43:
```suggestion

```

packages/stylelint-plugin/test-config.js
```diff
@@ -0,0 +1,12 @@
+const baseConfig = require("./dev-config")
```
ActuallyACat, line 1: This is in the package's `.gitignore` file. Should it be committed?

ActuallyACat: Welp - ignore me. This is fine.

packages/stylelint-plugin/src/index.ts
```diff
@@ -1,15 +1,15 @@
...
     true,
     { severity: "warning", disableFixing: false },
   ],
-  "kaizen/no-invalid-use-of-var-tokens-in-functions": [
+  "kaizen/no-invalid-use-of-var-tokens-in-equations": [
     true,
     { severity: "warning", disableFixing: false },
   ],
-  "kaizen/no-invalid-use-of-var-tokens-in-equations": [
+  "kaizen/no-deprecated-or-removed-tokens": [
```
dmisdm, line 12: Moving this rule made the stylelint plugin work more reliably.
The order of the rules here affects which order they are applied in.

packages/stylelint-plugin/src/rules/no-deprecated-or-removed-tokens.ts
```diff
@@ -37,27 +40,23 @@ const disallowedAtRules = new Set([
...
 ])
 
-const getReplacementForDeprecatedToken = (token: KaizenToken) =>
-  kaizenTokensByName[token.name.replace("kz", "kz-var")]
-
 /**
  * This linter will report any deprecated tokens, and replace them with their new CSS variable replacement IF POSSIBLE.
  * If the linter determines it can't migrate, it will report an unmigratable declaration.
  *
  * Note: There is definitely some weirdness here. In order to clean it up, we really need a better value parser, in particular one that has a better hierarchical structure (one which we can go up and down between parents and children) and a better understanding of SASS and LESS constructs like negation and string interpolation.
```
dmisdm, line 47:
```suggestion
 * This linter will report any deprecated or removed tokens, and potentially replace them with new ones.
 * If the linter determines it can't automatically do a replacement, it will report an unfixable declaration (perhaps the replacement would cause an error).
 *
 * Note: There is definitely some weirdness here. In order to clean it up, we really need a better value parser, in particular one that has a better hierarchical structure (one which we can go up and down between parents and children) and a better understanding of SASS and LESS constructs like negation and string interpolation.
```

packages/stylelint-plugin/src/rules/no-deprecated-or-removed-tokens.ts
```diff
@@ -148,14 +147,50 @@ export const preferVarTokens: RuleDefinition = {
 
     walkAtRulesWithKaizenTokens(
       styleSheetNode,
-      ({ postcssNode, kaizenVariables }) => {
+      ({ postcssNode, kaizenVariables, parsedValue }) => {
+        let replacedNodeParams = postcssNode.params
         kaizenVariables.forEach(variable => {
           // We only care about deprecated tokens
```
dmisdm, line 153:
```suggestion
          // We only care about deprecated or removed tokens
```

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -1,37 +1,14 @@
-import { Root } from "postcss"
-import { Language, Options } from "./types"
-import { allUsedTokensMustBeImported } from "./rules/all-used-tokens-must-be-imported"
-import { allTokenImportsMustBeUsed } from "./rules/all-token-imports-must-be-used"
-import { preferVarTokens } from "./rules/prefer-var-tokens"
-import { noInvalidUseOfVarTokensInEquations } from "./rules/no-invalid-use-of-var-tokens-in-equations"
-import { noInvalidUseOfVarTokensInFunctions } from "./rules/no-invalid-use-of-var-tokens-in-functions"
-import { noTokensInVariables } from "./rules/no-tokens-in-variables"
-import { getParser } from "./util/utils"
+import { lint } from "stylelint"
```
dmisdm, line 1: The changes in this file might be pretty hard to comprehend.
To summarise:
 - Rename variables within expected outputs so that they match our new design-tokens
 - Rename imports too (because we're moving from color-vars to color etc)
 - In some cases, input source was actually invalid, for example non existent variables we're used. These needed to change because improvements we're made to the rules.

dmisdm: Oh and also, the number of expected warnings sometimes changed too.

dmisdm: Another important bit about this file: Once we break design-tokens, these tests will break too.
Reason being that the stylelint plugin will behave differently for tokens that don't exist within design-tokens (for example, it won't recommend importing scss modules that don't exist, or it will remove scss imports for old variables like `kz-color-wisteria-100`)

packages/stylelint-plugin/src/types.ts
```diff
@@ -19,16 +19,39 @@ export type Options = {
...
 }
 
+/**  Represents a Kaizen token that is deprecated, but still exists. */
+export type DeprecatedKaizenToken = CurrentKaizenToken & {
+  deprecated: true
+  removed: false
+}
+
+/** Represents a Kaizen token that is both removed/non-existent and deprecated.
+ * We can't do much with these because we don't know anything about them apart from their name and a potential replacement. */
```
dmisdm, line 41:
```suggestion
 * We can't do much with these because we don't know anything about them apart from their name, value, and a potential replacement. */
```

packages/stylelint-plugin/src/util/kaizenTokens.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+      Object.entries(kaizenTokensByName).forEach(([key, value]) => {
+        if (!value) return
+        const replacement = getReplacementForDeprecatedOrRemovedToken(
+          value.name
+        )
+        if (!replacement) return
+        if (isKaizenTokenDeprecatedOrRemoved(replacement.name)) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `Kaizen token: ${value.name}, is deprecated and has a replacement of ${replacement.name}, but the replacement is also deprecated`
```
dmisdm, line 19: This warning is pretty good at describing what the expectation is.

packages/stylelint-plugin/src/util/variableUtils.ts
```diff
@@ -37,11 +42,13 @@ export const parseVariable = (node: WordNode): Variable | null => {
...
   if (firstChar === "@" || firstChar === "$") {
     const name = cleanedValue.substr(1)
+    const foundRemovedToken = removedKaizenTokensByName[name]
+    const foundKaizenToken = kaizenTokensByName[name]
     return {
       name,
       nameWithPrefix: cleanedValue,
       prefix: firstChar,
-      kaizenToken: kaizenTokensByName[name],
+      kaizenToken: foundKaizenToken || foundRemovedToken,
```
dmisdm, line 51:
```suggestion
      // Important bit here. If a token exists in the consumers version of `@kaizen/design-tokens`, it will take precedence over ones that we know are removed.
      kaizenToken: foundKaizenToken || foundRemovedToken,
```

packages/stylelint-plugin/test.scss
```diff
@@ -1,30 +0,0 @@
-@import "~@kaizen/design-tokens/sass/spacing";
```
ActuallyACat, line 1: Ah, glad to see this removed - it's been throwing false positives for me in the prepush hook.

jasononeil (changes requested):
I've left a big review 😅 

I've left comments and suggestions that I think would get this PR ready in its current form, but after going through it I'd like to explore an "Option 5" for how we know about deprecated tokens - a plain old hard-coded list (rather than trying to generate them programmatically).

There are draw-backs to hard coding (needing to create it manually, for one!), but I think we can reduce the complexity of this rule significantly by going for a simpler architecture - so it's worth seeing if that approach will work.

I'm not sure how much my other 16 comments will be relevant if we go down that road :)

packages/stylelint-plugin/package.json
```diff
@@ -23,16 +23,21 @@
     "postcss-less": "^4.0.1",
     "postcss-scss": "^3.0.5",
     "postcss-value-parser": "^4.1.0",
+    "prettier": "^2.3.2",
```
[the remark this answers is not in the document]
dmisdm: Oops thanks!

packages/stylelint-plugin/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "cli": "yarn build && node ./dist/src/cli.js",
     "test": "cd ../../ && yarn jest ./packages/stylelint-plugin",
-    "build": "yarn tsc --project tsconfig.dist.json",
+    "build": "./buildRemovedTokens.ts && yarn tsc --project tsconfig.dist.json",
```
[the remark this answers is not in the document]
dmisdm: yep i was half expecting this comment!

packages/stylelint-plugin/README.md
```diff
@@ -21,18 +21,23 @@ Using this preset allows you to adopt the Kaizen rules easily, and allows the pa
...
 These rules are included by default:
 
-- `kaizen/prefer-var-tokens` - Detects usages of non-themable tokens, and attempts to migrate them to our tokens that support theming with CSS variables.
+- `kaizen/no-deprecated-or-removed-tokens` - Reports and autofixes (if replacements exist) deprecated or completely removed tokens, for example: 
+  - `kz-color-cluny-100` -> `color-blue-100`
+  - `kz-var-color-wisteria-100 -> color-purple-100`
+  - `kz-var-spacing-md` -> `spacing-md`
+  - `kz-var-id-color-peach-100` -> `color-orange-100-id`
+  - `kz-var-color-seedling-500-rgb-params` -> `color-green-500-rgb`
+  - `kz-deprecated-color-lapis` -> no automatic replacement
```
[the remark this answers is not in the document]
dmisdm: I guess its a subjective thing hey!
Personally i don't think the README is large enough yet for us to be picky about real estate.
But, mulling over your comment a bit more i tend to agree. It's also more future proof to not include those specific examples.

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-functions.ts
```diff
@@ -69,51 +74,46 @@ const parsePercentage = (value: string): number | "NaN" => {
...
+  } else if (variable.kaizenToken.deprecated && replacement) {
+    kaizenTokenSubject = replacement
+  } else if (
+    !variable.kaizenToken.deprecated &&
+    !variable.kaizenToken.removed
+  ) {
+    kaizenTokenSubject = variable.kaizenToken
+  } else {
+    // Just bail if we reach here, because we couldn't decide on a kaizen token to use.
+    return undefined
```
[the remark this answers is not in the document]
dmisdm: This relates to the difficult separation of concerns within this rule and the `no-deprecated-or-removed-tokens` rule.
I've tried to make them only report on what they're responsible for, and this final condition of there not being a replacement for deprecated/removed tokens is not this rule's responsibility to report on anything.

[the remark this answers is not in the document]
dmisdm: I'll change this a bit to clear up the confusion

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-functions.ts
```diff
@@ -242,13 +242,13 @@ export const declContainsInvalidFunctions = (
...
 ) => {
-  let reported = 0
-  noInvalidFunctionsOnDeclaration(postcssNode, {
+  let reportedUnfixables = 0
+  noInvalidFunctionsOnDeclaration(postcssNode.clone(), {
     ...options,
-    fix: false,
-    reporter: () => {
-      reported++
+    fix: true,
```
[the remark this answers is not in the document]
dmisdm: Oh i think this is leftover from an experiment, thanks for pointing it out. It doesn't make a difference either way though.
The goal here was to only return true if something unfixable came up. Previously we we're returning true if any reports we're made at all, and setting `fix: true` accomplishes the same thing as line 250 where we condition on whether the report was unfixable.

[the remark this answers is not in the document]
dmisdm: The only place this is used in is the no-deprecated-or-removed-tokens rule to determine whether it should continue to replace a deprecated token.

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-functions.ts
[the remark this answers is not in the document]
jasononeil: Yeah the blurring of those rules is unfortunate.

I can't remember - did you previously try an approach to have the "no-deprecated-or-removed-tokens" rule do a naive find/replace in the first pass, and then have the "no-invalid-use-in-functions" rule detect where it ended up with something bad?

I feel like the reason we didn't do that before is because using the non-css-variable tokens in those functions remained an option. But soon we won't have those, so I wonder if keeping the rules more separate is possible.

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -660,31 +830,30 @@ describe("Codemod", () => {
...
-          }
-        },
       })
+      if (result.results[0]?.warnings.length !== expectedWarnings) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `Unexpected warnings for test: ${language}: ${testName}`,
+          result.results[0]?.warnings
+        )
+      }
```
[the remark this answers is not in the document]
dmisdm: This will tell you what was reported if there is a mismatch in expected warnings.
The `expect` doesn't tell you anything about the reason behind the warnings.
Still think it should go? I think it helps to debug failing tests.

packages/stylelint-plugin/src/util/kaizenTokens.spec.ts
dmisdm: I gave this a shot but it ended up with ~660 more tests - not necessarily a bad thing.

I interpreted this test as more of a _property based_ test, meaning, an assertion on the function over the input domain (not every possible string obviously, but just the ones that are the most likely which are all Kaizen token names).
I'm not sure what the most reasonable thing to do for Jest is. Property based happens in other languages/ecosystems i think it's a single test.
That being said, you could interpret this in the same way as the list of inputs + outputs in the spec that tests the whole plugin.
Because we're not hand coding the list of test cases, i.e. hand picking and declaring the inputs + outputs, i think it makes sense to keep this as a single test.

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -1,8 +1,46 @@
...
+ *          leaf: "some value"
+ *        }
+ *      }
+ *    }
+ * }
+ * ```
+ * ```
+ * Output:
+ * {
+ *    "--level1-level2-level3-leaf": "some value"
```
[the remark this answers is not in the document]
dmisdm: Oh good pickup thanks!

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -90,8 +111,78 @@ const parseCssVariableValue = (value: string): CSSVariable | undefined => {
...
+ * Ensure that any pattern on the left side represents a token that is deprecated (i.e `pattern.test(tokenName) === true` means deprecated)
+ * Also please ensure that idempotency is preserved when applying rules.
+ * I.e. if you applied every rule once on a string, you would get the same result if you re-ran the rules again.
+ * We want idempotency because, otherwise, applying each rule on a token won't always return a non-deprecated token (because of the first constraint)
+ * Order of the rules matters.
+ *
+ * The last value in the triple is a boolean that represents whether the rules should terminate at that rule, if it tests true.
+ *
+ */
+export const version2DeprecationRules: Array<[RegExp, string?, boolean?]> = [
```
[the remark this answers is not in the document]
dmisdm: I think i was attempting to test this at some point, agree it shouldn't be exported

packages/stylelint-plugin/src/util/variableUtils.ts
```diff
@@ -37,11 +42,14 @@ export const parseVariable = (node: WordNode): Variable | null => {
...
     const name = cleanedValue.substr(1)
+    const foundRemovedToken = removedKaizenTokensByName[name]
+    const foundKaizenToken = kaizenTokensByName[name]
     return {
       name,
       nameWithPrefix: cleanedValue,
       prefix: firstChar,
-      kaizenToken: kaizenTokensByName[name],
+      // Important bit here. If a token exists in the consumers version of `@kaizen/design-tokens`, it will take precedence over ones that we know are removed.
+      kaizenToken: foundKaizenToken || foundRemovedToken,
```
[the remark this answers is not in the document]
dmisdm: I could have really missed the mark on this PR i do apologise.

What i was trying to do was support both versions 2.10.x and 3.x of design-tokens, where version 3 won't contain the older tokens.
I may have lost sight of the fact that the only rule we want to know about removed tokens is the `no-deprecated-or-removed-tokens` rule, and i guess i thought being able to handle `CurrentKaizenToken | DeprecatedKaizenToken | RemovedKaizenToken` throughout every rule equated to more versatility and robustness.

dmisdm:
@jasononeil your comments really illuminated that i might have overblown the intended responsibilities of this plugin, so i'm just having a think about how to best refactor it. Hard coding the deprecations and replacements might be the best way for sure.

packages/stylelint-plugin/src/index.ts
dmisdm: This relates to the overarching issue with the PR: too much responsibility.
This reordering was because of the murkiness, which i'm trying to clear up.
Putting no-invalid-func before no-deprecated-or-removed-tokens meant that no-deprecated-or-removed-tokens wouldn't bail early because it detected an rgba function with an incompatible token.

Honestly, the splitting of these two rules was done purely because we didn't want no-deprecated-or-removed-tokens to be too un-cohesive. Now i kind of regret that as a premature attempt to make things cleaner.

dmisdm:
@jasononeil What do you think of this:
- In a new rule called `no-removed-tokens`, find all removed tokens and naively replace them. This would be the first rule in the pipeline, therefore any subsequent rules should only see existing tokens. Doing this means we don't need to know anything about variables apart from whether  they are removed, and what their replacement is. It also means `no-deprecated-tokens` is somewhat useless for V3 because there will be no deprecated ones at first; it will only be useful for people still on design-tokens version 2 which contains both (or only the old), and during migration which we are about to do.
- Don't support removed tokens throughout the rules, thus changing the `type KaizenToken` to just `CurrentKaizenToken | DeprecatedKaizenToken`, AND removing any need for knowing more information about removed tokens. We have the ability to determine whether a token is removed just by using the set of regular expressions and confirming it's non-existence in `@kaizen/design-tokens`. The only downside to that is: users may have their own variables declared with similar patterns, e.g. something like `$kz-var-color-wisteria-100-custom`. So, a hard coding or script-built list might be better.

The murkiness in `no-invalid-use-of-var-tokens-in-functions` is a rabit hole. If we remove any notion of deprecated tokens within it, the rule will only know how to go from `rgba($color-purple-100, x)` to `rgba($color-purple-100-rgb, x)`. It WON'T be able to go from `rgba($kz-var-color-wisteria-100)` to `rgba($color-purple-100-rgb, x)` like it currently does. In fact, it would go from `rgba($kz-var-color-wisteria-100)` to `rgba($kz-var-color-wisteria-100-rgb)`, then `no-deprecated-tokens` won't do any replacements because it will pickup that it's within an `rgba` function - they compete with each other a lot. Also, fixing this competition is only useful for people on version 2, given the first point where we would no longer be "supporting" removed tokens, and there won't be any deprecated ones in V3.

dmisdm:
Btw, I'm currently toying with the idea/implementation of having a single rule which handles 3 things:
- removed tokens
- deprecated tokens
- invalid tokens within functions

Do we really need 3 separate rules that work independently? Or do we just want maintainable and readable code? Because they don't necessarily mean the same thing.

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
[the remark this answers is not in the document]
jasononeil: Good point - I saw you do this elsewhere and think it can make sense, but forgot to come back to this comment

dmisdm:
Here's where i'm at now:
If we want to split the rules nicely, i think we would need to sacrifice the constraint that **no single rule should knowingly produce incorrect code**. To be more concrete: if the `no-deprecated-tokens` rule was able to transform `rgba($kz-color-wisteria-100, x)` to `rgba($color-purple-100, x)`, which fulfills it's responsibility, it would violate the constraint because the correct replacement would be `$color-purple-100-rgb`. However, the subsequent rule for detecting invalid tokens in functions would pick this up.
...hence the awkwardness of this: why would you have one rule enabled without the other? And, the tagline would be "if you don't enable the invalid functions rule, the no deprecated tokens rule might produce incorrect code sometimes".
Responsibilities also conflict when allowing the invalid functions rule to migrate deprecate tokens.

This is the core reason why i think we need to merge them into a single rule.

dmisdm:
While it's fresh in my head i thought i'd document something:

image

In this case, the equation containing `$kz-var-layout-*` could be safely replaced with `$layout-*` because the latter aren't CSS variables.

The equation rule logic does however know that `$layout-*` is allowed within equations, it's just not aware that it's potential replacement is valid, so it reports and bails out of everything.
This is something that should be fixed up when merging these two rules (no-invalid-use-of-tokens and no-invalid-use-of-var-tokens-in-equations) eventually, because the fact that the rules run seperately is the show stopper in making a clean connection.

dmisdm:
Another current limitation which is due to the way stylelint processes nodes:

image

There is supposefly an autofix for this, however:

image

This test fails, and when run with `--fix` nothing happens.

After debugging and stepping through, everything seems to work as expected, however from experience of this issue before it seems to be due to the order in which operations are made within the postcss walker, where some are overwritten.

dmisdm:
Another limitation due to the limitations of postcss-value-parser:

image

When interpolations have more than one node/word within them, they're not picked up as a Kaizen token, and therefore these instances are not migrated.

packages/stylelint-plugin/src/rules/all-token-imports-must-be-used.ts
```diff
@@ -1,7 +1,7 @@
 import { Root } from "postcss"
 import { unnecessaryKaizenImport } from "../messages"
-import { getCurrentImports, removeImport } from "../util/importUtils"
 import { KaizenToken, Options, RuleDefinition } from "../types"
+import { getCurrentImports, removeImport } from "../util/importUtils"
```
dmisdm, line 4: This is just me running "organise imports"

packages/stylelint-plugin/src/rules/lib/tokensInFunctions.ts
```diff
@@ -0,0 +1,202 @@
...
+// Does the value look like `123, 123, 123`? Used for determining whether an rgba function is used with the correct variable
+const isRgbTriple = (value: string) =>
+  /^\s*\d{1,3}\s*,\s*\d{1,3}\s*,\s*\d{1,3}\s*$/.test(value)
+
+/**
+ * Given a CSS declaration value, find any alpha modifying functions like add-alpha, transparentize, or rgba, then normalise them to either rgba or rgb,
+ * and ensure that a RGB triple Kaizen token is being used, e.g. `$color-purple-100-rgb` rather than `$color-purple-100`.
+ * It also knows how to migrate from a deprecated variable within a function, e.g. `$kz-color-wisteria-100` -> `$color-purple-100-rgb`
+ */
+export const fixAlphaModificationFunctions = (
```
jasononeil, line 103: This is the only export... should we call the file `fixAlphaModificationFunctions.ts`?

Then it avoids becoming a grab-bag utils file

jasononeil (commented):
Submitting WIP review, I'm still going

packages/stylelint-plugin/src/rules/lib/tokensInFunctions.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+
+/**
+ * Given a string like "50%", "40%", "30", "15" etc (you don't have to pass in the percentage sign), return a decimal like 0.5, 0.4, 0.3 etc
+ */
+const parsePercentage = (value: string): number | "NaN" => {
+  // Add-alpha can be used without a percentage sign, that won't fly.
+  // Fix the percentage sign, then convert to a decimal for usage in RGBA
+  const parsed = parseInt(value.replace(/\%/, ""), 10)
```
jasononeil, line 42: Should we do parseFloat?

```suggestion
  const parsed = parseFloat(value.replace(/\%/, ""))
```

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -0,0 +1,354 @@
...
+  stringifyVariable,
+} from "../util/variableUtils"
+import { walkVariablesOnValue } from "../util/walkers"
+import { fixAlphaModificationFunctions } from "./lib/tokensInFunctions"
+import { containsEquationThatDoesntWorkWithCSSVariables } from "./no-invalid-use-of-var-tokens-in-equations"
+
+const deprecatedKaizenTokenPattern = new RegExp(
+  Array.from(deprecatedTokenReplacements.keys()).join("|")
+)
+export function stringContainsDeprecatedKaizenToken(value: string) {
```
jasononeil, line 29: Check if this needs to be exported
```suggestion
function stringContainsDeprecatedKaizenToken(value: string) {
```

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * @returns a new string value if there we're changes, or undefined if there weren't any
+ */
+function detectAndFixInvalidTokens(
+  nodeValue: string,
+  postcssNode: Declaration | AtRule,
+  options: Options
+): string | undefined {
+  let parsedValue = postcssValueParser(nodeValue)
+  let newValue = nodeValue
```
jasononeil, line 100: I wonder if `currentValue: string` is better than newValue, because it's not necessarily different? Hard to say

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -0,0 +1,354 @@
...
+          replacementKaizenToken
+        )
+        variableNode.value = stringifyVariable(replacementVariable)
+
+        newValue = postcssValueParser.stringify(parsedValue.nodes)
+      }
+    }
+
+    // Now decide whether an autofix can occur using a series of rules.
+    // If we decide that it can't happen, report the reason then return so that we move on to the next node.
```
jasononeil, line 226: I'm appreciating the level of commentary in this function

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -0,0 +1,354 @@
...
+    // (because the only change which isn't safe is going from NonCSSVariable -> CSSVariable)
+    // The process will only reach here if a safe change is not available, due to the checks above.
+    // If there isn't, just report, and bail.
+    if (postcssNode.type === "decl" && isVariable(postcssNode)) {
+      report({
+        message: deprecatedTokenInVariableMessage(variable.name, replacement),
+        autofixAvailable: false,
+      })
+      return
+    }
```
jasononeil: Moving this rule up so it comes directly after the check about if the current value _is_ a CSS Variable, or if the new values _is not_ a CSS variable, will make it easier to match what the comment is saying to what the control flow in the code is doing.

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -0,0 +1,354 @@
...
+        message: genericContainsDeprecatedKaizenTokenMessage,
+        autofixAvailable: false,
+        node: postcssNode,
+      })
+    }
+  }
+
+  // Return undefined if no changes we're made
+  if (!changed) return undefined
+  else return newValue
```
jasononeil, line 321: You don't really need `else return` here, `return` is fine

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -537,57 +537,57 @@ const testExamples: TestExample[] = [
...
-      @import "~@kaizen/design-tokens/sass/color-vars";
-      $foo: $kz-var-color-wisteria-800;
+      @import "~@kaizen/design-tokens/sass/color";
+      $foo: $color-purple-800;
       .foo {
-        color: mix($kz-var-color-wisteria-800, $white, 80%);
+        color: mix($color-purple-800, $white, 80%);
       }`,
-    expectedWarnings: 3,
+    expectedWarnings: 1,
```
jasononeil, line 585: One thing I struggle with in this test suite... what's the 1 warning I'm expecting?

What are the two I'm no longer expecting? It's hard to know what behaviour change is expected by this diff

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -859,6 +859,75 @@ const testExamples: TestExample[] = [
...
+    language: "scss",
+    input:
+      '@import "~@kaizen/design-tokens/sass/layout-vars"; $foo: $kz-var-layout-breakpoints-large @media (max-width: $foo){}',
+    expectedOutput:
+      '@import "~@kaizen/design-tokens/sass/layout"; $foo: $layout-breakpoints-large @media (max-width: $foo){}',
+    expectedWarnings: 0,
+  },
+  {
+    testName:
+      "tokens within functions can be replaced when the replacement is not a CSS variable",
```
jasononeil, line 908: Should this be "when the existing token was also a CSS variable"?

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -164,13 +164,29 @@ const testExamples: TestExample[] = [
   },
   {
     language: "scss",
-    testName: "doesn't fix functions other than rgba, rgb, or add-alpha",
+    testName:
+      "knows not change a token to a CSS variable within a function that doesn't support them",
```
jasononeil, line 168: know not _to_ change

jasononeil (approved):
I'm done! 

It would be good to ensure any of my comments are addressed, but at this point they're all nitpicks, so I'm approving.

Keen to see this merged :shipit:

PR #1749, opened 2021-07-09
feat: Allow the Avatar to handle the case where the fullname cannot be provided

sugendran:
# Objective
Handle fallback case where the the fullname cannot be provided and the default icon should get shown.

# Motivation and Context
During an update of performance-ui it was found that there are a few places where the value may not exist.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
+++ b/draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
@@ -7,15 +7,17 @@ import styles from "./styles.module.scss"
 
 type AvatarSizes = "small" | "medium" | "large" | "xlarge"
 
-const getInitials: (fullName: string, max2Characters?: boolean) => string = (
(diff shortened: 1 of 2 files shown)


jasononeil:
Good catch!  

> Handle fallback case where the the fullname cannot be provided and the default icon should get shown.

What is this use case? I'm asking because we probably care about the alt text. Our [docs say](https://cultureamp.design/components/avatar/) we should use the name for alt text, but the fallback icon currently has no alt text.

We probably want to fix this, but I'm unsure why the name isn't known, so unsure what we should make the alt text in that case. 

Your PR doesn't make the status quo worse (fallback avatars already had no alt text) but it might make that broken case more common. I'm going to approve, but if you don't need this merged today it would be good to think through the alt text situation.

jasononeil (approved):
Approving, but we should follow through on the alt text - either in this PR or by creating an issue to follow up later

sugendran:
There's a few cases in here if you want some examples https://github.com/cultureamp/performance-ui/pull/3105

PR #1753, opened 2021-07-13
feat: Remove title from back of Tile component

dabarrell:
https://cultureamp.slack.com/archives/C0189KBPM4Y/p1623800537096700

# Objective

Remove the title from the back of the Title component.

# Motivation and Context

Removing the title from the back increases the available space for content. This was prompted by a desire to show a "View available languages" button for survey template tiles, but we're likely to run into space issues pretty regularly.

I've also left-aligned the text, at Marc's direction.

Context: https://cultureamp.slack.com/archives/C0189KBPM4Y/p1623800537096700

Before:
image

After:
image

UI Kit:
image

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/tile/KaizenDraft/Tile/__snapshots__/Tile.spec.tsx.snap
+++ b/draft-packages/tile/KaizenDraft/Tile/__snapshots__/Tile.spec.tsx.snap
@@ -205,24 +205,6 @@ exports[`<InformationTile /> snapshots renders InformationTile with information 
         </span>
       </div>
       <div
-        class="title"
-      >
-        <h3
-          class="heading heading-3 dark small"
-        >
-          Title
-        </h3>
-        <div
-          class="pt-0-point-25 m-0"
-        >
-          <p
-            class="paragraph small dark-reduced-opacity"
-          >
-            Metadata
-          </p>
-        </div>
-      </div>
-      <div
         class="information"
       >
         <p
--- a/draft-packages/tile/KaizenDraft/Tile/components/GenericTile.scss
+++ b/draft-packages/tile/KaizenDraft/Tile/components/GenericTile.scss
@@ -63,6 +63,7 @@ $tilePaddingTop: $kz-var-spacing-xl;
 .faceBack {
   padding-top: $tilePaddingTop;
   transform: rotateY(180deg);
+  background-color: $kz-var-color-cluny-100;
 }
 
 .isFlipped {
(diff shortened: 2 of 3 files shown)


draft-packages/tile/KaizenDraft/Tile/__snapshots__/Tile.spec.tsx.snap
```diff
@@ -177,10 +177,10 @@ exports[`<InformationTile /> snapshots renders InformationTile with information
         >
           <button
             aria-disabled="true"
-            aria-label="Information"
+            aria-label="Back"
             class="button iconButton"
             disabled=""
-            title="Information"
+            title="Back"
```
makenosound, line 183: "Back" feels a bit anaemic here (though better than "Information" which never really made sense), wondering if @dougmacknz has thoughts on how we could better communicate to screen readers what’s happening here.

makenosound (dismissed):
Left a small comment about the "Back" label, but otherwise the changes look fine to me, @dabarrell.

Wondering if we we should make this a breaking change is all, since it removes some context that was there before and we probably want to prompt people to think about what changed before blindly updating /cc @allybeaton

dabarrell:
@makenosound Your comments about breaking change makes sense. Not sure how the communication etc changes if it is made a breaking change, so will need some guidance there if we do that

makenosound:
Main thing there is posting in `#updates_design_system` with any details, and ensuring the commit message in the CHANGELOG make the things to consider clear.

dabarrell:
Okay, thanks. I'll wait for @allybeaton 's input :)

draft-packages/tile/KaizenDraft/Tile/__snapshots__/Tile.spec.tsx.snap
dougmacknz: I don't think there's any way to have this interaction be good for screen readers. The best solution I see here is that we hide the fact that there's an interaction at all, and just append the extra information to the DOM as _visually hidden_.

That would remove the need for this label at all, which means we don't have to worry about localising it as well.

makenosound: That seems fair, though perhaps outside the scope of @dabarrell’s PR. Reckon an `#accessibility` tagged issue for someone to follow-up could be a way to go, @dougmacknz?

ActuallyACat:
This is definitely a breaking change - Could you make sure one of your commits contain "BREAKING CHANGE: " (even by pushing an empty commit) so that conventional commits can version it correctly. 

I've looked through the threads, and it seems like the designers are all aligned on this change.

draft-packages/tile/KaizenDraft/Tile/__snapshots__/Tile.spec.tsx.snap
dougmacknz: Created https://github.com/cultureamp/kaizen-design-system/issues/1764.

I would keep this as "Information" for now @dabarrell, "Back" makes less sense I would say.

dabarrell: I've reverted this :)

dabarrell:
Thanks @allybeaton! I've added that to the commit message :)

makenosound (approved):

PR #1758, opened 2021-07-14
chore: Move brand moment stories out of src folder

dougmacknz:
Just noticed that I put the stories for this inside of the `src` folder, which is not how it is for the other packages. There's also a reference to a `docs` folder in the tsconfig file, so this will fix that here.


ellisjas (approved):

packages/brand/docs/Brand.stories.tsx
```diff
@@ -127,3 +127,16 @@ CollectiveIntelligenceReversed.story = {
     ),
   },
 }
+
+export const WithContainerWidth = () => (
```
dougmacknz, line 131: Oops, didn't mean to commit this.

ellisjas (approved):

PR #1763, opened 2021-07-15
fix: Constrain brand width to 100% of container

dougmacknz:
# Objective
Constrains the size of the image on the Brand component to 100% of the container width.

# Motivation and Context
When implementing the Brand component into the new minimal nav, I found that the logo was going over the container width. This fixes that.


lloydstubber (approved):
Good catch!

ActuallyACat (approved):

PR #1765, opened 2021-07-15
feat: forward refs through select component

talbet:
# Objective

We should allow refs to be passed through all components, but this is particularly important for elements that we are using in forms. Previously we had not allowed this in the Select component, even though the underlying library `react-select` does allow this.

Because `ref` is handled differently to other props, we need to wrap the components in a ForwardRef function. The forwardRef function provides a mutableRef type and `react-select` expects this to be an immutableRef, so we also have to type the arguments in a way that satisfies `react-select` rather than using generic params in forwardRef.

Because we are wrapping the component with forwardRef, we also need to set the displayName, as React will not be able to infer it. This will help with debugging, ensuring the components do not appear as `forwardRef(Unknown)` in the dev tools.

# Motivation and Context

Team Earthlings is testing `react-hooks-form` which tries to use uncontrolled form components wherever possible. When that is not possible (as in the case of `react-select` it still needs to pass a ref into the form component.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/select/package.json
+++ b/draft-packages/select/package.json
@@ -38,9 +38,9 @@
     "@kaizen/draft-form": "^3.6.3",
     "@kaizen/draft-tag": "^1.9.3",
     "@types/classnames": "^2.2.10",
-    "@types/react-select": "^3.0.22",
+    "@types/react-select": "^4.0.17",
     "classnames": "^2.2.6",
-    "react-select": "^3.1.0"
+    "react-select": "^4.3.1"
   },
   "devDependencies": {
     "elm-storybook": "cultureamp/elm-storybook#0.3.0",
--- a/draft-packages/select/KaizenDraft/Select/Select.tsx
+++ b/draft-packages/select/KaizenDraft/Select/Select.tsx
@@ -11,7 +11,7 @@ import clearIcon from "@kaizen/component-library/icons/clear.icon.svg"
 import { Tag } from "@kaizen/draft-tag"
 import styles from "./styles.react.scss"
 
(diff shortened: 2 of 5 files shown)


draft-packages/select/package.json
```diff
@@ -38,9 +38,9 @@
     "@kaizen/draft-form": "^3.6.3",
     "@kaizen/draft-tag": "^1.9.3",
     "@types/classnames": "^2.2.10",
-    "@types/react-select": "^3.0.22",
+    "@types/react-select": "^4.0.17",
     "classnames": "^2.2.6",
-    "react-select": "^3.1.0"
+    "react-select": "^4.3.1"
```
ActuallyACat, line 43: I was a little concerned about this change, but it doesn't appear any of the breaking changes are applicable to our implementation: https://github.com/JedWatson/react-select/blob/master/packages/react-select/CHANGELOG.md#400

ActuallyACat (commented):
Changes look okay to me - the lack of test coverage in this component does make me a little nervous though. 

Have you tested this with something like Verdaccio / `yarn link`?

ActuallyACat (approved):
I see that you've QA'd. LGTM.

lloydstubber:
@talbet is out today but I'm blocked on this. @ActuallyACat do you mind if I get this in now it's got the ✔️ ?

ActuallyACat:
> @talbet is out today but I'm blocked on this. @ActuallyACat do you mind if I get this in now it's got the ✔️ ?

I see no issue with that - I've just enabled automerge.

Issue #1725, opened 2021-06-25
Unclear what components belong to which package


ActuallyACat:
An import statement for each group of stories _used to_ exist as a string we provided to Storybook, e.g. 
https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/button/docs/Button.stories.tsx#L20-L24

This functionality disappeared when we moved to Storybook 6. 

In an ideal world, our code snippets and/or docs would include an example import so that consumers do not need to look at the source code to figure out what package a component comes from. A "quick" fix for this is to use [DocsPage to restore the previous functionality](https://storybook.js.org/docs/web-components/writing-docs/doc-blocks#docspage-1).

```
// Button.stories.js | Button.stories.ts

export const CustomSource = () => Template.bind({});

CustomSource.parameters = {
  docs: {
    source: {
      code: 'Some custom string here',
    },
  },
};
```

![Screen Shot 2021-06-25 at 3 12 28 pm](https://user-images.githubusercontent.com/1621182/123372895-ca0d2780-d5c7-11eb-9503-cf513f288636.png)

cc/ @lisatassone

dougmacknz:
@ActuallyACat - @sebpearce and I saw that you had updated the Box stories when you updated Storybook:
https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/stories/Box.stories.tsx#L21

We're thinking about copying that to all of the other components, but I wonder if there's a more ideal solution out there that we might want to explore first.

ActuallyACat:
Feels like there's a solid consensus that our "code docs" are missing a lot of details that leave folks guessing. 

I think there are two things we can do to improve the documentation here (without major changes to our build): 

## Lean into [react-docgen](https://github.com/reactjs/react-docgen) 
We're already using this library to automatically generate the props table for us. I noticed in their docs that it is also possible to get it to output a [general component description](https://github.com/reactjs/react-docgen#example). So we can continue to keep documentation within the code itself. This is going to be really useful if we only ever want to include superficial information in Storybook. 

As an example: 
```
export type MyComponentProps = {
  /* This appears in the props table in "docs" tab, and we're already encouraging this from contributors */
  id: string
}

/**
 * This should appear in the documentation 
 * We can add import information here, and any other information we want to appear in the "docs" tabs. 
 */
class MyComponent extends Component<MyComponentProps> { ... }
```

## A separate docs file
[...]

PR #1766, opened 2021-07-16
docs: Adjust import helpers on stories for Storybook 6 (first half)

dougmacknz:
Closes https://github.com/cultureamp/kaizen-design-system/issues/1725

# Objective
Small adjustment on our stories so that the import helper statements appear again.

# Motivation and Context
These disappeared after a Storybook 6 upgrade.

Seb and I agreed to do half each, you can see Seb's part here: https://github.com/cultureamp/kaizen-design-system/pull/1761

image


packages/component-library/stories/Layout.stories.tsx
```diff
@@ -1,225 +0,0 @@
-import * as React from "react"
```
ActuallyACat, line 1: Niiiiice

ActuallyACat (approved):
LGTM - very happy to see this change!

PR #1772, opened 2021-07-19
fix: remove loading state if avatar is already loaded

sugendran:
# Motivation and Context
In some cases the onLoad event does not fire if the image is from cache. This change checks if the image is in a complete state when it is rendered and set the loading state to success.

Navigating back and forth on the avatar with photos stories could repo this.


reetaa (commented):
We are talking about http caching right? Is there some other caching I'm not aware of 🤔

sugendran:
> We are talking about http caching right? Is there some other caching I'm not aware of 🤔

yes

reetaa (approved):

PR #1780, opened 2021-07-21
fix: Select types passing through to consumer

lisatassone:
When updating the Select draft package in performance-ui, type errors were discovered. It seems no types were being 'passed through' to the Select component so we were getting errors around function types accepting `any`.

On master
Kapture 2021-07-21 at 13 57 11

This change:
Kapture 2021-07-21 at 13 56 07

Looks like a recent change converting the component to a forwarded ref component caused the issue even though normally it wouldn't have been a problem. I think the complexity of some of `react-select`'s types caused them to become lost. However switching to the `NamedProps` interface worked as this is the type that is used under `Props` which we were originally using and is probably the one we should have been using.

# Checklist

# Additional Considerations


martindavid (approved):
TIL on `NamedProps` from react-select

talbet (approved):
Looks great, sorry this got missed when the forward ref was added

PR #530, opened 2020-06-03
feat: Rollout inter

ActuallyACat:
## About 
This PR updates Ideal Sans to Inter. This has been design reviewed by @katsambarcus. 

Where possible, I've removed unused imports in Sass.

## Other Stuff
The non-Inter related changes addressed were:

- ~Table - headings should not have bg~ Descoped
- Notifications - stories in Elm and React use different spacings in text
- Notifications - success text should not have an exclamation mark (success!)
- Notifications - Error state in stories should say "negative" not "error"
- Guidance Block - Spacing between buttons should be 12px, not 24px
- Select - Hover state on children should be cluny-100, hover text fg should be cluny-500
- ~Split Button - Soften the RHS border opacity to 70%~ Descoped
- Text area - Copy changes in story (Your Reply -> Your reply)

PR #1782, opened 2021-07-21
fix: remove input edit modal typography that's adding a gap

mbylstra:
# Objective

There is an unwanted white gap between the header and the body of an Input Edit Modal

I found that the gap is due to the baseline grid offset for our legacy typography. The body div of the modal should not have any typography styles applied, so this appears to have been unintentionally added.

** before **

image

** after **

image

# Motivation and Context
This is something I noticed when using the component (there is no git issue or Trello card)

# Checklist

# Additional Considerations


draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.scss
```diff
@@ -20,5 +20,4 @@
   background: $kz-var-color-stone;
   padding: $ca-grid $ca-grid $ca-grid * 1.5;
   border-bottom: 1px solid $ca-border-color;
```
ActuallyACat, line 22: This one looks like it was on me 😬 

I believe the rationale in [this PR](https://github.com/cultureamp/kaizen-design-system/pull/530/files) was to ensure no matter what a consumer put in the body (i.e. they forget to use the Heading / Paragraph component) it always rendered Inter typography.

ActuallyACat (approved):

draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.scss
mbylstra: Ah I see. I wonder if a css reset or base styles would be a better way of ensuring that - I believe you have a PR for that in the works. Otherwise it seems like we'd have to do that in every component that accepts children. I don't know if it's possible to force people via Typescript to not just use plain text nodes, or that would be another approach.

mbylstra: I _think_ this is safe enough to merge.

ActuallyACat: Yeah, I agree. It's low risk and low impact. The behaviour of engineers using Kaizen seems to be to reach towards Heading/Paragraph (or Text >.>) when needing copy. This is definitely something the shell component should take care off too 👍🏻

mbylstra: Yeah, I think people are pretty accustomed to using Heading/Paragraph now which is great.

Issue #1101, opened 2021-02-23
Ensure all deprecated components have been listed as deprecated


ActuallyACat:
We have the following deprecated components:
- [ ] component-library/Dropdown
- [ ] component-library/Layout
- [ ] component-library/NavigationBar
- [ ] component-library/MenuList
- [ ] component-library/OffCanvas
- [ ] component-library/Text
- [ ] draft-title-block/TitleBlock
- [ ] draft-zen-navigation-bar/ZenNavigationBar

We need to make sure that these are listed as deprecated in a few different ways.

## Adding the `@deprecated` JSDoc:

E.g.
```
/**
 * @deprecated Dropdown is deprecated. Please use draft-menu instead.
 */
class Dropdown extends React.Component<DropdownProps, DropdownState> {
```

Which results in strikethroughs in VSCode when importing:
![image](https://user-images.githubusercontent.com/1811583/125006372-54ae5600-e0a1-11eb-8e16-1aeb6b7eb8f4.png)

## Wrapping in `withDeprecatedComponent` from `react-deprecate-warning`

E.g.
```
export default withDeprecatedComponent(TitleBlock, {
  warning:
    "TitleBlock is deprecated. Use @kaizen/draft-title-block-zen instead.",
})
```

Which adds a console warning (in development environment only):
![image](https://user-images.githubusercontent.com/1811583/125214693-c386eb80-e2fb-11eb-9a41-53558a0662be.png)

## Adding (deprecated) in Storybook
Lastly, adding `(deprecated)` to the Storybook title also makes it clear which components are deprecated when engineers are looking through Storybook.

![image](https://user-images.githubusercontent.com/1811583/125214760-08128700-e2fc-11eb-9b69-49822e31c6d0.png)

PR #1787, opened 2021-07-26
fix: Add deprecated jsdocs and wrap with withDeprecatedComponent

JakePitman:
Closes https://github.com/cultureamp/kaizen-design-system/issues/1101

# Objective
Adds any missing `@deprecated` doc blocks to our deprecated components + wraps them in the `withDeprecatedComponent` HOC where that was missing as well.

# Motivation and Context
This helps to make it clear when engineers are using something that is deprecated.

### DIFF
--- a/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
+++ b/draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
@@ -43,7 +43,7 @@ class NavigationBar extends React.Component<Props, State> {
   static displayName = "NavigationBar"
   static Link = Link
   static Menu = Menu
-  static defaultProps = {
+  static defaultProps: Props = {
     environment: "production",
     loading: false,
     colorScheme: "cultureamp",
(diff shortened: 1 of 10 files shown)


packages/component-library/components/OffCanvas/OffCanvas.tsx
```diff
@@ -128,4 +128,6 @@ const withTrigger = (Component: React.ComponentType<any>) =>
     }
   }
 
-export default withContextProvider(withTrigger(OffCanvas))
+export default withDeprecatedComponent(OffCanvas, {
```
ActuallyACat, line 131: I believe you'll want to continue to wrap `OffCavas` with the `withTrigger` HOC as you're now changing the component that's exported. Unless there was a specific reason for doing it this way?

dougmacknz: Oops I think this was a mistake, nice catch!

draft-packages/zen-navigation-bar/KaizenDraft/ZenNavigationBar/NavigationBar.tsx
```diff
@@ -22,7 +23,7 @@ type Props = {
   loading?: boolean
   colorScheme?: ColorScheme
   badgeHref?: string
-  onNavigationChange: NavigationChange
+  onNavigationChange?: NavigationChange
```
dougmacknz, line 26: This was a weird one. There are a bunch of usages of this component that don't have this prop, but they weren't resulting in TS errors until we added the `withDeprecatedComponent` wrapping.

sebpearce (approved):
❤️

sebpearce (approved):

PR #1789, opened 2021-07-27
chore: Delete the draft-vertical-progress-step component

dougmacknz:
# Objective
Deletes the `@kaizen/draft-vertical-progress-step` component completely.

# Motivation and Context
This component was built in [TBC] (a while ago) and has still only been used in one place, which I'd imagine this was built for. The one usage is in our now unsupported Action Framework.

During the recent audit of Kaizen components, this component failed almost every quality attribute.
* Doesn't really have a design in the UI Kit. The 'progress stepper' in the UI Kit has a horizontal design only.
* Not RTL friendly.
* Hard coded English (not internationalised)
* Prescribes heading hierarchy which might not be right (accessibility issue).

I'm not saying that we don't need a vertical progress stepper component, but I believe that this one has enough issues, along with the fact that it hasn't been picked up in a couple of years, means it makes more sense to delete this and then build something with a fresh slate if and when that time comes.

# Notes
This will effectively lock murmur to design-tokens 3.x. Once there is a 4.x, Murmur will need to either find a way to remove this usage, or just copy over the code from here into `ca-ui`.


packages/design-tokens/docs/StoryBoard.stories.tsx
```diff
@@ -471,15 +466,15 @@ export const Everything: Story = () => {
...
       }}
     >
       <div
         style={{
           position: "fixed",
-          zIndex: 10,
-          right: theme.spacing.xxl,
-          top: theme.spacing.md,
+          zIndex: "10",
+          right: "theme.spacing.xxl",
```
ActuallyACat, line 476: I don't think these should be strings 👀

ActuallyACat (commented):
Question so we're all aligned - is the intention be eventually recreate this package as `kaizen/progress-stepper`? Additionally, the future package will have vertical and horizontal as variants of the same component, and not as separate packages?

packages/design-tokens/docs/StoryBoard.stories.tsx
dougmacknz: Hmm, this got auto formatted for some reason

site/docs/components/progress-stepper.mdx
```diff
@@ -9,11 +9,10 @@ needToKnow:
...
 - Aim for 3–4 steps.
-demoStoryId: deprecated-vertical-progress-step--completed-step
 health:
   designed: true
   documented: true
-  implemented: true
+  implemented: false
   latestDesign: false
   allVariants: false
   responsive: false
```
dmisdm, line 18: Probs not the right time to do it but could `deprecated: true` be useful?

dougmacknz:
> is the intention be eventually recreate this package as kaizen/progress-stepper? Additionally, the future package will have vertical and horizontal as variants of the same component, and not as separate packages?

I'm not planning to recreate this myself, but yes I would imagine that package would be created at some point. For now there's only designs for a horizontal one in the UI Kit. Hard to say at this point if they make sense in the same component (probably would), but at the very least would be in the same package.

site/docs/components/progress-stepper.mdx
dmisdm: Actually ignore more, for some reason i thought this component might be deprecated and thats why you were deleteing it

dougmacknz: The component _idea_ itself isn't deprecated though. There's designs for a horizontal one in the UI Kit that we'll want to build at some point, and possibly a rebuild of this vertical one as well.

ActuallyACat (approved):

PR #1792, opened 2021-07-27
chore: Enable Chromatic TurboSnap

dougmacknz:
# Objective
Enables [Chromatic TurboSnap](https://www.chromatic.com/docs/turbosnap) by adding the two required parameters to our storybook build and chromatic cli script.

# Motivation and Context
This will make Chromatic only create snapshots when there are changes, which will save us a ton of credits and allow us to disable the current functionality of only running it if you comment on the PR with "Run Chromatic"

### DIFF
--- a/.buildkite/scripts/run-chromatic-build.sh
+++ b/.buildkite/scripts/run-chromatic-build.sh
@@ -11,5 +11,5 @@ export CHROMATIC_APP_CODE
 CHROMATIC_APP_CODE=$(get_secret "chromatic-app-code") || exit $?
 
 yarn install --frozen-lockfile
-yarn storybook:build
-yarn chromatic --exit-zero-on-changes --storybook-build-dir storybook/public
+yarn storybook:build --webpack-stats-json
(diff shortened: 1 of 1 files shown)


mbylstra (approved):
FYI this only adds it to existing very manual buildkite setup (the one where you need to manually trigger the chromatic build and then look at the output logs to find the url that links to chromatic). I've approved it, because it's worth doing anyway to save credits.

However, what I think we really want is the github action which enables the full github integration (the status line green/red etc). This is the one we had an old PR open for, but it just needs to be tweaked a bit to pass in the `only-changed` parameter to the github action, and also to point it at our yarn build command. It wasn't possible to use the github integration with the github action in the past (for reasons), but now we can. I'm happy to help next monday with this, but you might not find it too difficult if you want to try it before then.

This was the PR: https://github.com/cultureamp/kaizen-design-system/pull/1601/files

We'll also be able to delete the chromatic buildkite step and also remove the `chromatic` dependency (because it will be inside the 3rd party github action)

PR #1794, opened 2021-07-28
docs: Group components in Storybook

dougmacknz:
# Objective
Adds some grouping/hierarchy to our Storybook menu.

See https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy

# Motivation and Context
The navigation on our Storybook is quite messy and difficult to scan for things. This will make it easier for folks that aren't super familiar with our components to easily scan and browse through components.

# Screenshots

Before:
image

After:
image


ActuallyACat (approved):
Thought I would have some nitpicks but this LGTM. Make sure you run Chromatic on these stories - I believe they treat renamed stories as new stories.

ActuallyACat (approved):

PR #1795, opened 2021-07-28
feat: Prefer color tokens rule

dmisdm:
This PR imports and merges in the stylelint rule which recommends Kaizen tokens as replacements for inlined color values like hex or rgba strings.

Prior art; the PR which we are migrating from: https://github.com/cultureamp/kaizen-design-system/pull/1582

- `src/rules/prefer-color-tokens.ts`
- `src/rules/prefer-color-tokens.spec.ts`

- Better equality matching of color strings. Will parse colors on both ends (the parsed value and the kaizen token) and compare them using a common stringification function (to hex). This means it recognises that `#fff`, `#ffffff`, and `rgb(255,255,255)` are all equal, and similar with any other color including ones that manipulate alpha.
- Will now replace with/recommend new $kz-var variables rather than the deprecated ones.

# Checklist

# Additional Considerations

### DIFF
--- a/packages/stylelint-plugin/CHANGELOG.md
+++ b/packages/stylelint-plugin/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [1.2.0](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/stylelint-plugin@1.1.2...@kaizen/stylelint-plugin@1.2.0) (2021-08-06)
+
+
+### Features
+
+* Deprecated and removed tokens rule uplift ([#1743](https://github.com/cultureamp/kaizen-design-system/issues/1743)) ([33411db](https://github.com/cultureamp/kaizen-design-system/commit/33411db1d02e09539dea98ec43e5183815c9974b))
+
+
+
+
+
 ## [1.1.2](https://github.com/cultureamp/kaizen-design-system/compare/@kaizen/stylelint-plugin@1.1.1...@kaizen/stylelint-plugin@1.1.2) (2021-07-02)
 
 **Note:** Version bump only for package @kaizen/stylelint-plugin
--- a/packages/stylelint-plugin/package.json
+++ b/packages/stylelint-plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kaizen/stylelint-plugin",
-  "version": "1.1.2",
+  "version": "1.2.0",
   "types": "dist/**/*.d.ts",
   "main": "dist/src/index.js",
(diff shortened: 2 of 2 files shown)


dmisdm:
Some things that are going through my head about next steps:
- This only works with colors at the moment, but, given the foundational work, it would be very easy to extend this to work with all other tokens e.g. knowing to replace `1.5rem` with `$spacing-md`
- Dataviz tokens in the Zen theme (which is still the "default" theme) have the same values as other colors, so i've put in a simple tie breaking sorter to choose "color-.." over "dataviz-.." if there is a conflict.
- The tokens which are recommended are all from the Zen theme, because it is currently the `defaultTheme` in `@kaizen/design-tokens`, and if you enter a hex value for something only in Heart, you won't get recommended a token replacement. I think we should force the use of Heart theme within the rule, then change it to `defaultTheme` after it changes @ActuallyACat what do you think?

dmisdm:
> Some things that are going through my head about next steps:
> 
> * This only works with colors at the moment, but, given the foundational work, it would be very easy to extend this to work with all other tokens e.g. knowing to replace `1.5rem` with `$spacing-md`
> * Dataviz tokens in the Zen theme (which is still the "default" theme) have the same values as other colors, so i've put in a simple tie breaking sorter to choose "color-.." over "dataviz-.." if there is a conflict.
> * The tokens which are recommended are all from the Zen theme, because it is currently the `defaultTheme` in `@kaizen/design-tokens`, and if you enter a hex value for something only in Heart, you won't get recommended a token replacement. I think we should force the use of Heart theme within the rule, then change it to `defaultTheme` after the switch, @ActuallyACat @jasononeil  what do you think?

Just making a decision on these:

- Lets not extend this rule to support other values just yet. To keep this PR smaller, lets just migrate Sugendrands rule across first.
- Using the heart theme rather than the default theme is going to blow this PR up, because we currently use `tokens/*.json` files to know about which tokens are available, and we would have to transition to using the TypeScript heart theme object within the stylelint plugin. Not only that, it's very unlikely IMO that someone has inlined a hex value of a heart theme token since it's release. Even if we assume the worst, we'll eventually flick the default theme over anyway.

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -129,7 +131,51 @@ export const kaizenTokensByName: Readonly<
...
+      : undefined
+    return {
+      ...acc,
+      [key]: existingValues ? [...existingValues, token] : [token],
+      ...(token.color && {
+        [token.color.toLowerCase()]: existingNewColorValues
+          ? [...existingNewColorValues, token]
+          : [token],
+      }),
+    }
```
dmisdm, line 178: Probably going to refactor this a bit to make it more readable

PR #1796, opened 2021-07-28
feat: Add an initial animation to the BrandMomentCaptureIntro component

ActuallyACat:
# Objective
This adds an "initial" animation (only ever played once) and an "ambient" animation (can be looped with the `loop` prop) to the BrandMomentCaptureIntro component. There were a few other things I decided to tidy up since I was in the area: 
* Remove the initial animation/ambient animation props from the video player. This was a premature decision. The video source swapping now happens in the BrandMomentCaptureIntro and the VideoPlayer's concerns are now just to stop/start/play/etc the video (what you would expect it to do). 
* Abstracted some types

# Motivation and Context
This component is unique to the other scene animations, as they contain one source and no restrictions around whether to loop or not. Our testing found that looping the diver about to jump was a cause of anxiety for some users who are about to begin a survey. 

# Checklist

# Additional Considerations


ActuallyACat:
Temporarily moved this to draft. When the asset is not cached (which will be the case for most our users) there jump between the players that's quite jarring. I'll preload both sources to minimise this.

dougmacknz (approved):
🚢 the animation looks awesome!

PR #1800, opened 2021-07-29
fix: show tooltips in chromatic tests

### LINKED ISSUE #1375: Automatically open tooltip for Chromatic tests
Chromatic is running visual regressions on the existing tooltip story, which is unopened when the snapshot is taken. 

I see two possible options: 
1. Create a stateless tooltip component (in addition to the existing stateful component). In the visual regression tests use the stateless components and default them to open. I can see use cases where a stateless variant would be useful for consumers. 
   1.a. Alternatively a "intiallyOpen" prop on the existing stateful component
2. Figure out a nice way to mock mouse interactions when running the story in Chromatic. 

Best explained with an image: 
Screen Shot 2021-04-06 at 2 34 19 pm

talbet:
# Objective

The visual regression tests around the tooltip component render the button that triggers the tooltip, but not the tooltip itself. This means that these tests provide little value, unless we can get the tooltip into the test.

The ideal way to do this would be to trigger a hover event in chromatic before the screenshot is captured, but the current chromatic API doesn't support any interaction or events. Instead, we are putting a new prop into the component to signal that the tooltip should be visible on the first render. This seems like it will only be useful for testing, but someone may find a valid use for it, so it hasn't been named obtusely (a name like `chromaticTestVisibility` would send a signal that this is the only valid usage)

# Motivation and Context

closes #1375 

Screen Shot 2021-07-30 at 9 36 02 am

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
+++ b/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
@@ -36,6 +36,11 @@ export type TooltipProps = {
    * `overflow: scroll` or `overflow: hidden` element.
    */
   portalSelector?: string
+  /**
+   * Should the tooltip be visible on the first render. Useful for visual
+   * regression testing.
+   */
+  isInitiallyVisible?: boolean
 }
 
 // Sync with Tooltip.scss
@@ -119,8 +124,9 @@ const Tooltip = ({
   position = "above",
   classNameAndIHaveSpokenToDST,
   portalSelector,
+  isInitiallyVisible = false,
 }: TooltipProps) => {
-  const [isHover, setIsHover] = useState(false)
+  const [isHover, setIsHover] = useState(isInitiallyVisible)
   const [isFocus, setIsFocus] = useState(false)
   const [
     referenceElement,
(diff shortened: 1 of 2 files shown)


talbet:
Here is a link to the chromatic build for this branch https://www.chromatic.com/test?appId=5e12c297b9d9020020789ca3&id=61033f5e581f7a0049ee812f

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -36,6 +36,11 @@ export type TooltipProps = {
    * `overflow: scroll` or `overflow: hidden` element.
    */
   portalSelector?: string
+  /**
+   * Should the tooltip be visible on the first render. Useful for visual
+   * regression testing.
+   */
+  initiallyVisible?: boolean
```
ActuallyACat, line 43: /Nitpick It feels more consistent to call this `isInitiallyVisible`, as we have Modal (has `isInitiallyOpen` prop), and FilterMenuButton (has `isInitialDropdownVisible` prop) following this convention.

ActuallyACat (changes requested):
Looks great - thanks @talbet and @aaronknol! I have a 🤏🏻 suggestion about the name of the new prop, but everything else LGTM.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
talbet: That makes sense. I've updated the variable name.

ActuallyACat (approved):

PR #1802, opened 2021-07-30
chore: Run design-tokens prepublish on dev builds

dmisdm:
We noticed that the release's compilation process broke in master, and therefore silently broke in development https://github.com/cultureamp/kaizen-design-system/pull/1801

This PR is not the most golden solution, but is a pragmatic one that would have detected the previous issue before merging the [problematic PR](https://github.com/cultureamp/kaizen-design-system/pull/1741) into master.

The golden solution unfortunately involves a bit more:
- We need to run a command in each package and we need it to not take a long time
- To not take a long time, we need some form of incremental building, so that each sub-package only compiles what it's responsible for. Running `yarn lerna run prepublish`, which doesn't do this, isn't time-feasible at the moment.
- We don't necessarily have to run a separate TypeScript build in each sub-package, but we probably do need to do something like `yarn lerna run prepublish` which would run prepublish in each of them, so that they all have the opportunity to perform customised build steps. This is what the release step does too, so it would make development builds more accurate.

The development build step never used to run this:
image

# Checklist

# Additional Considerations


ActuallyACat (approved):
This is one of the biggest bummers of compiling the repo as a whole, over compiling each package individually.

PR #1803, opened 2021-07-30
fix: Update Storybook background colour references

dougmacknz:
These names changed in https://github.com/cultureamp/kaizen-design-system/pull/1741/files#diff-b3e4a28d9c367ace4d3bc1ebf576bc4ee1bb1927a3d141a2cf4386224dde60c0L8, but the stories weren't updated. This updates the references to the updated background colour names in our stories.


ActuallyACat (approved):
Good find! I thought we had caught all of these in the original PR.

PR #1816, opened 2021-08-06
feat: Prefer color tokens rule

dmisdm:
This PR imports and merges in the stylelint rule which recommends Kaizen tokens as replacements for inlined color values like hex or rgba strings.

Prior art; the PR which we are migrating from: https://github.com/cultureamp/kaizen-design-system/pull/1582

- `src/rules/prefer-color-tokens.ts`
- `src/rules/prefer-color-tokens.spec.ts`

- Better equality matching of color strings by normalising them. Will parse colors on both ends (the parsed value and the kaizen token) and compare them using a common stringification function (to hex). This means it recognises that `#fff`, `#ffffff`, and `rgb(255,255,255)` are all equal, and similar with any other color including ones that manipulate alpha.
- Will only replace with/recommend non-deprecated ones.

# Checklist

# Additional Considerations


dmisdm:
Quoting comments from previous PR:
https://github.com/cultureamp/kaizen-design-system/pull/1795#issuecomment-888031358

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -129,7 +131,51 @@ export const kaizenTokensByName: Readonly<
...
+    return {
+      ...acc,
+      [key]: existingValues ? [...existingValues, token] : [token],
+      ...(token.color && {
+        [token.color.toLowerCase()]: existingNewColorValues
+          ? [...existingNewColorValues, token]
+          : [token],
+      }),
+    }
+  },
```
dmisdm, line 179: I'll probably refactor/rename some things here to make it more understandable

packages/stylelint-plugin/src/rules/prefer-color-tokens.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  },
+  {} as Record<string, KaizenToken | undefined>
+)
+
+export const messages = stylelint.utils.ruleMessages(ruleName, {
+  expected: (named, original) => `Expected "${original}" to be "${named}"`,
+})
+
+export const preferColorTokens: RuleDefinition = {
```
dmisdm, line 67: This probably needs some commentary on what it does

dmisdm:
Theres currently nothing in the README for this rule either

dmisdm:
I think it's important that this rule knows about color values from both Heart and Zen.
E.g. it would know how to replace both `#f0f1f4`(from Zen) and `#f4edf8` (from Heart) with `color-purple-100`.

Let's not do this in this PR as i don't think we have the apetite for it.

packages/stylelint-plugin/src/rules/prefer-color-tokens.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        background-color: $color-blue-500;
+        border-color: $color-purple-600;
+      }`,
+      warnings: [
+        messages.expected("$color-blue-500", defaultTheme.color.blue[500]),
+        messages.expected("$color-purple-600", defaultTheme.color.purple[600]),
+      ],
+    },
+  ],
+} as const
```
dmisdm, line 52: These tests are inconsistent with the others in the project, however, i actually like this format the best out of all of them, and its the stylelint idiomatic way.

@jasononeil you've called out in previous work that we want to know which warnings exactly should be reported, rather than just a number of `expectedWarnings`, and this could be a good solution.

packages/stylelint-plugin/README.md
```diff
@@ -31,6 +31,7 @@ These rules are included by default:
 - `kaizen/no-invalid-use-of-var-tokens-in-equations` - Detects invalid usages of Kaizen tokens within equations. It won't be able to automatically fix many cases, but it can for some simple cases for example negating a Kaizen variable without a runtime `calc()` function.
 - `kaizen/all-token-imports-must-be-used` - Reports and fixes unused Kaizen token imports.
 - `kaizen/all-used-tokens-must-be-imported` - Reports and fixes any Kaizen token imports that should exist but don't, based on the tokens that are used in the current stylesheet.
+- `kaizen/prefer-color-tokens` - Detects inlined color values (in any kind of form such as hex, rgba, hsl, hwb etc) and recommends (or automatically commits) replacements for them if they have the same value as an existing non-deprecated Kaizen token.
```
dmisdm, line 34:
```suggestion
- `kaizen/prefer-color-tokens` - Detects inlined color values (in any kind of form such as hex, rgba, hsl, hwb etc) and recommends and replaces Kaizen tokens for them if one is found with the same value. Only non-deprecated tokens will be recommended.
```

packages/stylelint-plugin/src/rules/prefer-color-tokens.ts
```diff
@@ -0,0 +1,119 @@
...
+function getTokenRank(_tokenName: string) {
+  const tokenName = _tokenName.toLowerCase()
+  return 1 - tokenName.indexOf("color") / tokenName.length
+}
+
+const ruleName = "prefer-color-tokens"
+
+/**
+ * This a map of colorHex -> KaizenToken.
+ * It allows you to determine which KaizenToken to use for a given color string in hex form.
```
dmisdm, line 29:
```suggestion
 * It allows you to determine which KaizenToken to use for a given color string, normalised using {@link normaliseColor}.
```

dmisdm:
@jasononeil 
Something that might be problematic about this rule is that it won't do anything smart to detect whether a replacement is a CSS variable and problematic within the context (such as a color manipulation function).
The nice thing is that our other rules should be able to pick up things like this, and because of this and the fact that we're only dealing with colors (nothing used in equations or AtRules), I think it's fine to release.

This PR was also just meant to be a simple migration from Sugendran's rule, and it's already increased in size due to wanting to fit it into our plugin and conventions, and improving how colors are detected.

Keen to know your thoughts though!

packages/stylelint-plugin/src/rules/prefer-color-tokens.spec.ts
dmisdm: FYI i've done something different in https://github.com/cultureamp/kaizen-design-system/pull/1836

I'm thinking the tests in this file should be changed to the style in #1836, with the function to do the testing split out to a shared testing lib file.

dmisdm: Optionally, we could open another PR to make the testing style more consistent across the project.

packages/stylelint-plugin/src/rules/prefer-color-tokens.ts
```diff
@@ -0,0 +1,119 @@
...
+    return {
+      ...acc,
+      [firstNonDeprecatedColorToken.color]: firstNonDeprecatedColorToken,
+    }
+  },
+  {} as Record<string, KaizenToken | undefined>
+)
+
+export const messages = stylelint.utils.ruleMessages(ruleName, {
+  expected: (named, original) => `Expected "${original}" to be "${named}"`,
```
jasononeil, line 64: I don't mind this being inlined here, but is there a reason you didn't include the messages in the shared messages file you've used for other rules?

(I think I prefer keeping it local actually)

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -129,7 +131,74 @@ export const kaizenTokensByName: Readonly<
...
+ *    "#fff": [{name: "$kz-var-color-white", color: "#ffffff", value: var(--kz-var-color-white, $fff),...}],
+ *    "#ffffff": [{name: "$kz-var-color-white", color: "#ffffff", ...}],
+ *    "230, 240, 247": [{name: "$kz-var-color-cluny-100-rgb-params", color: undefined, ...}],
+ *    "0 0 12px rgba(0, 0, 0, 0.19)": [{name: "$kz-var-shadow-large-box-shadow", color: undefined, ...}],
+ *    "400": [{name: "$kz-var-typography-paragraph-intro-lede-font-weight", ...}, {name: "$kz-var-typography-paragraph-body-font-weight", ...}, ...],
+ *     "1.5rem": [{name: "$kz-var-spacing-md", ...}]
+ * }
+ * ```
+ */
+export const kaizenTokensByValue: Record<
```
jasononeil, line 159: I'm questioning the value of this export... given it's not the right shape for us to use as-is in the new rule, is there any point for this record existing in this shape?

Why not just calculate `colorMap` inline in the `prefer-color-tokens.ts` file, based on kaizenTokensByName.

jasononeil (approved):
I'm happy to approve this as is, but also have raised some questions about a potentially unnecessary export (I don't think `kaizenTokensByValue` is useful in its current state).

If you decide to

packages/stylelint-plugin/src/rules/prefer-color-tokens.ts
dmisdm: This was from Sugendran's plugin that i didn't change, and i'm not 100% sure we should in this PR.

packages/stylelint-plugin/src/util/colors.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param color a string representation of a color, which can be in any kind of format like `#fff`, `blue`, `rgba(200, 60, 60, 0.3)`, `hsl(195, 100%, 50%)` etc. See https://www.npmjs.com/package/tinycolor2.
+ * @returns a string representation of the color (at the time of writing, it is a hex or rgba string), or null if the string you passed in wasn't a color.
+ */
+export function normaliseColor(color: string): string | null {
+  const parsedColor = tinycolor(color)
+  if (!parsedColor.isValid()) return null
+
+  // If alpha is less than 1, we should stringify as rgba so that the alpha channel isn't lost.
+  // Otherwise we should use hex.
```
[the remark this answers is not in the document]
dmisdm: @jasononeil yeah i went down that rabbit hole and tried many libraries lol!
There we're pros and cons to each (package size, support for multiple color functions like HSL), but one of the things i wanted was a return value that could also be used within CSS because i think this does happen, and 4 channel hex values aren't supported.

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -129,7 +131,74 @@ export const kaizenTokensByName: Readonly<
...
       },
     }
   })
 ).reduce((acc, next) => ({ ...acc, ...next }), {})
+
+/**
+ * This is a record of Kaizen CSS variable values -> Kaizen token SCSS variable names.
+ * It allows you to find the variable for a value (the inverse of getting the value of a variable).
+ * Because colors can be represented in multiple ways, additional keys have been added to here for every variable that holds a color, by running something along the lines of `colorString.to.hex(colorString.get(value).value)`.
+ * What this means is, if a value like $color-white holds (a CSS fallback) value of $fff, you will get the additional color of $ffffff, which is what colorString spits out.
```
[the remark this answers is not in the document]
dmisdm: Yep thanks!
```suggestion
 * What this means is, if a value like $color-white holds (a CSS fallback) value of #fff, you will get the additional color of #ffffff, which is what colorString spits out.
```

packages/stylelint-plugin/src/util/kaizenTokens.ts
dmisdm: You're right the value is questionable.
It's because at some point i decided to 'separate concerns' and this is what came out of it.
Also, though, this map allows us to make a more powerful rule that can recommend things like `1.5rem` be replaced with `$spacing-md`.
I think we should keep it.

jasononeil (approved):

packages/stylelint-plugin/src/util/kaizenTokens.ts
jasononeil: I'm okay with that. 👍

PR #1820, opened 2021-08-09
fix: Fix Chrome issue for BrandMomentCapture

ActuallyACat:
# Objective
When autoplay changes after the videos are mounted, this does not consistently trigger a play in all browsers. To ensure a consistent experience, we force a play() event when the autoplay prop changes.

This fixes an issue identified in Chrome where the "ambient" animation does not autoplay.


dmisdm:
I was having a go at verifying that your fix worked, but i wasn't 100% sure how; I opened the storybook preview of an animated scene illustration in Chrome to see if it autoplayed (e.g. https://dev.cultureamp.design/ally/safari-video/storybook/?path=/story/components-illustration-scene--brand-moments), is that about right?

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -105,7 +105,12 @@ export const VideoPlayer = ({
          */
       }
     }
-  }, [prefersReducedMotion])
+    /**
+     * Chrome seems to have an issue with changes to autoplay after the video
+     * has been mounted. If there is a change in autoplay we need to force the
+     * play() method to be called.
+     */
+  }, [prefersReducedMotion, autoplay])
```
dmisdm, line 113: I opened an animated scene illustration in Chrome (https://dev.cultureamp.design/ally/safari-video/storybook/?path=/story/components-illustration-scene--brand-moments) to see if it autoplayed, but it works intermittently. Are you getting the same behaviour? I.e. it autoplays on some page loads but not on others.

Looking at this component, and how there are `useEffect` hooks which depend on `videoRef`, my gut says `videoRef` should probably be from a `useState` rather than `useRef`?
Reason being, the latter (React refs) won't cause the consecutive `useEffect`s to be re-run because they don't re-render the component when they change. It might be working intermittently because the video element is sometimes mounted in time, so `videoRef` is not null during the effects.
Honestly just hunch could be wrong!

dmisdm: I had a quick crack, setting the `muted` prop on the video element did the trick for me. I was seeing an error like `play() failed because the user didn't interact with the document first` when looking closer.

image

ActuallyACat: Nice - thanks for flagging this, I completely missed the flakiness. You're right, I should probably also be setting muted on the `video` itself. Let me make those changes :loading:

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -105,7 +105,12 @@ export const VideoPlayer = ({
...
 
   useEffect(() => {
     // Add event listeners for the video element
```
dmisdm, line 116: A note here: 
The `videoRef` dependency in this `useEffect` hook is redundant because it's not the reference/box that changes, its the field `videoRef.current` that does.
The other thing is that this might not get re-run reliably because of how ref changes don't cause the component to get re-rendered.
There are a few ways to get around this, e.g https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

Seeing as this isn't related to your PR i've just approved so it's up to you!

dmisdm (approved):

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
ActuallyACat: I've just set the `muted` prop on the `video` and I can't see any intermittent issues anymore.

dmisdm: likewise it's always playing now!

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
ActuallyACat: This involves a little bit more refactoring than I accounted for - I'll create an issue and loop back to it. Thank you for raising 🙏🏻

PR #1825, opened 2021-08-10
docs: Add component health information to docs and display on website

dougmacknz:
# Objective
Adds health information to component readmes and displays this information in two ways:

On each component page, there's a new section showing these, added above the newly added 'Issues' section:
image

There's also a new page with a table showing the status for every component. This can be accessed at `/component-health`, or with a link from the Components -> Overview -> See also

image

# Motivation and Context
Coming out of the recent quality audit of our components, we want to continue tracking these statuses for each component and display them in a public place.

# Left out
I only filled out health data for the readmes that we have, which doesn't comprehensively cover all of the components that we have. There's also some readmes like 'Dropdown' that we might want to rethink to be in line with our actual implementations. I didn't get into this here, but it would be good to follow up with that at some point.


draft-packages/button/KaizenDraft/Button/README.mdx
```diff
@@ -8,6 +8,14 @@ needToKnow:
 - If it needs to navigate somewhere and can be opened in a new tab, use a link instead.
 headerImage: button
 demoStoryId: button-zen-react--default-kaizen-site-demo
+health:
```
ActuallyACat, line 11: :D Thanks, this looks good. 

👍🏻 American-English (internationalized)
👍🏻 internationalized category also represents RTL
👍🏻 designed and build categories are related to health. This holistically represents the component's health - including both code and design quality attributes.
👍🏻 "documented" omitted. This is redundant, as we don't typically create placeholder pages for upcoming components - if this page existed it's documented.

I could see a future where we _might_ want to add some nuance to these categories and even expand the categories themselves (has this component passed an accessibility audit, has this component undergone user testing, etc). I believe this format will support those needs nicely 👍🏻

ActuallyACat (commented):
This LGTM - pretty much exactly what I had in mind. 

How easy is it to query all components frontmatter data? I'm thinking in terms of populating the "overview health" page with all the component health data.  

It would probably be prudent to give @katsambarcus and @dlimiter a heads up about these changes as our design system owner and super advocate.

draft-packages/button/KaizenDraft/Button/README.mdx
jasononeil: On the topic of passing an accessibility audit, I can see us wanting to include some dates in future:

```
lastDesignReviewed: "2021-07-10",
lastAccessibilityCertification: "2020-10-16",
```

or similar

jasononeil:
How hard is it to get the frontmatter data from all of our components for a reports page?

I haven't really explored how Gatsby is reading the mdx files, but it looks like we could do a big old graphql query which is nice.

In which case, I'm a fan 👍

katsambarcus:
@dougmacknz I'm wondering if it might be clearer to refer to this section as 'Component health'?

Also realising how much this new section is contributing to pushing the 'meaty' bits of the content even further down the page. I'm wondering if there are some simple UI improvements we can make to address the rather generous spacing these pages currently employ?

cc: @jasononeil @allybeaton

ActuallyACat:
The header space is very generous and can be up to a quarter of the "above the fold" space. For a super quick fix, we could remove the snippet below the heading on component pages. That information is sometimes repeated in the "Need to know" section anyway. For example:
Screen Shot 2021-08-11 at 9 26 23 am

Another thought I had (which is less of a quick win) is to introduce a tab at the top of each component page. In the future, we're likely looking to move technical documentation to the Gatsby site. 
Screen Shot 2021-08-11 at 9 18 55 am

site/docs/components/overview.mdx
```diff
@@ -6,15 +6,13 @@ tags: ["Overview"]
 needToKnow:
 - Components are reusable code used to rapidly build pages. They are the source of truth for our shared User Interface elements.
 - These components are flexible, and we continuously develop and grow them to match our evolving product needs.
+- You can see an overview on the health status of all of the Kaizen components at [https://cultureamp.design/component_health](/component_health)
```
ActuallyACat, line 9: 😬 
Screen Shot 2021-08-13 at 2 29 48 pm

ActuallyACat:
```suggestion
- You can see an overview on the health status of all of the Kaizen components at [/component health](https://cultureamp.design/component_health)
===

File: site/src/components/SidebarAndContent.scss [hunk 3]
```diff
@@ -65,6 +66,16 @@ $tableOfContents-width: $ca-grid * 10;
   }
 }
 
+.healthContent {
+  margin-top: 2.25rem;
+
+  // Sorry for the hax but there's no way for me currently to add margin bottom
+  // to each tag without creating a parent element for each.
+  div[class^="Tag--"] {
```
ActuallyACat, line 74: I'd prefer the parent element approach as we can't always guarantee the classes on components won't change (e.g. if we changed the name of the Tag component to something else, we'd probably rename the classes too)

site/src/pages/component_health.tsx
```diff
@@ -0,0 +1,135 @@
+import { graphql } from "gatsby"
```
ActuallyACat, line 1: One nit - The unspoken convention in this repo is for all URLs to be kebab case, so this should be `/component-health`

site/docs/components/overview.mdx
dougmacknz: This is what happens when you speed code with laptop battery running out 😅

site/src/components/SidebarAndContent.scss
dougmacknz: F

site/docs/components/overview.mdx
dougmacknz: Oh, that doesn't work either because this is frontmatter 🤔 
image

ActuallyACat: Ah, bugger

dougmacknz:
Just documenting that I tried the health tags up in the header as well.

Firstly with the latest Kaizen tags, which don't look great on the darker background (plus the annoying margin bottom issue):
image

Then with the site specific tags, which looks better but doesn't have any visual distinction between positive and negative states 😐 
image

site/docs/components/overview.mdx
```diff
@@ -61,6 +61,10 @@ It is common for components to come with default margins surrounding them and ye
 Use "Small", "Medium", "Large": in code, `sm`, `md`, `lg`. For more information, see [Sizing guidelines](https://cultureamp.design/guidelines/sizing).
 
 
+# See also
+
+- [Component health](/component_health) for the statuses of all Kaizen components at a glance.
```
ActuallyACat, line 66:
```suggestion
- [Component health](/component-health) for the statuses of all Kaizen components at a glance.
```

Small one, noticed this hasn't been updated to the new URL

dougmacknz: 🤦 I checked that this rendered correctly but forgot to click it, nice catch

ActuallyACat (approved):
LGTM! Nice one!

PR #1813, opened 2021-08-05
feat: Migrate scss to use new design-tokens

dmisdm:
# Objective
We just added new colors to design-tokens, and a stylelint rule that automatically renames old to new.
We'd like the whole repository to be using the new names, in SCSS and JS/TS, so that we can soon remove the old from design-tokens.

This PR uses the new stylelint rule to automatically rename colors in SCSS.
A subsequent PR will do the same for JS/TS.

# Screenshots (if appropriate)

# Checklist

- [ ] If this contains visual changes, has it been reviewed by a designer?
- [ ] If this introduces either a new component or a breaking change to an existing component, has it been reviewed by an advocate? (Ask in #prod_design_systems)
- [ ] I have considered likely risks of these changes and got someone else to QA as appropriate
- [ ] I have or will communicate these changes to the front end practice

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?
- Have Storybook stories been updated?

**If you're new to Kaizen, please ask #prod_design_systems to set up an onboarding session to get you up to speed.** If you have an urgent PR to merge before that happens, it is safest to ask in #prod_design_systems to have a design system advocate review the PR to catch any issues.

PR #1833, opened 2021-08-11
feat: Removal of deprecated tokens

dmisdm:
# Objective
This PR is part of the goal to mass rename all Kaizen tokens.

We first added all new tokens:
- https://github.com/cultureamp/kaizen-design-system/pull/1741.
Made the stylelint plugin able to rename tokens:
- https://github.com/cultureamp/kaizen-design-system/pull/1743
- https://github.com/cultureamp/kaizen-design-system/pull/1836
 - NOTE this PR is in review and needs merging beforehand.
Then made migrations to all components so no old tokens we're used anymore:
- https://github.com/cultureamp/kaizen-design-system/pull/1823 
- https://github.com/cultureamp/kaizen-design-system/pull/1813
- https://github.com/cultureamp/kaizen-design-system/pull/1817

Now we need to actually break design-tokens by mass removing anything that was deprecated.

# Checklist

# Additional Considerations


packages/design-tokens/index.ts
```diff
@@ -4,5 +4,11 @@ export * from "./src/types"
...
-// Utils are namespaced because they are potentially useful but unrelated to direct usage of design tokens.
-export * as Utils from "./src/utils"
+export {
+  mapLeafsOfObject,
+  makeCSSVariablesOfTheme,
+  flattenObjectToCSSVariables,
+  makeCSSVariableTheme,
+  objectPathToCssVarIdentifier,
+  objectPathToCssVarReference,
+} from "./src/utils"
```
dmisdm, line 14: Being more explicit with these exports should make it easier to rename things in the future within utils, and potentially move stuff out of that single file.
Still thinking about which of these to export though...probably not everything here.

packages/design-tokens/tokens/color.ts
```diff
@@ -1,16 +1,10 @@
 import * as colorTokens from "./color.json"
```
dmisdm, line 1: @ActuallyACat what do you think about removing these neighbouring `*.ts` files within the tokens directory?

dmisdm: Since our preferred non-SCSS consumption is to use `useTheme`

dmisdm:
FYI some remaining build errors will be fixed when https://github.com/cultureamp/kaizen-design-system/pull/1836 is merged

packages/design-tokens/tokens/color.ts
ActuallyACat: I'd definitely like to see it go. Adding to this, it might also be useful to stop distributing the `json` token files for the same reason. I've had a peek at the monoliths and I don't think making this change would be terribly controversial.

dmisdm: So, i agree 100% we don't need the json tokens.
I removed them and started to see the effects:
- The use of these json tokens within components in Kaizen wasn't too hard to refactor, but it makes this PR increase in size.
- The stylelint plugin uses these files to know about Kaizen tokens, and fixing this bloated the PR even further, as it had a few trickle down effects such as requiring extra Jest configuration to be able to parse `*.scss` files (remember how we we're parsing `color.scss` etc at first within the stylelint plugin, and it needed that special rule in Jest?). There are alternative methods but it's still going to bloat this.

How do you feel about making another breaking change after this one?

dmisdm: Oh just to clarify...this doesn't relate to `tokens/*.ts` files. They're in the 🗑️ .

dmisdm:
There is one commit that is a refactor of `utils.ts` in design-tokens, that could probably not have been done in this PR.
Because this is all done in a single commit, and therefore hopefully easily reviewable, i've left it in for now, but please do let me know if this isn't preferred.
https://github.com/cultureamp/kaizen-design-system/pull/1833/commits/44bdaf3cfad19443193b47b9896b2f58270c3a3e

packages/design-tokens/tokens/color.ts
dmisdm: Omg brainwave:

image

image


image

image

ActuallyACat: Love that! That's certainly preferable if we're removing the .json tokens in a future breaking change.

dougmacknz (approved):
🚢

dougmacknz (approved):

PR #1836, opened 2021-08-11
fix: Support design-tokens v3 in stylelint plugin

dmisdm:
This change improves the no-invalid-use-of-tokens rule to handle both version 2 and 3 of design-tokens, by allowing for the possibility that a token might not exist anymore within it (but still existing in the deprecated tokens list). To do this we had to also slightly modify how `fixAlphaModificationFunctions` worked (as its ability was predicated on Kaizen tokens existing).
We also improve the way we test the stylelint plugin by allowing test cases to supply a `expectedWarningMessages` field, which causes a more specific assertion to made over what warning message texts are reported, rather than just a number of them.

**NOTE** that this branch was developed as a branch off https://github.com/cultureamp/kaizen-design-system/pull/1833 which contains the breaking change of design-tokens. This was done to further assert that the plugin is prepared for version 3 of design-tokens, and that no tests would break within #1833.
It was then rebased onto master.

# Checklist

# Additional Considerations

### DIFF
--- a/packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
+++ b/packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
@@ -7,7 +7,7 @@ import {
   deprecatedTokenInVariableMessage,
   deprecatedTokenUsageMessage,
   deprecatedTokenUsageWithoutReplacementMessage,
-  genericContainsDeprecatedKaizenTokenMessage,
+  genericContainsDeprecatedKaizenTokenMessage as containsDeprecatedKaizenTokenWithNoReplacement,
   invalidEquationContainingKaizenTokenMessage,
   replacementCssVariableUsedWithinUnsupportedFunction,
 } from "../messages"
@@ -24,10 +24,10 @@ import { fixAlphaModificationFunctions } from "./lib/fixAlphaModificationFunctio
 import { containsEquationThatDoesntWorkWithCSSVariables } from "./no-invalid-use-of-var-tokens-in-equations"
 
 const deprecatedKaizenTokenPattern = new RegExp(
-  Array.from(deprecatedTokenReplacements.keys()).join("|")
+  `(${Array.from(deprecatedTokenReplacements.keys()).join("|")})`
 )
-function stringContainsDeprecatedKaizenToken(value: string) {
-  return deprecatedKaizenTokenPattern.test(value)
+function getDeprecatedKaizenTokenPatternMatches(value: string) {
+  return value.match(deprecatedKaizenTokenPattern)
 }
 
 // Most CSS standard functions are allowed to contain CSS variables.
@@ -316,9 +316,14 @@ function detectAndFixInvalidTokens(
     and would silently cause a style regression when upgrading to v3 of design tokens if left unchanged.
   */
   if (!reported) {
-    if (stringContainsDeprecatedKaizenToken(newValue)) {
+    const deprecatedTokenMatches = getDeprecatedKaizenTokenPatternMatches(
+      newValue
+    )
+    if (deprecatedTokenMatches) {
       options.reporter({
-        message: genericContainsDeprecatedKaizenTokenMessage,
+        message: containsDeprecatedKaizenTokenWithNoReplacement(
+          deprecatedTokenMatches[1]
+        ),
         autofixAvailable: false,
         node: postcssNode,
       })
--- a/packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
+++ b/packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
@@ -138,8 +138,9 @@ function detectAndFixInvalidTokens(
     return
   }
 
-  const fixAlphaModificationFunctionsResult =
-    fixAlphaModificationFunctions(nodeValue)
+  const fixAlphaModificationFunctionsResult = fixAlphaModificationFunctions(
+    nodeValue
+  )
 
   if (fixAlphaModificationFunctionsResult.errors.length) {
     fixAlphaModificationFunctionsResult.errors.forEach(error =>
@@ -321,8 +322,9 @@ function detectAndFixInvalidTokens(
     and would silently cause a style regression when upgrading to v3 of design tokens if left unchanged.
   */
   if (!reported) {
-    const deprecatedTokenMatches =
-      getDeprecatedKaizenTokenPatternMatches(newValue)
+    const deprecatedTokenMatches = getDeprecatedKaizenTokenPatternMatches(
+      newValue
+    )
     if (deprecatedTokenMatches) {
       options.reporter({
         message: containsDeprecatedKaizenTokenWithNoReplacement(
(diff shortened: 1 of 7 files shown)


packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.ts
```diff
@@ -70,7 +70,8 @@ const noInvalidEquations = (
           options.reporter({
             message: negatedKaizenVariableMessage,
             node: postcssNode,
-            autofixAvailable: !isVariable(postcssNode),
+            autofixAvailable:
+              !isVariable(postcssNode) && options.language === "scss",
```
dmisdm, line 74: This was a bug.
We don't support autofixing negated variables in LESS, and even though the code does this correctly (look at line 55), we don't want to report that an autofix exists when in LESS.

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -17,11 +25,16 @@ const testStylelintConfig = {
...
   testName: string
-  expectedWarnings: number
+
   input: string
   expectedOutput: string
   only?: boolean
-}
+} & (
+  | {
+      expectedWarningMessages: string[]
```
jasononeil, line 34: Interesting! I think this is a good move for the usefulness of this test suite

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -221,23 +245,29 @@ const testExamples: TestExample[] = [
...
+        "darken"
+      ),
+    ],
   },
   {
     language: "scss",
     testName: "doesn't fix tokens within variables",
     input:
-      '@import "~@kaizen/design-tokens/sass/color"; $foo: $kz-color-wisteria-800;',
+      '@import "~@kaizen/design-tokens/sass/color"; $foo: $color-purple-800;',
```
jasononeil, line 260: Is the input supposed to change on this test?

I imagine you still want the old `kz-color-wisteria-800` token so that you can test it isn't fixed. Either that, or the `testName` feels misleading and I'm misunderstanding what the test is supposed to be :)

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -619,50 +670,30 @@ const testExamples: TestExample[] = [
     testName:
       "transitive tokens are fixed correctly when their usages are negated, and their replacements are simple",
     input:
-      '@import "~@kaizen/design-tokens/sass/spacing"; $foo: $kz-spacing-md; .foo { top: -$foo; }',
+      '@import "~@kaizen/design-tokens/sass/spacing"; $foo: $spacing-md; .foo { top: -$foo; }',
     expectedOutput:
-      '@import "~@kaizen/design-tokens/sass/spacing"; $foo: $kz-spacing-md; .foo { top: calc(-1 * #{$spacing-md}); }',
-    expectedWarnings: 1,
+      '@import "~@kaizen/design-tokens/sass/spacing"; $foo: $spacing-md; .foo { top: calc(-1 * #{$spacing-md}); }',
+    expectedWarnings: 0,
```
jasononeil, line 676: Interesting that the `expectedWarnings` used to be 1, this feels like a bug we missed

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
```diff
@@ -953,6 +993,26 @@ const testExamples: TestExample[] = [
...
+      .test {
+        color: var(--kz-var-color-wisteria-100);
+      }
+    `,
+    expectedWarningMessages: [
+      containsDeprecatedKaizenTokenWithNoReplacement(
+        "kz-var-color-wisteria-100"
+      ),
+    ],
+    only: true,
```
jasononeil, line 1014: This `only` looks out of place 😉

packages/stylelint-plugin/src/util/kaizenTokens.ts
```diff
@@ -1,4 +1,4 @@
-import { Utils } from "@kaizen/design-tokens"
+import { mapLeafsOfObject } from "@kaizen/design-tokens/src/utils"
```
jasononeil, line 1: In general we prefer to import from the root level rather than a module inside the package, as we want to be able to move modules around and not break any consumers. The exports from the package index should be the only public API.

See this recent Stack Overflow answer from Louis: https://stackoverflow.com/c/cultureamp/a/1641/63

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
```diff
@@ -24,10 +24,10 @@ import { fixAlphaModificationFunctions } from "./lib/fixAlphaModificationFunctio
 import { containsEquationThatDoesntWorkWithCSSVariables } from "./no-invalid-use-of-var-tokens-in-equations"
 
 const deprecatedKaizenTokenPattern = new RegExp(
-  Array.from(deprecatedTokenReplacements.keys()).join("|")
+  `(${Array.from(deprecatedTokenReplacements.keys()).join("|")})`
 )
-function stringContainsDeprecatedKaizenToken(value: string) {
-  return deprecatedKaizenTokenPattern.test(value)
+function getDeprecatedKaizenTokenPatternMatches(value: string) {
+  return value.match(deprecatedKaizenTokenPattern)
```
jasononeil, line 30: I'm curious: what was the reason for this change? Is `str.match(regex)` preferable to `regex.test(string)`?

jasononeil: Oh you're capturing them. I get it now :)

This might have been more obvious if there was an explicit return type on the function. But the name did change, I was just slow to notice :)

jasononeil (changes requested):
A few comments and questions, and 2 minor requests for changes.

Otherwise looking good!

packages/stylelint-plugin/src/util/kaizenTokens.ts
dmisdm: Hey i agree i don't want to do this but i wanted something that worked both with the current state and the next state of design-tokens - i should have been explicit with this change.
https://github.com/cultureamp/kaizen-design-system/pull/1833#discussion_r686477547

This is a very minor change anyway so ill put it back, then make the change in #1833 to not use the Utils namespace.

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
dmisdm: Gah sorry that was from debugging!

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
dmisdm: This is because we're not using a deprecated token anywhere in the new test anymore.
If the warning messages we're explicit in the prior this would have been a bit more obvious!

packages/stylelint-plugin/src/stylelintPlugin.spec.ts
dmisdm: Hmm this is a good point thanks, it probably overlooked it when renaming things.

packages/stylelint-plugin/src/rules/no-invalid-use-of-tokens.ts
dmisdm: Made it more explicit :)

packages/stylelint-plugin/src/util/kaizenTokens.ts
jasononeil: Nice, I appreciate that coming change :)

jasononeil (approved):
Nice! 🚢

PR #1839, opened 2021-08-12
feat: Animate empty state illustration

ActuallyACat:
- By default empty state components should animate and play once
- isAnimated and loop props exposed on EmptyState

https://dev.cultureamp.design/ally/animated-brand-moments/storybook/?path=/story/components-empty-state--neutral

### DIFF
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
@@ -1,12 +1,12 @@
 import classnames from "classnames"
 import * as React from "react"
-import { useTheme } from "@kaizen/design-tokens"
 import {
   EmptyStatesAction,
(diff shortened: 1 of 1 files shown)


draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -52,8 +52,10 @@ const EmptyState: EmptyState = ({
   children,
   straightCorners,
   useZenStyles,
+  isAnimated = true,
+  loop = false,
 }) => {
-  const theme = useTheme()
+  const animationProps = isAnimated ? { isAnimated, loop } : {}
```
ActuallyACat, line 58: This might look a bit strange at first, but it's working as expected. If the illustration is _not_ animated, then passing through a loop property (regardless of its value) does not make sense.

dougmacknz (approved):

PR #1840, opened 2021-08-13
feat: Bump design-tokens in peer dependencies to v3

dmisdm:
Change each peer dependency **on** design-tokens to v3.

# Motivation and Context

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/avatar/package.json
+++ b/draft-packages/avatar/package.json
@@ -38,7 +38,7 @@
     "react-textfit": "^1.1.0"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^3.0.0",
+    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0",
     "react": "^16.9.0"
   }
 }
--- a/draft-packages/avatar/package.json
+++ b/draft-packages/avatar/package.json
@@ -38,7 +38,7 @@
     "react-textfit": "^1.1.0"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3",
+    "@kaizen/design-tokens": "^3.0.0",
     "react": "^16.9.0"
   }
 }
--- a/draft-packages/badge/package.json
+++ b/draft-packages/badge/package.json
@@ -35,7 +35,7 @@
     "classnames": "^2.2.6"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3",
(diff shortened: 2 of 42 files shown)


draft-packages/avatar/package.json
```diff
@@ -38,7 +38,7 @@
     "react-textfit": "^1.1.0"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3",
+    "@kaizen/design-tokens": "^3.0.0",
```
jasononeil, line 41: Do we have any thoughts about supporting v2 and v3 concurrently for a while?

Once we switch only to v3, it would effectively mean people can't upgrade components until they upgrade the design tokens. Which is maybe a good forcing function, but could also be disruptive if a team needs a quick upgrade to ship something.

I see the options as:

1. Support both v2 and v3 side-by-side for a while (this will work until the next `feat:` bump in design tokens)
2. Merge this, and require everyone to switch to v3 before upgrading any other packages

With option (1) we could wait until everybody has upgraded, or at least, until the monoliths are upgraded, before we remove support for v2.

```suggestion
    "@kaizen/design-tokens": "^2.10.3 | ^3.0.0",
```

jasononeil (commented):
There's a discussion that needs to happen before we merge this (see my other comment)

draft-packages/avatar/package.json
dmisdm: Thats a good idea, i'm just thinkng about this a bit.

You may disagree with me here.

My initial thought is to leave it as just `^3.0.0` because this isn't going to block people, it's only going to warn them...and i like that; if they upgrade a component without upgrading design-tokens i want them to get a warning, whereas this change would cause them not to get one i think?

Also, updating the design-tokens peer dependency in each package once a new `feat` is released is quite easy to forget about and miss IMO. Although on second thought...i don't think it makes a difference either way here; even if peers are on `^3.0.0` or `^2.10.3 | ^3.0.0`, the next developer that adds a `feat` to design-tokens and uses such a feature within a component will still have to be dilligent with changing this.

dmisdm: @jasononeil seeing as using `^2.10.3 | ^3.0.0` is the path of least resistence, lets do that, as i'd like to get this merged soon and preferrably before posting about how to update design-tokens.

jasononeil (approved):

jasononeil (approved):
Good pickup on the double pipe 😓

PR #1846, opened 2021-08-16
chore: add tests for textarea and toggleswitch

lloydstubber:
# Objective
- Add tests to the TextArea and ToggleSwitch components

# Motivation and Context
- Future-proofs our components
- Closes this: https://github.com/cultureamp/kaizen-design-system/issues/911

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
@@ -18,9 +18,18 @@ const renderTextArea = (props?: TextAreaProps) => {
 }
 
 describe("<TextArea />", () => {
-  it("should render a value", () => {
+  it("should render a value when component is controlled", () => {
     const { queryByText } = renderTextArea()
     expect(queryByText(defaultTextAreaProps.value)).toBeTruthy()
+  })
+
+  it("should render a default value when component is uncontrolled", () => {
+    const { queryByText } = renderTextArea({
+      defaultValue: "default value",
+      value: undefined,
+    })
+    expect(queryByText(defaultTextAreaProps.value)).toBeFalsy()
+    expect(queryByText("default value")).toBeTruthy()
   })
 
   it("should call the `onChange` event when the value is updated", () => {
--- a/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
@@ -1,0 +1,56 @@
+import { cleanup, fireEvent, render, screen } from "@testing-library/react"
+import * as React from "react"
+import { TextAreaProps } from "./TextArea"
+import { TextArea } from "."
+
+afterEach(cleanup)
+
+const defaultTextAreaProps = {
+  id: "someTextAreaId",
+  value: "Some field value",
+  onChange: jest.fn(),
+}
+
+const renderTextArea = (props?: TextAreaProps) => {
+  const mergedTextAreaProps = { ...defaultTextAreaProps, ...props }
+
+  return render(<TextArea {...mergedTextAreaProps} />)
+}
+
+describe("<TextArea />", () => {
+  it("should render a value", () => {
+    const { queryByText } = renderTextArea()
+    expect(queryByText(defaultTextAreaProps.value)).toBeTruthy()
+  })
+
+  it("should call the `onChange` event when the value is updated", () => {
+    renderTextArea()
+
+    fireEvent.change(screen.getByRole("textbox"), {
+      target: { value: "Hello" },
+    })
+
+    expect(defaultTextAreaProps.onChange).toBeCalledTimes(1)
+  })
+
+  it("should render an `data-automation-id` attribute", () => {
+    const automationId = "someTextAreaAutomationId"
+
+    const { container } = renderTextArea({ automationId })
+    expect(
+      container.querySelector(`[data-automation-id="${automationId}"]`)
+    ).toBeTruthy()
+  })
+
+  it("should render an `id` attribute", () => {
+    const { container } = renderTextArea()
+    expect(
+      container.querySelector(`[id="${defaultTextAreaProps.id}"]`)
+    ).toBeTruthy()
+  })
+
+  it("should render a reversed text area", () => {
+    const { container } = renderTextArea({ reversed: true })
+    expect(container.querySelector(".reversed")).toBeTruthy()
+  })
+})
--- a/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
@@ -42,22 +42,6 @@ describe("<TextArea />", () => {
     expect(defaultTextAreaProps.onChange).toBeCalledTimes(1)
   })
 
-  it("should render an `data-automation-id` attribute", () => {
-    const automationId = "someTextAreaAutomationId"
-
-    const { container } = renderTextArea({ automationId })
-    expect(
-      container.querySelector(`[data-automation-id="${automationId}"]`)
-    ).toBeTruthy()
-  })
-
-  it("should render an `id` attribute", () => {
-    const { container } = renderTextArea()
-    expect(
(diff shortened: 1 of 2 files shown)


ActuallyACat (commented):
This LGTM. I have a question that I've posted on this PR's parent issue. I want to ensure these test cases cover what happened in the incident before I ✅

lloydstubber:
So I've tried replicating the issue that broke it. Where you specify a `defaultValue` and `value` but all I can repro is a console error stating this:

>Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.

```
// this passes
const { container } = render(<TextArea defaultValue="default" value="value" />)
expect(container).toBeInTheDocument
```

@dougmacknz could what is erroring in the browser be dependant on a library (autogrow) and not be possible to throw in a test env? Or was it just this error above?

lloydstubber:
@allybeaton just paired with @dougmacknz and he walked me through the initial issue and I've added a test for it

draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+  })
+
+  it("should render a default value when component is uncontrolled", () => {
+    const { queryByText } = renderTextArea({
+      defaultValue: "default value",
+      value: undefined,
+    })
+    expect(queryByText(defaultTextAreaProps.value)).toBeFalsy()
+    expect(queryByText("default value")).toBeTruthy()
+  })
```
lloydstubber, line 33: This will catch the issue that was raised initially

dougmacknz:
Documenting what we discussed on Slack:

I'm not a huge fan of the tests for `id` and `automation-id` (but since learnt that those exist on our other form components). I'd probs remove them, unless @ActuallyACat is opposed to that.

I like Kent C Dodd's philosophy of user behaviour focused testing, so testing the typical behaviour that we expect for actual users. Something like this:

TextArea
* can see the default value (already done)
* can edit the default value
* can see the controlled value (already done)
* can edit the controlled value
* can see an error message when it is in an error state

ToggleSwitch
* can see the default value
* can toggle the switch (already done)
* cannot toggle a disabled switch

ActuallyACat:
> I'm not a huge fan of the tests for id and automation-id (but since learnt that those exist on our other form components). I'd probs remove them, unless @ActuallyACat is opposed to that.

My vote is to allow any `data-*` attributes be spread onto the wrapper of the component, and to remain unopinionated about what automation/test attributes camps use. 

> I like Kent C Dodd's philosophy of user behaviour focused testing

👍🏻 The visual regression tests in Chromatic ought to cover the visual changes.

ActuallyACat (dismissed):

lloydstubber:
@dougmacknz @allybeaton So I've been down a bit of a rabbit hole of how to test these two:

TextArea
- can edit the default value
- can edit the controlled value

It seems we don't test the interactivity currently in the way we can test above. We can check the `onChange` fires, but we can't for instance do this:

```
  const defaultTextAreaProps = {
    id: "someTextAreaId",
    value: "Some field value",
    onChange: jest.fn(),
  }

  it("can edit a controlled component", () => {
    const { queryByText } = renderTextArea()
    expect(queryByText(defaultTextAreaProps.value)).toBeTruthy()
    fireEvent.change(screen.getByRole("textbox"), {
      target: { value: "Hello" },
    })
    expect(queryByText("Hello")).toBeTruthy()
  })
```

What we end up with is the value doesn't change, as there is no `onChange` event changing the input _unless_ we either mock one, or [use a library](https://testing-library.com/docs/ecosystem-user-event) that gives us a little more user interactivity. I have used this lib in other codebases and it works well for this kind of thing. Unsure if it's worth doing a sweep with it for all inputs though. 

WDYT?

dougmacknz:
Sorry I forgot to mention that I wouldn't be unhappy if some of those test cases weren't able to be done quickly. I'm thinking we should just roll with what you've got here for now, and we can have a follow up discussion about that at a later time, especially given we will want to revisit the tests on the other components as well. That sound alright to you?

lloydstubber:
Yep sounds good! Good to go for a ✅ now I've rebased @dougmacknz

draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  it("should call onToggle when toggle is changed", () => {
+    renderToggleSwitch()
+
+    fireEvent.click(screen.getByRole("checkbox"))
+    expect(defaultToggleSwitchProps.onToggle).toHaveBeenCalledTimes(1)
+  })
+
+  it("should show toggledStatus when passed through", async () => {
+    renderToggleSwitch()
+    expect(defaultToggleSwitchProps.toggledStatus).toEqual("off")
```
dougmacknz, line 31: Hmm, is there a way to gather this information from what is rendered instead? I can see that there's `on` and `off` classes that are put on the container element in the implementation. Could that be used?

lloydstubber: Oh there's classes? Should be able to swing something then. Let me have a poke

draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  })
+
+  it("has disabled attribute when disabled prop passed in", async () => {
+    const { container } = renderToggleSwitch({ disabled: true })
+    expect(container.querySelector("[disabled]")).toBeTruthy()
```
dougmacknz, line 36: I don't mind this too much, but a way to make this more user behaviour focused could be to use the same code as you have in the `onToggle` test above, and make sure that clicking this disabled one _doesn't_ fire the `onToggle` method.

lloydstubber: Yep I tried that but unfortunately our logic to handle if it's disabled etc. is _inside_ of the `onToggle` so it still gets called, then decides what to do. So this would be a good case again for the `userEvent` library and potentially mocking the function.

draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.spec.tsx
lloydstubber: Yep, that worked a treat. Updated.

draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.spec.tsx
dougmacknz: Oh right, fair enough 👍

draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
```diff
@@ -0,0 +1,49 @@
+import { cleanup, fireEvent, render, screen } from "@testing-library/react"
+import * as React from "react"
+import { TextAreaProps } from "./TextArea"
+import { TextArea } from "."
+
+afterEach(cleanup)
```
dougmacknz, line 6: This is an FYI, don't worry about fixing this now because we can do a sweep, but apparently these `afterEach(cleanup)` calls are obsolete now: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup

dougmacknz (approved):
Great stuff, thanks for working through all that 🚢

draft-packages/form/KaizenDraft/Form/Primitives/TextArea/TextArea.spec.tsx
lloydstubber: Good to know. When I was trying to write another test that fired a mock function it didn't actually clean it up and kept incrementing. So I changed it to this and it worked (but don't need it anymore as I had to scrap that test):

```
afterEach(() => {
  jest.resetAllMocks()
})
```

PR #1849, opened 2021-08-16
fix: Check for existing toast-notification-manager-portal before creating

dougmacknz:
Canary version of https://github.com/cultureamp/kaizen-design-system/pull/1848 so that I can check if this actually fixes the issue


ActuallyACat (approved):

makenosound:
I’m not sure this’ll fix the underlying issue because we still need to create a true singleton somehow to ensure that the place we store state is retained.

Perhaps we should chat about approaches @dougmacknz, I have an idea for how this can be resolved with the existing implementation but with our change of opinion about where React fits into Kaizen perhaps there’s a less problematic approach using providers.

PR #1850, opened 2021-08-16
chore(site): Add relevant issues to component pages

ActuallyACat:
# Objective
This queries https://api.github.com/repos/cultureamp/kaizen-design-system/issues?per_page=1000 on each Gatsby build and displays a list of relevant issues when user is on a component page.

For this to work, issues need to have an agreed-upon naming convention. In this PR, issues will need to be tagged with `component:${componentName}` in Github, where `componentName` is the name of the component that matches the `githubLabels` in Frontmatter. I've allowed `githubLabels` so we can choose to specify multiple labels (or no labels) as needed. For example, this documentation page (https://cultureamp.design/components/checkbox-field/) might be tagged with `component:checkbox-field` and `component:form`. 

~I'm not sure if the Github status ("open") is useful enough to put in the UI. My backup idea was to put a list of the Github tags within the table as those provide additional context about the issue without needing to click through to Github. This currently looks ugly with the current issue tag organisation in Github.~ I'm now using the Github labels. 

Here's where we could be if we rethought about our tagging (obvs hiding all the `component:*` tags from the UI) - WDYT? 
Screen Shot 2021-08-16 at 7 38 39 pm

Screen Shot 2021-08-16 at 6 52 02 pm


site/docs/components/avatar.mdx
```diff
@@ -3,6 +3,7 @@ title: Avatar
 navTitle: Avatar
 summaryParagraph: An avatar represents an individual or group visually.
 tags: ["Profile picture", "Display picture"]
+githubTags: ["component:avatar"]
```
ActuallyACat, line 6: Before I go ahead an update >30 documentation files, WDYT of this approach?

lloydstubber:
This is awesome! I think it'll be super helpful and it being dynamic with a low barrier to update is great. Will the conventional commits be forced in Moai?

IMO I don't think the "Open" label is the same level of label as the type of issue it is. Would it not just be if it's there, it's open?

ActuallyACat:
> IMO I don't think the "Open" label is the same level of label as the type of issue it is. Would it not just be if it's there, it's open?

Yeah, that's true. I've now updated this to include the github labels. It might be a bit messy until we organise issues 😓

lloydstubber (approved):
LGTM! Excited to see this go in

PR #1852, opened 2021-08-17
Mab/progress bar

mbylstra:
# Objective

# Motivation and Context

# Checklist

# Additional Considerations


packages/progress-bar/src/Kaizen/ProgressBar.elm
```diff
@@ -0,0 +1,93 @@
+module Kaizen.ProgressBar exposing (Mood(..), view)
```
mbylstra, line 1: The elm version is simplified compared to the React version (is missing variants). It just has the variant we needed. It doesn't seem as important to implement them due to Elm being contained.

site/docs/components/progress-bar.mdx
```diff
@@ -10,12 +10,13 @@ needToKnow:
 health:
   designed: true
   documented: true
-  implemented: false
-  latestDesign: false
-  allVariants: false
+  implemented: true
```
mbylstra, line 13: It's worth noting that the Figma file does not have a design for the "indeterminate" variant mentioned in  https://cultureamp.design/components/progress-bar/ so that has not been implemented.

site/docs/components/progress-bar.mdx
```diff
@@ -10,12 +10,13 @@ needToKnow:
...
+  latestDesign: true
+  allVariants: true
   responsive: false
```
mbylstra, line 16: I don't think anything would need to be done to make it responsive (it's naturally responsive in terms of width), but I'm also not sure if smaller screens call for a different design (such as font sizes), so I've left it as false.

site/docs/components/progress-bar.mdx
```diff
@@ -10,12 +10,13 @@ needToKnow:
...
   internationalized: false
   accessible: false
```
mbylstra, line 18: I _think_  it is accessible due to the aria attributes I added (which are also recommended in https://cultureamp.design/components/progress-bar/ ), but perhaps this is best left to false until it has been officially audited.

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return value.kind === "percentage"
+    ? value.value
+    : (value.value / value.max) * 100.0
+}
+
+function isAnimating({ variant, mood, value }: Props) {
+  if (variant === "static") {
+    return false
+  }
+  if (mood === "negative") {
```
mbylstra, line 43: This came from the legacy component. If it's in an error state, it's not going to be progressing, so it makes sense to stop the animation (sounds good to me)

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  value: Value
+  variant: Variant
+  mood: Mood
+  subtext?: string
+  labelExtraContent?: React.ReactNode
+}
+
+type Value =
+  | { kind: "percentage"; value: number }
+  | { kind: "fraction"; value: number; max: number }
```
mbylstra, line 16: This is a "discriminated union type", where the two options are discriminated by the "kind".

FWIW, this is what it would look like in Elm (it's the same idea):

```
type Value =
  Percentage Int
  | Fraction { value: Int, max: Int }
```

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+      ) : null}
+    </div>
+  )
+}
+
+function Label(props: Props) {
+  const percentage = calculatePercentage(props.value)
+  const content =
+    (props.value.kind === "percentage"
+      ? `${Math.round(percentage)}%`
```
mbylstra, line 86: I added Math.round because I think `33` looks much better that `33.3333333333` and I can't imagine we'd need more precision than that. We could add a "precision" prop later if it was really needed, and it wouldn't be a breaking change.

packages/progress-bar/docs/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+      description: {
+        component: 'Import { ProgressBar } from "@kaizen/progress-bar"',
+      },
+    },
+  },
+}
+
+export const DefaultStory = _ => (
+  <ProgressBar
+    value={{ kind: "percentage", value: 25 }}
```
mbylstra, line 20: This is what the "nested prop" with the "discriminated union type" mentioned earlier looks like in usage.

mbylstra: Alternatively it could be:

for percentage: 

```
<ProgressBar
    variant="loading"
    value={100}
    mood="positive"
 />
 ```

 for a fraction:
 ```
<ProgressBar
    variant="loading"
    value={1}
    max={10}
    mood="positive"
 />
 ```
    
which looks a little neater, but my point stands about the type errors being much easier to read with the nesting, and also it's really not clear that the first one will render as a percentage (the other one is more self-documenting)

File: site/docs/components/progress-bar.mdx [hunk 9]
```diff
@@ -37,11 +38,8 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
 - Static process or loading progress:
     - **Static**: a static progress bar shows current progress through a process that will not progress without user interaction.
     - **Loading**: a loading progress bar shows current progress through a loading process that the system will progress by itself without user interaction. We use animation to show that the system is operating.
-- Props:
-    - Min
-    - Max
-    - Initial value
-    - Infinite/indeterminate
+- render a Percentage (33%) or Fraction (1/10)
+- Infinite/indeterminate (not yet designed)
```
mbylstra, line 42: I updated the docs slightly to reflect how I implemented it. I think it's better to keep these docs high level design guides rather than an exact spec of the React props (so I removed the "props" label) - it's too easy for it to get out of sync. We can automatically generate React/Typescript props docs.

site/docs/components/progress-bar.mdx
dougmacknz: For now we are just doing our best to assess this ourselves, and we will probably introduce some data for whether these have been part of an audit or not in the future.

I can take a look tomorrow if you like and see if I can find any accessibility issues.

packages/progress-bar/ElmStories/ProgressBarStories.elm
```diff
@@ -0,0 +1,27 @@
+module ElmStories.ProgressBarStories exposing (main)
```
sentience, line 1: This file's path doesn't appear to be covered by the `"source-directories"` in the **elm.json** file. Is this a problem? Should this be moved under the **src** directory?

packages/progress-bar/src/Kaizen/ProgressBar.elm
sentience: Worth exposing `Configuration` just so that consumers don't have to redefine it if they need to pass a value of this type around?

packages/progress-bar/src/Kaizen/ProgressBar.elm
```diff
@@ -0,0 +1,96 @@
...
+import CssModules exposing (css)
+import Heading.Heading as Heading exposing (TypeVariant(..))
+import Html exposing (..)
+import Html.Attributes exposing (attribute, style)
+
+
+type alias Configuration =
+    { progressPercentage : Int
+    , mood : Mood
+    }
```
sentience, line 13: I gather you're deliberately preferring the simplicity of an plain record rather than an opaque `Configuration` type with exposed constructors. I guess the question is, will this lead to more breaking changes in the API as we add features to this component in the future?

Not strongly opinionated on this, but I think I'd have gone with the opaque type myself, knowing that there are already features designed for this component that aren't implemented here.

packages/progress-bar/src/Kaizen/ProgressBar.elm
```diff
@@ -0,0 +1,96 @@
...
+            [ box [ marginBottom 0.25 ]
+                [ Heading.view
+                    (Heading.p |> Heading.variant Heading4)
+                    [ text <| String.fromInt progressPercentage ++ "%" ]
+                ]
+            ]
+        , div
+            [ classList [ ( .progressBackground, True ) ] ]
+            [ div
+                [ styles.class (moodToClass mood)
```
sentience, line 70: For consistency:
```suggestion
                [ class (moodToClass mood)
```

sentience (approved):
The Elm is looking pretty good to me. A couple of things caught my eye that you might want to look at, but I'm happy to approve as-is and let you decide if they need changing.

packages/progress-bar/src/ProgressBar.tsx
OliBridgman: Personally, i'm not a fan of the nested prop. An object is likely going to be re-instantiated by the parent on a re-render, which means the progress bar is going to re-render each time. Now, that isn't the biggest deal in the world, it's not likely to cause performance issues or anything, but it could be a dangerous precedent to set for the design system. This component could get more complex over time, or other components could be built with this API as prior art. I believe the components from the design system should be as stable as possible, and if that comes at the expense of a more difficult to read type error, so be it.

But I think I have a seperate concern with the union itself. Max isn't a prop that can't exist when things are a percentage, rather, it _should_ be 100. If the progress bar were to have a default value of 100 for its max, then max can be an optional prop. With this, any value passed can be treated as a percentage and max can be ignored by the consumer in the default case. In the case of fractions, the max can be specified as anything else. The resulting progress is the same calculation and the rendered result is the same.

Where this breaks down is the label. But I don't think the type _should_ make the label change. The label seems like a perfectly reasonable thing for a consumer to want to control at render time. If we allow that to exist they can render:

* `${value}%`
* `${value} percent`
* `${value}/${max}`
* `${value} with any extra content you like`

The consumer can now also choose _not_ to render a label. We'd also no longer need to care about the `extraContent`.

packages/progress-bar/src/ProgressBar.tsx
OliBridgman: I don't believe we should block animations for consumers that want negative moods. I'd imagine they were called `moods` in part so that they were not necessarily associated with error states? I may be progressing to something bad and gosh darn it I want to animate while I get there! Can animation be its own prop instead?

packages/progress-bar/src/ProgressBar.tsx
dougmacknz: I think I like the simpler approach as well. I don't think it's unreasonable for this to only accept a percentage value rather than offering the ability to calculate the percentage from a fraction.

Moving the responsibility of the label onto the consumer seems like the most future proof solution, however I do wonder about the accessibility impact of no label - mainly because the colour contrast might be hard for some to see, and the label provides a way to see that information with sufficient contrast. It also means that consumers will need to support RTL themselves, which will probably be patchy.

Tough one, but I'm still leaning towards giving the consumer the responsibility of the label and crossing that bridge when we get there.

site/docs/components/progress-bar.mdx
dougmacknz: Unfortunately I'm pretty sure most of the colours are going to fail contrast ratio requirements. I'm not sure if the label is enough to invalidate that or not, but I guess we'll have this as `false` for now.

packages/progress-bar/docs/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { figmaEmbed } from "../../../storybook/helpers/figmaEmbed"
+import { ProgressBar } from "../index"
+
+export default {
+  title: `${CATEGORIES.components}/ProgressBar`,
+  component: ProgressBar,
+  parameters: {
+    docs: {
+      description: {
+        component: 'Import { ProgressBar } from "@kaizen/progress-bar"',
```
dougmacknz, line 12: /nit lowercase `i`

packages/progress-bar/src/Kaizen/ProgressBar.elm
mbylstra: TBH this is just what I copied from murmur and I didn't think of it. I agree, opaque types are generally better. I've pushed a change.

packages/progress-bar/src/ProgressBar.tsx
mbylstra: I've changed the API so it's just `?label`. Let me know if that's not what you meant (or feel free to add a commit with exactly what you were thinking).

packages/progress-bar/ElmStories/ProgressBarStories.elm
mbylstra: Interesting one. It works because `ProgressBarStories` is the `main`, so Elm doesn't need to find it in `source-directories`

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
+import { Box, Heading } from "@kaizen/component-library"
+import React from "react"
+import classnames from "classnames"
+import styles from "./ProgressBar.scss"
+
+type Props = {
+  value: Value
+  variant: Variant
```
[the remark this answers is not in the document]
mbylstra: I've removed `variant` and replaced it with `animating`. So now there's full flexibility (consumer can choose whether it's animated or not, but is also responsible for stopping the animation when it gets to 100%, and can choose to stop the animation when there's an error). I think having both `animated` and `animating` props would be pretty confusing so going for full flexibility here instead.

packages/progress-bar/ElmStories/ProgressBarStories.elm
sentience: I think even if the Elm compiler can accept it, this'll break some dev tools like elm-format and the Elm Language Server.

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,72 @@
+import { Box, Heading } from "@kaizen/component-library"
+import React, { ReactNode } from "react"
+import classnames from "classnames"
+import styles from "./ProgressBar.scss"
+
+type Props = {
+  value: number
+  max: number
+  animating: boolean
```
dougmacknz, line 9: I wonder if we should make this prop name consistent with Illustration's `isAnimated`

packages/progress-bar/src/ProgressBar.tsx
```diff
@@ -0,0 +1,72 @@
+import { Box, Heading } from "@kaizen/component-library"
+import React, { ReactNode } from "react"
+import classnames from "classnames"
+import styles from "./ProgressBar.scss"
+
+type Props = {
+  value: number
+  max: number
```
dougmacknz, line 8: Sorry I missed this at first, but having a required `max` prop seems like more trouble for consumers than we need, given I'd imagine most cases will be 100.

I just did a bit of searching and came across this as well, sounds like there _could_ be cases where screen readers don't respect a max value that isn't 100:
https://material-ui.com/components/progress/#non-standard-ranges

site/docs/components/progress-bar.mdx
```diff
@@ -38,11 +39,8 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
 - Static process or loading progress:
     - **Static**: a static progress bar shows current progress through a process that will not progress without user interaction.
     - **Loading**: a loading progress bar shows current progress through a loading process that the system will progress by itself without user interaction. We use animation to show that the system is operating.
-- Props:
-    - Min
-    - Max
-    - Initial value
-    - Infinite/indeterminate
+- Percentage (33%) or fraction (1/10)
```
dougmacknz, line 42: Do we actually want to support fractions as labels, or was this from when the value prop could accept a fraction?

packages/progress-bar/src/ProgressBar.tsx
mbylstra: There are plenty of cases in the product (actually more than when it's used as a loader) where the progress bar is not a percentage (see https://cultureamp.design/components/progress-bar/). If value is always a percentage then it shouldn't be named value, it should be `percentage`, or how would you know it's a percentage?  This is the same API as https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress. Btw, I'd considered using that for the implementation but I don't think it'd actually offer any advantage (I'd guess that screen readers would be more likely to respect aria elements than the progress element), and just the hassle of having to undo the browser default styles in  cross browser compatible way (although it'd be the semantically correct choice)

> I just did a bit of searching and came across this as well, sounds like there could be cases where screen readers don't respect a max value that isn't 100
If that's true we can easily normalise the value in the implementation to turn it into a percentage for the aria props (just seems like a limitation of screen readers that we can easily work around)

site/docs/components/progress-bar.mdx
mbylstra: It's from the original docs but the original docs were more specific in the implementation. I see the docs as design guidelines (as in "it is common or acceptable to use a fraction as a label"). So where are implementation is extremely flexible (people can put whatever they want in the label), the docs are there to provide guidance for common uses.

site/docs/components/progress-bar.mdx
```diff
@@ -10,12 +10,13 @@ needToKnow:
...
   designed: true
   documented: true
-  implemented: false
-  latestDesign: false
-  allVariants: false
+  implemented: true
+  latestDesign: true
+  allVariants: true
   responsive: false
   internationalized: false
```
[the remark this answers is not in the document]
mbylstra: I'm not sure what our common approach to internationalisation is (css or react context or prop drilling or what?) to be able to implement this, so I'd prefer to wait.

packages/progress-bar/src/ProgressBar.tsx
mbylstra: Actually I just realised I was already doing that (kind of unintentionally, it came from the original implementation, but it just made sense to normalise things anyway) - the aria max value is always 100: https://github.com/cultureamp/kaizen-design-system/pull/1852/files#diff-cd39c500d02e475cd6675f1db913712b4d0b876673722bb37e8518e9f59d4eb4R40

packages/progress-bar/src/ProgressBar.tsx
mbylstra: it seems like a different thing to me. "animating" is the current state of whether the bar is . It would literally be a bug if the consumer didn't stop animating the the progress bar after it reaches 100% (which consumers will probably do).  `isAnimated` sounds like the progress bar will do this for them, but it won't.

dougmacknz (approved):
Thanks for working through all of that!

mbylstra:
No probs. Thanks for the review. Just waiting on design review.

PR #1859, opened 2021-08-19
feat: Add Kaizen eslint plugin

ActuallyACat:
## About 
Add an Eslint plugin, which is a collection of ESLint rules for Kaizen. This plugin aims to solve the following problems:
- Large-scale migrations of Kaizen packages
- Encourage Kaizen best practices and anticipate common mistakes

Screen Shot 2021-08-25 at 7 02 15 pm


packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Install
+Add the package to your project's dependencies:
+
+`yarn add @kaizen/eslint-plugin`
+
+Add the plugin to your Eslint configuration: 
+```
+{
+  plugins: ["@kaizen"],
```
dmisdm, line 14: Just a thought:
In the same vein as the stylelint-plugin, we started with just a plugin but ended up also exposing a recommended config to extend from.
Do you see that happening with this package?

dmisdm (approved):
Awesome work!

dmisdm:
@ActuallyACat 
> This will (eventually) need a corresponding rule in Stylelint to catch Sass imports too.

0_o .... as in to pick up things like `@kaizen/draft-button/src/forbidden/file/why-u-do-dis.scss` ?

packages/eslint-plugin/src/rules/no-draft-imports.ts
```diff
@@ -0,0 +1,63 @@
...
+                      "kaizen/"
+                    )
+                    return fixer.replaceText(arg, newValue)
+                  }
+                  return null
+                },
+              })
+            }
+          })
+        }
```
dmisdm, line 38: Im curious, should we add a check here to see if the `callee` is `require`? Or, `require.resolve`, `require.resolve.paths` etc.  Otherwise (if i'm understanding correctly) will this potentially pickup any function that has `@kaizen/draft` anywhere in its parameters?

I wouldn't say this is a PR blocking problem as we can always amend later if what i say has value.

dmisdm: For extra dilligence we could also add a check to see if it's `import(...)`.

It's not like anyone at CA is doing this though so up to you.

ActuallyACat:
> 
> > This will (eventually) need a corresponding rule in Stylelint to catch Sass imports too.
> 
> 0_o .... as in to pick up things like `@import ~@kaizen/draft-button/src/forbidden/file/why-u-do-dis.scss` ?

Yeah, exactly :( Unfortunately, that's a scenario that's common enough that we'll probably need to consider.

packages/eslint-plugin/README.md
ActuallyACat: Yeah! I think that would be likely. For now we don't really have any rules but I imagine a future we'd want to use this plugin to encourage best Kaizen practices when working with components.

packages/eslint-plugin/src/rules/no-draft-imports.ts
ActuallyACat: Oh! I did not even think of this use case - thank you. I'll pop in a test for it while I'm at it.

dmisdm (approved):

dougmacknz (approved):

PR #1881, opened 2021-08-23
chore: add issue template to repo

lloydstubber:
After going through multiple issues and speaking with a few campers who have recently opened issues I’ve come up with an issue template. My main concern which @ActuallyACat brought up was to keep it brief or it’ll get ignored, so hoping this is a good start.

Could potentially trim down the labelling part and we handle that ourselves. WDYT?

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -5,7 +5,7 @@
 ## Expected Behavior
 
 
-## Actual Behavior (please provide screenshots)
+## Actual Behavior (please provide screenshots if it's a visual issue)
 
 
 ## Steps to Reproduce the Problem
@@ -23,4 +23,4 @@
 1. If related to a component, please add a label like so `component:mycomponent`.
 2. Then label with `bug`, `enhancement`/`feature`\* or `documentation`.
 
-**\* A feature is a new component/functionality and an enhancement is an improvement on existing component/functionality.**
+**\* A feature is a new component or piece of functionality and an enhancement is an improvement to an existing component or piece of functionality.**
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -17,7 +17,6 @@
 ## Specifications (please add any other notable libs/details)
 
 - Kaizen version:
-- Repo:
 
 ## Labelling
 
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,27 @@
+### If it's a discussion, please start one [here](https://github.com/cultureamp/kaizen-design-system/discussions).
+
+- [ ] Before opening this issue, I've searched open and closed issues for a fix.
+
+## Expected Behavior
+
+
+## Actual Behavior (please provide screenshots)
+
+
+## Steps to Reproduce the Problem
+
+1.
+2.
+3.
+
+## Specifications (please add any other notable libs/details)
+
+- Kaizen version:
+- Repo:
+
+## Labelling
+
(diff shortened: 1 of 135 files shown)


.github/issue_template.md
```diff
@@ -0,0 +1,27 @@
+### If it's a discussion, please start one [here](https://github.com/cultureamp/kaizen-design-system/discussions).
+
+- [ ] Before opening this issue, I've searched open and closed issues for a fix.
+
+## Expected Behavior
+
+
+## Actual Behavior (please provide screenshots)
```
jasononeil, line 8: Some of our issues are code only, should we clarify that? 

```suggestion
## Actual Behavior (please provide screenshots if it's a visual issue)
```

.github/issue_template.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Steps to Reproduce the Problem
+
+1.
+2.
+3.
+
+## Specifications (please add any other notable libs/details)
+
+- Kaizen version:
```
jasononeil, line 19: I like including this, should go a long way to identifying if an issue might have been subtly fixed by a recent release

.github/issue_template.md
```diff
@@ -0,0 +1,27 @@
...
+- Repo:
```
jasononeil, line 20: When I read this, my first though was "won't the repo always be kaizen-design-system?"

But I'm guessing this is about a bug where the problem might have been encountered? Is there a way to capture that in the name?

In our team we often talk about "consuming repos" but I am not sure that language would be obvious to people outside our team

.github/issue_template.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Labelling
+
+1. If related to a component, please add a label like so `component:mycomponent`.
+2. Then label with `bug`, `enhancement`/`feature`\* or `documentation`.
+
+**\* A feature is a new component/functionality and an enhancement is an improvement on existing component/functionality.**
```
jasononeil, line 27: Is functionality a plural? If I read "a new functionality" it sounds wrong. I'm tempted to reword this to avoid that :)

```suggestion
**\* A feature is a new component or piece of functionality and an enhancement is an improvement to an existing component or piece of functionality.**
```

lloydstubber: Good catch, moved it around and missed that.

.github/issue_template.md
lloydstubber: I was more going for "what repo you're working in?" for our context. Like `survey-summary-frontend` or maybe that's not useful to us?

.github/issue_template.md
lloydstubber: Yep I like that.

.github/issue_template.md
lloydstubber: Totally. Was also wondering if it's worth adding like a `Kaizen component version:` if it's like `"@kaizen/brand": "^1.1.1",` but maybe that's enough?

.github/issue_template.md
ActuallyACat: If it's an issue in production, the repo and/or location in the product is useful. If it's an issue someone has encountered in work that's in progress, a branch is useful. It would be hard to capture these things succinctly - I wouldn't be opposed to excluding this for brevity and let whoever opened the issue to provide whatever information they think is useful.

ActuallyACat (approved):
Looks good 👍🏻

PR #1889, opened 2021-08-24
feat: add new job to handle stale issues

lloydstubber:
# Objective
- To keep on top of our open issues and to nudge those that may have left them hanging.

# Motivation and Context
- As discussed, the more of these tasks that are automated, the easier it is to manage and spend time helping others contribute to the library.


lloydstubber:
The initial PR went in yesterday and the PR tagging worked a treat. There were a few warnings, for example the `only` field is incorrect, as it just runs on both pulls and issues.

I've added in the config for issues into the same job (annoying you can't double-barrel it as it's very repetitive) and upped the operations amount as it was 30 by default.

ActuallyACat (approved):

Issue #1759, opened 2021-07-14
Notification components should be moved out of component-library


dougmacknz:
Related to https://github.com/cultureamp/kaizen-design-system/issues/1623

We have moved away from the idea of having all of our components in a single package. Our notification components are one of a couple left that are still inside `@kaizen/component-library`. This caused some annoyances recently where we wanted to quickly update component-library in performance-ui, but we had to work through a bunch of breaking changes from the notification components

The 3 notification components are all inside the `packages/component-library/components/Notification` folder. They are:
* GlobalNotification
* ToastNotification
* InlineNotification

A new `@kaizen/notification` package can be created (please no `draft` 🙂)

PR #1901, opened 2021-08-30
feat: Move notification components out of component-library

dougmacknz:
**BREAKING CHANGE:** You will need update imports of the following components from `@kaizen/component-library` to `@kaizen/notification`:
- InlineNotification
- GlobalNotification
- ToastNotification
- addToastNotification/clearToastNotifications/removeToastNotification

Elm imports will need to change from `import Notification.Notification` to `import Kaizen.Notification`

# Objective
Moves our Notification components out of `component-library` and into a separate package.

# Motivation and Context
Closes https://github.com/cultureamp/kaizen-design-system/issues/1759


packages/notification/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@kaizen/notification",
+  "version": "0.0.0",
```
lloydstubber, line 3: Does this auto update from 0.0.0?

lloydstubber (approved):
LGTM just a little unsure of how the versioning works

packages/notification/package.json
dougmacknz: Yeah I'm hoping so, because I've got `BREAKING CHANGE` on this. Usually it just goes from 1.0.0 -> 1.1.0 without the `BREAKING CHANGE` part.

Not the end of the world if it doesn't work though.

PR #1910, opened 2021-08-31
docs: Add API guidance to component overview page

dougmacknz:
# Objective
Adds a bit of guidance around component API to the site.

# Motivation and Context
There's not a lot of guidance around this at the moment. This idea has been floating around for a while and Ally wrote up a bunch of stuff here: https://cultureamp.atlassian.net/wiki/spaces/~316784077/pages/1001396888/DRAFT+Kaizen+Consistent+Component+API

This is a compressed version of that, that we can build on in a separate page at a future time if we like

### DIFF
--- a/site/docs/components/overview.mdx
+++ b/site/docs/components/overview.mdx
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
 
 The status of these can be found on the documentation page for each component, or on the [Component health](/component-health) page, which shows the status of all components.
 
+### API
+- **Avoid language overloading**: Avoid prop names that are used in HTML, such as `type`.
+- **Lean on existing mental models**: Things like input `onChange` should not stray from the default behavior that React has defined.
+- **Encapsulation:** Components should only be concerned about their own form and function and not about how they appear in relation to other components.
+- **Favor composition**: This allows us to have a more succinct API that encourages code reuse.
+- **RTL support**: Avoid using left/right props or values, instead use start/end and support RTL automatically.
+- **Allowing ReactNode**: Favor allowing ReactNode values over strings where possible, for greater flexibility.
+
+#### Common props
+| Name | Description | Example |
+| --- | --- | --- |
+| `tag` | Specifies which HTML element the component should render as. | `<Heading tag="h2">` |
+| `mood` | Dictates the tone of the component, usually affecting colors. | `<BrandMoment mood="positive">` |
+| `primary` / `secondary` / `destructive` | Used to enable these common variants. | `<Button primary>` |
+| `reversed` | Enables a version of the component to be used on a dark background. | `<Button reversed>` |
+| `variant` | Only used if none of the above props makes sense.  | `<TextAreaField variant="prominent">` |
+
 ## Want to learn more about great components?
 
 - [Documenting Components by Nathan Curtis](https://medium.com/eightshapes-llc/documenting-components-9fe59b80c015).
--- a/site/docs/components/overview.mdx
+++ b/site/docs/components/overview.mdx
@@ -39,7 +39,7 @@ The status of these can be found on the documentation page for each component, o
 | Name | Description | Example |
 | --- | --- | --- |
 | `tag` | Specifies which HTML element the component should render as. | `<Heading tag="h2">` |
-| `mood` | Dictates the tone of the component, usually affecting colors. | `<BrandMoment mood="positive">` |
+| `mood` | [Dictates the tone of the component](https://dev.cultureamp.design/api-guidelines/guidelines/moods/), usually affecting colors. | `<BrandMoment mood="positive">` |
 | `primary` / `secondary` / `destructive` | Used to enable these common variants. | `<Button primary>` |
 | `reversed` | Enables a version of the component to be used on a dark background. | `<Button reversed>` |
 | `variant` | Only used if none of the above props makes sense.  | `<TextAreaField variant="prominent">` |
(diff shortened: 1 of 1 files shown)


dmisdm (approved):
This is nice!

ActuallyACat (commented):
Nice! Super happy that ya'll managed to condense that massive document down into a succinct format 👌🏻

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
 
 The status of these can be found on the documentation page for each component, or on the [Component health](/component-health) page, which shows the status of all components.
 
+### API
+- **Avoid language overloading**: Avoid prop names that are used in HTML, such as `type`.
+- **Lean on existing mental models**: Things like input `onChange` should not stray from the default behavior that React has defined.
+- **Encapsulation:** Components should only be concerned about their own form and function and not about how they appear in relation to other components.
+- **Favor composition**: This allows us to have a more succinct API that encourages code reuse.
+- **RTL support**: Avoid using left/right props or values, instead use start/end and support RTL automatically.
+- **Allowing ReactNode**: Favor allowing ReactNode values over strings where possible, for greater flexibility.
```
ActuallyACat, line 36: I know there's a few people that will be pleased about this change :D cc/ @OliBridgman ;)

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
 
 The status of these can be found on the documentation page for each component, or on the [Component health](/component-health) page, which shows the status of all components.
 
+### API
+- **Avoid language overloading**: Avoid prop names that are used in HTML, such as `type`.
+- **Lean on existing mental models**: Things like input `onChange` should not stray from the default behavior that React has defined.
+- **Encapsulation:** Components should only be concerned about their own form and function and not about how they appear in relation to other components.
```
ActuallyACat, line 33: I'm curious if this is enough information for non-design system folk. Should this include an example? The best one I can think of is no padding / margin around the component.

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
...
+- **Favor composition**: This allows us to have a more succinct API that encourages code reuse.
+- **RTL support**: Avoid using left/right props or values, instead use start/end and support RTL automatically.
+- **Allowing ReactNode**: Favor allowing ReactNode values over strings where possible, for greater flexibility.
+
+#### Common props
+| Name | Description | Example |
+| --- | --- | --- |
+| `tag` | Specifies which HTML element the component should render as. | `<Heading tag="h2">` |
+| `mood` | Dictates the tone of the component, usually affecting colors. | `<BrandMoment mood="positive">` |
```
ActuallyACat, line 42:
```suggestion
| `mood` | [Dictates the tone of the component](https://dev.cultureamp.design/api-guidelines/guidelines/moods/), usually affecting colors. | `<BrandMoment mood="positive">` |
```

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
 
 The status of these can be found on the documentation page for each component, or on the [Component health](/component-health) page, which shows the status of all components.
 
+### API
+- **Avoid language overloading**: Avoid prop names that are used in HTML, such as `type`.
```
ActuallyACat, line 31:
```suggestion
- **Avoid language overloading**: Choose names that are self-describing that continue to be clear in slightly different contexts. For example, `type`, has a specific meaning within the context of a form, but it could also be used to describe a different appearance of a component. A good rule of thumb is to avoid names that are used in HTML.
```

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
...
+- **Lean on existing mental models**: Things like input `onChange` should not stray from the default behavior that React has defined.
```
ActuallyACat, line 32:
```suggestion
- **Lean on existing mental models**: Lean into established naming conventions where it makes sense to do so. Things like input `onChange` should not stray from the default behavior that React has defined. 
```

site/docs/components/overview.mdx
```diff
@@ -27,6 +27,23 @@ We strive to achieve the following quality attributes on all of our components:
...
+- **Encapsulation:** Components should only be concerned about their own form and function and not about how they appear in relation to other components.
+- **Favor composition**: This allows us to have a more succinct API that encourages code reuse.
+- **RTL support**: Avoid using left/right props or values, instead use start/end and support RTL automatically.
```
ActuallyACat, line 35:
```suggestion
- **RTL support**: Avoid using left/right props or values, instead use start/end and support RTL automatically. See [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) for more information.
```

ActuallyACat: Some folks might not have come across logical properties just yet :)

dougmacknz: 🤔 this isn't necessarily tied to that. Most people are familiar with start/end over left/right with flexbox, and we don't actually have to use logical properties under the hood to support start/end either / we can't because of legacy Edge.

ActuallyACat: Ah! My mistake - ignore this.

ActuallyACat (approved):
LGTM - I appreciate you carrying this task 🙏🏻

Issue #1930, opened 2021-09-03
Tiles prescribe heading hierarchy that might not be correct


dougmacknz:
In `GenericTile` there is a `<Heading variant="heading-3">` which results in an `<h3>` being rendered. This is more than likely going to cause accessibility issues with the page heading hierarchy, as it did with the issue detailed below.

## Audit info

Type:WCAG 2
Principle: Perceivable
Guideline:1.3 Adaptable
Success Criteria: 1.3.1 Info and Relationships (A)
Level: A
Platform: Web
Severity:Low

### What's the issue?
The heading markup does not accurately represent the hierarchical relationships of the content.
Tasks is marked up as a H2 heading, while the card headings are marked up as H3 headings. This implies that the cards are a subsection of tasks.

### Why it matters
Screen reader users rely on headings to form a mental model of the content and understand how each piece of content relates to the whole.

### Our recommendation
Headings should be used to represent how the content is organised.

Use appropriate heading levels to create a hierarchical structure for the page.

Ensure that the relationship between topics and subtopics is reflected in the heading markup.

Note: It is best practice for heading levels to run in sequence.

PR #1936, opened 2021-09-06
feat: Create titleTag prop on Tile components

dougmacknz:
Closes https://github.com/cultureamp/kaizen-design-system/issues/1930

# Objective
Creates a new `titleTag` prop on our two Tile components (`InformationTile` and `MultiActionTile`) that allows consumers to change the tag that is rendered with the heading (currently hard-coded to `h3`).

Also changes `title` to accept a `ReactNode` rather than just a `string`.

# Motivation and Context
See https://github.com/cultureamp/kaizen-design-system/issues/1930

# Reasoning
@OliBridgman and I paired on this and explored a other options, before settling on this:
* Making the `title` prop optional and allow consumers to pass in their own heading via `children`
  * This didn't work so well because the title ended up vertically aligned.
* Allowing title to be `ReactNode`, while removing the `<Heading variant="heading-3">` wrapping when the title is *not a string*.
  * We decided this was too magical - how is a consumer meant to know that this switch happens based on whether they pass in a string or JSX?
* A render prop for title
  * We don't really like this pattern.

We also ruled out any options that would result in a breaking change, such as changing the `title` prop to be a `HeadingProps` object.

Ultimately we decided this new prop was the least evil of everything we tried, and it's not such a bad thing that we keep the heading styling locked to `heading-3`. It's not perfect, in an ideal world these components would be compositional, but there's a good chance that we'll be rewriting our Tile components at some point with the recent work the designers have done, so we didn't want to invest too much in our current ones.

### DIFF
--- a/draft-packages/tile/KaizenDraft/Tile/InformationTile.tsx
+++ b/draft-packages/tile/KaizenDraft/Tile/InformationTile.tsx
@@ -10,6 +10,7 @@ type InformationTile = React.FunctionComponent<InformationTileProps>
 
 const InformationTile: InformationTile = ({
   title,
+  titleTag,
   metadata,
   children,
   information,
@@ -17,6 +18,7 @@ const InformationTile: InformationTile = ({
 }) => (
   <GenericTile
     title={title}
+    titleTag={titleTag}
     metadata={metadata}
     information={information}
     footer={footer}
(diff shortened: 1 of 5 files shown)


lloydstubber (approved):
LGTM!

Issue #1454, opened 2021-04-19
Well component - missing variants


ellisjas:
## Missing stories / implementation of variants

* Informative borderless
* Positive dashed and borderless
* Negative dashed and borderless
* Cautionary dashed and borderless
* Assertive solid, dashed, and borderless
* Prominent solid, dashed, and borderless

## Figma design

https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=1929%3A14167

ActuallyACat:
This could be split into a few different tickets - I would suggest whoever picks up this issue create a ticket/s for any outstanding missing variants if they can't fit everything into a pairing session.

lloydstubber:
@ellisjas this is now fixed in this PR https://github.com/cultureamp/kaizen-design-system/pull/1939

PR #1939, opened 2021-09-06
feat: add new variants of well

lloydstubber:
# Objective
- Add examples of all the new variants of `<Well />`

# Motivation and Context
Closes this: https://github.com/cultureamp/kaizen-design-system/issues/1454

Too many screenshots so best to check the storybook
Screen Shot 2021-09-06 at 5 02 46 pm

# Checklist

### DIFF
--- a/draft-packages/well/styles.scss
+++ b/draft-packages/well/styles.scss
@@ -15,37 +15,37 @@
 
 .positive {
   background: $color-green-100;
-  border-color: var(#{$color-green-300-id}, $color-green-300);
+  border-color: $color-green-300;
   color: $color-purple-800;
 }
 
 .negative {
   background: $color-red-100;
-  border-color: var(#{$color-red-300-id}, $color-red-300);
+  border-color: $color-red-300;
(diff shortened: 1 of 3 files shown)


draft-packages/well/styles.scss
```diff
@@ -15,25 +15,37 @@
 
 .positive {
   background: $color-green-100;
-  border-color: var(#{$color-green-300-id}, $color-green-200);
+  border-color: var(#{$color-green-300-id}, $color-green-300);
```
dougmacknz, line 18: 🤔 do you know why these have the verbose id thing and not just `border-color: $color-green-300`?

dougmacknz: Oh, looks like this was only in place so that it could be `green-200` if the CSS var didn't exist (to support Zen). Now that these are the same, this can be simplified.

lloydstubber: Ahh that's why. Will update now.

dougmacknz (approved):

PR #1944, opened 2021-09-09
fix: Use min-height:100vh on small viewport as well

dougmacknz:
# Objective
See the first commit for the main objective here - just makes the `min-height: 100vh` that was only on large and up now on smaller view ports as well.

The extra stuff I did here was just revert any use of the logical properties `inline-size` and `block-size` back to `width` and `height` because of our pseudo support for legacy Edge, and given we're probably not going to be supporting Mongolian anytime soon :)

# Motivation and Context
@lloydstubber ran into this while implementing a `BrandMoment` (note the blank white space at the bottom). This happens when your browser height is beyond the height of the content.
image

I don't think there was any good reason for limiting the `min-height:100vh` to large and up - I probably did it because I originally had it as `height` and not `min-height`

### DIFF
--- a/packages/brand-moment/src/BrandMoment.scss
+++ b/packages/brand-moment/src/BrandMoment.scss
@@ -9,8 +9,8 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
 // --------------------------------
 
 .body {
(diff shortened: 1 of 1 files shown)


lloydstubber (approved):
LGTM!

PR #1945, opened 2021-09-09
fix: Disable event listeners on legacy Edge

dougmacknz:
# Objective
Stops legacy Edge blowing up when `useMediaQueries` is used.

Removes some functionality for legacy Edge as a result: it will no longer respond to changes after load. It's a legacy browser though so I don't think we care that much, and I wouldn't want to invest anymore in this. Having the values from page load gets you most of the way.

# Motivation and Context
We've noticed errors coming through: `"Object doesn't support property or method 'addEventListener'"`
https://cultureamp.slack.com/archives/C016N8Y1E10/p1630975772292000

### DIFF
--- a/packages/responsive/src/useMediaQueries.tsx
+++ b/packages/responsive/src/useMediaQueries.tsx
@@ -28,6 +28,10 @@ export const useMediaQueries = (
 } => {
   const theme = useTheme()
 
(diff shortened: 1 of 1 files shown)


lloydstubber (approved):
👌

PR #1951, opened 2021-09-13
chore: add snapshot test to brandmoment

lloydstubber:
# Objective
- After adding snapshot testing to the ErrorPages in BFR we can reuse that work to test the BrandMoment component

# Motivation and Context
- Prior to this, there was no testing. So hopefully this simple smoke test will catch any future bugs.


packages/brand-moment/src/BrandMoment.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onchange: null,
+  addListener: jest.fn(),
+  addEventListener: jest.fn(),
+  removeListener: jest.fn(),
+  removeEventListener: jest.fn(),
+}))
+
+// Stub video elements for JS-Dom to prevent console errors
+window.HTMLMediaElement.prototype.load = () => jest.fn()
+// @ts-ignore-next-line
```
lloydstubber, line 20: I know this looks bad, but for a stub I thought it'd be okay? Without it the type is a bit strange:
```
Type '() => jest.Mock<any, any>' is not assignable to type '() => Promise<void>'.
  Type 'Mock<any, any>' is missing the following properties from type 'Promise<void>': then, catch, finally, [Symbol.toStringTag]
```

jest.config.js
```diff
@@ -4,10 +4,11 @@ module.exports = {
   testMatch: ["**/*.spec.ts?(x)"],
   setupFilesAfterEnv: ["jest-canvas-mock"],
   moduleNameMapper: {
-    "\\.(jpe?g|png)$": "jest-static-stubs/$1",
+    "\\.(jpe?g|png|webm|mp4)$": "jest-static-stubs/$1",
     "\\.s?css$": "identity-obj-proxy",
     "\\.svg$": require.resolve("@kaizen/component-library/mocks/svgMock"),
   },
+  transformIgnorePatterns: ["[/\\\\]node_modules[/\\\\].+\\.(js|jsx|mjs)$"],
```
lloydstubber, line 11: @dmisdm does this look correct to you? It's just to stop it complaining about the unified nav package.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["es2018", "dom"],
+    "lib": ["es2018", "dom", "esnext"],
```
lloydstubber, line 3: @dmisdm added this at the top level. Not sure if it's worth having a package-level jest config? Let me know

lloydstubber:
Do you think it's worth adding more testing to this @ActuallyACat / @dougmacknz ? Just ran out of time due to config but could write more tests around props etc.

packages/brand-moment/package.json
```diff
@@ -41,6 +41,8 @@
     "classnames": "^2.3.1"
   },
   "devDependencies": {
+    "@cultureamp/frontend-core": "^2.1.1",
+    "@cultureamp/unified-navigation-ui": "^2.0.1",
```
ActuallyACat, line 45: Really glad I set up the private packages now :D

packages/brand-moment/src/BrandMoment.spec.tsx
ActuallyACat: Typescript is complaining because `play` returns a Promise, not a function. In this scenario, we can trust that the illustration package is properly testing that `play` is call on mount, so you don't need to mock a reject/resolve. Something like this would work: 
`window.HTMLMediaElement.prototype.play = () => new Promise(jest.fn())`

tsconfig.json
ActuallyACat: Let's have a package level config for this - it might introduce issues for downstream users of this package if we switch everything over to `esnext`.

ActuallyACat (changes requested):

packages/brand-moment/package.json
lloydstubber: Unsure what you mean by this? 😅

tsconfig.json
lloydstubber: So when I add it to the package level like so:
```
{
  "compilerOptions": {
    "allowJs": false,
    "declaration": true,
    "noEmit": false,
    "sourceMap": true,
    "lib": ["es2018", "dom", "esnext"]
  },
  "exclude": ["docs"]
}
```

I get this error:
```
➜  kaizen-design-system git:(ls/testBrandMoment) ✗ yarn test BrandMoment
yarn run v1.22.11
$ jest BrandMoment
 FAIL  packages/brand-moment/src/BrandMoment.spec.tsx
  ● Test suite failed to run

    node_modules/@cultureamp/unified-navigation-ui/lib/helpers/uiHelpers.ts:126:13 - error TS2339: Property 'flatMap' does not exist on type 'MenuItem[]'.

    126   menuItems.flatMap((item, idx) => {
                                 ~~~~~~~
```

lloydstubber: @ActuallyACat bump on this

lloydstubber: @ActuallyACat just got a stale notification so opening this back up. Do you have any way to solve the config issue above?

ActuallyACat: Sorry - let me look into this. Kaizen supports many different consumers with different needs, so we need to exercise some caution as using esnext features might cause issues for our users.

ActuallyACat: Gutfeel - is it worth adding `@cultureamp/unified-navigation-ui` to this repo?

lloydstubber: Unsure! What would that solve?

ActuallyACat: The issue is `unified-navigation-ui` is using `flatMap` - a function that is not supported in `es2018`. If we remove it the issue is avoided - which is a bandaid solution. I'm pretty hesitant to use `esnext` because there's no guarantee our users will correctly polyfill all `esnext` features (that, and esnext it's a moving target). 

If we bump to `es2019` that will successfully compile. I'm going to push a commit and chase up what's the minimum target environment for our users.

ActuallyACat: I removed support for private packages recently, as no one was actively using them in this repo. Now the build is broken because of that change. ಠ_ಠ

ActuallyACat (approved):

tsconfig.json
ActuallyACat: Removed unified-navigation. The component doesn't specify what shape of React node it expects.

PR #1960, opened 2021-09-17
feat: Remove all usages of var() in scss to provide Zen fallbacks

dougmacknz:
# Objective
Removes all usages of `var()` in `scss` that were being used as a way of doing `var([heart_value], [zen_value])`.

# Motivation and Context
This was a method we used to switch tokens based on Zen or Heart, because we knew that no CSS vars existed for Zen, and it would use the fallback value.

### DIFF
--- a/draft-packages/avatar/KaizenDraft/Avatar/decision-tokens.scss
+++ b/draft-packages/avatar/KaizenDraft/Avatar/decision-tokens.scss
@@ -1,10 +1,7 @@
 @import "~@kaizen/design-tokens/sass/color";
-@mixin avatar-theme-switching-tokens() {
-  --avatar--color-icon-color: rgba($color-purple-800-rgb, 0.7);
-}
 $dt-color-background-color: $color-gray-300;
 $dt-color-background-color-initials: $color-gray-300;
 $dt-color-background-color-personal: $color-orange-100;
 $dt-color-text-color: $color-purple-800;
 $dt-color-border-color: $color-white;
-$dt-color-icon-color: var(--avatar--color-icon-color);
+$dt-color-icon-color: rgba($color-purple-800-rgb, 0.7);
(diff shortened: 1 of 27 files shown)


draft-packages/avatar/KaizenDraft/Avatar/decision-tokens.scss
```diff
@@ -1,16 +1,10 @@
...
+$dt-color-background-color: $color-gray-300;
+$dt-color-background-color-initials: $color-gray-300;
+$dt-color-background-color-personal: $color-orange-100;
 $dt-color-text-color: $color-purple-800;
 $dt-color-border-color: $color-white;
-$dt-color-icon-color: var(
-  --avatar--color-icon-color,
-  rgba($color-purple-800-rgb, 0.3)
-);
+$dt-color-icon-color: var(--avatar--color-icon-color);
```
dmisdm, line 10: I'm a little bit iffy about the mixin + CSS var `--avatar--color-icon-color` here: does this need to be a CSS var or can we change `$dt-color-icon-color: var(--avatar--color-icon-color);` to `$dt-color-icon-color: rgba($color-purple-800-rgb, 0.7);` ?

This is the only place the mixin is being used: 
image

dmisdm: Its hard to gauge the intended behaviour, because it's totally reasonable to think that the fallback `rgba($color-purple-800-rgb, 0.3)` was intentionally used whenever the mixin `avatar-theme-switching-tokens` isn't.

dougmacknz: Got rid of the mixin :)

dmisdm (approved):
Stunning!

Issue #1918, opened 2021-09-02
Adjacent links lead to same destination


lloydstubber:
TLDR; IMO I would remove the link around the span so you just click the button to navigate back rather than combine them. Check with DST to double check this though.

**Type**: WCAG 2
**Principle**: Perceivable
**Guideline**:1.1 Text Alternatives
**Success Criteria**: 1.1.1 Non-text Content (A)
**Level**: Not applicable
**Platform**: Web
**Severity**: Low
**URL**: https://intopia.cultureamp.com/performance/1-1/6ce44465-ac60-426c-82f8-94d5d8c3ca92

### What's the issue?
Adjacent links lead to the same destination.

```
<a href="/performance/1-1" class="TitleBlockZen-TitleBlockZen__breadcrumb--2KFRt TitleBlockZen-TitleBlockZen__buttonReset--nWmBr" data-automation-id="TitleBlock__Breadcrumb" aria-label="Back to previous page">
  ... 
</a>
<a href="/performance/1-1" class="TitleBlockZen-TitleBlockZen__breadcrumbTextLink--V1NHM TitleBlockZen-TitleBlockZen__buttonReset--nWmBr" data-automation-id="TitleBlock__BreadcrumbText" aria-label="Back to previous page" tabindex="-1">
  <span class="TitleBlockZen-TitleBlockZen__breadcrumbText--3hErU">Back to 1-on-1s</span>
</a>
```

Note: although one of the links has a negative tabindex (so it won’t bother sighted keyboard users) it will still be announced by screen readers who are using arrow keys to navigate.

<img width="510" alt="Screen Shot 2021-09-02 at 10 16 50 am" src="https://user-images.githubusercontent.com/18339250/131761814-54f96e2d-4382-45d6-ad14-7babc39ede83.png">

### Why it matters
When adjacent links lead to the same destination it causes users to navigate more slowly, especially those using assistive technology, such as switch devices or screen readers. It increases the number of elements to navigate and screen reader verbosity.

### Our recommendation
Remove redundant links that led to the same destination or combine them into one.
Remove the aria-label to ensure the label in code will match the label on screen, “Back to 1-on-1s”.

```
<a href="/performance/1-1" class="TitleBlockZen-TitleBlockZen__breadcrumbTextLink--V1NHM TitleBlockZen-TitleBlockZen__buttonReset--nWmBr" data-automation-id="TitleBlock__BreadcrumbText">
[...]

PR #1966, opened 2021-09-22
fix: update title block back link and text to be one

brandonwinch:
## Objective
Update sibling links that go to the same place to be one link. This resolves a small accessibility problem as uncovered in the latest audit. There was a bit of CSS we could update while in here too!

## Motivation and Context
Fixes https://github.com/cultureamp/kaizen-design-system/issues/1918


dougmacknz (approved):
Awesome, thanks for sticking it out on this tricky first pairing task @brandonwinch!

PR #1975, opened 2021-09-28
feat: add SearchField and InputSearch to Kaizen

lloydstubber:
Closes https://github.com/cultureamp/kaizen-design-system/issues/1298

# Objective
Create the new `SearchField` component for the `FilterBar` work

# Work
- Created new primitive called `InputSearch`
- Created top-level component called `SearchField`
- Used styles from `Input` + got rid (hopefully all) of what I don't need
- Created stories for all use cases - @katsambarcus check the Storybook preview, there would be too many screenshots otherwise.

# Questions
- Do we need more tests?
- When it's loading, should the input be disabled?

# Checklist

# Additional Considerations


draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+    reversed = false,
+    loading,
+    startIconAdornment,
+    endIconAdornment,
+    primary,
+    secondary,
+    ...genericInputProps
+  } = props
+
+  const [value, setValue] = useState("")
```
lisatassone, line 40: I wouldn't have thought this would be a stateful component 🤔 

I can see that we can supply an `onChange` and `value` as props that will overwrite the ones specified such as `setValue`, but what about the onClear?

lloydstubber: I think the idea is we have both options but @dougmacknz might weight in on this. 

Good catch I haven't implemented the `onClear` overwrite. Will do now 👍

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+            [styles.disabled]: disabled,
+            [styles.secondary]: secondary,
+          },
+          classNameAndIHaveSpokenToDST
+        )}
+        {...genericInputProps}
+      />
+      {/* Inputs aren't able to have pseudo elements like ::after on them,
+          so we have to create an element ourselves for the focus ring */}
+      <div className={styles.focusRing} />
```
lisatassone, line 91: FYI: You can use multiple `box-shadow`'s to create an offset ring as an alternative. The pro is that we aren't having to do this, the con is that the offset color (the 'gap') needs to match the background. Though we have light and dark versions of where these inputs are used anyway so might not be such a con.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
lisatassone: Guess I'm also trying to work out when I wouldn't want to not know what the value is? We are just setting the internal state here and not telling the consumer anything about it 🤷‍♀️ What purpose is this serving?

dougmacknz: We offer both options for our form inputs to allow the option of doing native browser input (though it sounds like that's invalid here because of the `onClear` need) and using the value via form submission.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+export interface InputSearchProps
+  extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "className"> {
+  classNameAndIHaveSpokenToDST?: string
+  inputRef?: React.RefObject<HTMLInputElement>
+  reversed?: boolean
+  loading?: boolean
+  primary?: boolean
+  secondary?: boolean
+  startIconAdornment?: React.ReactNode
+  endIconAdornment?: React.ReactNode
```
dougmacknz, line 19: Don't think you need these `iconAdornment` props, they're not actually being used. On `Input` they are used to allow the `TextField` (or a direct usage) to pass an icon through to it, but we're not allowing that flexibility on this component.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
lisatassone: Interesting. I would have thought that you could achieve that behaviour by not supplying a `value` prop since inputs are inherently uncontrolled.

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  variant,
+  value,
+  defaultValue,
+  labelText,
+  ...genericInputProps
+}) => (
+  <FieldGroup
+    id={`${id}-field-group`}
+    inline={false}
+    automationId={`${id}-field-group`}
```
dougmacknz, line 34: Let's just get rid of all of these `automationId`'s, they're not used for your tests and we're not expecting consumers to use them anymore either.

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  >
+    <Label
+      id={`${id}-field-label`}
+      automationId={`${id}-field-label`}
+      htmlFor={`${id}-field-textarea`}
```
dougmacknz, line 39: This needs to match the `id` sent to `InputSearch`. It'd be worth chucking it into a variable to make sure they stay the same.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
dougmacknz: This needs a bit of work, `value` is a prop but it's being ignored and defaulted to an empty string here, and the clearing functionality doesn't work for uncontrolled. Let's pair.

dougmacknz: > I would have thought that you could achieve that behaviour by not supplying a value prop since inputs are inherently uncontrolled.

You're right, except this component is a bit more tricky because it has the 'clear' functionality, and we're creating our own button and hiding the native one (which only shows in some browsers). Anyway, we have ended up abandoning the idea of uncontrolled on this component now because it complicates things a fair bit, and as you mentioned - it's probably unlikely that someone will need it.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,97 @@
...
+  > {
+  classNameAndIHaveSpokenToDST?: string
+  reversed?: boolean
+  loading?: boolean
+  secondary?: boolean
+  onClear?: () => void
+}
+
+const InputSearch: React.FunctionComponent<InputSearchProps> = (
+  props: InputSearchProps
```
ActuallyACat, line 23: /meganit, this redeclaration of `props: InputSearchProps` is redundant as you've defined the prop types already in the line above.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,97 @@
...
+        className={classnames(
+          styles.input,
+          styles.search,
+          {
+            [styles.default]: !reversed,
+            [styles.reversed]: reversed,
+            [styles.disabled]: disabled,
+            [styles.secondary]: secondary,
+          },
+          classNameAndIHaveSpokenToDST
```
ActuallyACat, line 78: TIL when `classNameAndIHaveSpokenToDST` is applied to primitives form components, it's applied to the `input` not the wrapper (confirmed by looking at the other primitives).

ActuallyACat (commented):
Looking good to me! I'll leave the final approval to @dougmacknz as he's more across this work.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
dougmacknz: Wait, really? Setting it on the return type makes this obsolete?

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
dougmacknz: And they're named `className` rather than `classNameAndIHaveSpokenToDST`.

What would be interesting is if we allow `classNameAndIHaveSpokenToDST` on the container (`SearchField/TextField`) level. I would imagine it should also go down to the input level to be consistent with the other native props.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -0,0 +1,97 @@
...
+          [styles.withDisabled]: disabled,
+          [styles.withSecondary]: secondary,
+          [styles.withLoading]: loading,
+        }
+      )}
+    >
+      <div className={styles.startIconAdornment}>
+        {loading ? (
+          <div className={styles.loadingIcon}>
+            <LoadingSpinner accessibilityLabel="Loading comments" size="sm" />
```
dougmacknz, line 56: 🤔 What is `comments` in this context? I would probably go for an empty string here because having important information in an icon that precedes the input in the DOM isn't going to be helpful for assistive tech. Not sure exactly how we're going to tackle that issue just yet.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
dougmacknz: This has already been implemented in `TextField/Input`, so we'll just keep this for now, but that is good to know!

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { FieldGroup, Label } from ".."
+export interface SearchFieldProps
+  extends Omit<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    "className" | "defaultValue"
+  > {
+  reversed?: boolean
+  loading?: boolean
+  secondary?: boolean
+  labelText?: React.ReactNode
```
dougmacknz, line 13: Sorry just picked this up - this is usually required and I think we still want it required here. In the case that's it's `secondary` variant we'll want to put it as an `aria-label` on the input.

dougmacknz: Using it for an `aria-label` will also require this to be a string rather than a `ReactNode` as well.

draft-packages/form/KaizenDraft/Form/SearchField/SearchField.tsx
```diff
@@ -0,0 +1,54 @@
...
+  value,
+  labelText,
+  onClear,
+  ...genericInputProps
+}) => {
+  const searchFieldId = `${id}-search-field-input`
+
+  return (
+    <FieldGroup inline={false}>
+      <Label
```
dougmacknz, line 33: This is rendering a blank `label` element for secondary variants. Probs just wanna put this behind a `!secondary` conditional.

dougmacknz (dismissed):

dougmacknz (approved):

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
lisatassone: `React.FunctionComponent` accepts a generic which is used to type props, so Ally is correct. It's not needed again.

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
lisatassone: No worries

PR #1938, opened 2021-09-06
fix: Add static label to TitleBlockZen meatball IconButton

dan-lennox:
# Objective

Issue: https://github.com/cultureamp/kaizen-design-system/issues/1916

Add static label to TitleBlockZen meatball IconButton

_Have not considered localisation as part of this change. Identified that this might be best handled in a broader piece of work_

# Motivation and Context

Improves the accessibility of the TitleBlockZen component.

# Additional Considerations
- Does there need to be right-to-left (RTL) options for localization and internationalization?
- Has the test suite been updated?
- Have you done cross-browser testing for our [supported browsers](https://academy.cultureamp.com/hc/en-us/articles/204539569-Supported-browsers-for-Participants)?
- Have you updated any relevant documentation or left useful comments in the code?
- Have you reviewed Culture Amp's Web Accessibility guide [Current Compliance, Going Forward, Statement and Answers for Customers](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/428572998/Web+Accessibility)?
- If this contains substantial visual changes, especially far reaching ones, have you unlocked the Chromatic step in the pipeline?
- Have Storybook stories been updated?

Issue #1916, opened 2021-09-01
IconButton usage in TitleBlockZen does not have an accessible label


lloydstubber:
TLDR; In `TitleBlockZen`'s `SecondaryActions` sub-component, we’re passing an empty string to an `IconButton` label prop, which causes this issue. Create a new optional prop on `TitleBlockZen` for this string, with a default value (this will allow consumers to pass a localised string).

Related: https://github.com/cultureamp/kaizen-design-system/issues/1530 to make TS yell when you try to send in an empty string to avoid issue being created in the future.

## Audit info

**Type**: WCAG 2
**Principle**: Perceivable, Robust
**Guideline**:1.1 Text Alternatives, 4.1 Compatible
**Success Criteria**: 1.1.1 Non-text Content (A), 4.1.2 Name, Role, Value (A)
**Level**: A
**Platform**: Web
**Severity**: High
**URL**: https://intopia.cultureamp.com/performance/1-1/6ce44465-ac60-426c-82f8-94d5d8c3ca92

<img width="767" alt="Screen Shot 2021-09-01 at 4 39 26 pm" src="https://user-images.githubusercontent.com/18339250/131623923-dbb55ab1-51c7-4ec9-b673-32639f87e043.png">

### What's the issue?
The Other actions button does not have an accessible label.

<img width="604" alt="Screen Shot 2021-09-01 at 4 41 02 pm" src="https://user-images.githubusercontent.com/18339250/131624127-8a0ea31d-8f51-48b2-97fa-0305149e2ee9.png">

### Why it matters
If the button does not have a meaningful name, a screen reader user will not know what the button is for.

### Our recommendation
Give the button an accessible label. The button has an aria-label attribute, however it is empty. Add a descriptive label to this attribute, e.g. "Other actions".

PR #1984, opened 2021-10-06
fix: Add label to TiltleBlockZen overflow meatball IconButton

dougmacknz:
Issue: https://github.com/cultureamp/kaizen-design-system/issues/1916

Same thing as https://github.com/cultureamp/kaizen-design-system/pull/1938, but for the MainActions -> overflow button (basically the same button but only shows on smaller viewports). We didn't realise that there were two cases that needed this.


lloydstubber (approved):
that's a spicy meatball

PR #2127, opened 2021-10-27
ci: Remove precommit/prepush hooks

ActuallyACat:
# Objective
Remove precommit/prepush hooks

# Motivation and Context
Now the pipeline is running quicker with Github Actions, there is less of a need for pre-push hooks.


dmisdm (approved):
🚢🛳🚀

PR #2200, opened 2021-11-03
feat: Added new VisuallyHidden component to target only screen readers

christian314159:
# Objective
The `VisuallyHidden` component allows to wrap elements (usually text) so they are visually hidden and only picked up by screen readers, thus giving users of those additional information.

# Motivation and Context
* gives users of Kaizen an easy way of adding screen reader only information
* component is using the CSS class found here: https://css-tricks.com/inclusively-hidden/


packages/a11y/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@kaizen/a11y",
+  "version": "1.0.0",
+  "description": "A component to hide elements visually, but still make the accessible to screen readers",
```
dougmacknz, line 4: This doesn't matter a huge amount but this package is likely to hold other accessibility helpers, whereas this description is just for visually hidden.

packages/a11y/docs/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react"
+
+import { withDesign } from "storybook-addon-designs"
+import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
+
+import { VisuallyHidden } from ".."
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.a11y}/Visually Hidden`,
```
dougmacknz, line 9: There's a 'helpers' section where this would make more sense (rather than inside components)
image

packages/a11y/docs/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  component: VisuallyHidden,
+  parameters: {
+    docs: {
+      description: {
+        component: 'import { VisuallyHidden } from "@kaizen/a11y";',
+      },
+    },
+  },
+  decorators: [withDesign],
```
dougmacknz, line 18: /nit this is just for embedding figma designs, which this component will never have, so you can remove it

packages/a11y/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "!draft-templates",
+    "!stories",
+    "!docs",
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/component-library": "^11.5.0"
```
dougmacknz, line 35: I'm not seeing any usages of `component-library`, so looks like this doesn't need to be here. Less cross dependencies the better!

dougmacknz (commented):
Nice! Pulled up some very minor things. Mainly just want to get rid of that `component-library` dependency.

packages/a11y/docs/VisuallyHidden.stories.tsx
ActuallyACat: 🤔 IMO helpers are packages that don't necessarily have UI. I'm only questioning this because I also have a Pr that will belong in the a11y package but (I don't think) is a helper: https://github.com/cultureamp/kaizen-design-system/pull/2201

ActuallyACat:
One small thing 🤏🏻 We have an open issue for VisuallyHidden. Could you add "Closes https://github.com/cultureamp/kaizen-design-system/pull/2200" somewhere in your PR body so when this PR is merged the issue will be closed?

packages/a11y/docs/VisuallyHidden.stories.tsx
dougmacknz: Makes me wonder whether they do belong in the same package then, or if we want a general `utils` package instead.

packages/a11y/docs/VisuallyHidden.stories.tsx
christian314159: Thanks, I copy/pasted this setup from an existing one, not knowing what this does at all.

packages/a11y/docs/VisuallyHidden.stories.tsx
christian314159: Don't have strong opinions on this, but seeing there now will be 2 components related to A11y (`SkipLink` & `VisuallyHidden`) it feels like an OK grouping to me.

dougmacknz (approved):
Looking good to me, just +1 for @ActuallyACat's suggestion of linking this PR with the issue.

PR #2202, opened 2021-11-04
feat: added mood prop to InformationTile and MultiActionTile

arikark:
# Objective
- This change allows a 'mood' prop to be passed to the InformationTile and MultiActionTile 

# Motivation and Context
- See [here](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%9A%A7-UI-Kit%3A-Beta-Components?node-id=22354%3A71173)

image
image

# Checklist


katsambarcus:
Looking great, @arikark These moods work well and I'd be happy to ship it.

FWIW, a couple of things that @allybeaton and @dougmacknz should consider:

- I think this is a feature we would want to introduce to the Container component, so that any type of Content can benefit from switching the background color to reflect a mood. This would align with [our thinking on how we can offer validation on any block of content](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/?node-id=22354%3A71173), not just the empty state component.
- Re 'Prominent' mood name. Is this a good time to change the mood name to something else (e.g. emphasise, highlight, accentuate, etc, @allybeaton ? Or can we retrospectively do it, without too much trouble?
- @dougmacknz wondering how we may be able to combat the 'status' tag meeting minimum contrast ratios? Could we include a multiply effect to the background color of the tag?

ellisjas (approved):
🚢

Issue #1934, opened 2021-09-06
Tabs component does not communicate role or state


dougmacknz:
Our Tabs component could do a lot better at rendering more semantic markup. We're currently rendering `<a>` tags regardless of whether `hrefs` are actually provided, and there's some additional best practices that we can add (see links below).

## Audit info

Type:WCAG 2
Principle: Robust
Guideline:4.1 Compatible
Success Criteria: 4.1.2 Name, Role, Value (A)
Level: A
Platform: Web
Severity:Critical

### What's the issue?
The tabs component is marked up using a <a> element. It does not communicate its role (tab) or state (selected/unselected).

### Why it matters
If a tab does not have the tab role assigned to it then a screen reader user will not know how to interact with it.
A screen reader user will not know the current state of the tab

### Our recommendation

See the [Intopia ARIA Tabs codepen](https://codepen.io/intopia/pen/eYZmvrN) for an example.

See the [ARIA Practices Tabs example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/tabs.html) for information on the ARIA tab component including markup and keyboard functionality.

ActuallyACat:
How important is the order of the content? is it enough to markup the tabs with `aria-control=blah` with the related content as `id=blah`.

Issue #1922, opened 2021-09-02
Tab active state isn't seen in Windows high contrast mode


dougmacknz:
TLDR: the active state can't be seen in windows high contrast mode, we need to use a border, or use a media query to adjust this.

## Audit info

Type:Expert observation
Principle: Not applicable
Guideline:Not applicable
Success Criteria: Not applicable
Level: Not applicable
Platform: Web
Severity:High

### What's the issue?
The indicator to show which tab has been selected can't be determined when viewed with Windows high contrast mode enabled, because it relies on the background-color property which is removed in this mode.

### Our recommendation
Make sure that all informative styles are visible in Windows high contrast mode.
Use an icon or border style instead of relying on removed CSS properties such as box-shadow and background colours.

## Screenshots

Not high contrast mode:
![image](https://user-images.githubusercontent.com/1811583/131786154-1d37c956-d3a5-4097-8e63-684a67ce55b7.png)

High contrast mode:
![summary-tabs-hc(2)](https://user-images.githubusercontent.com/1811583/131786089-f97ed8b1-2136-42d8-b82d-f676b807cb55.PNG)

PR #2203, opened 2021-11-04
feat: Create new Tabs component

dougmacknz:
# Objective
Creates a new `tabs` package with components shimmed around [Reach UI's Tabs](https://reach.tech/tabs).

# Motivation and Context
The old tabs component had a plethora of issues, it was better to just rewrite from scratch.

This will close the following issues + adds keyboard accessibility.
Closes https://github.com/cultureamp/kaizen-design-system/issues/1934
Closes https://github.com/cultureamp/kaizen-design-system/issues/1922

Windows high contrast mode (not perfect but gets the job done):
image
image


packages/tabs/docs/Tabs.stories.tsx
```diff
@@ -0,0 +1,174 @@
+import React, { useState } from "react"
```
ActuallyACat, line 1: The alignment on the badge doesn't look right
Screen Shot 2021-11-05 at 5 09 41 pm

packages/tabs/src/styles.scss
```diff
@@ -0,0 +1,104 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/typography";
+
+.tabList {
+  border-bottom: 1px solid rgba($color-gray-600-rgb, 0.1);
+  padding: 0 $spacing-md;
```
ActuallyACat, line 8: By default, our components shouldn't include margin or padding around the wrapper of the component. Consumer can add spacing if they need it.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+   * Gets injected by TabList, no need to specify yourself
+   */
+  isSelected?: boolean
+  /**
+   * Adds a Kaizen Badge component to the tab.
+   * Comes with some logic baked in - changes variant based on active/focus/hover state.
+   */
+  badge?: string
+  disabled?: boolean
+  children: ReactNode
```
ActuallyACat, line 18: We might want to pre-emptively expose a couple of events for tabs - I imagine folks would want to use them for analytics and such. We expose `onClick`, `onBlur`, `onMouseDown` and `onFocus` for button.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const Tab = (props: TabProps) => {
+  const { isSelected, badge, disabled, children } = props
+  const [isHovered, setIsHovered] = useState<boolean>(false)
+  const [isFocused, setIsFocused] = useState<boolean>(false)
+  const showActiveBadge = isSelected || isHovered || isFocused
+  return (
+    <ReachTab
+      disabled={disabled}
+      className={classnames(styles.tab, { [styles.selected]: isSelected })}
+      onFocus={() => setIsFocused(true)}
```
ActuallyACat, line 33: e.g. 
```
onFocus={(evt) => { 
   setIsFocused(true);
   props.onFocus(evt);
}}
```

packages/tabs/src/TabList.tsx
```diff
@@ -0,0 +1,23 @@
+import { TabList as ReachTabList } from "@reach/tabs"
+import React, { ReactNode } from "react"
+import styles from "./styles.scss"
+
+export interface TabListProps {
+  /**
+   * Accessible name for the set of tabs
+   */
+  "aria-label": string
```
ActuallyACat, line 9: Yas, love that this is required.

packages/tabs/src/styles.scss
```diff
@@ -0,0 +1,104 @@
...
+    border-color: $color-blue-500;
+    border-radius: $border-focus-ring-border-radius;
+  }
+}
+
+@media (-ms-high-contrast: active) {
+  .tab,
+  .tabPanel {
+    // All of the transparent borders set above are all seen in high contrast mode which we don't want
+    border: 0;
```
ActuallyACat, line 93: I thought that buttons should always have a visible border in high contrast mode, which is why borders are always set with a transparent border?

packages/tabs/src/styles.scss
dougmacknz: I'm totally with you (I went through all of the ones we need to remove in https://github.com/cultureamp/kaizen-design-system/issues/1894). this is a tough one though. This isn't quite "around the wrapper" because of the bottom border. Marc has a particular margin he wants here, and said that too commonly people are leaving this out.

image

packages/tabs/src/Tab.tsx
dougmacknz: I've done this but left out `onMouseDown` because I'm not sure I want to be encouraging people to do things exclusive to mouse users.

packages/tabs/src/styles.scss
dougmacknz: > I thought that buttons should always have a visible border in high contrast mode

Good call, I've fixed this now.

lloydstubber (approved):
Tiny nit: but the first is our implementation and second is the design. I definitely like the edges where they connect more in the design. Might be hard to replicate in the browser though but just looks a little sharp.

Screen Shot 2021-11-08 at 2 56 30 pm
Screen Shot 2021-11-08 at 2 56 58 pm

PR #2204, opened 2021-11-05
fix: Avoid cursor jacking for Pagination buttons

dougmacknz:
# Objective
Removes CSS properties changing the cursor on pagianation buttons.

# Motivation and Context
https://adamsilver.io/blog/buttons-shouldnt-have-a-hand-cursor/


chy-m (approved):
Oh interesting, I didn't know that. Great article

lloydstubber (approved):

PR #2208, opened 2021-11-08
feat: Remove theme from Elm Modal and EmptyState components

christian314159:
# Objective
Remove the use of Kaizen.Theme from the two Elm components Modal and EmptyState

# Motivation and Context
Other than removing these legacy theming options, it will enable an easier package refresh on Murmur


mbylstra (approved):
Looks good! Thanks for doing this.

PR #2265, opened 2021-11-08
feat: Merge button and icon button stories and add reversed-destructive Button variant, and "working" state on IconButton 

### LINKED ISSUE #1988: Replace Button stories with sticker sheets + a single story w/ controls
Rather than having a million stories for every single combination of props for `Button`, let's create a couple of sticker sheets (one for light background + 1 for reversed), similar to what the [UI Kit in Figma](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=1929%3A17364) has (without separate columns for hover and focus):
image

Plus we can have one story where controls are activated so that the user can try out different combinations of props in a single story. Note: Button already has at least 1 story where controls are activated: https://cultureamp.design/storybook/?path=/story/components-button-button--default-kaizen-site-demo

This is something we will probably look at doing on the rest of our components as well, but lets start with Button where this issue is the most apparent because of all of the different variants and combinations.

Credit: @ActuallyACat's awesome idea, not mine 🙂

NatalieLoyzaga:
# Objective

- Replace Button stories with sticker sheets + a single story w/ controls
-  Add reversed-destructive Button variant, and "working" state on IconButton

# Motivation and Context
closes #1988

https://user-images.githubusercontent.com/48232362/140686441-cc18aa25-1169-4f50-968e-a44b2f313572.mov


draft-packages/button/docs/IconButton.stories.tsx
```diff
@@ -1,179 +0,0 @@
...
-import { IconButton } from ".."
-import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
-
-export default {
-  title: `${CATEGORIES.components}/${SUB_CATEGORIES.button}/Icon Button`,
-  component: IconButton,
-  parameters: {
-    docs: {
-      description: {
-        component: 'import { IconButton } from "@kaizen/draft-button";',
```
jasononeil, line 12: This bit of documentation tells users how to import the component.

It would be nice to copy paste this line so that the new combined story documents how to import both `Button` and `IconButton`.

Or show a combined import:

```
'import { Button, IconButton } from "@kaizen/draft-button";',
```

NatalieLoyzaga: Updated to have combined import description. Thanks!

draft-packages/button/docs/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+.buttonState {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  margin-right: 5rem;
+  @include ca-media-mobile {
+    flex-direction: column;
+    margin-top: 1rem;
+  }
+}
```
jasononeil, line 22: No change requested for this PR, but when we do this to another component, I'd love to think about how to make these re-usable so they're consistent across all our stories.

Perhaps they're new components for storybook only? Perhaps they're styles that we re-use in all of our stories?

draft-packages/button/docs/styles.module.scss
```diff
@@ -0,0 +1,22 @@
+@import "~@kaizen/component-library/styles/responsive";
+
+.buttonSection {
+  display: flex;
```
jasononeil, line 4: Did you consider using CSS grid for these?

It feels like a perfect opportunity to play with grid and learn how it works (given you're building a grid 😄)

Flex does do the job fine of course, no need to change it unless you want to

draft-packages/button/docs/styles.module.scss
NatalieLoyzaga: Thats a really good idea. I'll add that as another issue to create one and then refactor the buttons stories after.

draft-packages/button/docs/styles.module.scss
NatalieLoyzaga: Will defs use it for the new future stories component

jasononeil (dismissed):
This is great!

I have some thoughts about making this a more reusable component for other component stories, but I don't think they need to happen in this PR.

draft-packages/button/docs/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  @include ca-media-mobile {
+    flex-direction: column;
+    margin-bottom: 1rem;
+  }
+}
+.buttonState {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  margin-right: 5rem;
```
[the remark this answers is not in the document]
jasononeil: (Again, I think this can wait for when we make this a reusable component for other stories)

jasononeil (approved):

PR #2272, opened 2021-11-09
docs: Fixing Collapsible & Likert links

christian314159:
# Issue
- Link on https://cultureamp.design/components/overview/ for "Likert Scale Legacy" wasn't working, instead redirected to the "Collapsible" component
- URL for "Collapsible" component showed as "https://cultureamp.design/components/docs/" and not as expected as "https://cultureamp.design/components/collapsible/"

# Findings
- the slugs are generated from where the `mdx` file is located, using the folder name as the slug (see `gatsby-node.js`)
- both readme.mdx for Collapsible & Likert were located in a `docs` subfolder and thus had the same URL (`/components/docs/`), Collapsible winning out

# Solution
- move both readme file up one level into their respective packages main folder, so the slug will be generated from the appropriately named folder instead


lloydstubber:
Likert seems to be going to: `https://dev.cultureamp.design/fix-gatsby-component-links/components/legacy/` Is that the intended slug now?

christian314159:
@lloydstubber Thanks, I'll have a look why it's not using the full folder name - I think I was too happy about it actually showing the content, that I didn't check the URL slug ;)

lloydstubber (approved):

PR #2276, opened 2021-11-10
docs: Component health page to calculate percentage done per column

christian314159:
# Objective
Show not only the count of components that have achieved a certain health metric (e.g. "has all variants" or "responsive"), but calculate the overall percentage per column

# Motivation and Context
When looking at our OKRs we seem to calculate the % each time by hand, this makes it much easier.

# Screenshots

image

# Additional changes

## Sticky table header & footer
Noticed the table header looking weird, it actually overlapped the "Avatar" row:
image

Found that the `sticky` CSS needed some tweaking, now both table header & footer stick appropriately to the edge of the screen (tested in CH, SF & FF):
image

## Design tweaks

Implemented some additional design tweaks with @katsambarcus 's help. 

image

**Note** also the wording changes from **Matches latest design** to **UI Kit Parity** to fit the tight column header, this will flow through to the component pages themselves as well.

image


site/src/constants.ts
```diff
@@ -22,13 +22,13 @@ export const healthAttributeMap = [
   },
   {
     id: "latestDesign",
-    positive: "Matches latest design",
-    negative: "Doesn’t match latest design",
+    positive: "UI Kit parity",
+    negative: "No UI Kit parity",
```
ActuallyACat, line 26: nit/ We use the same language on the health page and for our OKRs so they match. If this is being updated, the OKR metric should match.

ActuallyACat (approved):

site/src/constants.ts
christian314159: Love me a good nit pickin' ;) That's a good point, fyi I am trying to reword/optimize our OKRs right now and will make sure it says the same thing. Cheers.

dougmacknz: If you want another nit: to me, "UI Kit parity" would mean latest design _and_ has all variants.

dougmacknz (approved):

PR #2286, opened 2021-11-14
feat: Column component

lloydstubber:
## Work
- @aaronknol and I have been working on this "Responsive" component (name TBC) to align with the [new responsive grids](https://www.figma.com/file/Xh4HCZB4IXUPLhyb8DOIPd/Responsive-Guidance%3A-Page-Layouts?node-id=151%3A6859) that @katsambarcus and the Design Advocates have been working on.
- The API is quite simple, we have props for each size and you pass the number (as a string at the minute but ideally a number) for the amount of columns at the sizes - sizes are hardcoded in the styles and determined by the Design Advocates in the Figma.
- Added in a hook so you can see the browser width change.

## Thoughts
- We'd need to document that 0px -> 664px will be one coloumn regardless and the `small="x"` kicks in between 664-1008px. 
- @katsambarcus the breakpoints seem quite close once you push over 1000 e.g. 1008, 1057, 1157. Is this intended?
- There needs to be more logic around min/max widths for columns as per the design doc but this is just a first pass.
- So far, it's quite simple but the amount of wrangling to get the CSS working isn't ideal. This would most definitely have a cleaner implementation working with a CSS-in-JS solution.

## How do I view it?
1. Go to [Storybook](https://dev.cultureamp.design/responsive-playground/storybook) -> Responsive to find it
2. I'd recommend closing the side bar ('S' key) so the story is full-width

Screen Shot 2021-11-12 at 4 34 01 pm


packages/column/Column.scss
```diff
@@ -0,0 +1,111 @@
...
+  .xlarge-3 {
+    @include three-column;
+  }
+
+  .xlarge-4 {
+    @include four-column;
+  }
+}
+
+@media (min-width: 1495px) {
```
christian314159, line 95: I don't think these classes will be used, unless you also add a `xxlarge` prop to the component.

packages/column/Column.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type ColumnProps = {
+  small?: "1" | "2"
+  medium?: "1" | "2" | "3"
+  large?: "1" | "2" | "3"
+  xl?: "1" | "2" | "3" | "4"
+  children?: React.ReactChild
+}
+
+export default function Column({ small, medium, large, xlarge, children }) {
```
christian314159, line 13: Was at first wondering, if this should be called `Row` instead (as the elements inside rather are the "columns"), but then I thought this could contain 12+ elements and it would give you a grid kinda layout with multiple rows - a grid rather, wouldn't it? So, in the end - I don't know, but might be worth putting out there.

christian314159 (approved):

PR #2287, opened 2021-11-15
fix: Restore inbuilt padding on TabList and add 'noPadding' prop

dougmacknz:
Undoes https://github.com/cultureamp/kaizen-design-system/pull/2281 and adds an `unpadded` prop to TabList to remove the padding.

Context: https://cultureamp.slack.com/archives/C016N8Y1E10/p1636932197107600


packages/tabs/docs/Tabs.stories.tsx
```diff
@@ -145,7 +139,7 @@ export const UsageInCard = () => (
       // eslint-disable-next-line no-console
       onChange={index => console.log("Tab changed to ", index)}
     >
-      <Box pt={0.25} pl={1} pr={1}>
+      <Box pt={0.25}>
```
dougmacknz, line 142: Question is, should this padding be built in as well 🤔

packages/tabs/src/TabList.tsx
```diff
@@ -1,18 +1,33 @@
...
 
 export interface TabListProps {
   /**
    * Accessible name for the set of tabs
    */
   "aria-label": string
+  /**
+   * Removes the built in margins on the sides
+   */
+  unpadded?: boolean
```
ActuallyACat, line 14: nitpicking, but I wonder if `noPadding` is more consistent with other components that have optional spacing. Form packages use the `no-${positionOfSpacing}${"padding"|"margin"} convention like `noBottomMargin`. Don't feel strongly about this though, as it's an uncommon enough scenario.

packages/tabs/docs/Tabs.stories.tsx
ActuallyACat: It would probably be easier to reason about if all padding was built in, especially as the prop is named "unpadded"

packages/tabs/src/TabList.tsx
dougmacknz: Yeah that's cool, I've renamed it now

ActuallyACat (approved):

PR #2291, opened 2021-11-16
fix: Use visually hidden instead of display:none for breadcrumb link name

dougmacknz:
# Objective
Replaces `display: none` with a set of styles used to hide something visually while keeping it accessible to assistive technology.

# Motivation and Context
This was pulled up in the accessibility audit retesting, here's what was said:
> The back to home link has now been combined into a single element. However the ‘Back to home’ text within the <a> is hidden via display:none; and only visible on hover or focus. This means a screen reader user navigating via read mode will not hear a name for the link (the URL is announced as the link’s name).
When the back to home text is hidden, do not use display:none; - instead use the clipping method utilised to hide text for screen reader users.


lloydstubber (approved):
👑  for the comments

PR #2306, opened 2021-11-23
feat: Update checkbox group design and stories

### LINKED ISSUE #2305: [Nat] [Checkbox Group] Update design
Update Checkbox Group Component to match the latest design.

Reduced the margin in-between checkboxFields to match Figma.

NatalieLoyzaga:
# Objective
Update the margin to match Figma design. Updated Stories to...
1.  Remove links from labels including the With Link story.
2. Use CheckboxField as the StoryBook control Component and update Demo story to use args.
3. Update Figma URL

# Motivation and Context

closes #2305


draft-packages/form/docs/CheckboxGroup.stories.tsx
```diff
@@ -44,15 +44,18 @@ class CheckboxGroupExample extends React.Component<Props> {
 
 export default {
   title: `${CATEGORIES.components}/${SUB_CATEGORIES.form}/Checkbox Group`,
-  component: CheckboxGroup,
+  component: CheckboxField,
```
dougmacknz, line 47: Was this intentional?

NatalieLoyzaga: Yep - I thought it made sense that you'd be able to control the checkboxField Props e.g checked on and off. Whereas the checkBoxGroup kinda just acts as the wrapper + a label and id. 

If you know how to have both checkboxField and checkboxGroup as a component plz let me know

dougmacknz: Ah right, makes me think that we shouldn't have a separate set of stories for CheckboxGroup/RadioGroup

jasononeil: It would be nice to be able to control the Group and the Field in one story, but 🤷 

I'm unsure which is best given we have to pick one. I think this is fine for now

jasononeil (approved):

PR #2321, opened 2021-11-29
chore: Removed deprecated ZenNavigationBar component

christian314159:
# Objective
Remove the ZenNavigationBar component that is not needed anymore seeing the Zen theme is phased out.


dougmacknz (dismissed):
🔪

dougmacknz (approved):

PR #2327, opened 2021-11-29
feat: Add feature detection for safari version < 13 to avoid addEvent…

### LINKED ISSUE #2299: useMediaQueries causing errors in production for some safari users
### If it's a discussion, please start one [here](https://github.com/cultureamp/kaizen-design-system/discussions).
```
TypeError
a.addEventListener is not a function. (In 'a.addEventListener("change",e,!0)', 'a.addEventListener' is undefined)
```
Tracked down to: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/responsive/src/useMediaQueries.tsx#L101

According to [MediaQueryLists browser compatibility](
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility),
> Before Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists.

- [x] Before opening this issue, I've searched open and closed issues for a fix.

## Expected Behaviour
I'm not sure, either `useMediaQueries` should gracefully degrade (similar to how we check for Edge) or we should support older versions with `addListener/removeListener`.

## Specifications (please add any other notable libs/details)

 For `performance-ui`:
 ```
 => Found "@kaizen/responsive@1.2.1"
info Reasons this module exists
   - "@kaizen#brand-moment" depends on it
   - Hoisted from "@kaizen#brand-moment#@kaizen#responsive"
   - Hoisted from "@kaizen#draft-title-block-zen#@kaizen#responsive"
```
Not sure about others.

## Labelling

1. If related to a component, please add a label like so `component:mycomponent`.
2. Then label with `bug`, `enhancement`/`feature`\* or `documentation`.

**\* A feature is a new component or piece of functionality and an enhancement is an improvement to an existing component or piece of functionality.**

## No Sensitive Information
As this is a public repo, please do not include screenshots that may share sensitive information about our customers.

### PR DESCRIPTION
**Pull request by NatalieLoyzaga**
# Objective

Update useMediaQueries to check fro safari version <= 13 by checking matchMedia.addEventListener method returns undefined when in an unsupported version of safari. 

# Motivation and Context
closes #2299

### DIFF
--- a/packages/responsive/src/useMediaQueries.tsx
+++ b/packages/responsive/src/useMediaQueries.tsx
@@ -31,16 +31,7 @@ export const useMediaQueries = (
   // The `addEventListener` calls blow up legacy Edge (<= v18/pre chromium),
   // so we disable the functionality of updating after page load.
   const isLegacyEdge = navigator.userAgent.match(/Edge/)
-
-  // To double check:
-  // Does Safari <13 consistently use addListener?
-  type windowAndDeprecatedEventListeners = {
-    addListener: any
-    removeListener: any
-  } & Window
-  const isUnsupportedSafari =
-    ((window as unknown) as windowAndDeprecatedEventListeners).addListener !==
-    undefined
+  const isUnsupportedSafari = (window.matchMedia('')).addEventListener !== undefined; 
 
   // ---------------------------------------
   // Create Kaizen breakpoint matches for initial state
--- a/packages/responsive/src/useMediaQueries.tsx
+++ b/packages/responsive/src/useMediaQueries.tsx
@@ -32,6 +32,16 @@ export const useMediaQueries = (
   // so we disable the functionality of updating after page load.
   const isLegacyEdge = navigator.userAgent.match(/Edge/)
 
+  // To double check:
+  // Does Safari <13 consistently use addListener?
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by ActuallyACat — APPROVED** [review]
🚢

File: packages/responsive/src/useMediaQueries.tsx [hunk 1]
```diff
@@ -31,6 +31,8 @@ export const useMediaQueries = (
   // The `addEventListener` calls blow up legacy Edge (<= v18/pre chromium),
   // so we disable the functionality of updating after page load.
   const isLegacyEdge = navigator.userAgent.match(/Edge/)
+  const isUnsupportedSafari =
+    window.matchMedia("").addEventListener === undefined
```
dougmacknz, line 35: This seems like it could be a replacement for the `isLegacyEdge` conditional?

NatalieLoyzaga: @ActuallyACat I merged already but did we want to revert and do this ^

dougmacknz: No need to revert, if anything we can do a follow up PR for it, but it's no big deal

PR #2330, opened 2021-11-30
feat: Add Profile varient to Tag

### LINKED ISSUE #790: Tag is missing the "profile" variant
image (28)

storybook: https://cultureamp.design/storybook/?path=/story/tag-react--default-medium-kaizen-site-demo
UI Kit: https://www.figma.com/file/GMxm8rvDCbj0Xw3TQWBZ8b/Zen-UI-Kit?node-id=1929%3A29917

NatalieLoyzaga:
# Objective

# Motivation and Context
closes #790 

image

# Additional Considerations

### DIFF
--- a/draft-packages/tag/KaizenDraft/Tag/Tag.tsx
+++ b/draft-packages/tag/KaizenDraft/Tag/Tag.tsx
@@ -1,7 +1,8 @@
 import { Icon } from "@kaizen/component-library"
+import { Avatar, AvatarProps } from "@kaizen/draft-avatar"
 import classNames from "classnames"
 import * as React from "react"
-import clearIcon from "@kaizen/component-library/icons/clear.icon.svg"
+import clearIcon from "@kaizen/component-library/icons/clear-white.icon.svg"
 import exclamationIcon from "@kaizen/component-library/icons/exclamation.icon.svg"
 import informationIcon from "@kaizen/component-library/icons/information.icon.svg"
 import successIcon from "@kaizen/component-library/icons/success.icon.svg"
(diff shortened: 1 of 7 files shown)


draft-packages/tag/KaizenDraft/Tag/Tag.tsx
```diff
@@ -60,7 +75,7 @@ const Tag = (props: TagProps) => {
   return (
     <div
       className={classNames(styles.root, {
-        [styles.default]: variant === "default",
+        [styles.default]: variant === "default" || "profile",
```
christian314159, line 78: This will always be a `truthy` expression. If the goal is to use this style when `variant` is either `default` or `profile`, this would need to be `variant === "default" || variant === "profile"`.

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
```diff
@@ -103,6 +118,13 @@ const Tag = (props: TagProps) => {
                 </span>
               )
             }
+            if (variant === "profile") {
+              return (
+                <span className={styles.validationIcon}>
+                  {avatar && renderAvatar(avatar)}
```
christian314159, line 124: I'm a TypeScript noob, but I think there is a way of making props depending on each other, as in requiring the `avatar` prop when `variant` is `"profile"`.
This would mean someone using the `Tag` component with `variant="profile"` would get immediate feedback if they didn't provide the `avatar` prop.
This would then also make this check here (`avatar && ...`) obsolete.

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
```diff
@@ -103,6 +118,13 @@ const Tag = (props: TagProps) => {
                 </span>
               )
             }
+            if (variant === "profile") {
```
christian314159, line 121: For future reference: You just added on to the pattern as it was, but I really like the `switch(true)...` pattern for these kind of scenarios, rather than doing all these `if(...){ return ...)` statements:
https://medium.com/trabe/using-switch-true-in-javascript-986e8ad8ae4f

christian314159 (changes requested):
Mostly just refactoring comments, but please have a look at the one about the classes (default/profile), that might be a bug.

Also, and I know there were no tests to begin with, I'd really love to see a basic test for at least the new variant that was added ❤️

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
NatalieLoyzaga: Thank you !! silly me

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
NatalieLoyzaga: I've given it a crack using "Discriminative Unions" to achieve this. Seems to work :)

NatalieLoyzaga:
@christian314159 Made those Typescript updates for the `profile` variant and the `avatar` prop. I have also add unit tests for default and all variants now.

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
```diff
@@ -122,7 +146,7 @@ const Tag = (props: TagProps) => {
               onMouseDown={onMouseDown}
               onMouseLeave={onMouseLeave}
             >
-              <span className={styles.background} />
+              {/* <span className={styles.background} /> */}
```
christian314159, line 149: If this is not needed, we should remove it, rather than leaving it here commented out.

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -204,14 +204,16 @@ const renderTag = (surveyStatus: SurveyStatus) => {
...
   }
-
-  return (
-    <div data-automation-id="survey-status-tag" className={styles.tag}>
-      <Tag variant={tagVariant} size="small">
-        {surveyStatus.text}
-      </Tag>
-    </div>
-  )
+  // TODO: explicitly omit/exclude profile as a variant type value
```
christian314159, line 207: 2 things here:

I don't actually understand how `tagVariant` could ever be `"profile"`, it's a local variable to the `renderTag` function and never set to `profile`?
 
It'd be nicer to have a "guard statement/clause" here instead of wrapping the `return` statement in an `if`:
```
if (tagVariant == "profile" ) { return }

return (
  <div>...
```

legacy-packages/title-block/KaizenDraft/TitleBlock/TitleBlock.tsx
```diff
@@ -106,11 +106,14 @@ class TitleBlock extends React.Component<Props, State> {
       variant = "statusLive"
     }
 
-    return (
-      <div className={styles.tag}>
-        <Tag variant={variant}>{surveyStatus.text}</Tag>
-      </div>
-    )
+    // TODO: explicitly omit/exclude profile as a variant type value
+    if (variant !== "profile") {
```
christian314159, line 110: Same here, unsure how this would ever be `"profile"`

christian314159 (dismissed):
👏

christian314159 (approved):

PR #2331, opened 2021-11-30
chore: Set timeout for each Buildkite step to 15min

christian314159:
# Objective
Each of our Buildkite steps should have a defined timeout in case it goes into an endless cycle, never exits, and never releases the build agent.

# Motivation and Context
We have seen occasions where builds were going on for days (!).


ActuallyACat (approved):
🚢 Ship it

PR #2351, opened 2021-12-08
feat: add company variants to Avatar, update storybook format and add test coverage

mcwinter07:
BREAKING CHANGE: sizes for avatars large and up will be updated which may cause breaking layout changes

# Objective
- This will add a control story so that developers can interact with the props
- This will add a 'isCompany' prop to Avatar and add missing stories for company avatars and design sheets
- This will update avatar sizes from `large`, `xlarge` and add `xxlarge`
- This will update some semantic structure of the avatar output to better support A11y
- This will add some tests to increase coverage over the component render logic
- Company Avatars will be strictly typed

# Motivation and Context
- To address the missing variants for company avatars and design sheet previews for default and reversed backgrounds.
- To update and standardise avatar sizes across company and user avatars after a design review
- To update the Avatar stories to make it easier to use and digest for other developers
The changes and updates to avatar sizes have come from collaboration with design to standardise avatar sizes on Large and up.

## Original
orginal stories

original story doc
f3.png)

## Updates
Screen Shot 2021-12-16 at 4 25 21 pm

The initial story now has props exposed to increase utility for developers

Screen Shot 2021-12-16 at 4 25 08 pm

Story coverage now includes the company stories and design sheets. With the introduction of the design sheets and controllable story we have removed older stories, and extended long initials and unicode as outliers.

# Checklist


katsambarcus:
@mcwinter07 lookin' good, some observations and thoughts:

**Controls**
image

- There appear to be two controls for disabling initials. The first (DisabledInitials) doesn't seem to do anything when I toggle between True/False. The second (disableInitials) worked.

- What does 'Edit String…' for the avatarSrc intend to do? Is this to add a URL for an image? Could this field's instructions be clearer for the consumer?

- Size control: Just want to confirm we're standardising our naming for our t-shirt sizes. I'm not sure what the preferred naming is but want to check that we're all aligned or leaning into other examples that exist in the system, e.g. is it Sm / S / small / Small? I don't have strong preferences, but would like to make sure there's consistency in naming across our code components and, ideally, what's defined in the UI Kit.

Will come back to you shortly with more around the design details (need to jump into a meeting)

draft-packages/avatar/KaizenDraft/Avatar/decision-tokens.scss
```diff
@@ -5,3 +5,10 @@ $dt-color-background-color-personal: $color-orange-100;
 $dt-color-text-color: $color-purple-800;
 $dt-color-border-color: $color-white;
 $dt-color-icon-color: rgba($color-purple-800-rgb, 0.7);
+
+// Proposed Values
+$dt-avatar-sm: 1.5rem;
```
dougmacknz, line 10: Don't worry about this now that you've done it—but FYI we're wanting to move away from these `decision-tokens` files. So if you were to do this again, I'd recommend that you just create these variables from within the `styles.module.scss` file.

draft-packages/avatar/KaizenDraft/Avatar/styles.module.scss
```diff
@@ -27,26 +27,41 @@
...
+    width: $dt-avatar-xl;
+    height: $dt-avatar-xl;
+  }
+
+  &.xxlarge {
+    width: $dt-avatar-xxl;
+    height: $dt-avatar-xxl;
+  }
+
+  abbr {
```
dougmacknz, line 62: I'd recommend we add a class to this element and style based on that instead. The reason I say this is because we've run into issues before where we've changed an element's tag and not realised that there's CSS attached to it. Adding a class makes it clear that there's styles attached, and it allows future developers to adjust the tag without needing to change the CSS selector.

dougmacknz (commented):
Looking pretty good! Just a couple small things to touch up. Has this had a design review yet?

dougmacknz:
One other thing is that it would be good to adjust the PR title to be more descriptive :)

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -98,21 +111,24 @@ export const Avatar = ({
...
           ref={image}
-          className={styles.avatarImage}
+          className={cx(styles.avatarImage, {
+            [styles.companyAvatarImage]: isCompany,
+          })}
           src={avatarSrc}
           onError={onImageFailure}
           onLoad={onImageSuccess}
-          alt={fullName}
+          alt={fullName || " "}
```
NatalieLoyzaga, line 120: In terms of accessibility - Wouldn't this need to be `"`" or does `" "` do the same thing? Question for @dougmacknz

dougmacknz: Ooh yeah good catch!

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -98,21 +111,24 @@ export const Avatar = ({
...
       {(avatarState === "none" || avatarState === "error") &&
         (disableInitials || initials === "" ? (
           fallbackIcon
         ) : (
-          <span
+          <abbr
             className={cx(styles.initials, {
               [styles.longName]: isLongName,
             })}
+            title={fullName || " "}
```
NatalieLoyzaga, line 131: Same here as above "" rather than " "

katsambarcus:
@mcwinter07 part 2 of review:

**Company:**
- For the isCompany boolean, is it possible for us to inject a company logo (.svg) so that peeps don't become confused as to what this component is for? 

- Also, it may not have been clear in the UI Kit, but I'd like for this component to include padding around the edges of it, so that there is always a little bit of pre-baked padding of a customers logo. See below for example:
image

  - Padding is 6px around all sides
  - Border radius of the logo container is 4px

- The `small` company progile has the letter H displayed (see visual) while the other sizes show the Oracle logo. Is the intent to remove the logo and show the intials of the organization? I quite like the idea of this becoming a psuedo-favicon, and I wonder if we should treat it as such by styling it as a plain old 20px favicon, with no bells and whistles (i.e. radius, shadows, etc)?
image

- Let's replace the `Oracle` logo with the `Hooli` logo, so that we're aligned with the UI Kit


**General questions/observations:**

- Is the intention to remove all the other stories and be left with `Default (Kaizen Demo)`, `Design Sheet (default)`, `Design Sheet (reversed)`? 
  - Do we need a sticker sheet for both default _and_ reversed? Could this be controlled with a true/false on one sticker sheet?

- The Small avatars are all 24px, however in the UI Kit they are 20px. This is so that they are an equal size to our icons, which are deliberately smaller than 24px, so that they can fit within other components and alongside our type styles in a balanced way. Can I get I get you to change these to 20px please?

- The alignment of initials/ icon on `small` and `medium` avatars is sitting a little high within the circle. Could you nudge these down by a pixel or two, so that the are centred on the y-axis, please?

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -29,7 +29,7 @@ type DistributiveOmit<T, K extends keyof any> = T extends any
 
 export const NON_REVERSED_VARIANTS = ["education", "admin"]
 
-type AvatarProps = Omit<DraftAvatarProps, "size">
+// type AvatarProps = Omit<DraftAvatarProps, "size">
```
NatalieLoyzaga, line 32: Can this comment be removed?

NatalieLoyzaga (commented):
Looks good - just made one comment about removing a comment.

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -98,32 +126,37 @@ export const Avatar = ({
...
-                {initials}
-              </Textfit>
-            ) : (
-              getInitials(fullName, size === "small")
-            )}
-          </span>
-        ))}
+      {!isCompany
+        ? (avatarState === "none" || avatarState === "error") &&
+          (disableInitials || initials === "" ? (
```
dougmacknz, line 140: This mega conditional was hard to read and understand before, and it's now become a bit harder 😅, you might want to consider extracting it out into a function.

draft-packages/avatar/KaizenDraft/Avatar/styles.module.scss
```diff
@@ -59,42 +93,58 @@
...
-    width: 62px;
+    width: 60px;
+  }
+  .large & {
+    width: 35px;
   }
   .medium & {
-    width: 20px;
+    width: 25px;
+    // TODO: confirm mb
```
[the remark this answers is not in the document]
mcwinter07: confirmed, will remove the comments :)

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
```diff
@@ -5,61 +5,82 @@ import { Textfit } from "react-textfit"
...
-   * There are 3 fixed avatar sizes. "small" will remove border and box shadow to save space.
-   * @default "medium"
+   * Renders Company Avatar variant - If true `fullName` and `avatarSrc` will be strictly typed.
    */
-  size?: AvatarSizes
+  isCompany?: false
 }
 
-export const Avatar = ({
+interface CompanyAvatarProps
```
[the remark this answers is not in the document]
mcwinter07: Good idea 👍

draft-packages/avatar/KaizenDraft/Avatar/Avatar.tsx
mcwinter07: I've split the initials into its own render function to make this a little less gnarly to read

christian314159 (approved):

jmitsios:
@mcwinter07 - Absolutely minor thing here, but could we re-order this sizing list in order of increasing size from **small** upwards to **XXLarge**?  It's just a bit out of order currently.
image

Outside of that I have no other issues to see!

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -29,8 +29,6 @@ type DistributiveOmit<T, K extends keyof any> = T extends any
 
 export const NON_REVERSED_VARIANTS = ["education", "admin"]
 
-type AvatarProps = Omit<DraftAvatarProps, "size">
```
dougmacknz, line 32: Curious why this was removed, 'cause it looks like it will still be forced to be `medium` size when passed into the title block

jmitsios (approved):

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
mcwinter07: I removed this because its was causing a TS compile error when it came up against updated Avatar Props. I think issue lies in trying to omit size from the union type of GenericAvatarProps and CompanyAvatarProps. I could potentially bring this back if you see it causing an issue by importing GenericAvatarProps over Avatar props but given size is set of ln 238 to medium as the default I'm not sure if that is necessary? Would be keen to get your thoughts

dougmacknz: I think it's to stop people passing in a size, because regardless of what is passed in, it will always become medium (see https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx#L240).

If you can make it work, I think it does make sense to keep it.

mcwinter07: Found a solution to get the best of both, I've pull in the individual interfaces and omitted the size prop
```
type AvatarProps =
  | Omit<GenericAvatarProps, "size">
  | Omit<CompanyAvatarProps, "size">
```

dougmacknz (approved):

PR #2355, opened 2021-12-10
Update slider documentation

lindsey-collins:
# Objective
Adding progress slider documentation to the overall slider documents.


site/docs/components/slider.mdx
```diff
@@ -51,6 +58,9 @@ Steps are selectable values along a slider range and can be visually indicated b
 
 Where the range of values is not obvious, sliders should be labeled with minimum and maximum values. An optional center label can also be used to denote a middle value or as a description.
 
+### Progress track
+
+A filled progress track is used in the progress slider. It follows the path of the regular track to indicate where the handle has been dragged from.
```
dlimiter, line 63: Is "where the handle has been dragged from" clear enough – could it be interpreted as "from it's last position" rather than "from the lowest/left value in the range"?

site/docs/components/slider.mdx
```diff
@@ -28,14 +28,21 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
 
 <iframe style="border: none;" width="744" height="450" src="https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FeZKEE5kXbEMY3lx84oz8iN%2F%25E2%259D%25A4%25EF%25B8%258F-UI-Kit-Heart%3Fnode-id%3D16080%253A4049" allowfullscreen></iframe>
 
+## Options
+
+- **Default:**
+    - In a default slider, the user can make an imprecise selection from a range of values, where there is no implication of progression from one side to the other.
+- **Progress:**
+    - In a progress slider, the user can make an precise selection from a range of values, aided by an exact indication of value. A sense of progress is implied, from one side to the other.
```
dlimiter, line 36: it's not super clear to me why in one of these cases the selection is precise and in the other imprecise? Is this because the handle is imprecise, but the current value field enables the exact current value to be set?

site/docs/components/slider.mdx
lindsey-collins: Good suggestion; I've updated the wording - hope it's a little more clear.

site/docs/components/slider.mdx
lindsey-collins: Yeah, that was the intention, but I definitely see that this could be confusing. Maybe I'm just making this more difficult than it needs to be; I can remove the words precise/imprecise and that could be better.

site/docs/components/slider.mdx
```diff
@@ -28,14 +28,21 @@ import WhenNotToUse from "docs-components/WhenNotToUse"
...
+
 ## To keep in mind
 
 Sliders are helpful for choosing subjective values.
 
 - Always indicate the current, selected value. Ideally with visible text associated with the handle, but at a minimum provide a text alternative for screen readers.
 - Provide a visible label describing what the slider changes, such as "Brightness".
 - Show the minimum value on the left and maximum value on the right, where these are reversed in Right-To-Left (RTL) languages.
-- We generally use sliders with no fill. You might use a fill when it's useful to give the impression of building towards the maximum value.
+- Default sliders have no full. The progress slider uses a fill to give the impression of building towards the maximum value.
```
katsambarcus, line 45: @lindsey-collins should we be referencing the ###Progress track section at this point? e.g. See [Progress track](/components/slider/#progress-track)

katsambarcus (changes requested):

site/docs/components/slider.mdx
lindsey-collins: Good call! Updated.

site/docs/components/slider.mdx
lindsey-collins: Have removed the precise/imprecise thing all together - @dlimiter and/or @katsambarcus, is that better?

PR #2365, opened 2021-12-14
fix: Chris/node engine support v16

ckychris:
# Notes

- Upgrading Node to v16 requires upgrading `node-sass` (we're several versions behind)
- Upgrading `node-sass` breaks a bunch of tests, which we haven't looked into yet
- We also want to upgrade `node-sass` in Front End Ops, it may be easier to do that first.

# Updates

Type errors appeared from Storybook `Meta` types and postcss node hierarchy changes, with the additional type `Document`. Working with @dmisdm to resolve util functions changes.

Prettier errors appeared which are resolved by linting every file once again.


packages/stylelint-plugin/src/util/variableUtils.ts
```diff
@@ -74,15 +81,17 @@ export const replaceTokenInVariable = (
  * Input: stylesheet,
  * Output: { "$foo": "red", "$other": "rgba(0, 0, 0, 0.1)". ...}
  */
-const getVariablesInBlock = (block: Container) =>
-  block.nodes
+const getVariablesInBlock = (block: ChildNode | Container | Document) => {
+  if (!("nodes" in block) || "toResult" in block) return {}
```
ckychris, line 85: @dmisdm 2 changes are made to add the `Document` type support into functions that process postcss variables as postcss parser and `parent` can return `Document` instead of `Container` or `ChildNodes`. 

I am not sure if my fix is valid as we might want to get `Doucment`'s global variables, which is causing a type error for a signature mismatch. As Document extends `Container<Root>`, it should able to be added without any issues.

The error stems from `.filter`

`
This expression is not callable.
  Each member of the union type '{ <S extends ChildNode>(predicate: (value: ChildNode, index: number, array: ChildNode[]) => value is S, thisArg?: any): S[]; (predicate: (value: ChildNode, index: number, array: ChildNode[]) => unknown, thisArg?: any): ChildNode[]; } | { ...; }' has signatures, but none of those signatures are compatible with each other.ts(2349)
`

```
return block.nodes
    .filter(
      (node): node is Declaration => node.type === "decl" && isVariable(node)
    )
    .reduce(
      (acc, next) => ({ ...acc, [next.prop]: next.value }),
      {} as Record<string, string | undefined>
    )
```

site/gatsby-config.js
```diff
@@ -95,7 +95,10 @@ module.exports = {
     {
       resolve: "gatsby-plugin-sass",
       options: {
-        precision: 9,
+        implementation: require("node-sass"), // TODO: Move to dart-sass once webpack has native plugin support
```
jasononeil, line 98: I believe dart-sass might be incompatible with some of our legacy codebases at the moment... last time we tried it caused build timeouts with no error messages 😬 Worth a shot again though, it'd be really nice to use it, but if we use it in Kaizen we need to be able to use it everywhere.

packages/design-tokens/docs/StoryBoard.stories.tsx
```diff
@@ -251,20 +251,20 @@ const StoriesContainer = (props: {
       {shouldRender && (
         <Stack key={key}>
           {Object.entries(props.storyModule).map(([k, V]) => {
-            // Ignore exports that are Story Metadata (which we can detect by the presence of a "title" prop)
-            if ("title" in V) return null
-            const parameters = "story" in V ? V.story.parameters : V.parameters
+            // !This is now irrelevant as "title" prop is now string | undefined
+            // (outdated) Ignore exports that are Story Metadata (which we can detect by the presence of a "title" prop)
```
jasononeil, line 255: Not sure if it's addressed in a later commit but we should clean up these outdated comments.

@dmisdm might be the best person to review for this file probably.

packages/stylelint-plugin/src/util/walkers.ts
```diff
@@ -116,6 +116,7 @@ export const walkDeclsWithKaizenTokens = (
     value: string
   }) => void | false
 ) => {
+  if (!("walk" in stylesheetNode)) return
```
jasononeil, line 119: @dmisdm can you review the stylelint-plugin changes in this PR?

packages/stylelint-plugin/src/util/variableUtils.ts
dmisdm: @ckychris just an FYI that this `if (!("nodes" in block) || "toResult" in block) return {}` broke tests because it changed the behaviour at runtime, but i totally get why you did it

dmisdm: at least i think it did...second guessing myself

ckychris: With the type restriction set to `Root` only. This line should be removed.

dmisdm: unfortunately that breaks it! i think Document needs to be here, just working through it

dmisdm: I think it should be ok now after the change i made, which checks for Document instances and bails out, for a good reason which is stipulated in the comments :)

draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.spec.tsx
```diff
@@ -18,11 +18,11 @@ afterEach(cleanup)
 
 describe("<AnimatedBase />", () => {
   describe("Loading", () => {
-    beforeEach(() => {
-      mockedGetAnimationData.getAnimationData.mockResolvedValue(
-        {} as LottieAnimation
-      )
-    })
+    // beforeEach(() => {
```
christian314159, line 21: Interesting, commenting out these `beforeEach` blocks didn't make the tests fail?
If they are really unnecessary, we should delete them, rather than leaving them here as comments.

packages/design-tokens/docs/StoryBoard.stories.tsx
```diff
@@ -251,20 +262,19 @@ const StoriesContainer = (props: {
       {shouldRender && (
         <Stack key={key}>
           {Object.entries(props.storyModule).map(([k, V]) => {
-            // Ignore exports that are Story Metadata (which we can detect by the presence of a "title" prop)
-            if ("title" in V) return null
-            const parameters = "story" in V ? V.story.parameters : V.parameters
+            if (isStoryMeta(V)) {
```
christian314159, line 265: In my PR for the Vulnerability Checking (KDS-99) I ran into some problems here as well, whoever merges in 2nd will likely have to make some changes again here ;)

christian314159 (changes requested):
Looking good, but see comments re comments ;) in LottiePlayer spec.
Also, this being about upgrading to Node v16, should the `.nvmrc` file be updated to reflect a v16 version so people using NVM use that version automatically?

draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.spec.tsx
ckychris: Yeah/ Although I can see the intention of the author was to expect it to have been called? However they just put down a line of function call there.

packages/design-tokens/docs/StoryBoard.stories.tsx
ckychris: Thanks, It's KDS-104 if you want to see the ticket.

draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.spec.tsx
ckychris: I have just put them back, but without `beforeEach`, and added `toHaveBeenCalled`.

ckychris:
> Looking good, but see comments re comments ;) in LottiePlayer spec. Also, this being about upgrading to Node v16, should the `.nvmrc` file be updated to reflect a v16 version so people using NVM use that version automatically?

Good spot. I have just updated all node version reference from 14 to 16, except the package.json range requirements.

christian314159 (approved):
👏

PR #2368, opened 2021-12-16
Update props 'title' types in `InlineNotification`

zoe0119:
# Objective
- When using `InlineNotification` component in murmur, we noticed the type of props `title` is mandatory. However, when we check the code in kaizen, the `title` props will be passed into `GenericNotification`, and the type here is optional, which means the props type are not consistent.

# Motivation and Context

-- Updating the prop type of `title` in InlineNotification to optional.

 

# Checklist

# Additional Considerations

### DIFF
--- a/packages/notification/src/InlineNotification.tsx
+++ b/packages/notification/src/InlineNotification.tsx
@@ -5,7 +5,7 @@ import GenericNotification, {
 
 type Props = {
   type: NotificationType
-  title: string
+  title?: string
   children?: React.ReactNode
   autohide?: boolean
   autohideDelay?: "short" | "long"


dougmacknz:
Hi @zoe0119, firstly, great to have you jump into Kaizen to make a contribution. However, I believe these two things are intentional inconsistent. `GenericNotification` is used across 3 components: `InlineNotification`, `ToastNotification`, and `GlobalNotification`. `GlobalNotification` doesn't have a title, and so `GenericNotification` must have title as optional. On the other hand, having checked with our designer @katsambarcus—he envisions titles as being a necessary part of the `InlineNotification` component.

If you have a design that has an Inline Notification which doesn't have a title, I would recommend starting a conversation with Marc and your designer to come to a conclusion on whether:
a) the design should change, or
b) this prop should be made optional

PR #2372, opened 2021-12-20
chore: Adding JS Vulnerability Scan step to pipeline [KDS-99]

christian314159:
# Objective
- Add a Buildkite step to check for vulnerable NPM packages
- Fail build when audit fails

# Notes
- Added an allow list file for all the packages I couldn't upgrade and also just were dev dependencies
- Some files are just Prettier changes
- Some TS errors came up that I fixed rather "crudely" as they were in some weird places
- Seems `resolutions` within a `package.json` only work in the main one, but not in the sub packages themselves (the ones in the Gatsby package.json didn't work)
- Had to pin yarn to 1.18.0 to fix an issue in one of the audit-fix steps

## Is this a breaking change?
Not 100% on this, but no, I don't think so. The only packages that were touched are the Gatsby site and the `generator`, which is actually not working and can be removed anyway.


ckychris (dismissed):
A good amount of effort to keep us safe! 💯

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.18.0.cjs"
```
jasononeil, line 5: I'm pretty interested in knowing what the bug that required this was.

The pattern of committing the whole yarn program to source control still feels very weird to me but I guess it is the advised way to do it, and we've not had issues with the approach in other repos (like `employee-imports`). 

I'm just curious about the bug.

packages/notification/src/components/GenericNotification.spec.tsx
```diff
@@ -27,7 +27,7 @@ test('Begins "hidden" but transitions out of it immediately', async () => {
 
 test("The cancel button hides the notification and triggers the onHide callback", async () => {
   const onHide = jest.fn()
-  const { container, getByTitle } = render(
+  const { container, getByTestId } = render(
```
jasononeil, line 30: Any idea why this broke, and if it will affect anything beyond these test cases?

(If the problem is with a change in react-testing-library, I guess we can assume its local to this repo. If the problem is with a change in behaviour to our component it might be more concerning)

jasononeil: The fact that at this commit you haven't even changed any dependencies is particularly weird to me.

packages/design-tokens/docs/StoryBoard.stories.tsx
```diff
@@ -253,14 +253,15 @@ const StoriesContainer = (props: {
           {Object.entries(props.storyModule).map(([k, V]) => {
             // Ignore exports that are Story Metadata (which we can detect by the presence of a "title" prop)
             if ("title" in V) return null
-            const parameters = "story" in V ? V.story.parameters : V.parameters
+            const parameters = "story" in V ? V.story!.parameters : V.parameters
             const storyElement = (
-              <V {...meta.args} {...("story" in V ? V.story.args : {})} />
+              // @ts-ignore
```
jasononeil, line 258: Can we use ts-expect-error instead?

js-audit-allow-list.json
```diff
@@ -0,0 +1,129 @@
...
+      "reason": "Dev dependency",
+      "reviewer": "christianklammer",
+      "whenSubDependencyOf": ["gatsby", "gatsby-source-filesystem"]
+    }
+  ],
+  "url-parse": [
+    {
+      "title": "URL Redirection to Untrusted Site",
+      "cve": "CVE-2021-3664",
+      "reason": "Dev dependency",
```
jasononeil, line 24: Do you know if this dependency is used in the generated site or not?

Given the nature of Kaizen (static site, separate domain, no user data) this feels sufficiently low risk to me but it's not clear that Gatsby is always a "devDependency"

jasononeil (approved):
Given the low risk nature of the Kaizen site I'm comfortable with this, though I think Gatsby is more than a devDependency (its more of a framework and includes runtime code) and if we were using it on a site with user data I'd want to be more careful before adding items to the allowList.

This whole exercise makes me want to move off Gatsby all the more.

.yarnrc
christian314159: I just removed this again to see if it works in CI. iirc It was an issue with `yarn-audit-fix` where it was unable to resolve something in anything other than 1.18.0 and lower.

packages/notification/src/components/GenericNotification.spec.tsx
christian314159: I had a couple of flakey tests locally - that all always worked in CI :( This might have just been one that only failed locally, but would have worked in CI without my changes.

packages/design-tokens/docs/StoryBoard.stories.tsx
christian314159: Sure.

js-audit-allow-list.json
christian314159: Yes, that's definitely a problem. My "hope" is, as we're using Gatsby as a static site generator, little to none of its own code is actually bundled in with the generated code.
The `url-parse` package is part of a websocket package which I think is just used for local dev. Downloaded the Gatsby repo and also had a look at the build artifact.

ckychris (dismissed):

packages/notification/src/components/GenericNotification.spec.tsx
dougmacknz: I've noticed this failing across everyone's local environments, thanks for fixing :)

dougmacknz (dismissed):

jasononeil (approved):
Sorry it took so long to review this. Looks good :+1:

dougmacknz (approved):

PR #2376, opened 2021-12-21
feat: Delete deprecated OffCanvas and NavigationBar components from component-library

dougmacknz:
BREAKING CHANGE: There are a few places (mainly in Murmur) that are importing the NavigationBar stylesheet `~@kaizen/component-library/components/NavigationBar/styles` for a `$ca-navigation-bar__height` variable. This can be replaced with the `$layout-navigation-bar-height` from `~@kaizen/design-tokens/sass/layout`.

The only usage of the NavigationBar component itself is on the Kaizen site, so that will need to update before this package can update there.

# Purpose

* Removes 2 deprecated components from `@kaizen/component-library`:
  * NavigationBar
  * OffCanvas (was only used in the NavigationBar component).
* Removes a bunch of usused dependencies from component-library

# Context
@sugendran reported that `@kaizen/draft-button` depends on `@kaizen/component-library` which depends on `@kaizen/draft-button`, creating a circular dependency. We can break this by removing the `draft-button` dependency in component-library by removing these 2 components.


draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -97,7 +96,7 @@ class GuidanceBlock extends React.Component<
     actions: GuidanceBlockActions,
     withActionButtonArrow?: boolean
   ) => (
-    <Media query={MOBILE_QUERY}>
+    <Media query="(max-width: 767px)">
```
christian314159, line 99: Is it worth leaving a `TODO` comment here to import this media query from a centrally managed point again later?

christian314159 (approved):

dougmacknz:
Superseded by https://github.com/cultureamp/kaizen-design-system/pull/2383

PR #2377, opened 2021-12-22
feat: Add AvatarGroup component to Kaizen Library

mcwinter07:
# Objective
To add a new AvatarGroup component to the Kiazen design system and reach parity with the UI tool kit

# Motivation and Context
This was created off the back of the revision of the Avatar components and designed to standardise the way avatar stack when grouped. 

Screen Shot 2021-12-22 at 10 54 45 am

Screen Shot 2021-12-22 at 10 54 55 am

# Checklist

# Additional Considerations
This is a good question - Would this benefit from having an inverted layout for RTL? and how should we show support from internationalisation re: the +3 counter token


draft-packages/avatar-group/KaizenDraft/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const AvatarGroup = ({
+  size = "medium",
+  maxVisible = 2,
+  avatars = [],
+}: AvatarGroupProps) => (
+  <ul
+    className={cx(styles.AvatarGroup, styles[size])}
+    aria-label="Avatar Group"
+  >
+    {renderAvatars(avatars, size, maxVisible)}
```
NatalieLoyzaga, line 74: Should there be a check to see whether avatar is not empty and then do the two render functions? It currently renders and empty `<ul>` list when avatar prop is empty. 

..I thought there might be a way to check for an non-empty array prop in typescript...I just did a quick google and found this??

```
type NonEmptyArray<T> = [T, ...T[]];
avatars: NonEmptyArray<AvatarProps>
```

draft-packages/avatar-group/KaizenDraft/AvatarGroup/AvatarGroup.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+    render(<AvatarGroup avatars={imgAvatars} maxVisible={4} />)
+    expect(
+      screen.getAllByRole("img", { name: "Adirana Aniseed" })
+    ).toHaveLength(4)
+  })
+  it("Overides the size of individual Avatars with the size set on AvatarGroup", () => {
+    const { container } = render(
+      <AvatarGroup avatars={largeAvatars} maxVisible={4} size="small" />
+    )
+    const avatarsWithLargeClass = container.querySelectorAll(".large").length
```
NatalieLoyzaga, line 58: Maybe in extension to this overriding test,  check that .small has been added as well and .large removed

draft-packages/avatar-group/KaizenDraft/AvatarGroup/styles.module.scss
```diff
@@ -0,0 +1,102 @@
...
+    height: $avatar-md;
+    font-size: 16px;
+    font-weight: $typography-heading-5-font-weight;
+    letter-spacing: $typography-heading-5-letter-spacing;
+  }
+
+  .large & {
+    width: $avatar-lg;
+    height: $avatar-lg;
+    font-size: 22px;
```
NatalieLoyzaga, line 81: I think this can be changed to use a font variable - $typography-heading-3-font-size

draft-packages/avatar-group/KaizenDraft/AvatarGroup/styles.module.scss
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  .xxlarge & {
+    width: $avatar-xxl;
+    height: $avatar-xxl;
+  }
+
+  .xlarge &,
+  .xxlarge & {
+    font-size: 34px;
```
NatalieLoyzaga, line 98: Same as above - $typography-heading-1-font-size

draft-packages/avatar-group/KaizenDraft/AvatarGroup/AvatarGroup.tsx
NatalieLoyzaga: Or even if you just guard it with a check before it renders another test would be good.

mcwinter07: Solid find, I went with this methodology to make it a little more readable without the generic:

avatars: [AvatarProps, ...AvatarProps[]]

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { v4 } from "uuid"
+import { Avatar, GenericAvatarProps, CompanyAvatarProps } from "./Avatar"
+import styles from "./AvatarGroup.module.scss"
+
+export type AvatarProps =
+  | Omit<GenericAvatarProps, "size">
+  | Omit<CompanyAvatarProps, "size">
+
+export type AvatarGroupSize = "small" | "medium" | "large"
+export type AvatarList = [AvatarProps, ...AvatarProps[]]
```
dougmacknz, line 12: Curious about this type, what does this do different from `AvatarProps[]`?

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
```diff
@@ -0,0 +1,109 @@
...
+  .medium & {
+    margin-left: calc(#{$avatar-md} / -2);
+  }
+
+  .large & {
+    margin-left: calc(#{$avatar-lg} / -2);
+  }
+
+  .small & [dir="rtl"] & {
+    margin-left: calc(#{$avatar-sm} / -2);
```
dougmacknz, line 31: Are these doing anything, they look the same as the LTR equivalents above?

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+.AvatarGroupItem {
+  &:first-of-type {
+    margin-left: 0;
+
+    .small & {
+      [dir="rtl"] & {
+        margin-left: calc(#{$avatar-sm} / -2);
```
dougmacknz, line 49: Again, this looks the same as what's set above?

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
```diff
@@ -0,0 +1,109 @@
...
+
+.AvatarCounter {
+  align-items: center;
+  background: $color-gray-300;
+  border: 3px solid $color-white;
+  border-radius: 100%;
+  box-sizing: border-box;
+  box-shadow: $shadow-small-box-shadow;
+  display: flex;
+  font-family: $typography-heading-1-font-family;
```
dougmacknz, line 75: This works fine for now, but for future proofing I would recommend specifying this for each size with the correct heading below instead, so that if we end up changing the heading-1 font, it doesn't effect the headings we've used here.

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
```diff
@@ -0,0 +1,109 @@
...
+    font-size: $typography-heading-5-font-size;
+    font-weight: $typography-heading-5-font-weight;
+    letter-spacing: $typography-heading-5-letter-spacing;
+  }
+
+  .large & {
+    width: $avatar-lg;
+    height: $avatar-lg;
+    font-size: $typography-heading-3-font-size;
+    font-weight: $typography-heading-1-font-weight;
```
dougmacknz, line 101: Can `$typography-heading-3-font-weight` be used here instead?

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
```diff
@@ -0,0 +1,109 @@
...
+    letter-spacing: $typography-heading-3-letter-spacing;
+  }
+
+  [dir="rtl"] & {
+    margin-left: 0;
+    direction: ltr;
```
dougmacknz, line 107: I take it the plus is meant to stay on the left for RTL?

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+   * Note that 'size' is ommited from the `AvatarProps` type so it will throw a type error if size is provided.
+   * */
+  avatars: AvatarList
+}
+
+const renderCounter = (remainingAvatars: number) => {
+  if (remainingAvatars <= 0) return
+  return (
+    <li
+      aria-label={`There is ${remainingAvatars} other memeber${
```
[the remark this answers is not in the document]
mcwinter07: Good idea, I'll bundle that into a ternary to handle that :)

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+        remainingAvatars > 1 && "s"
+      } of this group`}
+    >
+      <span
+        className={styles.AvatarCounter}
+        aria-hidden={true}
+        role="presentation"
```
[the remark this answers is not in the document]
mcwinter07: Nice catch, I think I might have started with one and updated it to aria-hidden instead but forgot to remove it

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
mcwinter07: This is a way to guard against a non empty array - it'll throw a ts error if an empty array is provided, where as AvatarProps[] won't. Not sure if it's overkill but good to know it exists

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
mcwinter07: I went back and forth over this one, I know with RTL layout, when a number is a negative that could be flipped onto the other side, but positive value usually just removes the symbol.

new Intl.NumberFormat("ar-EG").format(-1234);
would produce: ؜-١٬٢٣٤
but for a plus: ؜ ١٬٢٣٤

Since this isn't actually representative of a positive / negative relationship and is more a graphical element I opted to leave it as LTR.

I could potentially write up a function that runs the number through a formatter and let the output get handled that way but I think if that's the case it should be a new ticket branched off this since it may be a bit of a rabbit whole trying to get the formatting right.

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  avatars: AvatarList,
+  maxVisible: number,
+  size: AvatarGroupSize
+) => (
+  <>
+    {avatars?.map(
+      (avatarProps, index) =>
+        index < maxVisible && (
+          <li
+            key={`avatar-${v4()}`}
```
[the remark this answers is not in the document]
mcwinter07: Thanks for tip, will strip out the uuid and use the index

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
mcwinter07: Absolutely right, it's redundant. I think that was left over from trying to square away LTR styles

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
mcwinter07: I've been able to remove the previous but not with these. Ideally I'd use something like this (see below) to remove the need the double up but I haven't found a way to use :not with the rtl parent wrapper successfully.

:not([dir="rtl"]) & {
  &:first-of-type {
    margin-left: 0;
  }
}

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
mcwinter07: fair call, will update that to ensure each one is declared separately 👍

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+            className={styles.AvatarGroupItem}
+            role="presentation"
```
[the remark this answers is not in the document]
mcwinter07: I think this role="presentation" was left over from when I was experimenting with what a screen readers would read aloud when iterating over this loop to see if I could get a fluid read of all members. It shouldn't really be here since it detract from what a list should be read as so I'm going to remove it.

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.tsx
dougmacknz: Oh nice, I learned something :)

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
dougmacknz: I'm not yet understanding what's going on here, but I wonder if [Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) could help you here at all? i.e. using `margin-inline-start` instead of `margin-left`.

dougmacknz (approved):

draft-packages/avatar/KaizenDraft/Avatar/AvatarGroup.module.scss
mcwinter07: I've remove this and added the margin inline styling to the avatar group list. There's still a little bit of duplication but it drops the need to do the deep nesting and I think it's cleaner to read.

dougmacknz: Nice 😎

dougmacknz (approved):

PR #2380, opened 2021-12-24
feat: add isElevated prop to Card component wrapper and support for new variants backgrounds

mcwinter07:
# Objective
Add new card variants and changes the storybook story structure to reach closer parity with the UI toolkit. 

# Motivation and Context
- add missing variants
- add isElevated prop with default to ensure update is not a breaking change
- update documentation and exposed props in the Kaizen Demo

# Screenshots

Screen Shot 2021-12-24 at 4 45 17 pm

Screen Shot 2021-12-24 at 4 29 40 pm

Screen Shot 2021-12-24 at 4 29 35 pm


jmitsios:
@mcwinter07 Do you know where those 'Reversed' color palettes came from?  They don't seem to be in the UI Kit, and may need to be added in.

lloydstubber:
Do we need to update the Kaizen site to reflect these changes? It seems that the "Card" page is pretty out of date and there's no "Container" page AFAIK. Probably worth creating another card for that or just flagging with @katsambarcus for future ref.

lloydstubber:
Code looks good. Just will wait for the answer on the colours before ticking 👍

mcwinter07:
> 

@jmitsios Reversed colour palettes are the same just on a reverse background. Not sure if they're needed or not? I can remove the story if there isn't an instance of that

lloydstubber (approved):
LGTM

draft-packages/card/KaizenDraft/Card/Card.tsx
```diff
@@ -1,32 +1,58 @@
 import * as React from "react"
+import cx from "classnames"
 import styles from "./styles.module.scss"
 
+export type CardVariants =
```
NatalieLoyzaga, line 5: Update to use `mood` prop name rather than variant.

NatalieLoyzaga: Just something I've learnt recently that I don't think was clear initially

mcwinter07: I might split this into a seperate small task since we should also change the name of `variant` prop, which would cause a breaking change :)

PR #2382, opened 2022-01-04
feat: Create eslint custom rule to help with breaking changes

ckychris:
https://cultureamp.atlassian.net/browse/KDS-104

Create eslint custom rule to help with breaking changes (component-library)

- How does the consumer knows which version of eslint-plugin they should be using? (Adding peer dependencies?)
- `eslint-plugin` was not mentioned in the readme file for breaking changes upgrades


packages/eslint-plugin/src/rules/no-inline-notification-imports-from-component-library.ts
```diff
@@ -0,0 +1,251 @@
...
+
+// TODO - transform this rule into fixing all legacy packages, not just InlineNotification
+// TODO - kaizen version awareness
+
+module.exports = {
+  meta: {
+    type: "problem",
+    docs: {
+      description:
+        "Fix breaking changes from @kaizen >=11.0.0. InlineNotification component moved from @kaizen/component-library to @kaizen/notification. See more at https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/CHANGELOG.md#1100-2021-08-30",
```
dougmacknz, line 12: How hard would it be to expand this to cover the other notification components as well?
* GlobalNotification
* ToastNotification
* addToastNotification
* clearToastNotifications
* removeToastNotification

ckychris: Should be easy, just need sometime to make it generic, seems like we need to address those TODO comments from the get go!

packages/eslint-plugin/README.md
```diff
@@ -19,11 +19,13 @@ Enable the rules you want to use within rules property of your .eslintrc:
 {
   rules: {
     "@kaizen/no-draft-imports": "error",
+    "@kaizen/no-inline-notification-imports-from-component-library": "error"
```
jasononeil, line 22: For the stylelint plugin we ended up defining a simple preset that included all the recommended rules so that the consumers weren't expected to constantly be tweaking their Kaizen eslint rules every time we write a new one. (See readme: https://github.com/cultureamp/kaizen-design-system/blob/master/packages/stylelint-plugin/README.md#getting-started) 

I feel like it's probably worth us setting up something similar here so we have control over which rules are enabled and disabled by default.

jasononeil:
Closing this PR as we think a codemod solution will be more appropriate.

Feel free to reopen if that decision ever changes.

PR #2383, opened 2022-01-05
feat: remove navbar from site and fix storybook, fix SCSS

ckychris:


site/src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    const Badge = badges[environment] || namedBadge(environment)
+
+    return (
+      <Badge loading={loading} href={badgeHref} colorScheme={colorScheme} />
+    )
+  }
+}
+
+export default withDeprecatedComponent(NavigationBar, {
```
dougmacknz, line 144: I'd get rid of this `withDeprecatedComponent` HOC wrapper now, it's not relevant here and will just pollute the logs.

site/src/components/OffCanvas/OffCanvas.tsx
```diff
@@ -0,0 +1,136 @@
...
+              </button>
+              <Component {...this.props} />
+            </React.Fragment>
+          )}
+        </OffCanvasContext.Consumer>
+      )
+    }
+  }
+
+export default withDeprecatedComponent(
```
dougmacknz, line 131: Same here for removing this HOC wrapper.

storybook/header-preset/CopyAndPastedNavBar.md
```diff
@@ -0,0 +1,2 @@
+05/01/2022 - Chris Chan
+We would love to remove NavBar from Kaizen, the NavigationBar and OffCanvas components should be refactored and deal with when possible. It's currently being copied to both /site and /storybook
```
dougmacknz, line 2: This is no longer relevant since you found a way to avoid copying it to storybook as well

site/src/components/NavigationBar/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default as NavigationBar } from "./NavigationBar"
+export { default as Link } from "./components/Link"
+export { default as Menu } from "./components/Menu"
```
dougmacknz, line 3: The Kaizen site doesn't make use of `Menu` at all, so there's heaps of dead code that you can delete in this tree.

site/src/components/CopyAndPastedNavBar.md
```diff
@@ -0,0 +1,2 @@
+05/01/2022 - Chris Chan
+We would love to remove NavBar from Kaizen, the NavigationBar and OffCanvas components should be refactored and deal with when possible. It's currently being copied to both /site and /storybook
```
dougmacknz, line 2: Creating these kind of files isn't really a convention that we follow. What I would recommend is putting a comment on the `NavigationBar.tsx` file to explain that it was copy pasted over from an old Kaizen component.

site/src/components/NavigationBar/index.ts
dougmacknz: I'd also recommend doing a sweep of props and seeing what else can be deleted, including CSS. E.g. It doesn't look like the Kaizen site nav uses badges at all, so we can remove everything related to those.

site/src/components/NavigationBar/NavigationBar.tsx
ckychris: I will try to remove any application logic that is not related to the Kaizen site directly, as we know that the Culture amp apps has their own composition, where Kaizen will not provide the navigationBar.

site/src/components/CopyAndPastedNavBar.md
ckychris: Yeah, that make sense, I was trying to get a comment from you 😃

site/src/components/NavigationBar/index.ts
ckychris: Just did that, not able to completely remove badge CSS since the mobile menu is still using it, but it's now fixed to the default icon. Removed as much props as I can at first glance, without deep refactoring.

dougmacknz (commented):
Nice, this is looking about to good to go for me. There's just an error regarding uuid not being in the dependencies to fix.

I'd personally merge this straight to master rather than into my PR, but I'm happy to do it this way if you'd be happy to review my PR.

ckychris:
> Nice, this is looking about to good to go for me. There's just an error regarding uuid not being in the dependencies to fix.
> 
> I'd personally merge this straight to master rather than into my PR, but I'm happy to do it this way if you'd be happy to review my PR.

Just fixed the error and updated the branch, if you can approve and merge that will be great 😌 . (Target is now master)

dougmacknz (approved):

PR #2389, opened 2022-01-07
feat: tooltip placement for scroll and left/right

NatalieLoyzaga:
# Objective

- Add left and right placement variant. 
- Allow the popper to 'flip' depending on the space within the viewport -> see Scroll story example.

# Motivation and Context

- Match the UI kit.

image
image

### DIFF
--- a/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
+++ b/draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
@@ -19,7 +19,7 @@ $arrow-width: 14px;
   border-radius: $border-solid-border-radius;
   transition: opacity $animation-duration-fast,
     transform $animation-duration-fast;
-  padding: 10px 10px;
+  padding: 9px 9px;
   box-shadow: $shadow-small-box-shadow;
   text-align: center;
   max-width: 400px;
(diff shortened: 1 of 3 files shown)


jmitsios:
Awesome stuff Nat :) 

Could I suggest we reduce the distance between the tooltip and the asset to **6px** please?   I believe the spacing in the UI Kit is **12px**, but it's measuring it from the base of the _container_ (bubble) rather than the _tip_.

This may also help in the [Sticker Sheet](https://dev.cultureamp.design/KDS-70/Tooltip-placement/storybook/?path=/story/components-tooltip--sticker-sheet) layout, which currently has all the tooltips a touch too close to all the surrounding buttons.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.scss
```diff
@@ -83,6 +83,19 @@ $arrow-width: 20px;
...
     }
   }
+
+  [data-popper-placement^="left"] & {
+    right: 0;
+    .arrowInner {
+      transform: rotate(270deg);
+    }
+  }
+  [data-popper-placement^="right"] & {
```
mcwinter07, line 93: I didn't know about the `^=` selector, very handy

mcwinter07 (commented):
Looks pretty good to me. Just a quick query about `rtl` and `ltr` styling - it's to affect the placement but only on the sticker sheet - just want to make sure that's either a weird storybook outlier or intended?

mcwinter07 (approved):
I like the update to the sticker sheet 👍  Good to merge

PR #2401, opened 2022-01-14
feat: Guidance block container query

hqta:
# Objective
- Changes allow the guidance block to follow [Tile](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=1929%3A33867) design responsive behaviour based on the size of the guidance block instead of the screen size. 
- Utilizes `ResizeObserver` to add necessary css class names to allow us to modify the layout. 
- Additional RTL support to finesse spacings
- Improve button styling on mobile making them fullwidth per [Tile](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=1929%3A33867) designs
- Ensure text does not occupy widths less than `320px`
- Allow the secondary button to dismiss and hide the Guidance block
- Add `smallScreenTextAlignment` prop to left align text on mobile layout (center by default)

# Motivation and Context
- Changes were made as Guidance block side by side would behaves differently to a fullwidth Guidance block responsively. This allows the Guidance block to have consistent responsive behaviour.
- RTL spacing previously were not consistent
- Team Unify requires secondary button to allow the Guidance block to be dismissed and hidden on click.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
+++ b/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
@@ -180,17 +180,13 @@ $scene-illustration-size-mobile: 200px;
   align-items: unset;
   gap: $spacing-md;
   min-width: unset;
+  flex-wrap: wrap;
+
+  @include ca-media-tablet-and-under {
+    flex-direction: unset;
+  }
 
   &.stackButton {
-    .descriptionAndActions {
-      flex-direction: column;
-      align-items: unset;
-    }
-
-    .illustrationWrapper {
-      display: flex;
-    }
-
     .descriptionContainer {
       margin: 0 0 0 $spacing-sm;
       padding: 0;
@@ -207,12 +203,6 @@ $scene-illustration-size-mobile: 200px;
   }
 
   &.stackIllustration {
-    flex-direction: column;
-
-    .illustrationWrapper {
-      align-self: unset;
-    }
-
     .illustration {
       margin-right: auto;
 
@@ -242,6 +232,8 @@ $scene-illustration-size-mobile: 200px;
   }
 
   &.centerContent {
+    flex-direction: column;
+
     .illustrationWrapper {
       align-self: center;
     }
@@ -256,6 +248,7 @@ $scene-illustration-size-mobile: 200px;
 
     .descriptionContainer {
       text-align: center;
+      min-width: unset;
 
       [dir="rtl"] & {
         text-align: center;
@@ -301,6 +294,7 @@ $scene-illustration-size-mobile: 200px;
   .descriptionContainer {
     text-align: left;
     max-width: unset;
+    min-width: 320px;
 
     [dir="rtl"] & {
       text-align: right;
@@ -310,6 +304,10 @@ $scene-illustration-size-mobile: 200px;
   .buttonContainer {
     justify-content: end;
     width: unset;
+
+    @include ca-media-mobile {
+      flex-direction: row-reverse;
+    }
 
     > * {
       width: unset;
--- a/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
+++ b/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
@@ -316,6 +316,10 @@ $scene-illustration-size-mobile: 200px;
 }
 
 .stacked {
+  &.stackIllustration .illustrationWrapper {
+    margin: 0 auto;
+  }
+
   .descriptionAndActions {
     flex-direction: column;
     align-items: unset;
--- a/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
+++ b/draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
@@ -14,7 +14,6 @@ $ca-banner-collapse-height-delayed: margin-top $animation-duration-fast
   $animation-duration-slow ease;
 $illustration-size: 155px;
 $scene-illustration-size: 300px;
-$scene-illustration-size-mobile: 200px;
 
 .banner.noMaxWidth {
   max-width: inherit;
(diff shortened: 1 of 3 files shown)


lloydstubber:
👋  @jakubriedl, @hqta mentioned you weren't too keen on the resizeObserver solution, could outline your concerns when you have a chance? Cheers!

jakubriedl:
yeah we talked about it, the positioning could be done by calc but we couldn't find a easy css only option for changing the text-align on small container so we closed it with conclusion that it is too much effort to change that resizeObserver is good enough

jmitsios (dismissed):
Really great stuff Harry

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
```diff
@@ -177,57 +177,127 @@ $scene-illustration-size-mobile: 200px;
...
 
-  @include ca-media-mobile {
-    flex-direction: column;
+  @include ca-media-tablet-and-under {
+    flex-direction: unset;
   }
 
-  .illustrationWrapper {
-    padding: 0;
+  &.stackButton {
```
HeartSquared, line 189: Is the extra nesting necessary? CSS modules will actually ensure your classes differ between scss files (eg. `.stacked` here is not compiled to be the same as `.stacked` elsewhere). If you need the new class, it does not necessarily need to live within a parent.

Also, at some point, if there's a significant level of variation (especially since this file seems very long), we might be better making them separate components to keep them cleaner (I don't have full context though, so I could be wrong).

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -44,7 +45,9 @@ export type GuidanceBlockProps = {
     title: string
     description: string | React.ReactNode
   }
+  mobileTextAlignment?: TextAlignment
```
HeartSquared, line 48: Is this really only applicable to mobile? If we just happen to have multiple thinner columns on a desktop, does that mean this will not have an effect? From looking at Figma I don't see that context being a rule (but if it is, then sure).

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
hqta: Ideally, we create the new Tile component however, given the time urgency to deliver the project we've collectively decided to leverage off the Guidance Block until the Tile component is created acknowledging the tech debt. At that point we'll switch out. 

Since the variations rely on many existing functionality in the Guidance block, it didn't make sense to create entirely new separate components for this. Hence, there's overrides and nested CSS to override the existing styles.

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
hqta: Yup. Based on the tile design we need to provide centre and left alignment specifically for mobile - this is a design requirement.

Screen Shot 2022-01-18 at 2 10 11 pm

 Since we're adjusting the layout/styling based on the guidance block size as opposed to media queries, when the block becomes thin enough it will have the desired effect.

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
HeartSquared: You should still be able to create the variations without nesting usually :)
Sure if there has been discussion around the approach and it's urgent then it's understandable. Since I wasn't part of the discussion earlier, was the conclusion that this component will be deprecated and split into different components based on usage? Or perhaps just refactored since it looks somewhat old (class based as opposed to a function component).

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
HeartSquared: That...still sounds like it's not exclusively for mobile then? It's just for when the block is a smaller size? Or am I misunderstanding something?

hqta:
@jmitsios please do a final design review - I've just pushed up changes to simplify the implementation.

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.scss
hqta: Pushed up a change reducing a bunch of classnames and simplify to only use the `centerContent` for adjusting content when the card is due to show the centered content.

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
hqta: This is when the card reaches a size that's small enough for it to use the designated "mobile" design from the Tile design hence the naming.

lloydstubber: I think @HeartSquared is right here. Our language tries not to be devices specific and more based on the size of the viewport - although the old mixins are like that, it's hard not to follow suit. 

Ideally we'd use the word `small` as laid out in the UI Kit: https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=23035%3A118325

HeartSquared:
Jumped on a call last night with @hqta. From what I understood, this component has been agreed on to be deprecated in the near future, and as it is blocking progress for Team Unify, I am not too fussed on optimising the code as long as it works.
However, since I'm still new to the team, it is still probably best if someone else makes the approval :)

jmitsios (commented):
@hqta – I've found the Spot illustration and Scene illustrations disappearing once my viewport reaches a small size? (See screenshots)

image
image
image

hqta:
> @hqta – I've found the Spot illustration and Scene illustrations disappearing once my viewport reaches a small size? (See screenshots)

I believe this is the original functionality of the Guidance block. This should apply to all guidance block that are not the `inline` and `stacked` variants I've added.

hqta:
@jmitsios just checked out a commit prior to my first changes for guidance block and looks like the illustration is meant to hide on small viewport sizes. So this is expected :)

jmitsios (approved):

lloydstubber (approved):

PR #2402, opened 2022-01-17
feat: Add variants and update stories for Popover

NatalieLoyzaga:
# Objective

- Update Popover to include left and right Placement variant. 
- Allow for flip and fallback placements
- Update stories to show off clickable and hoverable story - Due to the makeup of this component being different to Tooltip, I decide to rather than re-invent the wheel but to illustration on the Kaizen Story of how you would use this component and make it clickable or hoverable.
- Include a 'Sticker sheet' story.

# Motivation and Context

- After the work on Tooltips we decided Popover was lacking variants and functionality.
- We have updated both the component but the UI Kit

image

image

image

### DIFF
--- a/draft-packages/popover/KaizenDraft/Popover/styles.scss
+++ b/draft-packages/popover/KaizenDraft/Popover/styles.scss
@@ -161,7 +161,7 @@ $negative-box-border-color: $color-red-300;
     color: $color-purple-800;
     outline: $color-blue-500 $border-focus-ring-border-style
       $border-focus-ring-border-width;
-    outline-offset: calc($border-focus-ring-border-width * -1);
+    outline-offset: calc(-1 * #{$border-focus-ring-border-width});
     border-radius: $border-borderless-border-radius;
   }
 }
--- a/draft-packages/popover/KaizenDraft/Popover/styles.scss
+++ b/draft-packages/popover/KaizenDraft/Popover/styles.scss
@@ -3,7 +3,7 @@
 @import "~@kaizen/design-tokens/sass/spacing";
 @import "~@kaizen/design-tokens/sass/typography";
 @import "~@kaizen/draft-button/KaizenDraft/Button/styles";
-
+@import "~@kaizen/deprecated-component-library-helpers/styles/type";
 // Sync with PopoverModern.tsx
 $arrow-width: 14px;
 $arrow-height: 7px;
@@ -109,12 +109,8 @@ $negative-box-border-color: $color-red-300;
 }
 
 .header {
-  font-family: $typography-paragraph-body-font-family;
-  font-size: $typography-heading-6-font-size;
-  font-weight: $typography-heading-6-font-weight;
-  letter-spacing: $typography-heading-6-letter-spacing;
-  line-height: $typography-paragraph-small-line-height;
   position: static;
+  @include kz-typography-heading-6;
   margin-bottom: $spacing-xs;
   display: flex;
   align-items: center;
@@ -124,11 +120,7 @@ $negative-box-border-color: $color-red-300;
 
 .container {
   position: static;
-  font-family: $typography-paragraph-body-font-family;
-  font-size: $typography-heading-6-font-size;
-  font-weight: $typography-paragraph-small-font-weight;
-  letter-spacing: $typography-paragraph-small-letter-spacing;
-  line-height: $typography-paragraph-small-line-height;
+  @include kz-typography-paragraph-small;
   white-space: pre-line;
 }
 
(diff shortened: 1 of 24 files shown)


jmitsios:
@NatalieLoyzaga - Few small things I noticed:
_UI Kit design is on the right_
image image
- Needs to tighten up the spacing between the header and the body text.
- What shadow size is being used?
- Space along the bottom is larger than the other sides, so we need to keep that consistent padding around all edges of the container.
- Could we see an example also with Icons next to the header please for the different moods?

jmitsios:
@NatalieLoyzaga 
- Looks to be missing the hover-state on the [x] close icon.
- _(Question for my knowledge)_ does the 'optional link' on Storybook have a way of inheriting our platforms link style + color?  Or does it just sit as the default browser hyperlink style?
- Sticker sheet page
  - is the only way to the popovers by clicking the 'Open all popovers' button? Could we also have them interactive/clickable individually?
  - Could we change the headers of each popover to match the mood (i.e: "Cautionary" instead of Header, etc..)

NatalieLoyzaga:
@jmitsios To respond to your last comment

- I have updated the button to have hover and focus state as per the UI kit.
- Sticker sheet...The way the component works which is different to Tooltip makes the sticker sheet story extremely convoluted. For every popover example it required its own component and it's own hook (controls). To be able to add the onClick functionality to each of them would then require additional controls to each of the individual hooks...I personally think it's a lot of extra code and rather the single button that controls all is a work around. ( of course if someone has a more autonomous idea/way of doing this I can explore )
- I have updated the headings.

draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -49,12 +45,13 @@ $large-width: 450px;
...
 
 %box {
-  background: white;
+  background: $color-white;
   border: $border-solid-border-width $border-solid-border-style $color-gray-300;
   filter: drop-shadow(0 0 7px rgba(0, 0, 0, 0.1));
   border-radius: $border-solid-border-radius;
   color: $color-purple-800;
   text-align: left;
+  padding: $spacing-sm 1rem 1rem;
```
lloydstubber, line 54: nit: I don't like how the UI Kit is using spacing that isn't in our variables 😬

draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -112,25 +109,32 @@ $negative-box-border-color: $color-red-300;
...
 .container {
-  @include ca-inherit-baseline;
-  @include kz-typography-paragraph-small;
-  padding: ($ca-grid / 6) ($ca-grid / 2) ($ca-grid);
+  position: static;
+  font-family: $typography-paragraph-body-font-family;
+  font-size: $typography-heading-6-font-size;
+  font-weight: $typography-paragraph-small-font-weight;
+  letter-spacing: $typography-paragraph-small-letter-spacing;
+  line-height: $typography-paragraph-small-line-height;
```
lloydstubber, line 131: Our type styles should be the one variable across the whole declaration. Guessing that first one is all `heading-6` and the second is all `paragraph-small`. If it looks incorrect in the browser we'll need to flag this in the UI Kit.

lloydstubber:
Screen Shot 2022-01-21 at 10 58 09 am

Screen Shot 2022-01-21 at 10 58 40 am

Here's your focus state vs the UI Kit. Needs to be a bit tighter around the icon.

draft-packages/popover/KaizenDraft/Popover/styles.scss
NatalieLoyzaga: The spacing is 12px in the UI kit but because of the line-height of the heading and body text, adding this 1rem padding around the whole container was the workaround to still getting a total spacing of 12px rather than adding additional padding in multiple spots. If that makes any sense.

draft-packages/popover/KaizenDraft/Popover/styles.scss
```diff
@@ -170,6 +161,7 @@ $negative-box-border-color: $color-red-300;
     color: $color-purple-800;
     outline: $color-blue-500 $border-focus-ring-border-style
       $border-focus-ring-border-width;
+    outline-offset: calc($border-focus-ring-border-width * -1);
```
lloydstubber, line 164: I can't seem to see this style in the browser... Is it working when you preview it from the Storybook link? @NatalieLoyzaga

NatalieLoyzaga: @lloydstubber I think because buildkite is failing the storybook preview hasn't updated...grr.

lloydstubber: Ahh true. Want to just throw a screenshot from your local dev here? Then I can give it a tick

NatalieLoyzaga: Yep done. Sorry should of done that anyways.

lloydstubber:
LGTM! Just let me know when it's green and I'll tick it

lloydstubber (approved):
Nice one! Got there in the end

PR #2412, opened 2022-01-23
feat: Bump design-tokens in peer dependencies to include v6

NatalieLoyzaga:
# Objective

 Bump design-tokens in peer dependencies to include v6

# Motivation and Context

Warnings for incorrect peer dependencies for design tokens when running yarn install.

### DIFF
--- a/draft-packages/avatar/package.json
+++ b/draft-packages/avatar/package.json
@@ -38,7 +38,7 @@
     "react-textfit": "^1.1.1"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^6.1.0",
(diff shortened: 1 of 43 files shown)


draft-packages/avatar/package.json
```diff
@@ -38,7 +38,7 @@
     "react-textfit": "^1.1.1"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0",
+    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^6.1.0",
```
dougmacknz, line 41: I think this could be `^6.0.0` rather than `^6.1.0`?

Also might be worth adding `5.0.0` to these as well

NatalieLoyzaga: That was a speedy review 😄 

I will update them all to be `^6.0.0` and include `5.0.0`

dougmacknz (approved):

PR #2415, opened 2022-01-24
fix: update all division in sass to use calc or built in math.div

lloydstubber:
# Objective
Update all usages of `number / number` to now use `calc(number / number)` or `math.div(number, number)`

### Note
One of these is left, as it's in a library we're using. I've subbed to the notifications so we can bump once they update it: https://github.com/foundation/motion-ui/issues/139

# Motivation and Context
We've bumped node/sass and now have this warning in the console.
image

# Checklist

# Additional Considerations


draft-packages/tag/KaizenDraft/Tag/Tag.scss
```diff
@@ -247,8 +247,8 @@ $pulse-size-after: $pulse-size-initial * $pulse-scale-factor;
   width: $pulse-size-after;
   height: $pulse-size-after;
   // this positioning has to be calculated as an exact value as scaling will mess with percentage values
-  left: -($pulse-size-after / 2) + ($pulse-size-initial / 2);
-  top: -($pulse-size-after / 2) + ($pulse-size-initial / 2);
+  left: calc($pulse-size-after / -2) + calc($pulse-size-initial / 2);
```
lloydstubber, line 250: Note: I had to swap the negative value to the non-variable value

packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -1,3 +1,8 @@
+// This is the only way to handle division for the strip-unit function.
+// calc() unfortunately doesn't work in this case.
+// Example: https://sass-lang.com/documentation/breaking-changes/slash-div#transition-period
+@use "sass:math";
+
```
lloydstubber, line 5: Seems like this is the way to solve the below as `calc()` doesn't work in the same way as `math.div`

ckychris:
This is some good effort!

draft-packages/badge/KaizenDraft/Badge/styles.module.scss
```diff
@@ -77,7 +79,7 @@ $small: $spacing-md;
   color: $color-white;
 }
 
-$badge-duration-timing: ($animation-duration-slow / 2)
+$badge-duration-timing: calc($animation-duration-slow / 2)
```
ActuallyACat, line 82:
```suggestion
$badge-duration-timing: calc(#{$animation-duration-slow}/ 2)
```
This is a CSS variable, so I believe you'll want to ensure it's interpolated.

packages/component-library/components/Dropdown/Dropdown.module.scss
```diff
@@ -6,7 +6,7 @@
 
 $width: 248px;
 $caButton-border-width: $border-solid-border-width;
-$caButton-verticalPadding: calc(#{$ca-grid / 2} - #{$caButton-border-width});
+$caButton-verticalPadding: #{calc($ca-grid / 2)} - #{$caButton-border-width};
```
ActuallyACat, line 9:
```suggestion
$caButton-verticalPadding: calc($ca-grid / 2) - #{$caButton-border-width};
```
This syntax might be confusing me, but I don't think you need to interpolate the whole equation. Unsure!

packages/deprecated-component-library-helpers/styles/type.scss
ActuallyACat: Super curious about learning more on this one! Is it because we're dividing two sass variables? It seems like the other divisions in these files are okay with just using calc.

ActuallyACat (commented):
Code wise there doesn't look to be anything out of the ordinary here. I have two small suggestions.

Let's do some visual regression tests before the final ✅ is given.

site/package.json
```diff
@@ -40,6 +40,7 @@
     "react-dom": "^16.9.0",
     "react-helmet": "5.2.1",
     "react-media": "^1.10.0",
+    "sass": "^1.49.0",
```
mbylstra, line 43: Upgrading to the latest version of the npm `sass` (which uses dart scss compiled to scss rather than "native" or "embedded" dart sass) is fine for compiling the Kaizen site because it's relatively small and the sass compile times aren't too bad.

However, If this PR is introducing a new scss feature that's not backwards compatible with the latest version of https://www.npmjs.com/package/node-sass (it stopped getting updates a few years ago), then this will cause a problem for our huge consuming codebases. Last time we tried to use dart scss (non embedded) in Big M for example, it took several minutes and the machine ran out of memory before it could even finish.

So the problem is that this would be blocking consuming repos from using these new versions of components until they have upgraded to dart-sass. Although a stable version of embedded/native dart sass was released very recently, I'm not sure that's everything - I think the webpack plugin also needs a release that has the option of using embedded sass (this may have already happened though). It just might not be that trivial and it's a bit unproven at this point.

On the other hand, it'd be a driving force for consumers to upgrade to dart sass. I'm not sure what's better or worse - I will have to leave your team to decide on that one, but I think it's worth consulting with the biggest consumers as to whether upgrading to dart-scss is something that's feasible for them.

PR #2416, opened 2022-01-24
feat: Force IconButton label and icon to be supplied

### LINKED ISSUE #1530: Consumers sometimes think that label is superfluous when using IconButton
I am seeing quite a few cases where people are using `<IconButton>` with an empty string for a label. I am hypothesizing  that people think it's pointless given it doesn't display visually, not realising the accessibility implications of providing no label.

I see a couple of things we could do to help this situation:
* ~Warn consumers in console when providing an empty string.~
* Rename the `label` prop to `accessibleLabel` (or similar) for `IconButton` to make it a bit more clear why you still need provide a label.

**UPDATE:** We've noticed that the similar console warnings for the Icon component haven't really worked in fixing the same issue over there (there are many warnings out there not being addressed). We need something better, and Jason has prototyped a way to force a non empty string below.

NatalieLoyzaga:
BREAKING CHANGE: IconButton will now generate a TypeScript error if you pass an empty string as the label or if you don't pass an icon.

# Objective

Adds type safety to the IconButton for the label prop. Ensures a non-empty string can not be passed in.

# Motivation and Context

closes #1530 

Note: we've not done the same thing for the `<Icon>` component. This will be a much more significant breaking change, and might be worth tackling alongside other breaking changes to `<Icon>`, like moving it into its own package.


NatalieLoyzaga:
@jasononeil Would you like me to remove the test for checking if an icon gets passed in?

jasononeil:
Yeah @dougmacknz mentioned that in the Slack thread, let's remove that test.

I would like to see the types for Button Props cleaned up and made easier to understand so mistakes like this are less easy to make, but that's a topic for another day.

If you don't mind removing it I think we can review this. The biggest thing to review I think is the fact that this is a breaking change to `@kaizen/draft-button` and we'll need to communicate that etc.

draft-packages/button/KaizenDraft/Button/IconButton.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from "react"
 import GenericButton, { IconButtonProps } from "./components/GenericButton"
 
-const IconButton: React.FunctionComponent<IconButtonProps> = (
-  props: IconButtonProps
-) => <GenericButton iconButton {...props} />
+const IconButton = <S extends string>(props: IconButtonProps<S>) => (
```
christian314159, line 4: I have no idea how this type declaration here works :( Maybe a Show & Tell topic?

draft-packages/button/KaizenDraft/Button/IconButton.spec.tsx
```diff
@@ -26,7 +34,15 @@ it("renders an accessible label when it's a custom component", () => {
...
       component={CustomComponent}
       label="Accessible label on the custom component version"
+      icon={configureIcon}
     />
   )
   screen.getByLabelText("Accessible label on the custom component version")
 })
+
+describe("type safety for icon button", () => {
+  it("generates a typescript error when passing an empty label", () => {
```
christian314159, line 44: 👏

christian314159 (approved):

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -70,8 +70,17 @@ type WorkingUndefinedProps = {
 export type LabelProps = LabelPropsGeneric &
   (WorkingProps | WorkingUndefinedProps)
 
-export type IconButtonProps = GenericProps &
+type NonEmptyString<T extends string> = T extends "" ? never : T
```
HeartSquared, line 73: Do we have somewhere we put shared types? I imagine this would be helpful in other components too :)

draft-packages/button/KaizenDraft/Button/components/GenericButton.tsx
```diff
@@ -70,8 +70,17 @@ type WorkingUndefinedProps = {
...
+
+type IconOverrideProps<S extends string> = {
+  label: NonEmptyString<S>
+  icon: React.SVGAttributes<SVGSymbolElement>
+}
+
+export type IconButtonProps<S extends string> = GenericProps &
+  IconOverrideProps<S> &
   (WorkingProps | WorkingUndefinedProps)
+
```
HeartSquared, line 83: I suggest if these props are being defined for `IconButton`, then these sit in the file with the component instead of in here

gyfchong:
Oh man this is a stale one haha.

I don't think this is exactly a breaking change though, the API remains the same, we're just stricter about the inputs? Happy for it to move forward as one, but as far as I'm aware, not everyone's builds fail on typescript errors cause it depends on how consumer's set things up.

gyfchong (approved):
Do we still need this given we're moving to "blessed" button?

Issue #2409, opened 2022-01-21
Radio Group should be marked up as a radiogroup and labelled


dougmacknz:
The markup that our `RadioGroup` renders looks like this:
![image](https://user-images.githubusercontent.com/1811583/150443108-7c54e7f5-9280-4b6d-a96c-dfd5726158e5.png)

There's a couple of improvements that should be made here for this to be more semantic, to convey that this is a group of controls, with a label.

Two options:
1. We use `fieldset` for the container and `legend` for the label.
  * This can be a bit annoying because of built in browser styles for these elements, coupled with the fact that this will almost definitely run into issues with perf-ui's materialize global styles.
2. We add `role=radiogroup` to the container, with `aria-labelledby` linking to the label (plus probably change the `label` tag to something else).

Despite the [first rule of aria](https://www.w3.org/TR/using-aria/#rule1): I'd recommend making an exception for this one and going for option 2.

References:
https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton
https://www.w3.org/WAI/tutorials/forms/grouping/

dougmacknz:
@lisatassone @HeartSquared @hqta If you end up finishing this quickly before the allotted time, `CheckboxGroup` could also use the same improvement, but just using `role=group` instead of `role=radiogroup`

jasononeil:
Fixed by #2426

PR #2426, opened 2022-01-25
feat: improve RadioGroup accessibility

lisatassone:
# Objective
To allow screen readers to connect the radio group to the label so it can be read out (also general accessibility).

# Motivation and Context
https://github.com/cultureamp/kaizen-design-system/issues/2409

This was originally discovered when trying to use RadioGroup to mark up a radio group in performance-ui. This change fixes the accessibility of the component, but it is still not able to be used in perform-ui. This is because the component enforces the use of the Label component that does not match what the perform product uses:

image

I raised a possible solution to allow the consumer to override the Label element by supplying a render function for it where a consumer could render custom UI with the appropriate ID, but that was deemed too flexible.

# Checklist

# Additional Considerations

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
@@ -1,7 +1,7 @@
 import { RadioField } from "@kaizen/draft-form"
 import { cleanup, render } from "@testing-library/react"
 import * as React from "react"
-import RadioGroup, { RadioGroupProps } from "./RadioGroup"
+import RadioGroup from "./RadioGroup"
 
 afterEach(cleanup)
 
@@ -63,6 +63,24 @@ describe("<RadioGroup /> ", () => {
         <RadioGroup labelText={title} children={null} />
       )
       expect(queryByText(title)).toBeTruthy()
+    })
+  })
+  describe("accessibilty", () => {
+    it("should markup aria label when supplied labelId", () => {
+      const labelId = "labelId"
+      const { getByText, getByRole } = render(
+        <RadioGroup
+          labelText="Label title"
+          labelId={labelId}
+          automationId="test"
+        />
+      )
+      expect(getByRole("radiogroup").getAttribute("aria-labelledby")).toEqual(
+        labelId
+      )
+      expect(
+        getByText("Label title").parentElement?.getAttribute("id")
+      ).toEqual(labelId)
     })
   })
 })
--- a/draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.tsx
+++ b/draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.tsx
@@ -8,6 +8,7 @@ import styles from "./styles.scss"
 export type RadioGroupProps = {
   automationId?: string
   labelText: string | React.ReactNode
+  labelId?: string
   noBottomMargin?: boolean
   reversed?: boolean
 }
@@ -18,6 +19,7 @@ const RadioGroup: RadioGroup = ({
   automationId = "",
   children,
   labelText = "",
+  labelId,
   noBottomMargin = false,
   reversed = false,
 }) => (
@@ -27,9 +29,12 @@ const RadioGroup: RadioGroup = ({
       [styles.noBottomMargin]: noBottomMargin,
       [styles.reversed]: reversed,
     })}
+    role="radiogroup"
+    aria-labelledby={labelId}
   >
     <div className={styles.radioGroupLabel}>
       <Label
+        id={labelId}
         automationId={`${automationId}-field-label`}
         labelText={labelText}
         labelType="text"
--- a/draft-packages/form/KaizenDraft/Form/RadioGroup/__snapshots__/RadioGroup.spec.tsx.snap
+++ b/draft-packages/form/KaizenDraft/Form/RadioGroup/__snapshots__/RadioGroup.spec.tsx.snap
@@ -4,6 +4,7 @@ exports[`<RadioGroup />  snapshots renders RadioGroup in a column and with a lab
 <div
   class="radioGroupContainer"
   data-automation-id=""
+  role="radiogroup"
 >
   <div
     class="radioGroupLabel"
@@ -26,6 +27,7 @@ exports[`<RadioGroup />  snapshots renders RadioGroup in a column and with a lab
 <div
   class="radioGroupContainer"
   data-automation-id=""
+  role="radiogroup"
 >
   <div
     class="radioGroupLabel"
@@ -48,6 +50,7 @@ exports[`<RadioGroup />  snapshots renders RadioGroup in a row and with a label 
 <div
   class="radioGroupContainer"
   data-automation-id=""
+  role="radiogroup"
 >
   <div
     class="radioGroupLabel"
@@ -70,6 +73,7 @@ exports[`<RadioGroup />  snapshots renders RadioGroup with radios 1`] = `
 <div
   class="radioGroupContainer"
   data-automation-id=""
+  role="radiogroup"
 >
   <div
     class="radioGroupLabel"
--- a/draft-packages/form/docs/RadioGroup.stories.tsx
+++ b/draft-packages/form/docs/RadioGroup.stories.tsx
@@ -62,9 +62,9 @@ const reversedBg = {
 export const DefaultKaizenSiteDemo = () => (
   <RadioGroupExample
     render={({ selectedOption, onChangeHandler }) => (
-      <RadioGroup labelText="Radio group label">
+      <RadioGroup labelText="Radio group label" labelId="RadioGroupLabel">
         <RadioField
-          labelText="Label"
+          labelText="Option one"
           name="radio"
           id="radio-1"
           selectedStatus={selectedOption === "radio-1"}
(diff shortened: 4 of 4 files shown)


lisatassone:
@HeartSquared I've attached the screenshot you were after on this PR 👍

draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
```diff
@@ -65,4 +65,25 @@ describe("<RadioGroup /> ", () => {
...
+    it("should markup aria label when supplied labelId", () => {
+      const labelId = "labelId"
+      const { getByText, container } = render(
+        <RadioGroup
+          labelText="Label title"
+          labelId={labelId}
+          automationId="test"
+        />
+      )
+      expect(container.firstElementChild?.getAttribute("role")).toEqual(
```
hqta, line 78: Nit: Just had a thought - could we perhaps use `getByRole("radiogroup")` (https://testing-library.com/docs/queries/byrole/)

hqta (approved):

draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
lisatassone: I knew I was forgetting something! Sure I'll give it a go

lisatassone: Done.

HeartSquared (approved):

hqta (approved):

draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
```diff
@@ -65,4 +65,22 @@ describe("<RadioGroup /> ", () => {
...
+    it("should markup aria label when supplied labelId", () => {
+      const labelId = "labelId"
+      const { getByText, getByRole } = render(
+        <RadioGroup
+          labelText="Label title"
+          labelId={labelId}
+          automationId="test"
+        />
+      )
+      expect(getByRole("radiogroup").getAttribute("aria-labelledby")).toEqual(
```
dougmacknz, line 78: I believe you could also just do `getByRole('radiogroup', { name: labelId })`

draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.tsx
```diff
@@ -8,6 +8,7 @@ import styles from "./styles.scss"
 export type RadioGroupProps = {
   automationId?: string
   labelText: string | React.ReactNode
+  labelId?: string
```
dougmacknz, line 11: I'm not a huge fan of making this a prop, particularly an optional prop, for a couple of reasons:
* Engineers won't care about this if it doesn't have any obvious effect on the implementation. It's too easy for people to miss or not care about. Ideally we want to make it impossible for people to create something that isn't accessible, or at the very least making something inaccessible should have the most friction.
* It won't retrospectively add this improvement to our existing usages, and it's very unlikely that engineers will touch old usages to add this improvement if they don't have to.

Did you explore generating a unique ID within the component with `uuid` at all? That's what I would go for, so that engineers don't need to worry about this and it's all just handled for them under the hood.

HeartSquared: We didn't consider `uuid`, but I did suggest adding `id` (for the container) and using it for the label, but the others said as consumers they prefer complete control over the label (I forget what the example given was - @hqta?).

Personally I would've preferred if we had refactored it to use `fieldset` and `legend`, but we were short on time, so we figured not introducing a breaking change might be the better option first? 

Perhaps I'm not understanding the intent of the pairing sessions yet. Are we meant to aim to get it to the most ideal state (because we didn't actually make it usable for @lisatassone's use case), or just find a fix for the immediate problem without breaking changes (in this case, slightly improve accessibility)? Maybe you and I can jump on a quick chat :)

dougmacknz: I'll hazard a guess that @hqta is going to say he might want to control the id for testing purposes? If that's the case, we could build on what we have here to generate a unique ID in the case that no `labelId` is provided, so that we get the best of both worlds.

The intent of the pairing sessions can be flexible, but generally it is to make a quick improvement or bug fix to the Kaizen codebase. Ideally it should only take 1 x 1 hour session, but sometimes we underestimate things and they need a follow up session, or to be broken out of being an advocate task. Whether it requires a breaking change or not isn't really a concern, and I'd favour getting to (or working towards) an ideal state over getting something done in the hour. With that said, making Radio Group work for Lisa's use case wasn't the intent for your session, the intent for your pairing session was to fix https://github.com/cultureamp/kaizen-design-system/issues/2409.

hqta: Read my mind on the `id` point @dougmacknz - I like the `uuid` suggestion for improved DX. Will keep this in mind in the future.

HeartSquared: Awesome. Great idea. I'm on board with the fallback option. I might give this a look (also to try and refactor it) on Friday since it's our P/L day.

lisatassone: I think defaulting to generating a UUID is a good improvement and just something we missed, so happy to add it. I think there's value to be able to override it and that the prop is useful (such as testing).

I was looking for ways to have more control over what I could do with this component for my use case. My approach when doing the new slice is to try and use as much Kaizen as possible and see where the gaps are and whether they are legitimate (as in, don't belong in Kaizen). 

We know that Perform's inputs are currently with design because there is a gap there that should be addressed. 

Kaizen currently dictates that all radio groups need to look a certain way because its using the Label component, but yet offers no way to enforce using the RadioField component. It doesn't accept a prop of options where it renders them out to actually create a group of radio inputs that look like what you see in the story. It stops short. Being so opinionated about the label seemed odd to me then. I'd honestly have it do that or not be there at all as the expectation that you would want from it, you don't get. Could just be me though 🤷‍♀️ 

If the idea was to have it be composable, we may not have the right abstraction.

dougmacknz:
> I raised a possible solution to allow the consumer to override the Label element by supplying a render function for it where a consumer could render custom UI with the appropriate ID, but that was deemed too flexible.

@lisatassone I would say a render prop would our last resort here. While I do agree we need to make our components more flexible, the label is pretty much the only thing this component is opinionated about at the moment, so at that point it feels a bit pointless to even use the component if you don't want that. Not to say that this component should stay that way, this component is pretty old (precedes my time at CA) and could use a rethink.

Looking at your design and ignoring the radio fields themselves for now—that looks like a variant that we support in some of our other components called `prominent`. The label becomes larger, and the description goes above the controls. You can see this pattern in our `TextAreaField` component for a very similar usage: https://cultureamp.design/storybook/?path=/story/components-form-text-area-field--default-prominent-desc

@katsambarcus it looks like we could use a `prominent` variant for Radio Group and Checkbox Group to cater for this use case. Now that I think of it, I had a need for the same design in Capture, but I opted not to bother using RadioGroup/CheckboxGroup at the time.

katsambarcus:
@dougmacknz agreed, there's something cooking. @ashleecharetteCA and I paired to ideate on this, this morning, and I think we could definitely introduce a prominent variant that serves this experience a lot better. @lisatassone once we've landed on something, either Ashlee or myself will let you know.

cc: @jmitsios

lisatassone:
> > I raised a possible solution to allow the consumer to override the Label element by supplying a render function for it where a consumer could render custom UI with the appropriate ID, but that was deemed too flexible.
> 
> @lisatassone I would say a render prop would our last resort here. While I do agree we need to make our components more flexible, the label is pretty much the only thing this component is opinionated about at the moment, so at that point it feels a bit pointless to even use the component if you don't want that. Not to say that this component should stay that way, this component is pretty old (precedes my time at CA) and could use a rethink.

Agreed. It did seem pretty pointless. I guess it was 'more of a look to the future'. But I agree, the component probably needs a complete rethink.

> 
> Looking at your design and ignoring the radio fields themselves for now—that looks like a variant that we support in some of our other components called `prominent`. The label becomes larger, and the description goes above the controls. You can see this pattern in our `TextAreaField` component for a very similar usage: https://cultureamp.design/storybook/?path=/story/components-form-text-area-field--default-prominent-desc
> 
> @katsambarcus it looks like we could use a `prominent` variant for Radio Group and Checkbox Group to cater for this use case. Now that I think of it, I had a need for the same design in Capture, but I opted not to bother using RadioGroup/CheckboxGroup at the time.

Currently the 'label' section of the design is rendered via our MarkdownText component, which takes Markdown and converts it to HTML. We have specific Kaizen components we use to render the HTML elements, such as `Paragraph` and `Heading`. So it is adhering to the design system in some way. 

The `prominent` variant though does look very similar. We don't actually have a story/documentation for `Label`. Is this because we don't want people using it outside of the form controls?

lisatassone:
> @dougmacknz agreed, there's something cooking. @ashleecharetteCA and I paired to ideate on this, this morning, and I think we could definitely introduce a prominent variant that serves this experience a lot better. @lisatassone once we've landed on something, either Ashlee or myself will let you know.
> 

Awesome 🎉

PR #2436, opened 2022-01-26
fix: extended props availability

mcwinter07:
# Objective
Omit exposed props to the Text Field Component that should not be available to the user

# Motivation and Context
The recent updates to the input and extension of its interface exposed several props that should not be available to the user in the Text Field component. These are either handled by logic within the component or should not be able to be set by the user. These have now been added into OmittedInputProps type to ease readability.

## Prev
Screen Shot 2022-01-26 at 11 25 49 am

## Update
Screen Shot 2022-01-26 at 11 29 23 am

# Checklist


HeartSquared (approved):
LGTM :)

draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
```diff
@@ -7,7 +7,14 @@ import { FieldGroup, FieldMessage, Input, InputProps, Label } from ".."
...
 
-export interface TextFieldProps extends InputProps {
+type OmittedInputProps =
+  | "startIconAdornment"
+  | "endIconAdornment"
+  | "ariaDescribedBy"
+  | "ariaLabel"
+  | "automationId"
+
+export interface TextFieldProps extends Omit<InputProps, OmittedInputProps> {
```
dougmacknz, line 17: I'd recommend looking at `TextAreaField/TextArea` and `Slider/InputRange` for our recommended way of defining form control prop interfaces. We are moving towards extending the native HTML/React props for the element. Happy to pair to explain more :)

dougmacknz: Just had a chat with @HeartSquared and realised you're already on top of all this with InputProps extending native props 😎

Issue #1012, opened 2021-02-08
delete any scss code that has been duplicated into deprecated-component-library-helpers (breaking change)


mbylstra:
This is the more ambitious follow up to https://github.com/cultureamp/kaizen-design-system/issues/1010 (just add deprecation warnings).

PR: https://github.com/cultureamp/kaizen-design-system/pull/1740

TODO
- [x] add a plugin for upgrading `@kaizen/component-library/style|layout|type.scss` imports, which defines some rules
- [x] preset imports the plugin and uses the rules

TODO (the actual plugin)
- [x] find usages of `@kaizen/component-library/styles/color|type|layout.scss`
- [x] replace with `@kaizen/deprecated-component-library-helpers/styles/color|type|layout.scss`
- [x] remove any duplicate `@kaizen/deprecated-component-library-helpers/styles/color|type|layout.scss` imports, making sure to keep the latest in the source code (to preserve any potential override behaviour)
- [x] write up how to use this (in the BREAKING CHANGE commit)

## Considerations:
If there are duplicate imports, and one is removed, could this cause a compilation error?
Eg:
```
@import "~@kaizen/component-library/styles/color";
@import "something.scss"
@import "~@kaizen/deprecated-component-library-helpers/styles/color";
```
would become
```
@import "something.scss"
@import "~@kaizen/deprecated-component-library-helpers/styles/color";
```
If "something.scss" is depending on `color.scss`, will this cause a compilation error (does the order of imports matter?)

mbylstra:
Assessing impact or removing these files:

https://github.com/search?q=org%3Acultureamp+%22%40kaizen%2Fcomponent-library%2Fstyles%2Fcolor%22+extension%3Ascss&type=Code&ref=advsearch&l=&l=

mbylstra:
I was hoping to be able to use github code search (limited to the cultureamp org) to find any imports to these files. However, it doesn't work as expected.

If I search for just "@kaizen/component-library" I get ~800 results (https://github.com/search?q=org%3Acultureamp+%22%40kaizen%2Fcomponent-library%22++extension%3Ascss&type=Code&ref=advsearch&l=&l=), but if I search for "@kaizen/component-library/styles/type" then I get 0 results, even though there are clearly lots of repos containing this.

What I realised just by searching "@kaizen/component-library" is that there are lots of repos using these files, so that question is already answered.

mbylstra:
[...]

PR #2437, opened 2022-01-26
fix: add missing imports in deprecated-component-library-helpers/styles/type.scss

mbylstra:
# Objective
To be able to upgrade consumers from `packages/component-library/styles/type.scss` to `packages/deprecated-component-library-helpers/styles/type.scss` (this is what we want, because we want to delete `packages/component-library/styles/type.scss`) the two files need to have the same imports, even if the import isn't actually used in `type.scss`. This is because the version of scss we use doesn't really have an import system (you implicitly import any import that a file you import imports). If `type.scss` imports `color.scss`, and the consumer imports `type.scss`, then the consumer has a implicitly imported `color.scss`, so removing `color.scss` from `type.scss` will break the consumer if it's using any vars from color.scss. This happens after running the `use-deprecated-component-library-helpers-scss-imports` stylelint fix on the Big M code base. Essentially, that stylelint fix depends on there being no breaking changes between `packages/component-library/styles/type.scss` and  `packages/deprecated-component-library-helpers/styles/type.scss` (and removing an import, even if it isn't used, is a breaking change in scss because of implicit imports). This is pretty confusing stuff, so please reach out to me on Slack if you have any questions, and I can try to explain and give more of a back story.

Other than unfortunately increasing build times and adding bloat, adding these two imports should be safe, as they only include ultra legacy `ca-xxxxx` sass vars (they shouldn't be overriding anything new)

# Motivation and Context
This is required to be able to https://github.com/cultureamp/kaizen-design-system/issues/1012

# Checklist

# Additional Considerations


packages/deprecated-component-library-helpers/styles/type.scss
```diff
@@ -1,6 +1,8 @@
 @import "~@kaizen/design-tokens/sass/spacing";
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/design-tokens/sass/typography";
+@import "~@kaizen/deprecated-component-library-helpers/styles/color";
+@import "~@kaizen/deprecated-component-library-helpers/styles/layers";
```
mbylstra, line 5: These are the two imports that are in `packages/component-library/styles/type.scss` but not in `packages/deprecated-component-library-helpers/styles/type.scss`

packages/deprecated-component-library-helpers/styles/layout.scss
```diff
@@ -1,4 +1,4 @@
-@import "type";
+@import "~@kaizen/deprecated-component-library-helpers/styles/type";
```
mbylstra, line 1: This is a bit unrelated to the goal of the PR, but I noticed this was using a relative import but I think it's safer to explicitly do an absolute import (otherwise, it's a bit  non deterministic whether this file will come from deprecated-component-library-helpers or component-library). I want this to be consistent with the other imports in this package.

I had to look up the docs on this, but relative imports don't need the ".". It checks the same directory as the file first, and if it's not there it checks other root paths that have been loaded. It seems a lot less magical to me to just make it fully explicit where it's coming from.

dougmacknz (approved):

PR #2488, opened 2022-01-27
chore: remove elm debug to get rid of elm optimise warnings

burritobill:
I removed elm debug so all the noisy warnings would go away inside of storybook

### DIFF
--- a/packages/progress-bar/src/Kaizen/ProgressBar.elm
+++ b/packages/progress-bar/src/Kaizen/ProgressBar.elm
@@ -94,7 +94,7 @@ view (Config config) =
             [ classList [ ( .progressBackground, True ) ] ]
             [ div
                 [ class (moodToClass config.mood)
-                , classList [ ( .animating, isAnimating |> Debug.log "isAnim" ) ]
+                , classList [ ( .animating, isAnimating ) ]
                 , progressBarStyle
                 ]
                 []
--- a/storybook/webpack/elm-webpack-loader-fix.js
+++ b/storybook/webpack/elm-webpack-loader-fix.js
@@ -7,24 +7,24 @@
 
 "use strict"
 
-var fs = require("fs")
-var path = require("path")
-var loaderUtils = require("loader-utils")
-var elmCompiler = require("node-elm-compiler")
-var yargs = require("yargs")
+const fs = require("fs")
+const path = require("path")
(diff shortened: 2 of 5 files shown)


storybook/webpack/elm-webpack-loader-fix.js
```diff
@@ -7,24 +7,24 @@
...
-var alreadyCompiledFiles = []
+let runningInstances = 0
+const alreadyCompiledFiles = []
 
-var defaultOptions = {
+const defaultOptions = {
   cache: false,
   forceWatch: false,
-  optimize: false,
+  optimize: true,
```
jasononeil, line 22: I'm curious if this will actually work. There was a reason we didn't turn `optimize: true` on when we upgraded to Elm 0.19 but that was 3 years ago and it escapes me now...

If it looks fine in storybook though it's probably safe to merge - these settings won't effect any consuming repos

packages/progress-bar/src/Kaizen/ProgressBar.elm
```diff
@@ -94,7 +94,7 @@ view (Config config) =
             [ classList [ ( .progressBackground, True ) ] ]
             [ div
                 [ class (moodToClass config.mood)
-                , classList [ ( .animating, isAnimating |> Debug.log "isAnim" ) ]
+                , classList [ ( .animating, isAnimating ) ]
```
jasononeil, line 97: I'm pretty sure Debug.log isn't required here, but I wonder if the value is supposed to be "inAnim" (a string) rather than "isAnimating" (a boolean).

@mbylstra feel like answering an elm question?

mbylstra: Oops, this looks like it was accidentally left in and can definitely be removed! 

(FWIW, in `|> Debug.log "isAnim"`, the pipeline operator puts the value on the left as the second argument to the Debug.log function, so it'd be logged as `isAnim True` or `isAnim False`. It's pretty handy for quickly putting a log statement in around an expression - something I miss from Elm. I often make a function similar to this in JS to half replicate it... eg "`function logExpression(exp) { console.log exp; return exp}`". We didn't have the no logging to console linter warning in Elm though 😅 )

HeartSquared:
Closing this as we will be removing Elm from Kaizen in health week :)

PR #2497, opened 2022-01-31
feat: text area uplift

mcwinter07:
# Objective
- Uplift Text Area component and align it to the UI library

# Motivation and Context
- This is part of the component uplift goals for the Kaizen design system
- Adding the new caution state
- Adding disabled styles
- Refactoring Storybook stories to demo / sticker sheet architecture

## Previous
old

## Changes
new
6165d0407c5.png)

# Checklist


ActuallyACat (approved):

draft-packages/form/KaizenDraft/Form/TextAreaField/TextAreaField.spec.tsx
```diff
@@ -30,3 +30,45 @@ it("renders a validation message when provided", () => {
...
+    />
+  )
+
+  const prominentLabel = container.getElementsByClassName("prominent")[0]
+  const siblingNode = prominentLabel?.nextElementSibling
+
+  expect(siblingNode?.className === "message").toBeTruthy()
+})
+
+it("Does not render a validation message if it is disabled", () => {
```
christian314159, line 63: When there is an `if` in a test,  it usually means there should be 2 tests at least. You are showing that it won't show the validation message when it is `disabled`, but for all I know it never shows it even when it is "enabled".
It might look verbose, but I'd so something like this here:
```
describe("Validation message", () => {
  describe("when the textarea field is NOT disabled", () => { // I like capitals here because it stands out against the other describe block
    it("renders the validation message", ...)
  })

  describe("when the textarea field is disabled", () => {
    it("does NOT render the validation message", ...)
  })
})
```
This also make it look quite nice in the output of the testrunner.

This structure is also really helpful when you have multiple tests for each `when...` block as you can have one `beforeEach` where you set up the state for all.

christian314159 (commented):
Looks good, but see my comment on how you could improve the tests.

jmitsios:
@mcwinter07 
- We might need to look at the spacing between the **Label** and **Container** across the board, in terms of other similar components (`select`, `text field`, etc..) The spacing here looks marginally tighter than that in the UI Kit.

**Storybook -** 
image
**UI Kit -** 
image

- Is there a way to get rid of the drag handles in the disabled states at all?  Nothing should be accessible or interactive in this state.
image

- There's a prop description of `success` under **status**. What does that point to?
image

jmitsios (dismissed):
Muy bueno @mcwinter07 👌

draft-packages/form/KaizenDraft/Form/TextAreaField/TextAreaField.spec.tsx
mcwinter07: True, good opportunity to test for happy and sad path :) I've updated this now to account for both test cases

christian314159 (approved):
Tests are looking ✨👌 now!

ActuallyACat (approved):

PR #2498, opened 2022-01-31
ci: Run Chromatic action on the master branch to auto accept baselines

ActuallyACat:
# Objective
The Chromatic baselines were not being updated when a branch was merged to master as there was an early exit in the steps. This change allows all master branches to update snapshots and autoaccepts them as the new baseline.

### DIFF
--- a/.github/workflows/chromatic.yaml
+++ b/.github/workflows/chromatic.yaml
@@ -28,7 +28,8 @@ jobs:
       - run: |
           echo "runChromatic=${runChromatic}" >> $GITHUB_ENV
         env:
-          runChromatic: ${{ github.event.review.state == 'approved' }}
+          # If action is running on master branch we would like to update baselines via autoAcceptChanges
+          runChromatic: ${{ github.event.review.state == 'approved' || github.ref == 'refs/heads/master' }}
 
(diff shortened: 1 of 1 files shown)


HeartSquared (approved):
🤞🏻

.github/workflows/chromatic.yaml
```diff
@@ -28,7 +28,8 @@ jobs:
       - run: |
           echo "runChromatic=${runChromatic}" >> $GITHUB_ENV
         env:
-          runChromatic: ${{ github.event.review.state == 'approved' }}
+          # If action is running on master branch we would like to update baselines via autoAcceptChanges
+          runChromatic: ${{ github.event.review.state == 'approved' || github.ref == 'refs/heads/master' }}
```
HeartSquared, line 32: Apparently we should be able to do something like this
```suggestion
          runChromatic: ${{ github.event.review.state == 'approved' || ${GITHUB_REF##*/} == 'master' }}
```

https://stackoverflow.com/questions/58033366/how-to-get-the-current-branch-within-github-actions

ActuallyACat: I have a slight preference for the longer form as it avoids bash magic. TIL though! Thanks for sharing.

PR #2499, opened 2022-01-31
feat: DatePicker setup basic single date picker

lloydstubber:
# Objective
This is the first piece of the DatePicker work. Which is creating a _very_ simple version of the component (bare bones API) and then the majority of the work being the styling of the input. The next piece will be styling the calendar and then getting into the meat of the API/documentation/tests side of things after that.

See entire epic for rest of work needing to be done: https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-62

# Motivation and Context
- Basic implementation of a single DatePicker with it's states styled to match the UI kit

# Notes
- Form primitive styling isn't the greatest. So we're just following suit here. We've tried to slim it down as much as possible and follow the TextField styling for hover/focus etc. as per @jmitsios 's instructions.
- Please disregard any API related work, this is just enough to get it rendering to style it. Although, the pattern of having it be a stateless component so we pass the date to it is how it'll work. So the story is very much how a consumer would use it.
- Label is closer in the implementation vs. the design but this matches how the TextField has it implemented. I think we'll need to update the label across the board rather than add padding to it per component. 

Default
image

Default on Hover
image

Default on Focus
image

Default w/ Value
image

Disbaled w/ Placeholder
image

Disabled w/ Value
image

### DIFF
--- a/packages/date-picker/src/DatePicker.scss
+++ b/packages/date-picker/src/DatePicker.scss
@@ -12,16 +12,10 @@
 
 // Vars
 $input-height: 48px;
-$input-base-padding-horizontal: $ca-grid / 2;
+$input-base-padding-horizontal: $spacing-sm;
 $input-icon-size: 1.25rem; // 20px
-$input-placeholder-opacity: 0.5;
-$input-disabled-background: $color-gray-300;
 $input-disabled-opacity: 0.3;
-$input-disabled-border-alpha: 50%;
-
-$search-input-placeholder-opacity: 0.7;
-$search-input-disabled-icon-opacity: 0.8;
-$border-solid-border-radius-curved: 7px;
+$input-disabled-color-opacity: rgba($color-purple-800-rgb, 0.7);
 
 ///////////////////////////////////////////////////
 // BASE STYLES
@@ -31,17 +25,13 @@ $border-solid-border-radius-curved: 7px;
   position: relative;
 }
 
-.withLoading .startIconAdornment {
-  margin-top: -2px;
-}
-
 //  is required to override performance-ui global materialize CSS >:|
-input {
+.input {
   @include form-input-reset;
   background-color: transparent;
   background-clip: padding-box;
   border: $border-solid-border-width $border-solid-border-style transparent;
-  border-radius: $border-solid-border-radius-curved;
+  border-radius: $border-solid-border-radius;
   display: block;
   font-size: 1rem;
   line-height: 1.5;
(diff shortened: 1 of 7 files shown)


packages/date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,131 @@
...
+  display: block;
+  font-size: 1rem;
+  line-height: 1.5;
+  padding: 0 $input-base-padding-horizontal;
+  height: $input-height;
+  width: 100%;
+  margin-bottom: 2px;
+
+  &.disabled {
+    &:not(.reversed) {
```
lloydstubber, line 44: There is no reversed/secondary design at the minute but no doubt will be. Unsure if it's worth removing.

packages/date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.withStartIconAdornment {
+  .startIconAdornment {
+    @include vertically-center-icon;
+    @include ca-position($start: $ca-grid / 2);
+  }
+
+  .input {
+    @include ca-padding(
+      $start: 1rem + $input-icon-size,
```
lloydstubber, line 91: 1rem is to handle the styling for icon at the start of input which isn't relative to `ca-grid`

ActuallyACat (changes requested):

jmitsios:
@lloydstubber

- _(See video attached)_ I'm noticing a situation where after I choose a date and I realise it wasn't the one I wanted.. I'd click back into the text field to re-trigger the calendar, but instead it just clears the focus state. Forcing another click to open it again.
- The calendar also seems to be overlapping the bottom of the focus ring too.
- I'm assuming to not make any design review comments about the calendar itself in the storybook link?
https://user-images.githubusercontent.com/95595231/151920540-075df21f-dba9-4325-b06a-33ecde218f0c.mov

jmitsios (approved):
The styling of the inputs is spot on 👍

packages/date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,131 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/deprecated-component-library-helpers/styles/layout";
+@import "~@kaizen/deprecated-component-library-helpers/styles/color";
+@import "~@kaizen/component-library/styles/mixins/forms";
+@import "~@kaizen/draft-form/KaizenDraft/Form/Primitives/base";
+@import "~@kaizen/draft-button/KaizenDraft/Button/styles";
```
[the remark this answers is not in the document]
NatalieLoyzaga: I have removed the button import as it was not being used

lloydstubber:
Closed in favour of https://github.com/cultureamp/kaizen-design-system/pull/2511

PR #2504, opened 2022-02-02
fix: resolve aspect ratio issue in guidance block

mcwinter07:
# Objective
- Add new prop lockAspectRatio to allow for more consistent layouts for side by side stacked guidance blocks using mixed scene illustrations.
- Clean up and reduce number to create a more consumer friendly stories

# Motivation and Context
Issue raised in [this ticket](https://cultureamp.atlassian.net/browse/KDS-125), which relates back to [this thread](https://cultureamp.slack.com/archives/C02NUQ27G56/p1642475620232200). T

"When we add enableAspectRatio prop to a scene illustration that appears to be portrait, we expect it to take on a landscape appearance adding white space on either side for a 4/3 aspect ratio however, this doesn’t happen."

# Solution
New prop has been added that will set a max height on the image so that scenes stacked and portrait sized scenes will produce more predictable layouts.

# Caveats
This is not the most elegant solution since tags are target in the CSS - It is inadvisable that we spend too much time making a perfect solution since this will be replace by the Tile Component, which should address these edge cases

# Screenshots:
- Portrait and scene illustrations have disproportionate heights when stacked next to each other
Screen Shot 2022-02-02 at 12 16 01 pm
- Portrait Illustrations can have excessive height if used
Screen Shot 2022-02-02 at 12 16 40 pm

# Checklist


hqta:
Nice work! I'd probably add a story for

- Layouts
  - Default layout
  - Inline with Scene Illustration

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -245,21 +253,20 @@ class GuidanceBlock extends React.Component<
   }
 
   bannerClassName(noMaxWidth): string {
-    return classnames(styles.banner, {
+    const {
+      variant = "default",
+      layout = "default",
```
HeartSquared, line 258: Something feels off about this. Isn't this already set in L87-88?

mcwinter07: It's a strange one, it is defined with the old school static defaultProps but I get a TS issue if I don't declare it there - It's been a while since I've used class components but maybe the function doesn't know what default props are set in the constructor?

mcwinter07: Screen Shot 2022-02-15 at 3 04 43 pm

@HeartSquared

HeartSquared: Gross :( Yeah I'm not too familiar with class based components either.

Could you possibly only destructure at the start of the component and use the destructured values directly within this method? It's strange that we are in a way overriding the values by reinitialising the default value. This would turn into something weird if later on someone only updated one of the two places.

mcwinter07: hmmmm maybe I can update the function to take props then pass it in when `bannerClassName` gets called, then I can de-structure / hopefully without a ts error

hqta (dismissed):

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -40,14 +39,24 @@ type VariantType =
 export type GuidanceBlockProps = {
   layout?: LayoutType
   illustration: React.ReactElement<SpotProps | SceneProps>
+  /*
+   ** Sets the how the width and aspect ratio will respond to the Illustration passed in.
```
christian314159, line 43: > Sets ~~the~~ how the ...

christian314159: Sorry, you know it's my special talent ;)

draft-packages/guidance-block/KaizenDraft/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -40,14 +39,24 @@ type VariantType =
...
   smallScreenTextAlignment?: TextAlignment
   actions?: GuidanceBlockActions
+  /*
+   ** This will still require the secondary object to be passed into the actions ie: {secondary: { label: "Dismiss action" }}`
+   */
   secondaryDismiss?: boolean
+  /*
+   ** persistent should always return true and will soon be deprecated.
+   ** The X close icon has been superseded with the pattern "dismiss" or "cancel" using the secondary action.
+   */
```
HeartSquared, line 59:
```suggestion
  /**
   * persistent should always return true and will soon be deprecated.
   * The X close icon has been superseded with the pattern "dismiss" or "cancel" using the secondary action.
   */
```

I think doc blocks are meant to look like this?
I'd probably also add backticks to `persistent` to indicate it's the prop name :)

christian314159 (approved):

PR #2511, opened 2022-02-06
feat: date picker input and calendar styling

NatalieLoyzaga:
# Objective
This is the first piece of the DatePicker work. Which is creating a _very_ simple version of the component (bare bones API) and then the majority of the work being the styling of the input and calendar. The next piece will be getting into the meat of the API/documentation/tests side of things after that.

See entire epic for rest of work needing to be done: https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-62

# Motivation and Context
- Basic implementation of a single DatePicker with it's states styled to match the UI kit

# Notes
- Feature branch PR (with basic setup of component) is [here](https://github.com/cultureamp/kaizen-design-system/pull/2489/files)
- We've moved away from using their input and using our Kaizen TextField input. This means we'll handle the popping up of the calendar ourselves for more flexibility (which isn't in this PR). This was decided as their accessibility around the input and moving focus to the popout wasn't great and we can handle this better.
- Please disregard any API related work, this is just enough to get it rendering to style it. Although, the pattern of having it be a stateless component so we pass the date to it is how it'll work. So the story is very much how a consumer would use it.
- Label is closer in the implementation vs. the design but this matches how the TextField has it implemented. I think we'll need to update the label across the board rather than add padding to it per component. 

Default
image

Default on Hover
image

Default on Focus
image

Default w/ Value
image

Disbaled w/ Placeholder
image

Disabled w/ Value
image


lloydstubber:
Updated the description to account for separated styling PRs that are now just all in here.

ActuallyACat:
Just a question about sequencing - which PR/issue will tackle the showing/hiding of the calender?

packages/date-picker/src/CalendarNav.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react"
+import { Icon } from "@kaizen/component-library"
+import arrowRight from "@kaizen/component-library/icons/arrow-right.icon.svg"
+import arrowLeft from "@kaizen/component-library/icons/arrow-left.icon.svg"
+import styles from "./DatePicker.scss"
+
+export type CalendarNavProps = {
+  onPreviousClick: () => void
```
ActuallyACat, line 8: /nit does `onPreviousClick` and `onNextClick` need to be documented in storybook? I can't see them in the props table.

packages/date-picker/src/CalendarNav.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onNextClick: () => void
+}
+
+export const CalendarNav: React.FunctionComponent<CalendarNavProps> = ({
+  onPreviousClick,
+  onNextClick,
+}) => (
+  <div className={styles.navbar}>
+    <button
+      className={styles.arrows}
```
ActuallyACat, line 18: There's nothing on these elements to indicate to a screen reader what these buttons do. Good use case for `aria-label`

packages/date-picker/src/CalendarNav.tsx
ActuallyACat: Or are these not available to the consumer? I think there's a lot of value in exposing them.

lloydstubber:
@ActuallyACat that will be the next PR which will be mostly polishing of the props (e.g. disabled days) and then functionality. It was simpler to style completely first as it's easier to check with the calendar rendered beneath vs. trying to inspect when popped up.

packages/date-picker/src/DatePicker.tsx
```diff
@@ -19,38 +19,34 @@ export const DatePickerWrapper: React.FunctionComponent<DatePickerProps> = ({
...
+      <DayPicker
+        selectedDays={selectedDate}
+        // TODO create disabledDays prop, this is to display disabled styling
+        disabledDays={{
+          daysOfWeek: [6],
         }}
+        onDayClick={onDayChange}
+        navbarElement={getNavbar}
+        classNames={defaultDatePickerClasses}
+        className={styles.calendar}
```
ActuallyACat, line 48: /nit `classNames` and `className`? Could these be combined or otherwise simplified?

packages/date-picker/src/DatePicker.tsx
```diff
@@ -19,38 +19,34 @@ export const DatePickerWrapper: React.FunctionComponent<DatePickerProps> = ({
...
-        {
-          [styles.withDisabled]: isDisabled,
+}) => {
+  const getNavbar = ({ ...navbarProps }: CalendarNavProps) => (
+    <CalendarNav {...navbarProps} />
+  )
+
+  return (
+    <>
+      <TextField
```
ActuallyACat, line 29: You'll need to be pretty careful about how people will use this component if you don't expose all these `TextField` props. Folks generally like the ability to do things like add their own automation/testing id, refs, and so on. Take a look at the [props on TextField](https://github.com/cultureamp/kaizen-design-system/blob/master/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx#L35-L50) and think about the use case for each prop and whether a consumer would like similar functionality from this component too.

ActuallyACat (changes requested):

jmitsios:
@lloydstubber + @NatalieLoyzaga 

Looking almost perfect!   Couple small things I've noticed:
Datepicker
- Is there a way I can easily see what style/font/weight/size is being used in Storybook?
  - The month **title** looks a bit larger than the **Heading 4** style.
  - Are the individual dates bolder? Or is it just how it's being rendered in the browser?
- When I press Tab in the calendar to focus on elements and then select a Date with spacebar, pressing Tab once more won't jump to the next available date and/or cycle through any further.  Is this expected behaviour?
- Looks like no one designed a hover-state in the UI Kit for the Prev/Next arrows.  Could we please mirror the same hover styling used for the dates? (i.e: Blue-100 background, with Blue-500 arrow)?

packages/date-picker/src/CalendarNav.tsx
christian314159: I know naming is hard and contentious ;) but imho `onPreviousClick/onNextClick` lack some context, maybe consider something like `onPreviousMonthClick` which is bit wordier but also clearer.

packages/date-picker/src/CalendarNav.tsx
```diff
@@ -0,0 +1,32 @@
...
+import arrowRight from "@kaizen/component-library/icons/arrow-right.icon.svg"
+import arrowLeft from "@kaizen/component-library/icons/arrow-left.icon.svg"
+import styles from "./DatePicker.scss"
+
+export type CalendarNavProps = {
+  onPreviousClick: () => void
+  onNextClick: () => void
+}
+
+export const CalendarNav: React.FunctionComponent<CalendarNavProps> = ({
```
christian314159, line 12: Just wondering, is `React.FunctionComponent` the one that has a `children` prop baked in - which we wouldn't want here, would we? I think @HeartSquared always uses a different one that I can't remember.

packages/date-picker/src/DatePicker.tsx
christian314159: Along the same lines, should we ever have a hardcoded `id`? See `id="date-picker"` below. I can see a datepicker used multiple times on one page and then we'd have the same ID multiple times which could potentially lead to weird, unexpected behavior.

packages/date-picker/src/DatePicker.tsx
```diff
@@ -19,38 +19,34 @@ export const DatePickerWrapper: React.FunctionComponent<DatePickerProps> = ({
...
-    {labelText && <Label disabled={isDisabled} labelText={labelText} />}
-    <div
-      className={classnames(
-        styles.wrapper,
-        styles.withStartIconAdornment,
-        styles.withSearch,
-        {
-          [styles.withDisabled]: isDisabled,
+}) => {
+  const getNavbar = ({ ...navbarProps }: CalendarNavProps) => (
```
christian314159, line 23: tbh I'd just inline that on line 46, makes it easier see what it is without having to scan the file up/down:
```
...
getNavbar={({ ...navbarProps }: CalendarNavProps) => <CalendarNav {...navbarProps} />}
...
```

packages/date-picker/src/DatePicker.tsx
```diff
@@ -19,38 +19,34 @@ export const DatePickerWrapper: React.FunctionComponent<DatePickerProps> = ({
...
-          disabled: isDisabled,
-          className: classnames(styles.input, {
-            [styles.disabled]: isDisabled,
-          }),
+        disabled={isDisabled}
+      />
+
+      <DayPicker
+        selectedDays={selectedDate}
+        // TODO create disabledDays prop, this is to display disabled styling
```
christian314159, line 41: Is this TODO referring to do away with the "magic numbers" and accept a descriptive string instead? (e.g. `mon/tue/wed/...`)?

christian314159 (commented):
👏 Great work, just left a few general comments. Up to you to action any of them, thanks!

packages/date-picker/src/CalendarNav.tsx
lloydstubber: Was leaving this kind of nitty gritty for the post-styling PR but happy to add these in as props with this PR.

packages/date-picker/src/CalendarNav.tsx
lloydstubber: Good catch, will update

packages/date-picker/src/DatePicker.tsx
lloydstubber: From our initial look, it's separated and we couldn't combine them but I'll take another look.

NatalieLoyzaga:
@jmitsios I have made comments for your feedback below. Let me know if there are any other issues :)

`Is there a way I can easily see what style/font/weight/size is being used in Storybook?`

Month Text: Heading-4 (1.125rem/18px font-size, 600 weight, normal spacing, 1.5rem/24px line-height) - I believe those match the UI kit so I'm unsure why it looks bigger
Day Text: Paragraph Body (1rem font-size, 400 weight, 1.5rem line-height) - The current colour seems to be just black where the UI is purple so I have updated that. 

`When I press Tab in the calendar to focus on elements and then select a Date with spacebar, pressing Tab once more won't jump to the next available date and/or cycle through any further. Is this expected behaviour?`

The expected behaviour for navigating the dates is to use the arrow keys rather than Tab. You press Tab to reach the first date then use the arrow keys to move around. 

`Looks like no one designed a hover-state in the UI Kit for the Prev/Next arrows. Could we please mirror the same hover styling used for the dates? (i.e: Blue-100 background, with Blue-500 arrow)?`

Updated now :)

jmitsios:
@NatalieLoyzaga — That's awesome, thanks for the helpful comments! And for updating that all so quick, it looks great.
The only thing I notice since that update is that it seems the current 'Today' date was previously Blue, now it's the same color as the other dates.  Is it okay to swap that back to Blue?  Thank you!

packages/date-picker/src/DatePicker.tsx
lloydstubber: Correct! That's in the next card.

packages/date-picker/src/CalendarNav.tsx
lloydstubber: I'm going to leave this for the next PR where we expose more of the API for consumers. This is mostly the styling/basic API to get a base in to work from.

packages/date-picker/src/DatePicker.tsx
lloydstubber: Thanks for that. I've passed through `id` along with the other props I think a consumer would want.

lloydstubber:
@jmitsios that style change has been updated by @NatalieLoyzaga and @ActuallyACat this is ready for re-review. I've worked through almost all the comments (and flagged some for the next stage of the work).

packages/date-picker/src/CalendarNav.tsx
lloydstubber: This does have `PropsWithChildren` baked in it seems. Keen to know if there's one we prefer, I always reach for this.

christian314159: If we don't expect `children`, then I'd say we shouldn't have them in the type declaration. Not that it really could do much harm if someone did
```
<DatePicker><h1>Hello World</h1></DatePicker>
```
but for cleanliness' sake.

Btw, I think it's `React.VoidFunctionComponent` that could be used, but I'd also ask @HeartSquared as I know she has some views on this.

HeartSquared: Yes `VoidFunctionComponent` aka `VFC` is the one without children. There was an article somewhere which mentioned that in a future major version they plan to remove children from `FC` and then deprecate `VFC`, but it hasn't been solidified yet, so in the meantime `VFC` is recommended as it is more type safe. If you do need `children`, then you would explicitly add it to your props since it adds more clarity and gives you full control of the type signature.

It's definitely on the list for Wednesday's meeting :)

lloydstubber: Thanks for that, it's now been updated.

christian314159:
Focus on previous month button:
image
I think UI Kit doesn't have the default dotted outline, just the blue one.

Focus on 6 Feb:
image
Not sure what UI Kit suggests here, the faint outline looks too minimal to me.

@jmitsios Your thoughts?

packages/date-picker/src/CalendarNav.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const CalendarNav: React.VFC<CalendarNavProps> = ({
+  onPreviousClick,
+  onNextClick,
+}) => (
+  <div className={styles.navbar}>
+    <button
+      className={styles.arrows}
+      onClick={() => onPreviousClick()}
+      type="button"
+      aria-label="Previous Month"
```
ActuallyACat, line 21: NB: This means that the component won't meet the `Internationalized` health metric. Not an issue at the moment as we don't have a solution for i18n in Kaizen right now. Something to keep in the back of your mind for the future.

ActuallyACat (approved):
LGTM (code wise - didn't poke at the design)

packages/date-picker/src/CalendarNav.tsx
HeartSquared: Here's a similar article to what I had read previously :) 
https://spin.atomicobject.com/2022/01/04/think-twice-react-fc/

packages/date-picker/src/CalendarNav.tsx
lloydstubber: Yep that's on the roadmap for the next iteration when we do the DateRange! Thanks for flagging though.

lloydstubber:
@christian314159 What browser are you using there? I can't replicate that. This is all the states in Chrome:
Screen Shot 2022-02-08 at 10 02 52 am

lloydstubber:
Ah yep that's in Firefox @christian314159. I'm guessing we'll need to prefix our focus states for that browser. Although I'm surprised we don't get this in our compile stage - @ActuallyACat is that usually the case?

Screen Shot 2022-02-08 at 10 04 30 am

lloydstubber:
@christian314159 okay that's fixed now, focus styles inside the calendar now match Chrome and FF. We just were using a mix of border vs. outline and were leveraging Chrome's inbuilt so FF was just showing it's native one.

lloydstubber:
## Firefox

Screen Shot 2022-02-08 at 10 31 14 am
Screen Shot 2022-02-08 at 10 31 08 am

## Chrome
Screen Shot 2022-02-08 at 10 32 27 am
Screen Shot 2022-02-08 at 10 32 58 am

christian314159 (approved):
Looking good in Firefox now! 👍

PR #2514, opened 2022-02-07
feat: autogenerate id for aria labels if not supplied

lisatassone:
This is a follow up to https://github.com/cultureamp/kaizen-design-system/issues/2409. We are auto-generating the label id now if it is not explicitly supplied to ensure that the group is always accessible.

I took the liberty of removing the snapshot tests as they weren't testing anything that seemed useful or couldn't be done explicitly with RTL. Without giving them a `labelId` they would also have always created a false negative due to the uuid generation. If we want to keep them though, I can add them back in and add the `labelId` to the snapshots.

# Checklist

# Additional Considerations


draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.tsx
```diff
@@ -22,27 +23,31 @@ const RadioGroup: RadioGroup = ({
   labelId,
```
lisatassone, line 23: I didn't just default it to a uuid here because I wasn't sure whether we cared that the id's could change between renders when not supplied. Seems not great, so I'm making sure we have a stable one for the lifecycle of the component.

hqta (approved):

PR #2515, opened 2022-02-07
refactor: Duplicate draft-button component

mcwinter07:
# Objective
- update storyboard
- duplicate draft-button component and move to packages folder with git history
- remove elm from button package
- update file paths
- reset version to 0.0.0
- remove change log to start clean

# Motivation and Context
Transition draft components over to packages

PR #2516, opened 2022-02-07
feat: duplicate draft button and move it to packages

mcwinter07:
# Objective
- Move duplicated draft-button to packages folder and update imports
- Update storyboard to use new Button package

# Motivation and Context
Step 2 in deprecating the draft-button package and moving it over into packages

# Note:
Do not merge until the Deprecate PR is merged #2515


packages/button/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@kaizen/button",
+  "version": "0.0.0",
```
HeartSquared, line 3: Since we're resetting the version, we should probably delete the changelog.
Alt if we want to keep the changelog, we should revert the version to match it.

packages/button/docs/IconButton.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+    label: "Label",
+  },
+  parameters: {
+    actions: {
+      argTypesRegex: "^on.*",
+    },
+    docs: {
+      description: {
+        component:
+          'import { IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen-draft/button".',
```
HeartSquared, line 21:
```suggestion
          'import { IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen/draft-button".',
```

Haha oops neither of us noticed at the time 🙃

packages/button/docs/IconButton.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+    ),
+  },
+  decorators: [withDesign],
+}
+
+export const DefaultKaizenSiteDemoIcon = args => (
+  <IconButton {...args} icon={configureIcon} />
+)
+
+DefaultKaizenSiteDemoIcon.storyName = "Default Icon (Kaizen Site Demo)"
```
HeartSquared, line 35:
```suggestion
DefaultKaizenSiteDemoIcon.storyName = "Default (Kaizen Demo)"
```

Alt put `Button` in it. Since it's not actually an Icon story.

packages/button/docs/Button.stories.tsx
```diff
@@ -0,0 +1,1020 @@
...
+    label: "Label",
+  },
+  parameters: {
+    actions: {
+      argTypesRegex: "^on.*",
+    },
+    docs: {
+      description: {
+        component:
+          'import { Button, IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen-draft/button"',
```
HeartSquared, line 29:
```suggestion
          'import { Button, IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen/draft-button"',
```

HeartSquared (changes requested):

packages/button/docs/IconButton.stories.tsx
ActuallyACat: nit/ Don't rename the `(Kaizen Site Demo)`. We can use it to filter out all the stories we embed on the Gatsby site.

HeartSquared: Hmm...That's an interesting one. Pretty sure we've been renaming the newer ones to be without `Site` because no one else knew that was a thing 🙃

ActuallyACat: Ah. We should stop doing that 😬

HeartSquared: I mean, as long as it's consistent moving forward it should be okay right? Or is there something set up to look for it?

PR #2518, opened 2022-02-08
chore: Kaizen Site scrolls to top of page when selecting different animation preset on animation guidelines [KDS-190]

christian314159:
# Issue

image

I noticed that selecting a different animation preset on our [Animation Guidelines](https://cultureamp.design/guidelines/animation/) would make the page jump to the very top and I had to scroll all the way back again :(

# Investigation

Turns out each option in the dropdown is using the deprecated `MenuItem` component, which is essential an `A` tag with `href="#"` by default - hence it jumping to the top.

# Solution

I amended to `onClick` function passed to `MenuItem` with a good ol' `e.preventDefault()` and now it doesn't jump anymore but you can immediately see the change in animation :)

[See in action here on the branch build for the Kaizen site](https://dev.cultureamp.design/kds-190-fix-kaizen-site-scrolling-issue-on-animation-guidelines/guidelines/animation/)

### DIFF
--- a/site/src/docs-components/animation/AnimationExample.tsx
+++ b/site/src/docs-components/animation/AnimationExample.tsx
@@ -48,12 +48,13 @@ class AnimationExample extends React.PureComponent<AnimationExampleProps> {
           {this.presets.map(item => (
             <MenuItem
               key={item.preset}
-              action={() =>
+              action={e => {
                 this.setState(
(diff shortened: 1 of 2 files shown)


HeartSquared (approved):
Seems okay as a quick fix. Is it much more work to change it to use the new `Menu` component?

PR #2519, opened 2022-02-09
refactor: deprecate draft-button component

HeartSquared:
Pairing with @mcwinter07 

# Objective
- update storyboard
- update storybook to exclude deprecated stories
- begin deprecation of draft-button

# Motivation and Context
Transition draft components over to packages


draft-packages/button/KaizenDraft/Button/Button.tsx
```diff
@@ -1,13 +1,18 @@
...
   ButtonProps,
   ButtonRef,
 } from "./components/GenericButton"
 
 const Button = forwardRef(
-  (props: ButtonProps, ref: Ref<ButtonRef | undefined>) => (
-    <GenericButton {...props} ref={ref} />
-  )
+  (props: ButtonProps, ref: Ref<ButtonRef | undefined>) => {
+    logComponent(
```
HeartSquared, line 10: The `forwardRef` prevented us from using `withDeprecatedComponent` so we implemented the logger directly instead

storybook/main.ts
```diff
@@ -23,7 +23,7 @@ const getStoryPathsFromEnv = (): string[] | false => {
 const defaultStoryPaths = [
   "../packages/**/*.stories.tsx",
   "../packages/**/*.stories.mdx",
-  "../draft-packages/**/*.stories.tsx",
+  "../draft-packages/**/!(deprecated.)*.stories.tsx",
```
HeartSquared, line 26: This introduces a new pattern where we exclude stories that are prefixed with `deprecated.` from being loaded into storybook

mcwinter07 (approved):
Looks good to merge! 👍

PR #2521, opened 2022-02-10
feat: Updated the allowable React version range to include React 17

ActuallyACat:
# Objective
We're releasing this for two reasons: 
* We have been testing all of our components in React 17 for some time and have not come across any issues thus far. 
* We would like to roll all packages forward due to some issues publishing when Lerna was bumped to v4

### DIFF
--- a/draft-packages/avatar/package.json
+++ b/draft-packages/avatar/package.json
@@ -39,6 +39,6 @@
   },
   "peerDependencies": {
     "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^5.0.0 || ^6.0.0",
-    "react": "^16.9.0"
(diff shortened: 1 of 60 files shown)


mcwinter07 (commented):
I noticed that Hierarchical menu and select didn't get added because their react version is 16.13.1 - Going to do a 1 to 1 check to make sure anything else wasn't missed 👍

ActuallyACat:
> I noticed that Hierarchical menu and select didn't get added because their react version is 16.13.1 - Going to do a 1 to 1 check to make sure anything else wasn't missed 👍

Thank you! This prompted me to look closer and I found a couple more too (ignoring legacy-packages)

PR #2526, opened 2022-02-11
feat: DatePicker 

lloydstubber:
# Objective
- Build out the new DatePicker (single date) component using the Kaizen TextField and rendering react-day-picker inside of a Popper.js instance.

# Notes
- We've created an enum to map the numbered days of `disabledDaysOfWeek` for ease of use
- We've split the `disabledDays` object a part into separate props for the consumer. This will make it easier than them needing to structure it in the way [react-day-picker does it](http://react-day-picker.js.org/examples/disabled).
- The component is stateless, so the consumer handles the date and passes the value to it.
- We've hardcoded the format (to match design) for now. `en-US` is the format with the `,` and `en-GB` is just `12 Feb 2020`. At a later date when we upgrade and add `date-fns` to react-day-picker v8 we can have format as a prop.
- It isn't localised nor has LTR support - although has 80% of LTR baked in already. This will come with the DateRange component.
- We've done a few simple interactive tests to test the opening, closing and selecting of dates.
- Doug brought up being able to type into the input for accessibility reasons. We've decided against this for the first pass as we'd need to look at how best to handle the formatting of the dates on type vs. when it renders. The way it is built that would not be a hard feature to implement but for Team Act/Perform this isn't necessary right now.

# Design notes
- As the UI Kit is a little inconsistent with spacing on this one we've gone as close as possible while trying to use our own variables.
- We took the liberty of styling the disabled focused days, this may need finessing. Keen for feedback!

Mouse interactions:

https://user-images.githubusercontent.com/48232362/153533642-0fdaa163-1370-40fd-98ca-ccb1ea60376b.mov

Keyboard interactions:

https://user-images.githubusercontent.com/48232362/153533681-fb8a9dd1-0ef4-423c-988d-991a6f9d032f.mov

Disabled Dates:
- Disabled all dates before today
- disabled dates between 14/2/2022 and 16/2/2022
- disabled date 25/2/2022
image

# Checklist

# Additional Considerations


mcwinter07:
Just a quick flagon the docs tab - it looks like Storybook isn't smart enough to determine the minimum height for the  stories. What do you think about adding a decorator or wrapper to the component story with a min height? 

Screen Shot 2022-02-15 at 10 36 21 am

vs

Screen Shot 2022-02-15 at 10 59 38 am

packages/date-picker/package.json
```diff
@@ -0,0 +1,51 @@
...
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "react-day-picker": "^7.4.9",
+    "classnames": "^2.3.1",
+    "@kaizen/draft-form": "^5.2.2",
+    "@kaizen/component-library": "12.2.0",
+    "@kaizen/deprecated-component-library-helpers": "^2.5.0",
```
mcwinter07, line 39: Having a quick squiz through and I don't think the `@kaizen/deprecated-component-library-helpers` helpers are used anywhere? Could possibly drop it from dependencies if not?

lloydstubber: Good catch! Not sure when that got introduced, will remove.

jmitsios:
@lloydstubber Absolutely solid work — I think you've ticked off all the things we've flagged recently. 👍 Can't find anything else to complain about

dougmacknz:
Accessibility wise, there's an issue with having this as a text field that can't be used. Currently, a screen reader user lands on a text field, but it's read-only and the user has no idea that there's a dialog associated with it that must be used to input a date. Focus should be getting moved to the dialog, but it's not kosher to do that on an input focus, instead you want a button for this.

There's also the issue of forcing users to navigate a datepicker in order to input a date. For us that don't use a screen reader and can use a mouse, we might not think it's that hard, but for some people/situations—it may be considerably easier to input a date directly over having to navigate a date picker.

I highly recommend that we change the design slightly (CC @jmitsios) so that the calendar icon becomes an icon button, and the text field beside can be used without a date picker. This would follow the recommended [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html).

The tricky part becomes localisation and formatting. If we're rushing to get this out the door—I would just hard code this to `mm/dd/yyyy` for now, given how US-centric our customer base (and the internet as a whole) is, much like we default to US spelling for our copy. Then we can follow up later to see what we can do to localise it.

mcwinter07 (commented):
Just been playing around with keyboard nav and seems pretty solid. The one thing I will flag is that despite having the readonly prop passed in osx screen reader still announces "editable text" 🤔 @dougmacknz not sure if you've come into this issue before?

NVM I just read Doug's comment above *facepalms*

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    // Make sure calendar popup is not in the DOM
+    expect(screen.queryByRole("dialog")).not.toBeInTheDocument()
+
+    // Focus input and test calendar popup is showing
+    await act(async () => element.focus())
+    expect(element).toHaveFocus()
+    expect(screen.getByRole("dialog")).toBeVisible()
+  })
+
```
mcwinter07, line 29: Could also potentially add a test for focus out here too if you wanted to cover both bases :)

packages/date-picker/src/DatePicker/CalendarNav/CalendarNav.tsx
```diff
@@ -0,0 +1,35 @@
...
+  NavbarElementProps,
+  "onPreviousClick" | "onNextClick"
+>
+
+export const CalendarNav: React.VFC<CalendarNavProps> = ({
+  onPreviousClick,
+  onNextClick,
+}) => (
+  <div className={styles.navbar}>
+    <button
```
dougmacknz, line 18: Out of curiosity, did you explore using an `IconButton` here instead of a custom button + `Icon`?

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,224 @@
...
+  /** Accepts a DayOfWeek value to start the week on that day. By default,
+   * it's set to Monday.
+   */
+  firstDayOfWeek?: DayOfWeek
+
+  // Accepts a date to display that month on first render.
+  initialMonth?: Date
+
+  // The date passed in from the consumer that renders in the input and calendar.
+  selectedDate?: Date
```
dougmacknz, line 36: We may want this to be named `value` to be consistent with our other form components. WDYT?

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  // Event passed from consumer to handle the date on change.
+  onDayChange: (day: Date) => void
```
dougmacknz, line 39: Similar to above, we might want this to just be `onChange` for consistency with our other components.

packages/date-picker/docs/DatePicker.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+      id="date-picker"
+      labelText="Label"
+      selectedDate={selectedDate}
+      onDayChange={onDayChange}
+      isDisabled
+    />
+  )
+}
+
+export const DisabledInputWithValue = () => {
```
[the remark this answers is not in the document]
lloydstubber: Old habits die hard! Want me to do a sticker sheet instead? Or for the next round of changes?

packages/date-picker/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@kaizen/date-picker",
```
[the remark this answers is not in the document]
lloydstubber: I did think that but thought once range is in here it'll be a pretty big package.

packages/date-picker/src/DatePicker/CalendarNav/CalendarNav.tsx
lloydstubber: Yep we did look at that too start with, but it's too big and design didn't want a super small version in the UI kit.

packages/date-picker/src/DatePicker/DatePicker.tsx
lloydstubber: Yeah good catch. Updated now!

dougmacknz (approved):
🚢

jmitsios:
> I highly recommend that we change the design slightly (CC @jmitsios) so that the calendar icon becomes an icon button, and the text field beside can be used without a date picker. This would follow the recommended [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html).

@dougmacknz — So the recommendation being that we split them into two seperate assets, with the Calendar icon being its own standalone button, sitting beside the text field?

I'm happy to look into that and explore it if it's an accessibility consideration, but, I'd just need to question what purpose the text field would serve on its own, if we're not allowing the user to input their own dates at the moment.  It feels like in this instance the text field would merely be a display of the users selection from the calendar?

dougmacknz:
> So the recommendation being that we split them into two seperate assets, with the Calendar icon being its own standalone button, sitting beside the text field?

> I'm happy to look into that and explore it if it's an accessibility consideration, but, I'd just need to question what purpose the text field would serve on its own, if we're not allowing the user to input their own dates at the moment. It feels like in this instance the text field would merely be a display of the users selection from the calendar?

You'd only do that alongside allowing the user to input the date. @lloydstubber and I had a discussion and decided that we'll start with the whole thing as a button for the first pass (no ability to input dates manually), and then reassess at a future time. So I'd say hold off for now on any investigation on making the icon a button.

PR #2527, opened 2022-02-11
feat: Install eslint-plugin-jsx-a11y

dougmacknz:
# Objective
Installs [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) and remediates/ignores any issues after implementing the recommended set of rules.

# Motivation and Context
Helps to catch some common a11y anti-patterns. The main ones I came across while remediating the issues that surfaced here were:
* onClick being put onto non-interactive elements (div)
* Images with no/obsolete alt text
* Links being used where it should be a button

### DIFF
--- a/draft-packages/button/docs/deprecated.Button.stories.tsx
+++ b/draft-packages/button/docs/deprecated.Button.stories.tsx
@@ -987,6 +987,8 @@ const MultipleButtons = args => (
 
 const CustomComponent = args => {
   const CustomLink = (buttonProps: CustomButtonProps) => (
+    // Disabling because it's a false positive
+    // eslint-disable-next-line jsx-a11y/anchor-has-content
     <a href={buttonProps.href} {...buttonProps} {...args} />
   )
   // ^ In actual usage - this would be a react-router <Link> component or similar
--- a/draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.tsx
@@ -97,7 +97,7 @@ export const AnimatedBase = ({
     </svg>
   )
 
-  const FailedState = <img src={assetUrl(fallback)} aria-hidden={true} />
+  const FailedState = <img src={assetUrl(fallback)} alt={alt} />
 
   return (
     <figure className={wrapper}>
--- a/draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
+++ b/draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
@@ -113,6 +113,7 @@ export const StatelessMenu: React.FunctionComponent<StatelessMenuProps> = ({
   ) : null
 
   return (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
     <div data-automation-id={automationId} onClick={onClick}>
       <div className={styles.buttonWrapper} ref={setReferenceElement}>
         {menuButton}
--- a/draft-packages/popover/docs/Popover.stories.tsx
+++ b/draft-packages/popover/docs/Popover.stories.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
 import React, { useState } from "react"
 import { usePopover, Popover as PopoverRaw } from "@kaizen/draft-popover"
 import { withDesign } from "storybook-addon-designs"
@@ -84,11 +85,10 @@ const InlineBlockTargetElement = ({
       display: "inline-block",
       marginTop: "100px",
     }}
-    onClick={openPopover}
     onMouseEnter={openPopover}
     onFocusCapture={openPopover}
   >
-    <IconButton label="Label" icon={informationIcon} />
+    <IconButton label="Label" icon={informationIcon} onClick={openPopover} />
   </div>
 )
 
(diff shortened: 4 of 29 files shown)


draft-packages/menu/KaizenDraft/Menu/StatelessMenu.tsx
```diff
@@ -113,6 +113,7 @@ export const StatelessMenu: React.FunctionComponent<StatelessMenuProps> = ({
   ) : null
 
   return (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
     <div data-automation-id={automationId} onClick={onClick}>
```
dougmacknz, line 117: Bit of a weird one, this `onClick` prop is only used twice (both within Kaizen) and it's only used to `stopPropagation`. I'm just going to ignore this, because I have plans to rewrite Menu :)

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -372,6 +372,7 @@ export const TableRowCell: TableRowCell = ({
 }) =>
   href != null ? (
     <a
+      // eslint-disable-next-line jsx-a11y/no-interactive-element-to-noninteractive-role
```
dougmacknz, line 375: This component is a mess so I'm just not gonna bother trying with this one

draft-packages/button/docs/deprecated.Button.stories.tsx
```diff
@@ -987,6 +987,8 @@ const MultipleButtons = args => (
 
 const CustomComponent = args => {
   const CustomLink = (buttonProps: CustomButtonProps) => (
+    // Disabling because it's a false positive
```
lloydstubber, line 990: How's it a false positive?

draft-packages/illustration/KaizenDraft/Illustration/Players/LottiePlayer.tsx
```diff
@@ -97,7 +97,7 @@ export const AnimatedBase = ({
     </svg>
   )
 
-  const FailedState = <img src={assetUrl(fallback)} aria-hidden={true} />
+  const FailedState = <img src={assetUrl(fallback)} alt={alt} />
```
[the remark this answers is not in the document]
mcwinter07: True, whether its animated or not the alt should still help describe the image :)

draft-packages/popover/docs/Popover.stories.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
lloydstubber, line 1: Why the whole file?

packages/design-tokens/docs/DocsComponents.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
lloydstubber, line 1: What's the TLDR on why these are in a few files at the top?

site/src/components/NavigationBar/components/Link.tsx
```diff
@@ -34,13 +34,13 @@ class Link extends React.PureComponent<LinkProps> {
     } = this.props
 
     return (
+      // eslint-disable-next-line jsx-a11y/click-events-have-key-events
```
lloydstubber, line 37: What is this one taking care of?

lloydstubber (commented):
LGTM just a few questions on what we're disabling in places

draft-packages/button/docs/deprecated.Button.stories.tsx
dougmacknz: It thinks the link is going to have no content, i.e. `<a href="http://www.google.com">[nothing here]</a>`, but it will end up with content eventually because of how it's being used (passed through to the `component` prop on Button).

draft-packages/popover/docs/Popover.stories.tsx
dougmacknz: Because it's a stories file, and I think it's fine to do fake links (`href="#"`) for any stories.

site/src/components/NavigationBar/components/Link.tsx
dougmacknz: This old component has an `onClick` prop that gets used on a link, which is a no-no. What makes it tricky to fix is that it also has an `href` prop, so it can't just be blanket changed to a `button`. I'd have to switch the element based on what was passed in, and I don't think it's worth the effort for this legacy stuff.

packages/design-tokens/docs/DocsComponents.tsx
mcwinter07: Was curious about that too, I'm assuming the pattern is if its at the top it applies to the whole component? @dougmacknz is my read on that right?

dougmacknz: That's right, this will disable the rule for the whole file. Also notice it's `eslint-disable` instead of `eslint-disable-next-line`.

lloydstubber: Ah my bad, I know they're disabling the whole file. I just meant why are there so many disabling the whole file?

dougmacknz: Oh I was replying to @mcwinter07. Hmm, on second look this one should just be on the line, but for the others it's really common to do invalid links in stories so I just added it to the top once instead of having to do it for each line.

lloydstubber (approved):

PR #2535, opened 2022-02-16
docs: audit notification update

jmitsios:
• Deleting old Figma embed for usage example
• Added 3 new usage examples for Inline, Global, Toast


HeartSquared:
>Deleting old Figma embed for usage example

@jmitsios nothing looks deleted, am I misunderstanding the changes I'm looking for?

HeartSquared:
image

Looking at the preview, it seems like the Figma files might need unrestricted access, unless the intention is that it is hidden for outsiders?

dlimiter (approved):

jmitsios:
> > Deleting old Figma embed for usage example
> 
> @jmitsios nothing looks deleted, am I misunderstanding the changes I'm looking for?

Thanks for calling that out! :) I should have expanded on that comment a little further to say that there was an older Figma embed pointing to a deprecated example of usage, which I deleted.  It's now been replaced by new, and up-to-date examples instead.

jmitsios:
> image
> 
> Looking at the preview, it seems like the Figma files might need unrestricted access, unless the intention is that it is hidden for outsiders?

Could you shed any light on this @katsambarcus ? Is this the intended use?

katsambarcus:
@jmitsios @HeartSquared  we've previously tried offering unrestricted access so that everyone outside of Culture Amp could view the Figma Embeds, but sadly Figma's restrictions aren't fine-grain enough, and what we were experiencing was an avalanche of non-Campers requesting edit access. I worked with Figma support to see if this was solvable, but it currently isn't, so I had to shut it down. 

For now, Figma Embeds are only viewable if you login with your CA credentials until Figma can improve the controls of what an outsider can see, but not touch.

HeartSquared (approved):
LGTM :)

PR #2538, opened 2022-02-16
feat: duplicate and move draft loading spinner to packages

mcwinter07:
# Objective
Duplicate draft spinner, update package structure and retain git history.

# Motivation and Context
This is step two in deprecating draft-packages


packages/loading-spinner/LoadingSpinner/src/LoadingSpinner.tsx
```diff
@@ -0,0 +1,79 @@
...
+type size = "sm" | "md"
+
+export interface LoadingSpinnerProps {
+  accessibilityLabel: string
+  /**
+   * Generally use "md" unless spinner is inside a form field
+   * @default "md"
+   */
+  size?: size
+  children?: React.ReactNode
```
HeartSquared, line 13: Perhaps to fix later, but it looks like `children` doesn't do anything for the component so we should be able to remove it (but technically it's a breaking change I guess)

HeartSquared:
We can probably remove the empty CHANGELOG (can't comment on it since it's empty 🙃 )

packages/loading-spinner/LoadingSpinner/src/LoadingSpinner.tsx
mcwinter07: Yeah I was thinking that was an odd one, was going to confirm with peeps before axing it but will definitely do it in a separate PR :)

HeartSquared (approved):

PR #2540, opened 2022-02-17
fix: deprecate loading spinner draft-package

mcwinter07:
# Objective
* add deprecate console and doc block warning
* add deprecate notation to stories and update descriptions
* update storyboard to use deprecated notation

# Motivation and Context
This is the first step in deprecating the loading spinner and moving it across to draft packages


draft-packages/loading-spinner/KaizenDraft/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -77,3 +78,11 @@ export const LoadingSpinner = ({
     )}
   </div>
 )
+
+/**
+ * @deprecated draft-packages LoadingSpinner is deprecated. Please use Button from "@kaizen/loading-spinner" instead.
```
HeartSquared, line 83: Hmm...shouldn't this still go above the component itself instead of this wrapper?

mcwinter07: vs
Screen Shot 2022-02-17 at 1 30 51 pm

mcwinter07: Seemed strange me but I opted to go with this method since it provided the most feedback to the consumer - happy for recommendations otherwise :) @HeartSquared

dougmacknz: Also, "Please use Button"—copy pasta? 😄

mcwinter07: hahahha - yep! well I needed another commit to fix my moai issue!

HeartSquared: All good :) Although it makes me wonder if we could have the docblock as part of the HOC 🤔

mcwinter07: A Docblock in a HOC - Sounds almost Doctor Seuss-esque 😆 We could though, however there are instances where we'll still have to add them in separately like in Button that doesn't used the withDeprecatedComponent

HeartSquared: Of course, but those should be more rare instances so it should be okay :)
Although now that I think of it, can we actually put a param from the HOC in the comment? 🤔 Maybe it's not actually possible to do

draft-packages/loading-spinner/docs/deprecated.LoadingSpinner.stories.tsx
```diff
@@ -8,13 +8,13 @@ import { figmaEmbed } from "../../../storybook/helpers"
 import { CATEGORIES } from "../../../storybook/constants"
 
 export default {
-  title: `${CATEGORIES.components}/Loading Spinner`,
+  title: `${CATEGORIES.components}/Loading Spinner (deprecated)`,
```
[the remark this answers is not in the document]
dougmacknz: For other cases where you do want to keep deprecated stories, there is a `CATEGORIES.deprecated` you can use here instead of `CATEGORIES.components` to group deprecated stories.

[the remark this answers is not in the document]
HeartSquared: We're actually keeping the files to keep the historical structure. We added some config so that files we rename to start with `deprecated.` are actually excluded from storybook and chromatic :)

draft-packages/loading-spinner/KaizenDraft/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: I'd be keen to try just not within this PR, since we'd be changing the `withDeprecatedComponent` :) Might create a separate ticket to investigate we can go back in an make an amendment to the deprecated files

HeartSquared, line 83: It still says Button 😆 

```suggestion
 * @deprecated draft-packages LoadingSpinner is deprecated. Please use LoadingSpinner from "@kaizen/loading-spinner" instead.
```

draft-packages/loading-spinner/index.ts
```diff
@@ -1 +1 @@
-export * from "./KaizenDraft/LoadingSpinner/LoadingSpinner"
+export { default as LoadingSpinner } from "./KaizenDraft/LoadingSpinner/LoadingSpinner"
```
HeartSquared, line 1: This no longer exports the props. Would we be able to just directly place the HOC around the component in L17? Although it's possible it might be less readable doing that, in which case we will need to add the props export in here.
Alt, rename the component itself (`LoadingSpinnerComponent`?) which you pass into the HOC and have that be exported as `LoadingSpinner`.

draft-packages/loading-spinner/KaizenDraft/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: I could have sworn I commited this and pushed it up... Must have been a fever dream 😢

draft-packages/loading-spinner/index.ts
mcwinter07: I might opt for the second options for readability, I tried the first one but it wasn't the prettiest when it came to readability

HeartSquared (approved):
LGTM :)

PR #2543, opened 2022-02-18
feat: convert date picker input to button for better a11y support

lloydstubber:
# Objective
- Change the current DatePicker input to be a button, as it's currently an input a user cannot type in. Making it a strange pattern for screen readers.
- Update structure and split a part component to suit our new guidelines - although Calendar and CalendarNav sit at the same level, I could nest but definitely don't want to split it out any further.
- A lot of this work is around trapping the focus of the popup when it opens and focusing on the current day _or_ selected day.
- WIP: I'm still adding more unit tests to this, mainly just around the focus functionality I've added.

# Motivation and Context
- This is an in between step of having an input you can type in (but would involve a lot more work) to satisfy our a11y needs

# Checklist

# Additional Considerations


jmitsios:
@lloydstubber 
The disabled input state is still showing a hand-cursor when hovering over it _(can't show the cursor in the screenshot)_
image
Our regular Disabled button states keep the cursor as an arrow instead, so I'm guessing we should be mimicking that interaction style.
image

dougmacknz:
@lloydstubber I'd remove any cursor CSS and stick with the default

packages/date-picker/src/DatePicker/components/Calendar/Calendar.spec.tsx
```diff
@@ -0,0 +1,47 @@
+// import { act, render, screen } from "@testing-library/react"
```
christian314159, line 1: Is this test obsolete now? Or going to be re-enabled later?

lloydstubber: Writing these now! Didn't mean for this to slip in.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,140 @@
...
+}) => {
+  const calendarRef = useRef<HTMLDivElement>(null)
+
+  const getNavbar = ({ ...navbarProps }: CalendarNavProps) => (
+    <CalendarNav {...navbarProps} />
+  )
+
+  useEffect(() => {
+    if (value) {
+      if (calendarRef && calendarRef.current) {
```
christian314159, line 71: Looks like `if (calendarRef && calendarRef.current)` is a condition for both the `if` and the `else` block, you should be able tp pull it out as a guard clause instead and remove the nestes if statement.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  it("show focus on selected day", async () => {
+    render(<Calendar {...defaultProps} />)
+
+    const selectedDate = screen.getByRole("gridcell", {
+      name: "Tue Mar 01 2022",
+    })
+
+    expect(selectedDate).toHaveFocus()
+  })
+})
```
lloydstubber, line 27: I also wanted to write one for focus on `today` when there's no value but that will change every day.

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -149,71 +144,66 @@ export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
...
+
+  const handleReturnFocus = () => {
+    if (wrapperRef && wrapperRef.current) {
+      const buttonElement = wrapperRef.current.getElementsByClassName(
+        datePickerStyles.button
+      )[0] as HTMLElement
+
+      buttonElement?.focus()
+    }
+  }
```
lloydstubber, line 159: Frustratingly (is that a word?) the `returnFocus` prop in `react-focus-lock` doesn't work out of the box all the time (and people have complained) so it accepts a function.

Also, I would've loved to use `buttonRef` here but as it's something a consumer can pass in, I've had to do the outer wrapper.

packages/date-picker/src/DatePicker/components/CalendarNav/CalendarNav.tsx
```diff
@@ -3,7 +3,7 @@ import { Icon } from "@kaizen/component-library"
 import arrowRight from "@kaizen/component-library/icons/arrow-right.icon.svg"
 import arrowLeft from "@kaizen/component-library/icons/arrow-left.icon.svg"
 import { NavbarElementProps } from "react-day-picker/types/Props"
-import styles from "../DatePicker.scss"
+import styles from "../../DatePicker.scss"
```
dougmacknz, line 6: WDYT about creating a new scss file for the styles that are used for this component? That's generally what we don't end up with one unruly scss file.

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -149,71 +145,63 @@ export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
...
-            onDayClick={handleOnDayChange}
-            navbarElement={getNavbar}
-            classNames={defaultDatePickerClasses}
-            onKeyDown={e => handleKeyDown(e)}
+            disabledDates={disabledDates}
+            disabledDaysOfWeek={disabledDaysOfWeek}
+            disabledRange={disabledRange}
+            disabledBefore={disabledBefore}
+            disabledAfter={disabledAfter}
+            disabledBeforeAfter={disabledBeforeAfter}
```
dougmacknz, line 200: If you wanted to remove the amount of props you have to send and receive, you could write a util that takes all of these and returns the array that you end up passing into DayPicker's `disabledDays` prop. Then this would just become a single `disabledDays` prop.

packages/date-picker/src/utils/handleDisableDays.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { DayOfWeek } from "../DatePicker/DatePicker"
+import { handleDisabledDays } from "./handleDisabledDays"
+
+beforeEach(() => {
+  // Temporarily allow us to alter timezone calculation for testing
+  Date.prototype.getTimezoneOffset = jest.fn(() => 0)
```
lloydstubber, line 6: This sets it to UTC so we can make sure the mock matches

packages/date-picker/src/DatePicker/components/CalendarNav/CalendarNav.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.arrows {
+  @include button-reset;
+  display: inherit;
+  padding: 0.25rem $spacing-xs;
+  border: $border-solid-border-width $border-solid-border-style transparent;
+  border-radius: $border-radius;
+
+  &:hover {
+    cursor: pointer;
```
dougmacknz, line 27: Missed this one (no need for changing the cursor)

packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -3,119 +3,130 @@
 @import "~@kaizen/design-tokens/sass/spacing";
 @import "~@kaizen/design-tokens/sass/border";
 @import "~@kaizen/design-tokens/sass/shadow";
-@import "~@kaizen/draft-button/KaizenDraft/Button/styles";
 
 // Vars
-$border-radius: 3px;
+$border-radius: 7px;
```
dougmacknz, line 8: Design token you can use instead `$border-solid-border-radius`

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -0,0 +1,91 @@
+@import "~@kaizen/design-tokens/sass/typography";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/shadow";
+
+// Vars
+$border-radius: 7px;
```
dougmacknz, line 8: Design token you can use instead `$border-solid-border-radius`

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }
+      | undefined
+  }
+  classNameAndIHaveSpokenToDST?: string
+  value?: Date
+  initialMonth?: Date
+  firstDayOfWeek: number
+  disabledDays?: Modifier | Modifier[]
+  handleOnDayChange: (day: Date, modifiers: DayModifiers) => void
+  handleKeyDown: (e: any) => void
```
dougmacknz, line 26: Given these are props, I believe we agreed on `on` being the prefix instead of `handle`

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -0,0 +1,91 @@
...
+    line-height: $typography-heading-4-line-height;
+    letter-spacing: $typography-heading-4-letter-spacing;
+    color: $color-purple-800;
+    margin: 0 0 $spacing-xs;
+  }
+
+  .weekday {
+    font-size: $typography-paragraph-body-font-size;
+    font-weight: $typography-paragraph-body-font-weight;
+    line-height: $typography-paragraph-body-line-height;
```
dougmacknz, line 44: Usually when we use these typography tokens, we use all of them (with the odd case where the `line-height` doesn't work well). In order words, this one is missing `letter-spacing`, `font-family` and possibly `line-height` (and there's other similar cases in this file).

It may not seem important right now because everything that's not `display-0` has Inter as the font-family, but if we ever decide to change fonts for headings but leave Inter for paragraph, this will end up with the heading font family.

packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -3,119 +3,130 @@
...
+$button-height: 48px;
+$button-base-padding-horizontal: $spacing-sm;
+$button-icon-size: 1.25rem; // 20px
+
+$disabled-opacity: 0.5;
```
dougmacknz, line 13: Usually we use 0.3 for disabled opacity. I haven't had a good look but just want to make sure this is right?

packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -3,119 +3,130 @@
...
-      outline: $color-blue-500 $border-focus-ring-border-style
-        $border-focus-ring-border-width;
-    }
-  }
-
-  .month {
-    margin: $spacing-md;
-    border-collapse: separate;
-    border-spacing: 1px;
+  &:focus:not([disabled]) {
```
dougmacknz, line 55: This can be addressed later if we want, but usually we use `focus-visible` instead of `focus` so that the border/outline only shows when navigating with a keyboard (this will show on click).

We have a polyfill that adds a `focus-visible` class to elements when they are focused using a keyboard (Safari doesn't support :focus-visible). You just need to use `:global(.focus-visible)` to target so that you target the class name as it is, instead of a CSS Module uniquified class name.

packages/date-picker/src/DatePicker/components/CalendarNav/CalendarNav.scss
```diff
@@ -0,0 +1,37 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/draft-button/KaizenDraft/Button/styles";
```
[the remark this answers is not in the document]
lloydstubber: Yep no worries. As in a `_mixins.scss` file in this package?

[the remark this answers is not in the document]
dougmacknz: Yup

packages/date-picker/src/DatePicker/DatePicker.scss
lloydstubber: Yeah this is correct, the `InputSearch` has the same. It's a bit weird but the calendar icon is less opaque than the rest so you need to split it up.

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -80,7 +81,7 @@ export enum DayOfWeek {
 
 export const DatePicker: React.FunctionComponent<DatePickerProps> = ({
```
dougmacknz, line 82: As per the recent discussions about using VFC instead of FC

```suggestion
export const DatePicker: React.VFC<DatePickerProps> = ({
```

packages/date-picker/src/DatePicker/DatePicker.scss
lloydstubber: Is this for every instance or just this line? We're using focus quite a bit in combination with hover etc.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { Dispatch, SetStateAction, useEffect, useRef } from "react"
+import DayPicker from "react-day-picker/DayPicker"
+import cx from "classnames"
+import { NavbarElementProps } from "react-day-picker/types/Props"
+import { Modifier, DayModifiers } from "react-day-picker/types/Modifiers"
+import { CalendarNav } from "../CalendarNav/CalendarNav"
+import { defaultCalendarClasses } from "./CalendarClasses"
+import calendarStyles from "./Calendar.scss"
+
+export type CalendarProps = {
```
dougmacknz, line 10: Might want to extend native div props here, and use `style` instead of a custom `styles` prop?

packages/date-picker/src/DatePicker/DatePicker.scss
dougmacknz: Every instance

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
lloydstubber: I went down this road when I was originally trying to type everything after moving the calendar into its own file. The problem is this isn't a native style, it's created by Popper in the parent and passed down, so I needed to type it accordingly.

packages/date-picker/src/mixins/_button-reset.scss
```diff
@@ -0,0 +1,10 @@
+@mixin button-reset {
```
dougmacknz, line 1: Sorry I meant a `_mixins.scss` file rather than a file for each mixin, see https://cultureamp.slack.com/archives/C02NUQ27G56/p1645591324648029. But we can adjust this later if you want

packages/date-picker/src/utils/handleDisabledDays.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { DayOfWeek } from "../DatePicker/DatePicker"
+import { daysToNumbers } from "./daysToNumbers"
+
+/**
+ * To save the consumer from passing an array of disabledDay options (each being a different object).
+ * We have split the options out so the consumer can pass in an individual disabledDay option for ease of use.
+ */
+
+export const handleDisabledDays = (
```
dougmacknz, line 15: Somewhat of a nit pick, but I think `handle` is the wrong word here. `calculateDisabledDays`. `determineDisabledDays`, or something like that would make more sense IMO

dougmacknz (approved):
Couple of optionals but otherwise looking good 👍

dougmacknz:
Oh one more thing—can you make sure that each component has a `displayName`?

dougmacknz (approved):

PR #2546, opened 2022-02-20
refactor(deps): update dependencies from draft button to packages button

mcwinter07:
Warning:
Do not merge this until the button package has been update to version 1.0.0

# Objective
- Update all import and package.json for Kaizen components to use the new packages button instead of the draft

# Motivation and Context
With draft button deprecated there are several Kaizen components that use the old package. These need to be update to use the new button to clean up the console warnings and proceed with the deprecation tasks.


dougmacknz (dismissed):

draft-packages/badge/docs/Badge.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react"
 
-import { Button } from "@kaizen/draft-button"
+import { Button } from "@kaizen/button"
```
HeartSquared, line 3: Why does this (and a few others) not have a `package.json` update?

draft-packages/select/package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "@kaizen/component-library": "^12.4.2",
     "@kaizen/deprecated-component-library-helpers": "^2.5.0",
-    "@kaizen/draft-button": "^5.6.2",
+    "@kaizen/button": "^1.0.1",
```
HeartSquared, line 37: This one is interesting. Seems like it's only used in the `.elm` and not in the main component 🤔
```elm
-- L43
import KaizenDraft.Button.Button as Button
```

The options I see for this would be:
1. Remove the dependency if the elm file does not use it
2. Update the elm file to point to the new Button
3. Revert this change in `package.json` and leave a comment that it's only used for elm

packages/button/docs/IconButton.stories.tsx
```diff
@@ -18,7 +18,7 @@ export default {
     docs: {
       description: {
         component:
-          'import { IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen/draft-button".',
+          'import { IconButton } from "@kaizen/button". This IconButton supersedes "@kaizen/button".',
```
HeartSquared, line 21: Yeah we definitely don't want to change this one 😂

packages/component-library/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@kaizen/deprecated-component-library-helpers": "^2.5.0",
-    "@kaizen/draft-button": "^5.6.2",
+    "@kaizen/button": "^1.0.1",
```
HeartSquared, line 35: I think we might need to dig into this one. The usages of `Button` don't appear to be actual usages, but there's certainly button-ish references within the files (eg. the readme)

HeartSquared (changes requested):

packages/button/docs/IconButton.stories.tsx
mcwinter07: omg yep, that would not have be smart, nice catch🧤

draft-packages/filter-menu-button/package.json
```diff
@@ -33,7 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-library": "^12.4.2",
-    "@kaizen/draft-button": "^5.6.2",
+    "@kaizen/button": "^1.0.1",
```
mcwinter07, line 36: No elm components so should be be good to bump

draft-packages/guidance-block/package.json
```diff
@@ -33,7 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-library": "^12.4.2",
-    "@kaizen/draft-button": "^5.6.2",
+    "@kaizen/button": "^1.0.1",
     "@kaizen/draft-illustration": "^3.6.3",
```
mcwinter07, line 37: Should be fine since there's no elm version

draft-packages/menu/package.json
```diff
@@ -33,7 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-library": "^12.4.2",
-    "@kaizen/draft-button": "^5.6.2",
+    "@kaizen/button": "^1.0.1",
```
mcwinter07, line 36: Should be fine since there's no elm version

draft-packages/select/package.json
mcwinter07: Based off the catch up today I'll add both in this pattern:

```
    "@kaizen/button": "^1.0.1",
    "//": "The draft dependency should remain here until this is moved out of draft packages ",
    "@kaizen/draft-button": "^5.6.2",
```

HeartSquared: I don't believe this will be treated nicely (apparently things explode; I guess it will try to install a `//` package with this string version).

Here's the pattern from root's `package.json` :)

```
  "dependencies": {
    "node-sass": "6.0.1",
  },
  "dependenciesComments": {
    "node-sass": "version locked to keep consistency with gatsby site",
  },
```

draft-packages/badge/docs/Badge.stories.tsx
```diff
@@ -1,5 +1,5 @@
-import React from "react"
-import { Button } from "@kaizen/draft-button"
+import * as React from "react"
```
HeartSquared, line 1:
```suggestion
import React from "react"
```

mcwinter07: ah drat, I think that got override in a conflict change, will fix that up 👍

draft-packages/guidance-block/package.json
```diff
@@ -33,13 +33,17 @@
...
+    "@kaizen/button": "^1.0.4",
     "@kaizen/draft-button": "^5.6.5",
     "@kaizen/draft-illustration": "^3.6.5",
     "@kaizen/draft-tooltip": "^3.7.3",
     "@types/classnames": "^2.3.1",
     "classnames": "^2.3.1",
     "react-media": "^1.10.0"
   },
+  "dependenciesComments": {
+    "@kaizen/draft-button": "The draft dependency should remain here until this is moved out of draft packages"
```
HeartSquared, line 45: I suggest mentioning the purpose of this is for Elm.

```suggestion
    "@kaizen/draft-button": "The draft dependency should remain here (for Elm) until this is moved out of draft packages"
```

draft-packages/menu/package.json
```diff
@@ -33,6 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-library": "^12.4.5",
+    "@kaizen/button": "^1.0.4",
     "@kaizen/draft-button": "^5.6.5",
```
HeartSquared, line 37: Missing a comment for this one? 🤔

draft-packages/tile/package.json
```diff
@@ -37,7 +37,10 @@
   },
   "dependencies": {
     "@kaizen/component-library": "^12.4.5",
-    "@kaizen/draft-button": "^5.6.5",
+    "@kaizen/button": "^1.0.4",
     "classnames": "^2.3.1"
+  },
+  "dependenciesComments": {
+    "@kaizen/draft-button": "The draft dependency should remain here until this is moved out of draft packages"
```
HeartSquared, line 44:
```suggestion
```

_Nothing to see here (there's no draft-button)._ 👀

HeartSquared (changes requested):

draft-packages/tile/package.json
mcwinter07: This is not the draft-button you're looking for ⭐ 🧙 🤖 Consider it gone 😆

draft-packages/guidance-block/package.json
mcwinter07: I'm going to opt for the latter, short and concise

draft-packages/menu/package.json
mcwinter07: No I think that one just should have a draft button, will remove it 👍

HeartSquared (approved):
Seems legit 📦

HeartSquared (approved):

PR #2547, opened 2022-02-21
fix: Changed icon color for `validationPositive` variant of tag to darker green [KDS-220]

christian314159:
### Issue

Icon for `validationPositive` variant of `Tag` was much lighter in comparison than the icons of other variants.

image

### Solution

Changed it to darker color ;)


HeartSquared (approved):
LGTM :)

PR #2548, opened 2022-02-21
fix: give calendar popup a z-index and background-color

lloydstubber:
# Objective
- Fix issue where the background was transparent on popup and not the highest element

# Motivation and Context
image

# Checklist

# Additional Considerations

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.scss
+++ b/packages/date-picker/src/DatePicker/DatePicker.scss
@@ -9,7 +9,9 @@
 $border-radius: 3px;
 
 .calendar {
+  background-color: $color-white;
   font-family: $typography-heading-4-font-family;
+  z-index: 1;
 
   .wrapper {
     padding-bottom: 0;


packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -9,7 +9,9 @@
 $border-radius: 3px;
 
 .calendar {
+  background-color: $color-white;
   font-family: $typography-heading-4-font-family;
+  z-index: 1;
```
lloydstubber, line 14: Interested to know if we do this for any other popups or if it could introduce bugs?

HeartSquared: Any possibility of adding some sort of test/verification for this? Maybe even just a story.

lloydstubber: Yep, can do. Just to clarify though: a story that has the calendar open by default over the top of another element that has a `z-index: 0` so Chromatic can check it?

HeartSquared: Hmm... Mostly yes, although I'm wary of just catering for `z-index: 0` since our teams might actually give the other elements a higher z-index right? Perhaps we should do that crazy thing where we set our z-index to 9999 just to be above everything else constantly (unless you can think of a case where we wouldn't want that? It does make me wonder why it's not something out of the box)

lloydstubber: Yeah I was wondering how we generally handle the tooltip/modal etc. with z-indexes. I don't see any current tests for them but there is a `layers.scss` file with this in there:
```
$ca-z-index-body: 0;
$ca-z-index-dropdown: 1000;
$ca-z-index-sticky: 1020;
$ca-z-index-fixed: 1030;
$ca-z-index-modal-backdrop: 1040;
$ca-z-index-modal: 1050;
$ca-z-index-popover: 1060;
$ca-z-index-tooltip: 1070;
```

@dougmacknz what's your take on how high the index should be?

lloydstubber: Sidenote, the datepicker handles its own open/closing internally so our Chromatic is just rendering the input, not the calendar. So we can't test the z-index currently.

I think this PR should be the two small fixes once we decide on what index to set it to - as perf is waiting on the update - and then I'll add another card for the open/close prop along with some more unit and visual tests to cover our bases.

HeartSquared: >Sidenote, the datepicker handles its own open/closing internally so our Chromatic is just rendering the input, not the calendar. So we can't test the z-index currently.

Oh that's a shame. Hopefully there's actually a prop somewhere we could utilise which tells it to be always open for us.

Happy to approve this to get it in first as long as the z-index of 1 is enough for them (otherwise just up it).

HeartSquared (approved):

packages/date-picker/src/DatePicker/DatePicker.scss
dougmacknz: Yeah in the past we've just used that `layers.scss` file from `component-library`, which hasn't been perfect but got us by for the most part. Unless we can think of a better idea, we might want to create the same sort of thing in `design-tokens`.

PR #2549, opened 2022-02-21
docs: add package/component recipes and code standards

HeartSquared:
# What

Add docs for package structure, a basic component, and other general code standards.

[The easier to read version here :)](https://github.com/cultureamp/kaizen-design-system/tree/KDS-198/docs--code-standards/docs)

# Why

Docs based on our discussion #2413

# Notes

- I copied the `package.json` and `tsconfig.dist.json` templates from the old generator
- I have currently left the `docs` folder structure for stories, however I left a suggestion of what we might be able to do to remove it in the future (so that the stories can sit with their components)
- Since Doug wasn't a fan of destructuring props, I made it a loose rule (still added it as a default)

## Next steps

- [ ] Create a Confluence page and link to these docs

Next discussion topics/tasks (in no particular order):
- Figure out our patterns for testing (and whether we can get rid of snapshots)
- How many props on a component are too many?
- What do we want our docs to look like?
- Stricter linting rules? (eg. do not allow `any`)
- Recipe for dot notation components
- Recipe for generic components
	- Do we want to prefix the generic with `T` within the usage? (I find it increases readability)
- Component generator
- What about generic utils/types which don't belong specifically to the component?
- `const` > `let` (prevent reassignment for clearer understandability of usage for the variable)
	- https://github.com/jonaskello/eslint-plugin-functional/blob/master/docs/rules/no-let.md


docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+
+Subcomponents live in the `components/` directory and should not have subcomponents of their own. If you are finding that you want subcomponents for your subcomponents, then one (or both) should be a component instead.
+
+Aside from the above, subcomponents should follow the same structure as a normal component.
+
+Subcomponents living within their own directories have the benefits of keeping files small and clean, and allows for mocking for tests.
+
+### Hooks and Utils
+
+Hooks and Utils go into their own respective folders, with their filename reflective of either the function name or a collective name based on the functionality of functions within the file.
```
lloydstubber, line 62: nit: I don't think we need to capitalise utils here

lloydstubber (approved):
Nice one! This looks great. Just a small nit but happy for this to go in once there's some more eyes on it.

docs/basic-component.md
HeartSquared: Fixed :)

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+
+Style files should named to match the component name (eg. the Pancake component will be named `Pancake.scss`) and live in the same directory.
+
+Keep these clean by separating styles for subcomponents in their own respective files.
+
+### Tests
+
+Test files should named to match the component/function name (eg. `Pancake.spec.tsx` and `functionName.ts`) and live in the same directory.
+
+Ideally we would want to avoid snapshot tests as they do not have meaningful assertions/expectations and are very fragile (they can break with insignificant changes). We should instead write unit and functional tests.
```
hqta, line 92: Just wondering if we can have guidelines on what we should be testing here. Eg. interaction testing, accessibility, style changes associated with functionality

Also on how we write tests to be standardised Eg. Given, when, then

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+  - You don't need to extend HTML attributes as they would come from the child props (this also ensures consistency)
+
+```tsx
+export interface NewComponentProps extends PancakeStackProps {}
+
+export const NewComponent: React.VFC<NewComponentProps> = ({ ...props }) => <PancakeStack {...props} />
+```
+
+- Declare the `children` prop if you require it
+  - Usually it is `React.ReactNode`, however you can customise it to your needs
```
hqta, line 186: Might be helpful to have this to reference:
https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+- Add `classNameOverride` to replace the omitted `className`
+- Prefix boolean props with `is` or `has`
+- Declare a default value for optional boolean props
+  - Unless you specifically have a need to differentiate between `false` and `undefined`, this allows you to have the type safety of only needing to cater for a boolean value in any usages (eg. in a util)
+  - Aim to name the prop so that the default value is `false` which will allow the consumer to only need to include `isBoolean` as opposed to needing to negate it using `isBoolean={false}`
```
hqta, line 191: Not sure if this is needed if omitting the prop will effectively have false unless I've missed something. I'm thinking along the lines of how `<Component isBoolean />` is equivalent to `<Component isBoolean={true} />`

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+      {...props}
+    >
+      {children}
+      {doSomething(hasOptionalBooleanProp) && <SubComponent />}
+    </div>
+  );
+}
+```
+
+- Write and directly export a `React.VFC` (React VoidFunctionComponent)
```
hqta, line 230: Probably include "if there is no children prop" for readability instead of having to click the link to find out.

docs/code-standards.md
```diff
@@ -0,0 +1,42 @@
...
+
+Following the naming conventions for [JavaScript](https://google.github.io/styleguide/javascriptguide.xml?showone=Naming#Naming) / [TypeScript](https://www.typescriptlang.org/docs/handbook/enums.html):
+
+- Constant values should be upper snake case (`CONSTANT_VALUES_LIKE_THIS`)
+- Enums should be named in singular PascalCase with values also in PascalCase
+
+```ts
+// Example enum
+enum Direction {
+  Up,
```
hqta, line 18: Just wondering if this should be
```
enum Direction {
  Up: "Up"
}
```
since this would just default fo 0,1,2,3

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+  children,
+  classNameOverride,
+  isBooleanProp,
+  hasOptionalBooleanProp = false,
+  onCustomFunction,
+  ...props
+}) => {
+  const [hasSyrup, setHasSyrup] = useState<boolean>(false)
+
+  const handleCustomFunction = () => {
```
hqta, line 122:
```
const handleCustomFunction = (): boolean => {
```
As the code standard suggests.

docs/basic-component.md
```diff
@@ -0,0 +1,247 @@
...
+  - [Styles](#styles)
+  - [Tests](#tests)
+- [Component.tsx template](#componenttsx-template)
+  - [Props](#props)
+  - [The component](#the-component)
+  - [Display name](#display-name)
+
+## Component structure
+
+Given the component `PancakeStack`, the component structure will follow this:
```
hqta, line 19: Where would we put common types used between components?

docs/code-standards.md
HeartSquared: I pulled that example from the TS docs, but it just depends on what you want your enum values to be. Technically there's nothing wrong with it being int values since it's the default, but you're right in that in most cases we would probably actually assign string values.

I don't mind amending it, but this was more about how to format an enum rather than how to use one (since there are official docs for that).

docs/basic-component.md
HeartSquared: The point of `VFC` is that it will never include `children`, thus if you need it, you must always put it yourself. This increases type safety as `children` would not exist if not explicitly included (the flaw of `FC`).
I can put a summary of the difference between `VFC` and `FC` in here if that is clearer, but I had excluded it because I wasn't sure if it would be too much (since the docs are already kinda long).

docs/basic-component.md
HeartSquared: Probably just needs rewording. The sentence is meant to mean that we want the consumers to only need to do `<Component />` (default false) and `<Component isBoolean />` (set to true) as opposed to `<Component />` (default true) `<Component isBoolean={false} />` (set to false)

docs/basic-component.md
HeartSquared: That will be addressed separately. We have not had a thorough discussion on testing yet.

docs/basic-component.md
HeartSquared: Haha whoops! Nice pick up :) Thanks!

docs/basic-component.md
HeartSquared: That is a very good question... Perhaps we'd need to have a separate package just for those generic types. I was certainly thinking we could introduce a type which deals with the `className` rename eg.
```
interface ClassNameOverride<TBaseAttributes> extends Omit<TBaseAttributes, "className"> { 
  classNameOverride?: string
}
```
assuming it works.

Which now that I look at it, we need to figure out if we are happy to prefix generics with a `T` or if that's just a me thing haha.

docs/basic-component.md
hqta: I think if we make it an onboarding item for all front end engineers to go through this then it should be a once off read and a refresher when things change. It could be a simple description of VFC if you don't have children FC if you do.

docs/code-standards.md
hqta: Fair enough - more of a consistency thing than anything else

docs/basic-component.md
HeartSquared: Nope. Never `FC` is the idea :) If you ever need `children`, then you still use `VFC` and include `children` yourself in the props

docs/basic-component.md
```diff
@@ -0,0 +1,261 @@
...
+  isBooleanProp: boolean
+  hasOptionalBooleanProp?: boolean
+  onCustomFunction: () => void
+}
+```
+
+- Create and export an `interface` for your props in the format of `{PascalComponentName}Props`
+  - You can restrict a prop's type signature by redeclaring it (eg. changing a prop from `string | undefined` to `string`)
+  - If you need to change the type signature of an extended prop, ensure you `Omit` it otherwise it will become `never` (eg. changing a prop from `string | undefined` to `number`)
+- `Omit` `"className"` as we re-add it with an alias (`classNameOverride`)
```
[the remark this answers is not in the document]
HeartSquared: Ah. I thought it was considered condescending (at least, my vague notes say that), so was aiming for these docs to be more the "moving forward we should do this for new components (or ones without it)". @christian314159 what are your thoughts (I believe you were the one who raised it)?

If we do plan to deprecate the old naming, I can certainly still add a note in here about it. Otherwise I can just change the docs as Doug suggested.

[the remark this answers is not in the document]
HeartSquared: As discussed in stand up this morning, we'll move forward with `classNameOverride` for any newer usages, but I will document that it supersedes `classNameAndIHaveSpokenToDST`

mcwinter07 (commented):
awesome stuff. Did have a quick quandary on building out packages that could have multiple sub packages like the form, but chuffed with the rest of it 👍

docs/basic-component.md
```diff
@@ -0,0 +1,266 @@
...
+      - groupedFunctions.spec.ts
+      - functionName.ts
+      - functionName.spec.ts
+    - _variables.scss
+    - index.ts
+    - PancakeStack.tsx
+    - PancakeStack.scss
+    - PancakeStack.spec.tsx
+```
+
```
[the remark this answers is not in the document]
HeartSquared: Hmm... I would actually think following on our patterns, it would be something like this:

```
typography/
    docs/
		- DataHeader.stories.tsx
		- DataHeader.stories.scss
		- Heading.stories.tsx
		- Paragraph.stories.tsx
		- SomeStoryHelperComponent.tsx
		- SomeStoryHelperComponent.scss
    src/
		DataHeader/
	        components/
	        hooks/
	        utils/
	        - _variables.scss
	        - index.ts
	        - DataHeader.tsx
	        - DataHeader.scss
	        - DataHeader.spec.tsx
        Heading/
			...
		Paragraph/
			...
    - index.ts
    - package.json
    ... other stuff
```

I think it will look less strange once the stories can actually live with their components, but in the meantime it's quite similar to the existing structure where the docs are detached.
 
I believe the `package.json` config I copied from the old template relies on `docs` being on the top level to be excluded.
But if it isn't tied into the top level, since we'd still need it in a `docs` directory in the meantime, it would look like this:
```
typography/
    src/
		DataHeader/
			docs/
				- DataHeader.stories.tsx
				- DataHeader.stories.scss
				- SomeStoryHelperComponent.tsx
				- SomeStoryHelperComponent.scss
	        components/
	...etc
```

Normally the top level `docs/` would be for documentation about the repo as opposed to component docs (which would live within the component directories), which is why the components should always be within the top level `src/` and not switched.
Might be easier to remember this: there is one `package` and multiple `components`, and `src/` belongs to the `package` :)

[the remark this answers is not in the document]
mcwinter07: Gotcha, so single src, top level docs for actual documentation not stories would produce something like this:
```
typography/
  docs/
    // markdown if relevant but not stories
  src/
    DataHeader/
      docs/
        - DataHeader.stories.tsx
      components/
      hooks/
      utils/
      - _variables.scss
      - index.ts
      - DataHeader.tsx
      - DataHeader.scss
      - DataHeader.spec.tsx
    otherComponent/
  - index.ts
  ...etc
```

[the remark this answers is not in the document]
HeartSquared: Ideally I think it would be good for it to look like this (without the need for a `docs/` in the component), but it needs thought on how to deal with those helper things.

```
typography/
  docs/
    // markdown if relevant but not stories
  src/
    DataHeader/
      components/
      hooks/
      utils/
      - _variables.scss
      - index.ts
      - DataHeader.tsx
      - DataHeader.scss
      - DataHeader.spec.tsx
      - DataHeader.stories.tsx
    otherComponent/
  - index.ts
  ...etc
```

But maybe the in house `docs/` is a good middle ground 🤔 

Or perhaps we have a `__docs/` or `__stories/` (`__something` is a common pattern I've to signify internal things) which you only put in when you need components beyond the `*.stories.*` files.

```
typography/
  docs/
    // markdown if relevant but not stories
  src/
    DataHeader/
      __docs/
        - SomeStoryHelperComponent.tsx
        - SomeStoryHelperComponent.scss
      components/
      hooks/
      utils/
      - _variables.scss
      - index.ts
      - DataHeader.tsx
      - DataHeader.scss
      - DataHeader.spec.tsx
      - DataHeader.stories.tsx
      - DataHeader.stories.scss
    otherComponent/
  - index.ts
  ...etc
```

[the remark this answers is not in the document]
mcwinter07: Sounds good. There's probably a bigger conversation on how we want to handle dev docs outside of storybook, which would tie nicely into retooling how we split out our stories, but for the moment I think the original proposed pattern works 👍

mcwinter07 (approved):
Thanks for the points of clarification 👍 Every looks hunky dory to that's a ✅ on my end

PR #2552, opened 2022-02-21
docs: replace design token StoryBoard with ColorToken story

mcwinter07:
# Objective
- Remove storyboard component from design tokens
- Create Color Tokens story to replace the storyboard

# Motivation and Context
Dependencies within StoryBoard was triggering major releases for design tokens on any component with a breaking change. This PR aims to reduce the risk of Lerna accidentally bumping the version of design tokens up any further.


packages/design-tokens/docs/ColorTokens.stories.tsx
```diff
@@ -0,0 +1,261 @@
+/* eslint-disable import/order */
```
HeartSquared, line 1: This is a very strange rule to disable 🤔 Does it conflict with another rule or something?

packages/design-tokens/docs/ColorTokens.stories.tsx
```diff
@@ -0,0 +1,261 @@
...
+  parameters: {
+    layout: "fullscreen",
+    docs: {
+      description: {
+        component:
+          "This boarding is a representation of the design tokens available",
+      },
+    },
+  },
+} as Meta
```
HeartSquared, line 20: I'd probably avoid type casting if there's no need for it (since it can hide errors that TS normally picks up on)

packages/design-tokens/docs/ColorTokens.stories.tsx
```diff
@@ -0,0 +1,261 @@
...
+              style={{
+                color: props.color,
+                filter: "invert(1) grayscale(1) contrast(100)",
+              }}
+            >
+              {colorString.to.hex(parsedColor.value)}
+            </span>
+          </Heading>
+        )}
+        {parsedColor && (
```
HeartSquared, line 112: This condition is the same as L100 so it's unnecessary (just merge them together with a fragment)

packages/design-tokens/docs/ColorTokens.stories.tsx
```diff
@@ -0,0 +1,261 @@
...
+        padding: theme.spacing.md,
+      }}
+    >
+      {memoizedChildren}
+    </div>
+  )
+}
+ColorTokens.storyName = "Color Tokens"
+ColorTokens.parameters = {
+  chromatic: { disable: true },
```
HeartSquared, line 260: Hmm...I wonder if maybe we'd actually want to test the colour palette?

HeartSquared (changes requested):

HeartSquared:
I'd also classify this as `docs` as opposed to a `refactor` :)

HeartSquared:
Looking at the preview, we can actually take it out of being a story and convert it into a page instead :)

packages/design-tokens/docs/ColorTokens.stories.tsx
mcwinter07: This is just left over from the storyboard component - I mostly just copied across the components needed to render out the color chips and didn't focus on clean up. Happy to remove and run through these comments though 👍

packages/design-tokens/docs/ColorTokens.stories.tsx
mcwinter07: Actually that might me a good move, just to make sure we don't grunk up the colors at any point. Though if we want to do this maybe we leave it as a story instead of a page

mcwinter07: Also hesitant to make this a doc since we don't have pre-written documentation on it

HeartSquared: Maybe we could keep it as a story but remove the canvas? I wonder if that's possible 🤔 But if it's too much trouble we can leave it for a separate PR

mcwinter07: actually I stand corrected, there [totally is](https://cultureamp.design/guidelines/color/) ❗ 🧠  maybe we just leave this one as a test for colors

HeartSquared (approved):
The Australian in me is sad about the misspelling of `colour` 😢 

As discussed, let's review the story and make it look prettier later since it's not a quick fix we can do here :)

PR #2554, opened 2022-02-21
chore: Adding some sub-dependencies of Gatsby to allow list [KDS-229]

christian314159:
### Objective

Adding some vulnerabilities found in our Gatsby sub-dependencies to the allow list.


lloydstubber (approved):
Thanks for this. For context if anyone wonders why; it's displaying large warnings in our Buildkite UI and we can't bump them without a ton of work because... Gatsby.

HeartSquared (approved):

PR #2555, opened 2022-02-22
chore(chromatic): disable chromatic snapshots as a default

HeartSquared:
# What

Disable sending stories to Chromatic as a default and only add them in when we want them to be.

I selected unique stories (or at least the ones that made sense to stay) and excluded ones which didn't seem worth visually testing.

# Why

We've been unnecessarily sending stories to Chromatic and it's costing us without providing us benefit.

# Changes

- Disable uploading stories to Chromatic as a default
- Enable stories individually with `chromatic: { disable: false }`
- When testing a single component (eg. wip), you can add `disableSnapshot: true` to `storybook/preview.tsx` and then add `chromatic: { disableSnapshot: false }` to your individual story

## Components

See Chromatic build here: https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=259

### Visually Hidden
- Included ✅ :
	- `Hidden text embedded in visible text`

#### Avatar
- Included ✅ :
	- `Design Sheet (default)`
	- `Design Sheet (reversed)`
	- `Initials Long`
	- `Initials Unicode`
- Excluded ❌ :
	- `Default (Kaizen Demo)`

#### Avatar Group
- Included ✅ :
	- `Design Sheet (default)`
	- `Design Sheet (reversed)`
- Excluded ❌ :
	- `Default (Kaizen Demo)`

#### Company Avatar
- Included ✅ :
	- `Design Sheet (default)`
	- `Design Sheet (reversed)`
- Excluded ❌ :
	- `Default (Kaizen Demo)`

### Badge
- Included ✅ :
	- `Light Badges`
	- `Reversed Badges`
- Excluded ❌ :
	- `Default (with args)`
	- `Animated Badge`

### Box
- Included ✅ :
	- `Box With Margin`
	- `Box With Padding`
	- `Box With X And Y Padding`
	- `Box With Rtl Support`
- Excluded ❌ :
	- `Default`

### Brand
- All included ✅ 

### Brand Moment
- Included ✅ :
	- `Informative intro`
	- `Positive outro (customer focused)`
	- `Error`
- Excluded ❌ :
	- `Positive outro`
	- `Informative intro (customer focused)`

#### Button
- Included ✅ :
	- `Light Buttons`
	- `Reversed Buttons`
- Excluded ❌ :
	- `Default Button (Kaizen Site Demo)`

#### IconButton
- Included ✅ :
	- `Default Icon (Kaizen Site Demo)`

### Card
- Included ✅ :
[...]

PR #2561, opened 2022-02-24
docs: Update avatar stories to use storyWrapper

NatalieLoyzaga:
# Objective
Update Avatar and Avatar Group stories to use StoryWrapper for sticker sheets.

# Motivation and Context
Reduce the number of stories and consolidate to one sticker-sheet.

image
image


HeartSquared:
The headings in the screenshot are misaligned from the actual content 🤔

draft-packages/avatar/docs/Avatar.stories.tsx
```diff
@@ -40,121 +36,261 @@ DefaultStory.args = {
...
-            <>
-              <Avatar {...avatarProps} />
-              <br />
-            </>
-          )
-        })}
-      </div>
-    ))}
-  </div>
+export const StickerSheetDefault = () => (
```
HeartSquared, line 39: Firstly, definitely an improvement on the example code by moving the data into the story, so good work!

I think we can clean up the duplication a little though (which means changes per column will be easier) by using some constants. Make sure the constants are still within the story so that it surfaces in the code sample for our consumers :)

```
export const StickerSheetDefault = () => {
  const PROPS_PHOTO_PERSONAL = {
    fullName: "Jane Doe",
    disableInitials: false,
    isCurrentUser: true,
    avatarSrc: "https://www.cultureampcom-preview-1.usw2.wp-dev-us.cultureamp-cdn.com/assets/slices/main/assets/public/media/chapters-card-1@2x.05e547444387f29f14df0b82634bf2b6.png",
  }

  const PROPS_INITIALS_PERSONAL = {
    fullName: "Jane Doe",
    disableInitials: false,
    isCurrentUser: true,
  }

  return (
    ...
    <Avatar {...PROPS_PHOTO_PERSONAL} size="xxlarge" />
    <Avatar {...PROPS_INITIALS_PERSONAL} size="xxlarge" />
    etc...
  )
}
```

The downside of course is that we have an object for the props as opposed to having them declared within the component tags _however_ we should have separate stories for proper usage anyway (I'm almost thinking perhaps we should disable the docs for sticker sheets, but we'll revisit that separately).

draft-packages/avatar/docs/Avatar.stories.tsx
```diff
@@ -40,121 +36,261 @@ DefaultStory.args = {
...
+        />
+        <StoryWrapper.Row rowTitle="XX-Large">
+          <Avatar
+            fullName="李存信"
+            disableInitials={false}
+            isCurrentUser
+            size="xxlarge"
+          />
+          <Avatar
+            fullName="Spicy Jalapeno Bacon Ipsum Dolor Amet Aute Elit Chicken Mollit"
```
HeartSquared, line 199:
```suggestion
            fullName="Spicy Jalapeno Taco Bacon Ipsum Pretzel Dolor Amet Nacho Elit Chicken"
```

I changed the string in #2555 because it was still having false positives before, but I battle tested the new one so it shouldn't happen again

draft-packages/avatar/docs/Avatar.stories.tsx
```diff
@@ -40,121 +36,261 @@ DefaultStory.args = {
...
-              <Avatar {...avatarProps} />
-              <br />
-            </>
-          )
-        })}
-      </div>
-    ))}
-  </div>
-)
+StickerSheetDefault.storyName = "Sticker Sheet (Default)"
```
HeartSquared, line 296: We should probably still have a Reversed sticker sheet so we can visual test it both manually and in Chromatic.

One thing I looked into with the code sample and reusing story content is that we can't just make a content component then put it in, you actually need to create a Template and use the `Template.bind` pattern. Feel free to reach out if you need more clarification :)

HeartSquared (changes requested):

draft-packages/avatar/docs/Avatar.stories.tsx
NatalieLoyzaga: There are no isReversed or reversed prop/styles for the component so I assumed it was not necessary to create a ReversedSheet/ test within chromatic?

draft-packages/avatar/docs/Avatar.stories.tsx
NatalieLoyzaga: Awesome I have done this! looks a lot cleaner

draft-packages/avatar/docs/Avatar.stories.tsx
HeartSquared: Hmm...that's true too. It certainly makes it seem redundant. I just checked Figma and we don't have an example there either. Maybe it's okay to not have it? I might pose the question in Slack :)

draft-packages/avatar/docs/AvatarGroup.stories.tsx
```diff
@@ -31,84 +31,23 @@ DefaultStory.args = {
...
-        width: "auto",
-      }}
-    >
-      <div>
-        <Heading tag="h4" variant="heading-4">
-          Large
-        </Heading>
-        <br />
+    <StoryWrapper>
+      <StoryWrapper.RowHeader headings={["Deafult"]} />
```
HeartSquared, line 38:
```suggestion
      <StoryWrapper.RowHeader headings={["Default"]} />
```

:)

draft-packages/avatar/docs/AvatarGroup.stories.tsx
```diff
@@ -31,84 +31,23 @@ DefaultStory.args = {
   avatars: avatarGroupData["users"],
 }
 
-export const DesignSheetDefault = () => {
+export const StickerSheetDefault = () => {
   const data = avatarGroupData["users"] as AvatarList
```
HeartSquared, line 35: I would probably do a similar thing to the other story where we include the data structure in the example here :)

HeartSquared (changes requested):

draft-packages/avatar/docs/AvatarGroup.stories.tsx
```diff
@@ -23,92 +22,134 @@ export default {
...
+  }
+
+  const AVATARS = [
+    EXAMPLE_USER_1,
+    EXAMPLE_USER_2,
+    EXAMPLE_USER_3,
+    EXAMPLE_USER_4,
+    EXAMPLE_USER_5,
+    EXAMPLE_USER_6,
+  ] as AvatarList
```
HeartSquared, line 70:
```suggestion
  const AVATARS: AvatarList = [
    EXAMPLE_USER_1,
    EXAMPLE_USER_2,
    EXAMPLE_USER_3,
    EXAMPLE_USER_4,
    EXAMPLE_USER_5,
    EXAMPLE_USER_6,
  ]
```

HeartSquared (changes requested):

HeartSquared (approved):
LGTM 🌟

PR #2562, opened 2022-02-24
docs: Upgrade badge stories to use storyWrapper

NatalieLoyzaga:
# Objective

Uses StoryWrapper component to consolidate the stories into one sheet.

image
image


draft-packages/badge/docs/Badge.stories.tsx
```diff
@@ -27,7 +27,7 @@ export default {
   decorators: [withDesign],
 }
 
-const BadgeStoryWrapper: React.FunctionComponent<{
+const BadgeAnimationStoryWrapper: React.FunctionComponent<{
```
HeartSquared, line 30:
```suggestion
const BadgeAnimationStoryWrapper: React.VFC<{
```

draft-packages/badge/docs/Badge.stories.tsx
```diff
@@ -56,141 +56,59 @@ const BadgeStoryWrapper: React.FunctionComponent<{
...
-          Large
-        </Heading>
-        <Badge size="large" variant="default">
+const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
+  isReversed,
+}) => (
+  <StoryWrapper isReversed={isReversed}>
+    <StoryWrapper.RowHeader headings={["Default", "Active", "Dot"]} />
+    <StoryWrapper.Row rowTitle="Small">
+      <BadgeStoryWrapper>
```
HeartSquared, line 82: I think it's fine to just not add the extra padding around the small badges. It actually throws off the alignment a little the more I look at it (since it's neither left nor centre aligned) 🤔 
Also, and this may only apply to me, but I would question whether the excess white space would be baked into the component when I see it, which is a check I shouldn't need to do.

I think we should only need to decorate our stories if they are not clear without them :)

HeartSquared (changes requested):

PR #2564, opened 2022-02-25
chore: add commitizen

HeartSquared:
# What

Set up [`commitizen`](https://github.com/commitizen/cz-cli).

# Why

Conventional commits can sometimes be confusing (eg. is my doc fix really a fix?). Commitizen will help provide you prompts with descriptions to help you through it, and then format your commit for you based on your input :)

# Recording

https://user-images.githubusercontent.com/25891850/155635289-705c7ad3-7819-492b-b93f-922d39201d2c.mp4

# How do I use this?

1. `yarn install` (to ensure you have the changes)
2. `git add` some changes
3. `yarn commit` (or `yarn cz` if you prefer; the former is a script alias)
4. Profit!

# Notes

As a first pass, I want it to be available, but still optional for people to use.
Apparently you can make it hook into `git commit`, but there’s some extra steps, and I don’t want to stop people who are comfortable with just plain writing their own commits. Also I didn’t find it that great of an experience with the code sample they provided, so it would probably mean more digging into making it a better experience.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,6 +175,8 @@ This workflow requires that all pull requests have a title formatted as follows:
 <type>: <description>
 ```
 
+To help you, we have [`commitizen`](https://github.com/commitizen/cz-cli) installed, so you can run `yarn commit` and follow the prompts and commitizen will format your commit for you.
+
 #### Description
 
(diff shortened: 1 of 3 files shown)


mcwinter07 (approved):
I think this could be a pretty handy tool to have in our belt 👍 Might be worth to showcasing this in show and tell before we merge it in but other than than its a ✅ from me

mcwinter07 (approved):
👍 LGTM - merge away!

PR #2565, opened 2022-02-25
feat: allow for margin top and bottom

NatalieLoyzaga:
# Objective

Add a mt and mb boolean prop to allow a marginTop or marginBottom to be added.

# Motivation and Context

For cases where there are multiple storyWrappers stacked.

### DIFF
--- a/storybook/components/StoryWrapper/StoryWrapper.scss
+++ b/storybook/components/StoryWrapper/StoryWrapper.scss
@@ -13,6 +13,14 @@ $grid-gap: 1.5rem;
   }
 }
 
+.marginTop {
+  margin-top: 4rem;
+}
+
+.marginBottom {
+  margin-bottom: 4rem;
+}
+
 .topHeader {
   grid-template-areas: ". stories";
   grid-template-columns: 100px 1fr;
--- a/storybook/components/StoryWrapper/StoryWrapper.scss
+++ b/storybook/components/StoryWrapper/StoryWrapper.scss
@@ -13,12 +13,8 @@ $grid-gap: 1.5rem;
   }
 }
 
-.marginTop {
+.storyRowContainer + .storyRowContainer {
   margin-top: 4rem;
-}
-
-.marginBottom {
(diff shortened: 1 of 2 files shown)


HeartSquared:
In this situation, as opposed to adding more props, you can actually use scss to target the class being next to the same class
eg. `.storyWrapper + .storyWrapper` then add the gaps

storybook/components/StoryWrapper/StoryWrapper.tsx
```diff
@@ -11,6 +11,8 @@ export interface StoryWrapperProps {
   hasNoRowTitles?: boolean
   isReversed?: boolean
   children: Array<React.ReactElement<StoryRowHeaderProps | StoryRowProps>>
+  mt?: boolean
+  mb?: boolean
```
mcwinter07, line 15: Bit of a nit pick, but I'd generally prefer hasMarginTop or hasMarginBottom for a boolean - I know its a little long but it just makes it clear in the props that user's won't be setting a value

mcwinter07 (commented):
Looks good, bar one small nit pick

storybook/components/StoryWrapper/StoryWrapper.scss
```diff
@@ -13,6 +13,10 @@ $grid-gap: 1.5rem;
   }
 }
 
+.storyRowContainer + .storyRowContainer {
+  margin-top: 4rem;
+}
```
HeartSquared, line 18: You should be able to put this into the existing `.storyRowContainer` :)

```
.storyRowContainer {
  grid-area: stories;
  display: grid;
  gap: $grid-gap;

  & + .storyRowContainer { // I wonder if `& + &` works? Never tried
    margin-top: 4rem;
  }
}
```

NatalieLoyzaga: Yep I have updated to use & + & :)

HeartSquared (approved):
LGTM 😄

PR #2572, opened 2022-03-01
docs: refactor StoryWrapper to allow mapping rows

HeartSquared:
# What

Refactor `StoryWrapper` to allow mapping `StoryRow`.

# Why

`StoryWrapper` was using a stricter type for `children`, which would normally be fine, but it turns out it would then prevent you from using `.map`, despite actually having only valid children inside your callback.

# Changes

- Refactor StoryWrapper `children` to be `React.ReactNode`
	- Note: When we receive `children` which do not match one of the expected subcomponents, I let it still render them. Alt we can make it do something else like printing a message to say what to use
- Update Avatar and AvatarGroup docs


NatalieLoyzaga:
Nice! 

I had noticed a bug prior to this - when you remove rowHeader and only have one row component as the child you get a typescript error. So this fixes that which is good! 

Avatar looks clean - I will use this structure on similar stickersheets.

NatalieLoyzaga (approved):
Keen to start using maps

PR #2575, opened 2022-03-01
docs: update inline notification stories to use storyWrapper

NatalieLoyzaga:
# Objective

Uses StoryWrapper component to consolidate the stories into one sheet.

Note: The component is missing variants compared to UI kit - separate card made.

image

image


HeartSquared (changes requested):
You'll want to fix your commit messages and PR title to include the subject. Makes it much easier to find :)

---

I think you might want to make sure we still have the multiline/forced multiline stories, since they actually do have visual differences. You could possibly incorporate it into your sticker sheet.

You'll also want to still leave a story for the `autoHide` prop (separately, since it shows functionality; just don't include it for Chromatic).

The `persistent` prop does not differ visually between variants, so it could just be a Row as opposed to a column.

The link is unnecessary since that is just part of `children` from what I can see.

Also, I'm pretty sure `automationId` is a testing id, so that's unnecessary to define in our stories :)

NatalieLoyzaga:
@HeartSquared I have updated the story to include multi line, autoHide separate story and removed the Persistent col.

packages/notification/docs/InlineNotification.stories.tsx
```diff
@@ -49,245 +20,134 @@ export default {
...
+    <>
+      <StoryWrapper isReversed={isReversed}>
+        <Heading
+          variant="heading-3"
+          tag="h1"
+          color={isReversed ? "white" : "dark"}
+        >
+          Prominent
+        </Heading>
+        <StoryWrapper.RowHeader headings={["Dismissible"]} />
```
HeartSquared, line 70:
```suggestion
```

Since `Persistent` is now a row, this doesn't make sense to be the column heading.

packages/notification/docs/InlineNotification.stories.tsx
```diff
@@ -49,245 +20,134 @@ export default {
...
     title="Success"
-    persistent
-    automationId="notification1"
+    {...props}
+    autohide
+    autohideDelay="short"
   >
     New user data, imported by mackenzie@hooli.com has successfully uploaded.{" "}
     <a href="/">Manage users is now available</a>
   </InlineNotification>
```
HeartSquared, line 151:
```suggestion
  </InlineNotification>
  <p>Content below the notification</p>
```

So that it doesn't look like it goes completely blank :)

HeartSquared (changes requested):
Just some small changes :)

HeartSquared (approved):
LGTM :)

PR #2576, opened 2022-03-02
feat: Create new layout-content-max-width-with-sidebar design token

dougmacknz:
# Objective
Creates a new layout token for content max width for pages that have a sidebar.

# Motivation and Context
We currently have a `layout-content-max-width` design-token for most of our pages, but we've found we need another one to support layouts that have a sidebar.


HeartSquared (approved):
Curious to know where we got the value from though.

dougmacknz:
> Curious to know where we got the value from though.

It's the width of the content area on our pages that have a sidebar. https://cultureamp.slack.com/archives/C0189KBPM4Y/p1646023890646609

PR #2579, opened 2022-03-02
feat: Datepicker chromatic stories

NatalieLoyzaga:
# Objective

- Update the DatePicker stories to use stickersheet format
* Update all `new Date()` to use a past date
  * Past date to ensure that the bold blue styling for today's date does not get added
- Add the chromatic parameter to each of the stories (not including the form example)
- Use `userEvent` to display focus state on a `day` for Calendar example

Note: Form Example story... This is to test the z-index in comparison to other pop up style components. The z-index is currently set for the calendar at `1010` this is based on the [layers.scss 
](https://github.com/cultureamp/kaizen-design-system/blob/master/packages/component-library/styles/layers.scss) variables. Made the decision to hardcode until further discussion (CSS DACI)

# Motivation and Context

[KDS-226](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-226)

image


NatalieLoyzaga:
run chromatic

NatalieLoyzaga:
chromatic

packages/date-picker/docs/DatePicker.stories.tsx
```diff
@@ -23,88 +30,154 @@ export const KaizenDefault = props => {
...
-    <DatePicker
-      id="make-me-unique-5"
-      labelText="Label"
-      value={new Date(2022, 1, 14)}
-      onChange={onDayChange}
-      isDisabled
-    />
+    <>
+      <p>
+        This story is used to test the z-index in comparision to other pop up
```
dougmacknz, line 170: Is the intention to keep this story? If it's not going on Chromatic and it's just testing to see if z-index works, I'd say it's just adding clutter?

NatalieLoyzaga: yeah I would agree, I will remove :)

HeartSquared: If it shows the z-index works, I would send it to Chromatic instead of removing it :)

NatalieLoyzaga: Its a weird one -  Since both of the components don't have a prop to set the initial state to be opened, to get into the state where the popups overlap you need to click a few times back and forth. We would need to use userEvents and testing-library to get it to open which will take a bit of work.

NatalieLoyzaga: Im thinking for now it is not worth the effort in terms of what is required for this card and since this PR is required for the next date picker card. The z-index was a _nice to have_ extra of this card so I can note it in another card.

HeartSquared: I actually had requested the story as something to have when Lloyd was implementing the z-index fix, so that we could have a visual test in Chromatic.

The issue at the time was that the DatePicker, when open, was actually sitting below some other form elements (I think it was a text input?). Even if it doesn't work for Chromatic yet, it's worth us being able to verify that it's not opening up under other elements.

NatalieLoyzaga: I'll create another card for it and i'll add your comments above - Ideally this PR gets merged as we want to use chromatic for the next card that potentially breaks the focus styling.

HeartSquared: Sure, go for it :)

dougmacknz (approved):

PR #2580, opened 2022-03-02
feat: Create ToggleIconButton primitive for rich-text-editor

dougmacknz:
# Objective
Creates a new component called `ToggleIconButton` inside a newly created `@kaizen/rich-text-editor` package.

# Motivation and Context
This component will be one type of `ToolbarItem` that could be included in a Rich Text Editor `Toolbar`. It will be the only one we create for now, for use with bold, italics, underline, and list functionality.


yarn.lock
```diff
@@ -16928,7 +16986,7 @@ moo-color@^1.0.2:
   dependencies:
     color-name "^1.1.4"
 
-motion-ui@cultureamp/motion-ui:
+motion-ui@cultureamp/motion-ui, "motion-ui@github:cultureamp/motion-ui":
```
HeartSquared, line 16989: This is strange

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -0,0 +1,52 @@
+@import "@kaizen/design-tokens/sass/animation";
+@import "@kaizen/design-tokens/sass/border";
+@import "@kaizen/design-tokens/sass/color";
+@import "@kaizen/design-tokens/sass/spacing";
+
+.button {
```
HeartSquared, line 6: Do we need to change the cursor? I noticed that Slack's RTE buttons use a pointer

HeartSquared (changes requested):
Mostly questions and clean up - overall it looks fine.
Also the `index.ts` is empty.

packages/rich-text-editor/docs/ToggleIconButton.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+      },
+    },
+    ...figmaEmbed(
+      "https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=12317%3A92573"
+    ),
+  },
+  decorators: [withDesign],
+}
+
+export const DefaultStory = (args: ToggleIconButtonProps) => (
```
HeartSquared, line 31: Not sure if you've tested other icons, but might be worth including the relevant ones in a story to ensure they all render correctly (eg. alignment next to each other, whether the current styles work with all icons or if it looks off when changed).

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -0,0 +1,52 @@
...
+  height: 3rem;
+  border-radius: $border-solid-border-radius;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: $color-purple-800;
+  position: relative;
+  transition: background-color $animation-duration-immediate;
+
+  &:hover:not(.active),
```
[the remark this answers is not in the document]
dougmacknz: It's one of the questions we have for the designers at the moment, waiting for hear back.

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
dougmacknz: I'd defer to https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html and just the general fact that browsers don't put pointers on buttons.

Related: https://adamsilver.io/blog/buttons-shouldnt-have-a-hand-cursor/

packages/rich-text-editor/docs/ToggleIconButton.stories.tsx
mcwinter07: Might hold off on producing more stories for this individual component for now until we can the Toolbar and ToolbarSection (also design feedback on the button itself)

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
HeartSquared: Interesting. Especially since it's much more common to see the pointer on a button on websites than not. I wonder how long it will take for people to follow the convention, considering that the more common pattern is actually to have the pointer (as he also mentioned he associated it himself).

I just read part 2 of the blog too, but I still have a question.

In the case where we have an `<a>` that looks like a button, do we still have the pointer? What about the opposite case where `<button>` looks like a link?

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+          })}
+          {...nativeButtonProps}
+        >
+          <div className={styles.iconWrapper}>
+            <Icon icon={icon} role="presentation" inheritSize />
+          </div>
+        </button>
+      </Tooltip>
+    )
+  })
```
HeartSquared, line 37: Oops forgot in my first review.

```suggestion
  })
  
ToggleIconButton.displayName = "ToggleIconButton"
```

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  icon: React.SVGAttributes<SVGSymbolElement>
+  label: string
+  /*
+   * determines the active or inactive state along with the "aria-pressed" attribute
+   */
+  isActive?: boolean
+}
+
+export const ToggleIconButton: React.VFC<ToggleIconButtonProps> =
```
HeartSquared, line 17: Just saw the file structure too. Looking at our docs, it should be:
`src/RichTextEditor/components/ToggleIconButton`
as the `components` folder is only necessary for when you have subcomponents.

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
mcwinter07: In theory, if it does act as a link I would assume so? Though I think that A11y best practice is to have your links visually distinct from your buttons - I'm not 100% sure if it would directly violate a rule (ie: not having your links visually distinct from text) but ideally links have unique styles that are consistent :/

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.tsx
mcwinter07: Mind if I get a sense check from the both of you @HeartSquared @dougmacknz before I push this change:

Screen Shot 2022-03-04 at 3 35 02 pm

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.scss
dougmacknz: Looks like our Button component currently does a pointer regardless, I'd definitely be in favour of changing that.

The button that looks like a link situation feels a bit more icky, but it's not a case we've run into yet (that I'm aware of anyway).

Related: https://github.com/cultureamp/kaizen-design-system/discussions/1868

packages/rich-text-editor/src/components/ToggleIconButton/ToggleIconButton.tsx
HeartSquared: Yep that looks right! :)

HeartSquared (approved):
LGTM :D

PR #2582, opened 2022-03-03
feat: use focus-visible for focus styling

NatalieLoyzaga:
# Objective

Update all instances of `:focus` to use `&:global(.focus-visible)` to ensure the focus styling shows on keyboard interactions only. 

Add
```
 :focus {
  outline: none;
}
```
to all instances

# Motivation and Context
Closes [KDS-261](https://cultureamp.atlassian.net/browse/KDS-261)


packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -33,15 +33,15 @@ $placeholder-opacity: 0.7;
   border-radius: $border-solid-border-radius;
   padding: 0 $button-base-padding-horizontal;
 
-  &:focus:not([disabled]),
-  &:hover:focus:not([disabled]),
+  &:global(.focus-visible):not([disabled]),
+  &:hover:global(.focus-visible):not([disabled]),
```
dougmacknz, line 37: I've never seen this before, for when it's focused and hovered at the same time? 🤯

packages/date-picker/src/DatePicker/DatePicker.scss
```diff
@@ -33,15 +33,15 @@ $placeholder-opacity: 0.7;
...
   &:hover:not([disabled]) {
     background-color: $color-gray-200;
     border-color: $color-gray-600;
     color: $color-purple-800;
   }
 
-  &:focus:not([disabled]) {
+  &:global(.focus-visible):not([disabled]) {
```
dougmacknz, line 44: I believe you'll also want to do on top of this, so that the browser default outline doesn't show on click focus.

```
:focus {
  outline: none;
}
```

(This applies to all of the ones you've done here)

NatalieLoyzaga: Updated this :) thanks

packages/date-picker/src/DatePicker/DatePicker.scss
NatalieLoyzaga: haha I removed this - I don't think its a thing

NatalieLoyzaga: Must of been added as overkill

dougmacknz (approved):

PR #2587, opened 2022-03-07
feat: update to use FocusOn

NatalieLoyzaga:
# Objective

Use [react FocusOn ](https://github.com/theKashey/react-focus-on) to handle the following...

- FocusLock on the calendar/input(button) when open
- Set aria-hidden to other elements other than the Datepicker when open (toggles when open/closed)
- Use `onClickOutside` method to handle the clicking outside to set isOpen to false rather than custom hook.
- Use `onEscapeKey` method to handle keyDown event for escape rather than custom function.


packages/date-picker/package.json
```diff
@@ -38,6 +38,8 @@
     "classnames": "^2.3.1",
     "react-day-picker": "^7.4.9",
     "react-focus-lock": "^2.8.1",
+    "react-focus-on": "^3.5.4",
+    "aria-hidden": "^1.1.3",
```
lloydstubber, line 42: Now we're using `react-focus-on` I don't think we need `aria-hidden`

NatalieLoyzaga: yep will remove

NatalieLoyzaga: and focus lock

lloydstubber:
Nice work! LGTM but get another see of eyes on this too

dougmacknz (approved):

PR #2592, opened 2022-03-08
refactor: remove deprecate warning package

mcwinter07:
Breaking Change:
- Remove the react-deprecate-warning package

# Objective
Remove the react deprecate warning package

# Motivation and Context
As this package is no longer going to be used as a method to identify deprecated packages, we have decided to remove this.


HeartSquared (approved):
LGTM :)

PR #2593, opened 2022-03-08
docs: update progressbar stories to use storyWrapper

NatalieLoyzaga:
# Objective

Use storyWrapper to create stickersheet stories

Note: I have created a card to add reversed variant for Progress Bar

image


packages/progress-bar/docs/ProgressBar.stories.tsx
```diff
@@ -1,7 +1,9 @@
 import React from "react"
+import { Story } from "@storybook/react"
 import { CATEGORIES } from "../../../storybook/constants"
 import { figmaEmbed } from "../../../storybook/helpers/figmaEmbed"
 import { ProgressBar } from "../index"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 
 export default {
   title: `${CATEGORIES.components}/ProgressBar`,
```
dougmacknz, line 9: This has been bugging me

```suggestion
  title: `${CATEGORIES.components}/Progress Bar`,
```

NatalieLoyzaga: haha so fair, I've updated

packages/progress-bar/docs/ProgressBar.stories.tsx
```diff
@@ -1,96 +1,83 @@
...
-  name: "Positive (Kaizen Site Demo)",
-  parameters: {
     ...figmaEmbed(
       "https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=1929%3A20890"
     ),
   },
 }
 
-export const PositiveSubtext = () => (
+export const DefaultStory = () => (
```
HeartSquared, line 23:
```suggestion
export const DefaultStory: ComponentStory<typeof ProgressBar> = args => (
  <ProgressBar {...args} />
)
DefaultStory.storyName = "Default (Kaizen Demo)"
DefaultStory.args = {
  value: 25,
  max: 100,
  mood: "positive",
  isAnimating: true,
  label: "25%",
}
```

Couldn't highlight the whole section so the suggestion looks weird.
Realised there weren't any controls for ProgressBar, so this will add it in while also binding the values into them :)

packages/progress-bar/docs/ProgressBar.stories.tsx
```diff
@@ -1,96 +1,83 @@
...
-    isAnimating={false}
-    value={25}
-    max={100}
-    mood="cautionary"
-    label="Label"
-  />
+const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
+  isReversed,
+}) => (
+  <StoryWrapper isReversed={isReversed}>
```
HeartSquared, line 37: Are you going to add a reversed sticker sheet? If not:

```suggestion
const StickerSheetTemplate: Story = () => (
  <StoryWrapper>
```

HeartSquared (changes requested):

packages/progress-bar/docs/ProgressBar.stories.tsx
NatalieLoyzaga: I've added reversed sheet and creating card to create reversed variant

HeartSquared: We might not need a reversed variant of the component design-wise, but seeing how it looks on a reversed background is nice nonetheless 😄

NatalieLoyzaga: The labels (percentage and subtext)  are black so the reversed will have to make them white

HeartSquared: Oh! Yes that makes sense. Good pick up 👍🏻 We're reaping the benefits of having both sticker sheets already 😆

HeartSquared (approved):
Great work! 🎉

PR #2597, opened 2022-03-09
feat: Update new start and end icons

chy-m:
Update 2 new icons for pagination: `start` and `end`. 
Needed to be black fill to be able to inherit colours. 

image


dougmacknz (approved):
thanks!

PR #2603, opened 2022-03-09
docs: update buttons story name and controls

NatalieLoyzaga:
# Objective

- Update Default Buttons (kaizen demo) to Buttons
- Update Icon (kaizen demo) to Icon Button
- Remove controls on stickersheets.

image

### DIFF
--- a/packages/button/docs/Button.stories.tsx
+++ b/packages/button/docs/Button.stories.tsx
@@ -39,13 +39,13 @@ export default {
 
 export const DefaultKaizenSiteDemo = args => <Button {...args} />
 DefaultKaizenSiteDemo.story = {
-  name: "Default (Kaizen Demo)",
+  name: "Button",
 }
 
 export const DefaultKaizenDemoIcon = args => (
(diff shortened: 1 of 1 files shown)


packages/button/docs/Button.stories.tsx
```diff
@@ -39,13 +39,13 @@ export default {
 
 export const DefaultKaizenSiteDemo = args => <Button {...args} />
 DefaultKaizenSiteDemo.story = {
-  name: "Default (Kaizen Demo)",
+  name: "Button",
```
HeartSquared, line 42: Doesn't this change the embed on the site page?

NatalieLoyzaga: Nope, it's based on the component name rather than the story name (I double checked this incase).

There will be a mass search and replace PR once we decide on the naming conventions but for now I just updated the story display name to be less convoluted and confusing.

HeartSquared (approved):

PR #2607, opened 2022-03-11
feat: Add error state to Select component

JakePitman:
- Adds an 'error' variant to the Select component
- Adds a story displaying 'error' variant
- 
image

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.tsx
+++ b/draft-packages/select/KaizenDraft/Select/Select.tsx
@@ -39,7 +39,7 @@ export interface SelectProps extends ReactSelectProps<any, boolean> {
   fullWidth?: boolean
 }
 
-export type VariantType = "default" | "secondary" | "secondary-small"
+export type VariantType = "default" | "secondary" | "secondary-small" | "error"
 
 export const Select = React.forwardRef<any, SelectProps>((props, ref) => {
   if (props.fullWidth === false && props.variant !== "secondary") {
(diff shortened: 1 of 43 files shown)


draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -39,7 +39,7 @@ export interface SelectProps extends ReactSelectProps<any, boolean> {
   fullWidth?: boolean
 }
 
-export type VariantType = "default" | "secondary" | "secondary-small"
+export type VariantType = "default" | "secondary" | "secondary-small" | "error"
```
HeartSquared, line 42: I wouldn't consider an error state to be a variant, so it would make more sense to have a`hasError?: boolean` prop instead.

Alt, if it makes more sense to bake in the error message, have something on the lines of `errorMessage?: string` which sets the error styles with the message when an error exists.

@dougmacknz you may be more familiar with our patterns to know which makes more sense to implement.

dougmacknz: The way this has been done with all of our other form component is a `status={"default" | "error" | "success" | "caution"}` prop, so I'd go for that (we'll only be adding default and error for now).

Plus a `validationMessage` prop that uses the `FieldMessage` primitive, which on second thought, wouldn't be too hard to add here as well if you're happy to Jake.

JakePitman: Yeah good point. At the moment I'm handling displaying the error message in our consuming repo. I figured there may be instances where we'd want the message to be separate from this field. Eg. if we ever just want one error message for a form, instead of on each input field, or if we need to display it somewhere other than right below the Select component

JakePitman: Changed error state handling from `variant` to `status`, and added a validation message prop. `<FieldMessage/>` is just inheriting the same status as the parent `<Select/>` component

dougmacknz (approved):

PR #2608, opened 2022-03-11
refactor: change animated illustration scenes to not animate in Chromatic

HeartSquared:
# What

Change animated illustration scenes to not animate in Chromatic.

[New Chromatic snapshots](https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=339)

Refactored stories file to use templates and adjust the controls to be more relevant (it's not perfect as there's a bit of manual overriding required since animated/non-animated use different props, so having them share a file is a little conflicting by nature).

# Why

Animated scene illustrations were not showing up in the Chromatic snapshots (eg. [here](https://www.chromatic.com/component?buildNumber=316&historyLengthAtIndex=44&distanceToMoveBack=-13&appId=60a1e4a102f0cb003b5d19d6&name=Components%2FIllustration%2FScene&specName=Brand+Moments&componentInspectorKey=6227e05c07cc90003a2f97dd-1200-snapshot-true)).

# Changes

- Add display name to illustration scenes and changed to use VFC
	- Changed during wip as I was using the values; figured I just leave the changes instead of reverting
- Turn off animations for Chromatic snapshots

### DIFF
--- a/draft-packages/illustration/KaizenDraft/Illustration/BrandMomentCaptureIntro.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/BrandMomentCaptureIntro.tsx
@@ -1,4 +1,4 @@
-import * as React from "react"
+import React from "react"
 import { AnimatedSceneProps } from "./Scene"
 import { Base } from "./Base"
 import { VideoPlayer } from "./Players/VideoPlayer"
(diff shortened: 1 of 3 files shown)


mcwinter07 (dismissed):
Good to merge 🦖

Issue #2596, opened 2022-03-09
Cautionary color update for Error messages


jmitsios:
## Problem
The **error message container** color for the Cautionary mood is currently `yellow-200`, which we've found to be too prominent when sitting alongside other cautionary components (i.e: inline notifications) which were using `yellow-100`

## Changes made
We have updated the cautionary error message color in the UI Kit from `yellow-200` to `yellow-100` for the following components:

### Text Field
![image](https://user-images.githubusercontent.com/95595231/157367098-5dd35598-8b34-4488-bb9d-3f6185210698.png)
### Text Field Area
![image](https://user-images.githubusercontent.com/95595231/157367134-463b963e-3175-4bb2-b67e-e9d44e0de0c1.png)
### Select
![image](https://user-images.githubusercontent.com/95595231/157367156-717d2ec2-32eb-4f84-8094-bd5b422e0052.png)
### Rich Text Editor
![image](https://user-images.githubusercontent.com/95595231/157367186-c7b855f8-96c8-449f-80dc-aced036d31c5.png)

PR #2609, opened 2022-03-11
docs: Component Health Update Mar 11

dougmacknz:
Component health movements:

* CheckboxGroup to latestDesign: true
* Select to allVariants: false (missing cautionary and others)
* Datepicker to allVariants, responsive, and documented: true

Rather than updating health to false for a few components, I updated `FieldMessage` cautionary styling to `yellow-100`. Closes https://github.com/cultureamp/kaizen-design-system/issues/2596

Moved infinite scroll to guidelines and removed component health data for it (we don't see this as a component that needs to be built in Kaizen).

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
+++ b/draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
@@ -14,7 +14,7 @@ health:
   designed: true
   documented: true
   implemented: true
-  latestDesign: false
+  latestDesign: true
   allVariants: true
   responsive: true
   internationalized: true
--- a/draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
+++ b/draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
@@ -15,7 +15,7 @@ health:
   documented: true
   implemented: true
   latestDesign: true
-  allVariants: true
+  allVariants: false
   responsive: true
   internationalized: true
   accessible: true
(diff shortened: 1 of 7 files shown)


site/docs/guidelines/infinite-scroll.mdx
```diff
@@ -1,20 +1,11 @@
 ---
-title: Infinite Scroll with a “View more” button
+title: Infinite Scroll
 navTitle: Infinite Scroll
 summaryParagraph: Infinite Scroll separates large bodies of content into separate sections. You can access each extra section via a "View more" button.
 tags: ["Infinite scrolling", "Continuous scroll", "Load more", "Async loading"]
 githubLabels: ["component:infinite-scroll"]
```
HeartSquared, line 6: Looks like guidelines should have these properties:

```md
// from guidelines/checkbox-radio-toggle-switch.mdx
title: "Checkbox, Radio, and Toggle Switch"
navTitle: "Checkbox, Radio, and Toggle Switch"
summaryParagraph:
tags:
needToKnow:
inComparingSection: true
```

So we should remove `githubLabels`.
`inComparingSection` appears to be optional though.

dougmacknz: Yeah looks like the `inComparingSection` attribute is used when you want the page to appear in the bottom 'Comparing Components' section:
image

dougmacknz: I've removed the `githubLabels` frontmatter, and deleted the label from Github as well

draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
```diff
@@ -14,7 +14,7 @@ health:
   designed: true
   documented: true
   implemented: true
-  latestDesign: false
+  latestDesign: true
```
HeartSquared, line 17: If the error state is missing, does it still count as meeting the latest design?

HeartSquared (dismissed):
May or may not need a change pending your answer to the question. I'll just approve it first in the case it isn't an issue.

draft-packages/form/KaizenDraft/Form/CheckboxGroup/README.mdx
dougmacknz: That would be considered as part of allVariants, so maybe I should go to false on that.

HeartSquared (approved):

PR #2610, opened 2022-03-11
feat: fix EmptyState chromatic snapshots

HeartSquared:
# What

- Refactor embed of scene illustration in EmptyState
- Update AnimatedSceneProps to allow extensions to have isAnimated as boolean
	- Update EmptyState to extend AnimatedSceneProps as opposed to AnimatedProps which only allowed value to be `true`
- Add some better controls to EmptyState stories
- Update EmptyState stories to not animate scenes in Chromatic
	- https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=389 (ignore the Button ones, Chromatic baselines were messed up)

# Why

Chromatic snapshots for `EmptyState` was not showing the illustrations


mcwinter07 (commented):
It may not be in the scope of this fix but I noticed that the rtl spacing is off on the image. It looks like this is applying a margin right regardless of the parent rtl class. Might be worth splitting off in its own ticket?
Screen Shot 2022-03-23 at 10 34 58 am
Screen Shot 2022-03-23 at 10 35 09 am

Other than that this looks good to go - Happy to approve and split the rtl stuff off into another ticket

mcwinter07 (approved):
Looks good 👍 Merge away

PR #2612, opened 2022-03-15
feat: Copy Heading and Typography components into new @kaizen/typography package

dougmacknz:
## Objective

Creates a new `@kaizen/typography` package holding a copy of our `Heading` and `Paragraph` components from `component-library`.

Adds a `@deprecated` docblock to the `component-library` versions of these components.

Stories and READMEs have been moved instead of copied to avoid double ups.

## Context

We are looking to abolish `component-library`, moving ahead with our multi-package structure.

## Follow ups

Following this:
* Adjust all internal Kaizen packages to import from the new package instead of component-library, checking for any cases where `component-library` can be removed as a dependency.
* Create a codemod to help port consuming repos over to the new package.
* Remove these components from `component-library` with a breaking change.


packages/typography/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "!draft-templates",
+    "!stories",
+    "!docs",
+    "!tsconfig.dist.json"
+  ],
+  "author": "",
+  "private": false,
+  "license": "MIT",
+  "dependencies": {
+    "@kaizen/component-library": "^12.6.0",
```
HeartSquared, line 35: Seems like this is only used by the stylesheets just like `@kaizen/design-tokens`, so perhaps it can be moved into `peerDeps`? It would probably mean that `~kaizen/deprecated-component-library-helpers` should be added too though.

packages/typography/src/Paragraph/README.mdx
```diff
@@ -0,0 +1,15 @@
+---
```
HeartSquared, line 1: Are these READMEs going to conflict with the ones still in `component-library`?

dougmacknz: Oops I was meant to delete those, nice catch

packages/typography/package.json
dougmacknz: I'm gonna have a quick go at removing any usages of component-library-helpers to see if I can remove the dependency on this completely.

dougmacknz: Ok that seems to have worked well. I just copied over the `fonts.scss` file, which makes sense to live here. Not sure if it should be in a folder 🤔

The only `component-library` reference left is a Box import in one of the stories files. Doesn't seem to care that it's not listed in the deps, but wonder if I should add to devDeps?

HeartSquared: Hmm... I wonder how it works to begin with. Does lerna actually create internal aliases for the packages in the repo? Of which case, then the dep probably doesn't need to exist at all. Sounds kinda funny to put it in a devDep though, since it actually is "needed" per se to build storybook. Maybe trying to build storybook might give us our answer?

dougmacknz: Storybook builds fine, so maybe we just leave it out 👍

HeartSquared (dismissed):
LGTM :)

dougmacknz:
For some reason, the site is no longer picking up the READMEs for the 2 components 👀

HeartSquared:
> For some reason, the site is no longer picking up the READMEs for the 2 components 👀

You're right. It seems that `packages` that aren't in `component-library` aren't magically included in the site (we actually also forgot to remove the README for `draft-button` when we deprecated it, which is why we didn't realise it sooner).

Looking at the `packages` folder, there appears to be quite a few packages which aren't components. Probably explains why `component-library` existed. Would it make sense to have `packages/components` to house the component packages?

HeartSquared (approved):

HeartSquared (approved):

PR #2613, opened 2022-03-15
chore: upgrade kaizen to use webpack 5

lloydstubber:
# Objective
- Upgrade Kaizen to use Webpack 5 as a part of fixing the node-sass vulnerabilities

# Motivation and Context
- https://cultureamp.slack.com/archives/CFSSXB50E/p1646962183290229?thread_ts=1646954180.498179&cid=CFSSXB50E


christian314159 (approved):
Checked the branch previews, all worked as far as I can tell. Would be nice to see Chromatic 👍

PR #2615, opened 2022-03-15
fix: draft-card classNameAndIHaveSpokenToDST prop

Moskvyak:
# Objective

classNameAndIHaveSpokenToDST property was incorrectly resolved in `classNames` as `classNameAndIHaveSpokenToDST` instead of the value because it was passed as an object (aka `{ classNameAndIHaveSpokenToDST: true }`

# Motivation and Context

The custom styles were ignored so it resulted in broken UIs in murmur after updating the package to the latest version
https://cultureamp.atlassian.net/browse/MO-109


HeartSquared (approved):
Thanks for the fix :)

PR #2616, opened 2022-03-16
refactor(deps): Update all internal usages of Heading and Paragraph to typography package

dougmacknz:
Following https://github.com/cultureamp/kaizen-design-system/pull/2612, this moves any Kaizen usages of `Heading` and `Paragraph` to come from the new `@kaizen/typography` packages instead of `component-library`.


christian314159 (approved):
👍

PR #2618, opened 2022-03-16
chore: update gatsby to use sass instead of node-sass

lloydstubber:
# Objective
- Switch our Gatsby site to use sass instead of node-sass due to a vulnerability
- We were going down the path of trying to upgrade all of Gatsby but had issues with plugins
- This seemed like the easiest way forward and only required altering some scss but not with new sass features that aren't backwards compatible

# Motivation and Context
- To get rid of node-sass vulnerability


NatalieLoyzaga (approved):
Approve for chromatic

site/src/docs-components/animation/AnimationSequencesExample.scss
```diff
@@ -1,8 +1,8 @@
 @import "~@kaizen/component-library/styles/animation";
 @import "~@kaizen/design-tokens/sass/animation";
 
-$duration: calc(-1 * #{$animation-duration-deliberate}); // 700ms
-$delay: calc(-1 * #{$animation-duration-slow}); // -400ms (stagger effect)
+$duration: -#{$animation-duration-deliberate}; // -700ms
```
lloydstubber, line 4: This comment was incorrect, the variable is 700ms but we're using it as a negative so I've updated accordingly.

dougmacknz (approved):

dougmacknz (approved):

PR #2620, opened 2022-03-16
Revert "chore: upgrade kaizen to use webpack 5"

HeartSquared:
# What

Reverts cultureamp/kaizen-design-system#2613

# Why

Local storybook is having some trouble - noticeably with Brand Moments in the `docs` tab bricking the app.

This branch is to test if webpack 5 might be the cause.


HeartSquared:
Seems to not be the cause (at least, not these changes by themselves)

PR #2627, opened 2022-03-17
chore: Improved deprecation warning for Text component in Storybook [KDS-204]

christian314159:
## Why

In the process of updating the deprecation guidelines, I needed a better example of how to mark something as deprecated in Storybook.

## What

Changed the story for Text, which is deprecated, to include a proper description with an alternative option.


mcwinter07 (approved):
Nice 👍 merge away in good health

mcwinter07 (approved):
LTGM 👍

PR #2628, opened 2022-03-17
feat: add refs to form components

lloydstubber:
# Objective
- Add refs to our form wrapper components down into their primitives

# Motivation and Context
- So a consumer can add in their own ref

# Notes
I've had to leave out TextAreaField and Checkbox as they already have their own internal refs. If they're needed I can look into adding them so they have an external also but there seems to be a bit of logic built into Checkbox's ref which is difficult to pick a part.


HeartSquared:
Someone correct me if I'm wrong, but I believe we should use `forwardRef`?
https://reactjs.org/docs/forwarding-refs.html
https://felixgerschau.com/react-forwardref-explained/

draft-packages/form/KaizenDraft/Form/Primitives/InputRange/InputRange.tsx
```diff
@@ -67,6 +69,7 @@ const InputRange: React.FunctionComponent<InputRangeProps> = (
           setStep(1) // Put the stepper to 1 to avoid floating value
           onChange && onChange(e)
         }}
+        ref={ref}
```
christian314159, line 72: Traditionally I'd want to see a test here, that the external `ref` prop is passed down to right internal element. How hard would it be to add one like that (for all files here)?

lloydstubber:
@HeartSquared yep good catch. If I understand correctly though that'd just be for the parents, right?

draft-packages/form/KaizenDraft/Form/Primitives/InputRange/InputRange.tsx
lloydstubber: Yeah I did have a scout but we don't have any examples of testing refs. I'll have a look into how to do that as I haven't before - other than testing focus states etc.

HeartSquared:
> @HeartSquared yep good catch. If I understand correctly though that'd just be for the parents, right?

I believe it goes on the component that wants to accept the ref (which should be our components) :)
The `ref` prop is therefore not declared in our own props (unless we have a good reason, but usually with a different name), but instead is its own param.

PR #2633, opened 2022-03-21
chore: Remove Heading and Paragraph components from component-library

dougmacknz:
**BREAKING CHANGE:** Heading and Paragraph have been moved to the `@kaizen/typography` package. See https://github.com/cultureamp/node-packages/tree/master/packages/jscodeshift for a codemod to bulk replace the imports in your repo.

## What
Removes the `Heading` and `Paragraph` components from the `component-library` package. 

## Why
These 2 components have been moved to a new `@kaizen/typography` package

## Notes
I haven't removed the Elm versions of these components, which means I've had to leave the CSS files in. Depending on which goes in first, either this PR or https://github.com/cultureamp/kaizen-design-system/pull/2633 will need to remove the CSS files.


mcwinter07 (commented):
Deletes looks good but the branch preview appears to be partially bust? I get the toolbar on initial load but breaks on any form of navigation. Local compile doe storybook seems to work fine so might be some kind broken reference in gatsby somewhere? 🤔

Screen Shot 2022-03-24 at 11 36 56 am

dougmacknz:
I'm gonna try just opening another PR to see if that magically fixes the issue with the branch preview, seeing as a bit has changed since this was created.

dougmacknz:
Narrator: _it didn't_

PR #2635, opened 2022-03-21
fix: add class override for select input and loading elements to use use inter

mcwinter07:
## Why
Select searchable input the input text and loading text is rendered as a default serif, instead of inter. 

## What
- update `.control` class to use `base-font-style()` helper and add `.loadingMessage` class to control font styles
- add `LoadingMessage` component primitive to add style override


draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -174,6 +176,10 @@ const DropdownIndicator: typeof components.DropdownIndicator = props => (
   </components.DropdownIndicator>
 )
 
+const LoadingMessage: typeof components.LoadingMessage = props => (
+  <components.LoadingMessage {...props} className={styles.loadingMessage} />
+)
+
```
mcwinter07, line 182: I've added this in so we can control  the styles of the loading message wrapper. It will still render the default "Loading..." but just give us the ability to set the default styles to inter.

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -108,6 +108,7 @@ export const Select = React.forwardRef<any, SelectProps>((props, ref) => {
           IndicatorsContainer,
           ClearIndicator,
           IndicatorSeparator: null,
+          LoadingMessage,
```
HeartSquared, line 111: I imagine `LoadingMessage` is probably exclusive to the async version so probably isn't needed here

HeartSquared (approved):
Just one comment on whether the non-async needs the prop, but otherwise looks fine :)

draft-packages/select/KaizenDraft/Select/Select.tsx
mcwinter07: My assumption is that if there's a big list of data it may not instantaneously find the results from a search? Ultimately its just updating the primitive that's already baked into react-select so I figured it was safer to leave it in

HeartSquared: Nah it's all preloaded when it's not async :) But that's okay, if TS isn't complaining (although we aren't using the generic (which we probably should, but later), so not sure if that's the cause) then we can just leave it in until we make it tighter. If it doesn't do anything it's just useless code, but nothing will break. We can dig into it properly later.

mcwinter07: gotcha, just took a deeper look and your completely right :)  Following YAGNI so its gone now

HeartSquared (dismissed):

draft-packages/select/KaizenDraft/Select/Select.tsx
```diff
@@ -221,6 +228,10 @@ const IndicatorsContainer: typeof components.IndicatorsContainer = props => (
     className={styles.indicatorsContainer}
   />
 )
+
+const ValueContainer: typeof components.ValueContainer = props => (
+  <components.ValueContainer {...props} className={styles.valueContainer} />
+)
```
mcwinter07, line 234: Turns out targeting the who of the controls for the font update was a bad idea and was throwing off the layout on the reversed backgrounds. I've added this primitive in to supersede the original container class

HeartSquared (approved):

PR #2637, opened 2022-03-22
refactor(herocard): soft deprecate HeroCard and remove from health metrics

HeartSquared:
## Why

As per UI Kit, `HeroCard` is to be superseded by `Tile` when it is rebuilt.

## What

- Soft deprecate `HeroCard`
- Update site component health and page to accept `deprecated` frontmatter


christian314159:
One last thing, as the story has been moved the link has changed and doesn't work anymore on the Kaizen site where Storybook is included: 
image

christian314159 (approved):
👍 👍 👍

PR #2640, opened 2022-03-23
feat: create loading-skeleton package and loadingParagraph component

NatalieLoyzaga:
## Why

## What

- Create a new loading-skeleton package under packages
- Create a new component within the package called LoadingParagraph
- Utilised existing props of the loadingPlaceholder variant="paragraph" with some naming refactors
- Transferred over existing styles from loadingPlaceholder.
- Created new sticker sheet style story

image
image

### DIFF
--- a/packages/loading-skeleton/docs/LoadingSkeleton.stories.tsx
+++ b/packages/loading-skeleton/docs/LoadingSkeleton.stories.tsx
@@ -1,8 +1,11 @@
 import React from "react"
+import { Story } from "@storybook/react"
 import { withDesign } from "storybook-addon-designs"
+import { Paragraph } from "@kaizen/typography"
 import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
 import { LoadingParagraph } from "../src/LoadingSkeleton"
 import { figmaEmbed } from "../../../storybook/helpers"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 
 export default {
   title: `${CATEGORIES.components}/${SUB_CATEGORIES.loadingSkeleton}`,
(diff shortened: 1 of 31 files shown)


packages/loading-skeleton/src/LoadingSkeleton/components/LoadingParagraph/LoadingParagraph.scss
```diff
@@ -0,0 +1,72 @@
...
+
+.link {
+  background-color: rgba($color-blue-500-rgb, 0.2);
+}
+
+.link.reversed {
+  background-color: rgba($color-white-rgb, 0.65);
+}
+
+.inline {
```
christian314159, line 49: As the stories are essentially our tests for this component via Chromatic, I think we should add an `inline` example story, as well as maybe something that is using `width` (or all in one?).

christian314159 (commented):
Great start, just see my comment around adding more stories.

packages/loading-skeleton/docs/LoadingSkeleton.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Story } from "@storybook/react"
+import { withDesign } from "storybook-addon-designs"
+import { Paragraph } from "@kaizen/typography"
+import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
+import { LoadingParagraph } from "../src/LoadingSkeleton"
+import { figmaEmbed } from "../../../storybook/helpers"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.loadingSkeleton}`,
```
HeartSquared, line 11: Based on the specs of your task, since the Loading components will have different props, you'll probably want different stories files so that the ArgsTable can correctly display the props you have access too, so I suggest renaming this file to `LoadingParagraph.stories.tsx`.

We can look at trying to combining and optimising stories in Storybook later :)
I think if you keep the `title` the same, they should combine in the tree at the very least.

packages/loading-skeleton/src/LoadingSkeleton/components/LoadingParagraph/LoadingParagraph.scss
NatalieLoyzaga: I have created a new row within the sticker sheet to show inline. Do you think any other props need to be shown as well?

packages/loading-skeleton/docs/LoadingSkeleton.stories.tsx
NatalieLoyzaga: Updated to LoadingParagraph and I agree once I've built them out I intend on finding a nice way to put them together.

packages/loading-skeleton/src/LoadingSkeleton/components/LoadingParagraph/LoadingParagraph.scss
christian314159: One could argue, we would need a story representing every prop - possibly just for the sake of having them tested in Chromatic.
The ones I mentioned had the most value to me, where the others might be a bit "niche".

NatalieLoyzaga: That additional story is also using width so I've just updated the row header to specify it as well as inline

christian314159: I'd be pretty sure that the below is also a problem with the current component: when it wraps there's some left margin on the first item of the 2nd row :(
I wouldn't worry about it right now, but I think it should probably be put in a flex container instead. If we only had one handy.... :)

image

christian314159 (approved):

PR #2644, opened 2022-03-24
feat: loadingheading component and loadingSkeleton stories

NatalieLoyzaga:
## Why

## What

- Create LoadingHeading component within loading-skeleton
- Create stories for each Heading variant
- Ensure the height and margin reflects typography font-size and line-height respectively
- Create loading skeleton stories using Guidance block to showcase IRL example (NOTE: this will get continuously refactored during this work)
image
image
image


christian314159:
Happy to move this to a different card, but is there anything we can do about the vertical alignment of the left hand title of each row of the stickersheet? The examples on the right line up relatively nicely with the loading placeholders, but those on the left look completely misplaced :(

packages/loading-skeleton/src/LoadingSkeleton/components/LoadingHeading/LoadingHeading.scss
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.heading-1 {
+  height: $typography-heading-1-font-size;
+  margin-bottom: $typography-heading-1-line-height;
+}
+
+.heading-2 {
+  height: $typography-heading-2-font-size;
+  margin-bottom: $typography-heading-2-line-height;
```
christian314159, line 37: I think this might be wrong, see the screenshot below where I made the heading go to 2 lines.
image
`margin-bottom` shouldn't be the full lineheight but rather `lineHeight - fontSize = marginBottom`.
And be aware, this might work here, but `line-height` could in some cases be a "unit-less" value, for example just `1.5` (i.e. 1.5 times the font size).

packages/loading-skeleton/package.json
```diff
@@ -33,6 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@kaizen/component-base": "^1.1.0",
+    "@kaizen/typography": "2.1.0",
```
christian314159, line 36: Hm, `typography` is just used for the stories, isn't it? Could we make it a dev dependency then? Would that work for storybook?

christian314159 (commented):
Please have a look at the margins and the package.json as per comments.  Thanks!

packages/loading-skeleton/package.json
NatalieLoyzaga: I am using it for the `HeadingVariants`  within LoadingHeading.tsx as well.

packages/loading-skeleton/src/LoadingSkeleton/components/LoadingHeading/LoadingHeading.scss
NatalieLoyzaga: I have updated the values to be the calculated value.

christian314159 (approved):
🚢

PR #2646, opened 2022-03-25
chore: add component generator

HeartSquared:
## Why

Helps making a new package much simpler.

https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-347

## What

Install [plop](https://github.com/plopjs/plop) and add templates.

Run `yarn plop` and follow the prompt (or `yarn plop NewComponentName`).

You can:
- create new packages with a component
- add a component to an existing component
- add a subcomponent to an existing component
	- if you add a subcomponent to a non-existent component, it will work, but the parent component will not be created


plop-templates/basic-component/src/ComponentName/ComponentName.spec.tsx.hbs
```diff
@@ -0,0 +1,19 @@
...
+
+const render{{pascalCase componentName}} = (customProps?: Partial<{{pascalCase componentName}}Props>) => {
+  const props = { ...DEFAULT_PROPS, ...customProps }
+  return render(<{{pascalCase componentName}} {...props} />)
+}
+
+describe("<{{pascalCase componentName}} />", () => {
+  it("test case description goes here", () => {
+    const { container } = render{{pascalCase componentName}}()
+    /** @todo: Fill in test case */
```
christian314159, line 17: What about adding a `expect(true).toBe(false)` here? That way you cannot forget about this file ;)

plop-templates/basic-component/package.json.hbs
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@kaizen/{{kebabCase componentName}}",
+  "version": "1.0.0",
+  "description": "The {{titleCase componentName}} component",
```
dougmacknz, line 4: One issue we've had in the past is that there's been a tendency to create a new package for every new component, without thinking about whether it belongs in an existing package, or whether it makes sense to create a package that will allow for grouping.

I'm not sure if this is feasible at all, but would there be a way to give users a prompt to check if they want to add the component to an existing package, or create a new package when they use this?

HeartSquared: We can definitely add more prompts and further template configs to cater for that :) I just figured that would be the next iteration. But I could experiment on that next learning day to make sure it works.

plop-templates/basic-component/docs/ComponentName.stories.tsx.hbs
```diff
@@ -0,0 +1,59 @@
...
+          'import { {{pascalCase componentName}} } from "@kaizen/{{kebabCase componentName}}".',
+      },
+    },
+    ...figmaEmbed("REPLACE_THIS_WITH_FIGMA_URL"), /** @todo: Replace with Figma frame url */
+  },
+  decorators: [withDesign],
+} as ComponentMeta<typeof {{pascalCase componentName}}>
+
+export const DefaultKaizenSiteDemo = args => <{{pascalCase componentName}} {...args} />
+DefaultKaizenSiteDemo.storyName = "Default (Kaizen Site Demo)"
```
NatalieLoyzaga, line 25: I think we have decided that the naming of the demo story should just be the component name rather than Default (Kaizen Site Demo).

NatalieLoyzaga: Would there be away to create the .mdx file for the component within the /site/components folder? and from there that demo story url/id would link up to the `demoStoryId` parameter - I believe the previous generator created this but also this was when the .mdx was within the component folder...What are your thoughts?

HeartSquared: >I think we have decided that the naming of the demo story should just be the component name rather than Default (Kaizen Site Demo).

So would it look something like this?

```suggestion
DefaultKaizenSiteDemo.storyName = "Component Name" // title case
```

Or rename the story itself from `DefaultKaizenSiteDemo` to `ComponentName`?

Although I wonder if that would look a bit redundant in the tree since the directory itself is the component name 🤔 

>Would there be away to create the .mdx file for the component within the /site/components folder? and from there that demo story url/id would link up to the demoStoryId parameter - I believe the previous generator created this but also this was when the .mdx was within the component folder...What are your thoughts?

We absolutely can! :) Good suggestion. I had completely forgot about it. Ideally we'd also have a proper README once we remove the magic that is currently turning them into pages.

mcwinter07: > I think we have decided that the naming of the demo story should just be the component name rather than Default (Kaizen Site Demo).

But would that be on the story with the args exposed or just a base example that we would reference in the mdx docs? Just wondering if we need some kind of notation to describe that it is a demo in storybook

plop-templates/site/docs/components/{{kebabCase componentName}}.mdx.hbs
```diff
@@ -0,0 +1,110 @@
+---
```
HeartSquared, line 1: Migrate this file to site repo

plop-templates/basic-component/docs/{{pascalCase componentName}}.stories.tsx.hbs
```diff
@@ -0,0 +1,59 @@
...
+          'import { {{pascalCase componentName}} } from "@kaizen/{{kebabCase componentName}}".',
+      },
+    },
+    ...figmaEmbed("REPLACE_THIS_WITH_FIGMA_URL"), /** @todo: Replace with Figma frame url */
+  },
+  decorators: [withDesign],
+} as ComponentMeta<typeof {{pascalCase componentName}}>
+
+export const DefaultKaizenSiteDemo = args => <{{pascalCase componentName}} {...args} />
+DefaultKaizenSiteDemo.storyName = "Default (Kaizen Site Demo)"
```
HeartSquared, line 25: Need to change this

docs/package-structure.md
```diff
@@ -91,11 +103,15 @@ The base package.json will look like this, where:
...
+  "dependencies": {
+    "@kaizen/component-base": "^1.1.0",
+    "classnames": "^2.3.1"
+  },
   "devDependencies": {
-    "@kaizen/component-base": "^1.0.0"
+    "rimraf": "^3.0.2"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0",
```
HeartSquared, line 98: Should include ^9.0.0 (also in template)

NatalieLoyzaga (approved):
This is awesome !!! Nice work

mcwinter07 (approved):

dougmacknz (approved):
Looks legit

PR #2650, opened 2022-03-29
feat: Create RichTextEditor component (sans controls)

dougmacknz:
## What
Creates the `RichTextEditor` component, without any controls for now (just paragraphs and hard breaks).

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
@@ -1,0 +1,65 @@
+import React, { useState, useEffect, HTMLAttributes, ReactNode } from "react"
+import { v4 } from "uuid"
+import classnames from "classnames"
+import { history } from "prosemirror-history"
+import { keymap } from "prosemirror-keymap"
+import { EditorState } from "prosemirror-state"
+import { Node } from "prosemirror-model"
+import { Label } from "@kaizen/draft-form"
+import { baseKeymap } from "prosemirror-commands"
+import { buildKeymap, useRichTextEditor } from "@cultureamp/rich-text-toolkit"
+import { OverrideClassName } from "@kaizen/component-base"
+import { EditorContentArray } from "./types"
+import { createSchemaFromControls } from "./schema"
+import styles from "./RichTextEditor.scss"
+
+export interface RichTextEditorProps
+  extends OverrideClassName<Omit<HTMLAttributes<HTMLDivElement>, "onChange">> {
+  onChange: (content: EditorContentArray) => void
+  value: EditorContentArray
+  labelText: ReactNode
+}
+
+export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
+  const { onChange, value, labelText, classNameOverride, ...restProps } = props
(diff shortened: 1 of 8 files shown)


packages/rich-text-editor/src/RichTextEditor/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const nodes: NodeSpec = {
+  doc: coreNodes.doc,
+  paragraph: coreNodes.paragraph,
+  text: coreNodes.text,
+  // eslint-disable-next-line camelcase
+  hard_break: coreNodes.hard_break,
+}
+
+export const marks: MarkSpec = {
```
dougmacknz, line 18: I've left some marks stuff in here from my copy+paste, even though we're not really using it yet, because `RichTextEditor` is using the `createSchemaFromControls` function below to create the schema.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { OverrideClassName } from "@kaizen/component-base"
+import { EditorContentArray } from "./types"
+import { createSchemaFromControls } from "./schema"
+import styles from "./RichTextEditor.scss"
+
+export interface RichTextEditorProps
+  extends OverrideClassName<Omit<HTMLAttributes<HTMLDivElement>, "onChange">> {
+  onChange: (content: EditorContentArray) => void
+  value: EditorContentArray
+  labelText: ReactNode
```
dougmacknz, line 20: I haven't added the ability to do `aria-labelledby` here yet instead because I'd quite like to do a discriminated union if I can, which I'll save for a separate piece of work.

dougmacknz: Card created for that here: https://cultureamp.atlassian.net/browse/KDS-356

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
+  const { onChange, value, labelText, classNameOverride, ...restProps } = props
+  const labelId = v4()
```
christian314159, line 25: This should use a `useMemo` or similar, otherwise it'll rerun and create a new ID on each render.

dougmacknz: Nice catch

dougmacknz: I did a quick search regarding `useMemo` vs `useState` for this and opted for `useState` after reading this https://stackoverflow.com/questions/57664450/usememo-vs-usestate-for-react-hooks-constants

mcwinter07 (dismissed):
Looks good to me - interesting read on the useState vs useMemo for the labelId, will have to pocket that nugget away for next time. Merge away in good health 🦖

mcwinter07 (approved):

PR #2653, opened 2022-03-29
ci: Support GitHub registry access from CI workflows

zioroboco:
## Why
This repo is required to consume a package published to the GitHub package registry from [cultureamp/rich-text-toolkit](https://github.com/cultureamp/rich-text-toolkit). A GitHub token is needed to access the registry, but the token which had previously been added to this repo is not available for use in public repos.

## What
This change updates the GitHub actions CI workflows to:
- use the workflow's own GitHub token to pull packages from the registry
- clean up a bunch of noise in the workflow definitions

I'm a bit murky on the chromatic workflow, but I'm reasonably confident I haven't broken anything.


dougmacknz (approved):
As mentioned in Slack, I only half understand what's going on here, but seems fine to me.

.github/workflows/ci.yml
```diff
@@ -8,114 +8,72 @@ jobs:
...
+          cache: yarn
+          node-version-file: .nvmrc
+          registry-url: https://npm.pkg.github.com/
+      - run: yarn install --frozen-lockfile
         env:
-          PACKAGE_REGISTRY_ACCESS_TOKEN: ${{ secrets.PACKAGE_REGISTRY_ACCESS_TOKEN }}
-        run: |
-          npm config set "//npm.pkg.github.com/:_authToken" "$PACKAGE_REGISTRY_ACCESS_TOKEN"
-          yarn install --frozen-lockfile
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
zioroboco, line 18: These install steps had previously been attempting to use the `PACKAGE_REGISTRY_ACCESS_TOKEN` org secret to fetch internal packages from the GitHub package registry. However, this secret is (currently) not available to public repos.

This change sets the GitHub actions workflow's own GitHub token as the registry auth token, which will enable the step to install any public packages on the GitHub registry.

Note that internal packages will still not be available — we can cross this bridge when (if?) we come to it.

.npmrc
```diff
@@ -0,0 +1 @@
+@cultureamp:registry=https://npm.pkg.github.com/
```
zioroboco, line 1: This lets yarn know which registry to use for `@cultureamp`-scoped packages (i.e. the GitHub registry).

.github/workflows/ci.yml
```diff
@@ -8,114 +8,72 @@ jobs:
...
+      - run: yarn install --frozen-lockfile
         env:
-          PACKAGE_REGISTRY_ACCESS_TOKEN: ${{ secrets.PACKAGE_REGISTRY_ACCESS_TOKEN }}
-        run: |
-          npm config set "//npm.pkg.github.com/:_authToken" "$PACKAGE_REGISTRY_ACCESS_TOKEN"
-          yarn install --frozen-lockfile
-      - name: Run unit tests
-        run: USE_REACT_17=false yarn test --ci
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - run: USE_REACT_17=false yarn test --ci
```
zioroboco, line 64: Maybe a bit subjective, but I prefer that steps with a fairly readable one-line implementation (e.g. invoking a package script) don't include a name, since I find it easier to scan the output from the pipeline and understand the specific commands each step corresponds to.

zioroboco (commented):
Hey @dougmacknz — left comments to clarify the specific changes, because I don't want to be rearranging things in ways that don't make sense to the people that work in this repo. Feel free to merge if you're comfortable with these changes.

dougmacknz:
Thanks @zioroboco, that helped a lot.

PR #2654, opened 2022-03-29
feat(splitbutton): add SplitButton package

HeartSquared:
## Why

The draft `SplitButton` component was very different from what we have in the UI kit. Rather than trying to change the existing one, rebuilt it from scratch.

## What

Add new (refactored) `SplitButton` package.
Includes tests, stories, and extra stories for Chromatic snapshots.

- I introduced some new patterns to get better Chromatic snapshots (eg. simulating pseudo states) and also to improve Storybook controls (found out there’s a subcomponents key which is nifty, although very limited in what it does)
- Extended our testing library expect methods
- Ideally I’d have been able to use the Button component, but I didn't want to mess with it to try and get it working for my needs. Built a simplified version reflecting the same styles (maybe it can be used as a base to refactor the existing one?).

## Screenshots

New Chromatic snapshots [here](https://www.chromatic.com/build?appId=60a1e4a102f0cb003b5d19d6&number=434)

|Sticker Sheet - Default|
|---|
|image|

|Sticker Sheet - Reversed|
|---|
|image|

### Edit 4 Apr:

Slight changes to the pseudo states after speaking with @jmitsios.
The focus state will also no longer appear on mouse clicks, but will still appear when navigating using the keyboard.

|Pseudo styles|
|---|
|image|

### DIFF
--- a/packages/split-button/docs/SplitButton.stories.tsx
+++ b/packages/split-button/docs/SplitButton.stories.tsx
@@ -1,0 +1,221 @@
+import React from "react"
+import { ComponentMeta, Story } from "@storybook/react"
+import { action } from "@storybook/addon-actions"
+import { MenuItem, MenuList } from "@kaizen/draft-menu"
+import { withDesign } from "storybook-addon-designs"
+import isChromatic from "chromatic"
+import duplicateIcon from "@kaizen/component-library/icons/duplicate.icon.svg"
+import editIcon from "@kaizen/component-library/icons/edit.icon.svg"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
+import { figmaEmbed } from "../../../storybook/helpers"
+import { CATEGORIES } from "../../../storybook/constants"
+import { SplitButton, SplitButtonProps } from "../"
+import { ActionButton, DropdownButton } from "../src/SplitButton/components"
+
+const IS_CHROMATIC = isChromatic()
+
+const ACTION_BUTTON_PROPS__BUTTON = {
+  label: "Edit Survey",
+  onClick: action("clicked"),
+}
+const ACTION_BUTTON_PROPS__ANCHOR = {
+  label: "Edit Survey",
+  href: "//example.com",
+}
+
+const DROPDOWN_CONTENT__ENABLED = (
+  <MenuList>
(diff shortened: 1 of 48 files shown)


packages/split-button/docs/SplitButton.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+      </StoryWrapper>
+
+      <StoryWrapper isReversed={isReversed}>
+        <StoryWrapper.RowHeader headings={["Hover", "Focus"]} />
+        <StoryWrapper.Row rowTitle="Action button">
+          <SplitButton
+            isReversed={isReversed}
+            actionButtonProps={{
+              ...ACTION_BUTTON_PROPS__BUTTON,
+              classNameOverride: "__hover",
```
mcwinter07, line 140: Just wary on this since we don't usually showcase hover or focus states - do think this is something we should start doing? or is it more relevant here because split button has two focusable elements

mcwinter07 (approved):
Nice update to a component that needed some love 👍  I left a small comment on the focus / hover states in the sticker sheets - I get why they might be there but not sure if needed. Either way happy its a green tick if you think it'll help the consumers 🦖

jmitsios:
@HeartSquared Very nice update! 👍   Only minor feedback I've got - 

- After clicking the button it puts it into the focus state.  I don't think this should be happening?
image

- Hovering over the drop-down menu arrow isn't highlighting (changing the color) of the separator between the button and arrow:
image

HeartSquared:
So after a chat with @jmitsios:
- There are a few minor style changes to be made to the dropdown button (I replicated what Figma was, but it has been revised since).
- I'll look into possibly adjusting the stories to not leave the focus state when clicking the action button since it isn't something that happens in other stories
- Anchor behaviour is normal :)

packages/split-button/docs/SplitButton.stories.tsx
HeartSquared: We should actually be doing this more often so we can have Chromatic help us pick up if there are changes between the snapshots :)

mcwinter07 (approved):
The rebuild looks good, especially digging the thorough documentation in the args table! You can probably make a codemod for this too. Merge away in good health 🦖

PR #2665, opened 2022-03-31
feat: LoadingGraphic and update loading skeleton story

NatalieLoyzaga:
## Why

## What

- Create LoadingGraphic component and stories.
- Update LoadingSkeleton story to include loadingGraphic.
- Update other LoadingSkeleton stories to disable control on stickersheet.
- Refactor styles to be within one common stylesheet across the package.

image

image


packages/loading-skeleton/docs/LoadingGraphic.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { LoadingGraphic } from "../src/LoadingSkeleton"
+
+import { figmaEmbed } from "../../../storybook/helpers"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.loadingSkeleton}/Loading Graphic`,
+  component: LoadingGraphic,
+  parameters: {
+    chromatic: { disable: false },
```
HeartSquared, line 18: This will enable all the stories (so you wouldn't need to put them on the individual ones). If you want to only send a select few, then you'd want to remove this :)

packages/loading-skeleton/src/LoadingSkeleton/components/LoadingGraphic/LoadingGraphic.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export type GraphicSizes =
+  | "small"
+  | "medium"
+  | "large"
+  | "xlarge"
+  | "xxlarge"
+  | "scene"
+
+const graphicSizesRem = new Map<string, number>([
```
christian314159, line 14: Just a question really: Could `<string, number>` here be `<GraphicSizes, number>`? It is one of those, isn't it?

christian314159 (dismissed):

HeartSquared:
I have quite a few questions and likely suggestions, so let's have a chat when you have time :)

mcwinter07 (approved):
Look good to me! We may want to create a story with some A11Y suggestions with native HTML attributes but I think that would need a little more investigation and warrants a seperate PR 🦖 merge in good health

PR #2667, opened 2022-04-04
feat: Add inputrules to RTE

dougmacknz:
## What
Adds 'inputrules' to the RichTextEditor, including:
* Smart quotes
* Two hyphens into an emdash
* Three full stops into an ellipsis

## Why
Because these things are expected from rich text editors, and sets us up for adding list functionality.

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/keymap.ts
+++ b/packages/rich-text-editor/src/RichTextEditor/keymap.ts
@@ -1,0 +1,78 @@
+import { EditorState, Transaction } from "prosemirror-state"
+import { Schema } from "prosemirror-model"
+import {
+  chainCommands,
+  exitCode,
+  joinDown,
+  joinUp,
+  lift,
+  selectParentNode,
+  setBlockType,
+  toggleMark,
+} from "prosemirror-commands"
+import { redo, undo } from "prosemirror-history"
+import { undoInputRule } from "prosemirror-inputrules"
+
+const mac =
+  // eslint-disable-next-line ssr-friendly/no-dom-globals-in-module-scope
+  typeof navigator != "undefined" ? /Mac/.test(navigator.platform) : false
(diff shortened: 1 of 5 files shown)


packages/rich-text-editor/src/RichTextEditor/keymap.ts
```diff
@@ -0,0 +1,78 @@
...
+export function buildKeymap(schema: Schema) {
+  const keys: {
+    [key: string]: (
+      state: EditorState<any>,
+      dispatch?: ((tr: Transaction<any>) => void) | undefined
+    ) => boolean
+  } = {
+    "Mod-z": undo,
+    "Shift-Mod-z": redo,
+    Backspace: undoInputRule,
```
dougmacknz, line 29: This is the one line I added to this file

packages/rich-text-editor/src/RichTextEditor/keymap.ts
```diff
@@ -0,0 +1,78 @@
+import { EditorState, Transaction } from "prosemirror-state"
```
dougmacknz, line 1: I've changed my mind and decided that this keymap file belongs in Kaizen instead of the toolkit. My reasoning is that it's quite tied to things that we currently have in Kaizen, like the name of marks in the schema. It's also consistent with what I've done with the inputrules file here.

Perhaps we could revisit this in the future and see if there's something we could extract out here to the toolkit so make for consistent keymaps across custom RTEs.

mcwinter07: I think that's fair. As you said we can extract it later when we've built out the functionality of the out of the box solution - I think by then we'll have a firm understanding of what needs to be and can be passed into the more composable solution

mcwinter07 (approved):

PR #2668, opened 2022-04-04
feat: update interactive state styles and add toggle mood support

mcwinter07:
## Why
Update interactive states and adding moods to align to UI kit

## What
- adds mood prop to toggle icon button
- updates toggle icon button styles with new moods


packages/rich-text-editor/docs/ToggleIconButton.stories.tsx
```diff
@@ -28,11 +30,68 @@ export default {
...
+          isActive={false}
+          {...defaultButton}
+        />
+        <ToggleIconButton
+          mood="secondary-destructive"
+          isActive={false}
+          {...defaultButton}
+        />
+      </StoryWrapper.Row>
+      <StoryWrapper.Row rowTitle="Selected">
```
dougmacknz, line 73: Not ideal that we don't have this name consistent between implementation and UI Kit. @jmitsios can we convince you to call this `Active` instead of `Selected`? Reasoning here: https://cultureamp.slack.com/archives/C02NUQ27G56/p1646181677465939?thread_ts=1646181455.286509&cid=C02NUQ27G56

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -41,12 +35,78 @@
...
+
+  &.active {
+    background-color: $color-red-200;
+  }
 }
 
 .iconWrapper {
-  width: 1rem;
-  height: 1rem;
+  width: 15px;
```
dougmacknz, line 110: Could we do 0.9375rem instead of a px value to make these scale up with font size?

dougmacknz:
image
Tooltip placement a lil concerning 🤨

Also, should `base` be `default`?

packages/rich-text-editor/docs/ToggleIconButton.stories.tsx
jmitsios: > Not ideal that we don't have this name consistent between implementation and UI Kit. @jmitsios can we convince you to call this `Active` instead of `Selected`? Reasoning here: https://cultureamp.slack.com/archives/C02NUQ27G56/p1646181677465939?thread_ts=1646181455.286509&cid=C02NUQ27G56

Consider it done.
These have been updated in the UI Kit now to be `Active` instead :)

jmitsios:
Couple small points:
- Can the `Tooltip` be set to only show _Above_ or _Below_, so that it isn't blocking the other icons while hovering around? _(see image)_
- Could we put a small delay on the `Tooltip` appearing on hover (try 500ms?)
- While a button is in the `Active` state, we would use the same Hover as already implemented.

image

mcwinter07:
> Couple small points:
> 
> * Can the `Tooltip` be set to only show _Above_ or _Below_, so that it isn't blocking the other icons while hovering around? _(see image)_
> * Could we put a small delay on the `Tooltip` appearing on hover (try 500ms?)
> * While a button is in the `Active` state, we would use the same Hover as already implemented.
> 
> image

Hey @jmitsios, re the above I've explicitly set the tooltip to show above, but it is using the Tooltip component, which will shift it position based on the size of the wrapper. To showcase this I've made a new story to indicate a more real world example of how it should render on an actual page with content.

RE: the delay, since we're using the tooltip component we don't have a way to bake an animation delay into it, we could potentially create something custom that resembles the tooltip? I'd be inclined to hold off on that until we tackle those keyboard shortcuts that we were exploring as an extension to this

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
mcwinter07: Done 👍

packages/rich-text-editor/docs/ToggleIconButton.stories.tsx
mcwinter07: Updated 🦖

mcwinter07:
> image
> 
> Tooltip placement a lil concerning 🤨
> Also, should `base` be `default`?

I've updated the stories to get them a little more inline with figma - had to make a presentational wrapper since the toggle icon won't usually just be in a grid with `fr` to determine its containers width, but it should have a more accurate feel now

dougmacknz (approved):

jmitsios (approved):
Awesome work! Only one thing I've noticed here -
- After clicking on one of these examples, the Tooltip remains active on the screen without switching to the Active state _(Also happens in Default (Kaizen Demo)_
image

PR #2669, opened 2022-04-04
fix: Set pointer-events to none on RTE focus ring

dougmacknz:
## What
Removes pointer events from the psuedo element created for the RTE focus ring.

## Why
This element is getting in the way of selecting text.


christian314159 (approved):
👍 I can select things again! :)

PR #2672, opened 2022-04-04
feat: Add list functionality to RichTextEditor

dougmacknz:
Adds the ability to create bullet and numbered lists (only via inputrules for now without a toolbar).

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
@@ -37,11 +37,9 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
       schema,
       plugins: [
         history(),
-        keymap({
-          ...baseKeymap,
-          ...buildKeymap(schema),
-        }),
-        buildInputRules(),
+        keymap(buildKeymap(schema)),
+        keymap(baseKeymap),
+        buildInputRules(schema),
       ],
     }),
     { "aria-labelledby": labelId }
--- a/packages/rich-text-editor/src/RichTextEditor/keymap.ts
+++ b/packages/rich-text-editor/src/RichTextEditor/keymap.ts
@@ -12,6 +12,14 @@ import {
 } from "prosemirror-commands"
 import { redo, undo } from "prosemirror-history"
 import { undoInputRule } from "prosemirror-inputrules"
+import {
+  wrapInList,
+  splitListItem,
+  liftListItem,
+  sinkListItem,
+} from "prosemirror-schema-list"
+
+// Most/all of this file is copied from https://github.com/ProseMirror/prosemirror-example-setup/blob/master/src/keymap.js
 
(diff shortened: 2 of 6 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -37,11 +37,9 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
       schema,
       plugins: [
         history(),
-        keymap({
-          ...baseKeymap,
-          ...buildKeymap(schema),
-        }),
-        buildInputRules(),
+        keymap(buildKeymap(schema)),
+        keymap(baseKeymap),
```
dougmacknz, line 41: When these two things were combined into a single `keymap()` — it meant that some things were getting overwritten, e.g. "Enter". The order also matters, the base keymap needed to come last.

mcwinter07: Interesting, I suppose it doesn't hurt to delineate between what we're pulling in from prose mirror and what we're defining ourselves

packages/rich-text-editor/src/RichTextEditor/keymap.ts
```diff
@@ -70,8 +78,22 @@ export function buildKeymap(schema: Schema) {
...
+  if (schema.nodes.ordered_list) {
+    keys["Shift-Ctrl-9"] = wrapInList(schema.nodes.ordered_list)
+  }
+
+  if (schema.nodes.list_item) {
+    const type = schema.nodes.list_item
+    keys["Enter"] = splitListItem(type)
+    keys["Mod-["] = liftListItem(type)
+    keys["Mod-]"] = sinkListItem(type)
   }
```
mcwinter07, line 97: Nice! Definitely think it worth investing time after the toolbar has been integrate to pass these commands into the tooltips

mcwinter07 (approved):
Awesome stuff! Not that I wasn't sold on prosemirror before, but really seeing its value shine through with the indentation helpers 👍  Merge in good health 🦖

packages/rich-text-editor/src/RichTextEditor/schema.ts
```diff
@@ -47,5 +48,19 @@ export const createSchemaFromControls = controls => {
...
+
+  // TODO: This is hard coded to always enable lists for now,
+  // but once we have the toolbar we can use the controls to determine this.
+  const listsEnabled = true
+
+  const schema = new Schema({ nodes, marks: newMarks })
+
+  if (listsEnabled) {
+    return new Schema({
+      nodes: addListNodes(schema.spec.nodes, "paragraph block*", "block"),
```
dougmacknz, line 60: Bit of a gotcha here, this `addListNodes` function has to be run on a schema that's built, rather than just the nodes object. 

https://discuss.prosemirror.net/t/nodes-append-is-not-a-function/2216

PR #2679, opened 2022-04-06
feat: Add rows prop to RichTextEditor for setting min-height

dougmacknz:
## What
Adds a `rows` prop to `RichTextEditor`, which will set a min-height on the editable area, in units of body paragraph line height. Pretty much the same as we've done on `TextAreaField`, but this component has autogrow enabled all the time. 

The default value has been set to 3, in line with `TextAreaField` and roughly matching the UI Kit.

## Why
Gives users a good way to set a min height on the text area.

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
@@ -52,3 +52,9 @@
     bottom: calc(-1 * #{$focus-ring-offset});
   }
 }
+
+@for $i from 1 through 20 {
+  .editor.rows#{$i} > :global(.ProseMirror) {
+    min-height: calc($typography-paragraph-body-line-height * $i);
+  }
+}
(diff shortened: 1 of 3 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
```diff
@@ -52,3 +52,9 @@
     bottom: calc(-1 * #{$focus-ring-offset});
   }
 }
+
+@for $i from 1 through 20 {
+  .editor.rows#{$i} > :global(.ProseMirror) {
+    min-height: calc($typography-paragraph-body-line-height * $i);
+  }
+}
```
mcwinter07, line 60: Nice, this feels almost CSS in JS-esque, just in SCSS

mcwinter07 (approved):
Nice, I like the solution to add row props like `<textarea/>` but with our line height 🦖  Merge in good health

PR #2680, opened 2022-04-06
chore(docs): Made interactive CheckboxGroup story actually interactive [KDS-378]

christian314159:
## Why
I noticed that the ["interactive" story](https://cultureamp.design/storybook/?path=/story/components-form-checkbox-group--interactive-kaizen-site-demo) for `CheckboxGroup` wasn't actually that interactive, it could only be toggled from `mixed` to `off`, but not `on`.

## What
- removed the `actions` config option that caught the event callbacks that would set the state
- refactored the story helper component from a class to a functional component

### DIFF
--- a/draft-packages/form/docs/CheckboxGroup.stories.tsx
+++ b/draft-packages/form/docs/CheckboxGroup.stories.tsx
@@ -15,18 +15,20 @@ interface RenderProps {
   onCheckHandler: (event: React.ChangeEvent<HTMLInputElement>) => any
 }
 
-interface Props {
+interface CheckboxGroupExampleProps {
   render: (props: RenderProps) => JSX.Element
 }
 
-function CheckboxGroupExample({ render }) {
+const CheckboxGroupExample: React.VFC<CheckboxGroupExampleProps> = ({
+  render,
+}) => {
   const [checkedStatus, setCheckedStatus] = useState("mixed")
   const onCheckHandler = () => {
     const newStatus = checkedStatus === "on" ? "off" : "on"
     setCheckedStatus(newStatus)
   }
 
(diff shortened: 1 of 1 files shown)


draft-packages/form/docs/CheckboxGroup.stories.tsx
```diff
@@ -19,42 +19,20 @@ interface Props {
   render: (props: RenderProps) => JSX.Element
 }
 
-class CheckboxGroupExample extends React.Component<Props> {
-  public state = {
-    checkedStatus: "mixed",
+function CheckboxGroupExample({ render }) {
```
HeartSquared, line 22:
```suggestion
const CheckboxGroupExample: React.VFC<Props> = ({ render }) => {
```

I would probably change `Props` to `CheckboxGroupExampleProps` too, but up to you

draft-packages/form/docs/CheckboxGroup.stories.tsx
```diff
@@ -19,42 +19,20 @@ interface Props {
...
-    return (
-      <div>
-        {render({
-          checkedStatus: this.state.checkedStatus,
-          onCheckHandler: this.onCheckHandler,
-        })}
-      </div>
-    )
-  }
+  return <>{render({ checkedStatus, onCheckHandler })}</>
```
HeartSquared, line 29:
```suggestion
  return render({ checkedStatus, onCheckHandler })
```

draft-packages/form/docs/CheckboxGroup.stories.tsx
christian314159: Thanks, I'll have a look. As it's a story I wasn't too concerned with "TS'ing" it, as I probably should have been.

PR #2681, opened 2022-04-07
feat: Rich Text Editor toolbar integration

mcwinter07:
## Why
Adding toolbar to the Kaizen Rich Text Editor that utilizes UI primitives

## What
- adds toolbar UI and functionality into Kaizen RTE
- add tests for toolbar navigation
- add styles to support toolbar wrapper


packages/rich-text-editor/docs/RichTextEditor.stories.tsx
```diff
@@ -26,4 +26,21 @@ export const Default = () => {
     </>
   )
 }
+
+Default.storyName = "Default (Kaizen Demo)"
+
+export const WithControls = () => {
```
dougmacknz, line 32: FWIW I don't think we need to keep the example with no toolbar, but lets address this later by replacing these with a sticker sheet when we add the moods.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -3,45 +3,56 @@ import { v4 } from "uuid"
...
 import { OverrideClassName } from "@kaizen/component-base"
 import { EditorContentArray } from "./types"
 import { createSchemaFromControls } from "./schema"
+import { createInitialState } from "./state"
 import { buildKeymap } from "./keymap"
 import { buildInputRules } from "./inputrules"
 import styles from "./RichTextEditor.scss"
+import { Toolbar, ToolbarSection, ToggleIconButton } from "./"
 
+type ToolbarControls = "strong" | "em" | "underline"
```
dougmacknz, line 24: I'm trying to remember if we decided against this for some reason, but I'm thinking we should go for `bold` and `italic` instead of `strong` and `em` to make these more user friendly.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -3,45 +3,56 @@ import { v4 } from "uuid"
...
-      schema,
-      plugins: [
-        history(),
-        keymap(buildKeymap(schema)),
-        keymap(baseKeymap),
-        buildInputRules(schema),
-      ],
-    }),
+  const [editorRef, editorState, dispatchTransaction] = useRichTextEditor(
+    createInitialState(value, schema, [
```
dougmacknz, line 50: I had inlined this stuff so that we didn't need that extra `state.ts` file. Did you intend to change this back to the old way?

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -18,7 +18,8 @@
     outline: none;
   }
 
-  &:global(.focus-visible)::after {
+  &:global(.focus-visible)::after,
+  &:focus-visible:after {
```
dougmacknz, line 22: The `.focus-visible` class polyfill should cover us for all cases, unless you discovered something I'm not aware of?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -54,11 +65,34 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
...
       {labelText && <Label id={labelId} labelText={labelText} />}
       {/* TODO: add a bit of margin here once we have a classNameOverride on Label */}
-      <div
-        ref={editorRef}
-        className={classnames(styles.editor, classNameOverride)}
-        {...restProps}
-      />
+      <div className={styles.editorWrapper}>
+        {controls && (
+          <Toolbar aria-controls="toolbar-ref-id" aria-label="Text formatting">
```
dougmacknz, line 70: This `toolbar-ref-id` isn't referencing anything yet I think? We'll want to make sure it's unique as well in the case of multiple RTEs on one screen, probably with `uuid`.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
mcwinter07: That sounds good to me 👍 Happy to make that change

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
mcwinter07: Oh yes, so this was a weird one. There's actually a case where when you toggle the button with keyboard it looses the .focus-visible class entirely despite still retaining focus on the button, native focus-visible seems to catch this but the polyfill doesn't

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
mcwinter07: True its not entirely necessary, my thinking was to have it export more like a helper function so it could be more easily digestible to a consumer / someone referencing this if they need to build a custom RTE.

dougmacknz (approved):
Woop woop! Looks good, ship in great health

PR #2689, opened 2022-04-13
test: clean up flaky tests and console errors

mcwinter07:
## Why
- There has been several occasions where the notification package has cause a build failure at the jest step. This only occur when the entire test suite it run and not on a component level.
- This PR aims to clean up this issue, along with several updates to tests to reduce the noise of the console errors.

## What
- fix unstable_flushDiscreteUpdates warning caused by [react issue](https://github.com/facebook/react/issues/10389)
- fix wrap "not wrapped in act(...)" warning
- fix flaky test for generic notifications that occasionally brakes jest build step


HeartSquared:
As we discovered, GenericNotification test is still flaky :(

packages/date-picker/src/DatePicker/DateRangePicker.spec.tsx
```diff
@@ -27,6 +27,7 @@ describe("<DateRangePicker />", () => {
   })
 
   it("is able to select date range and shows in button", async () => {
+    jest.setTimeout(10000)
```
mcwinter07, line 30: I've added this here because running the full test suite locally sometime fails when the time to complete this test exceeds jests standard timeout
Screen Shot 2022-04-21 at 1 44 21 pm
s

mcwinter07:
Calling out that there are 3 error that will remain after this (see screen below) - I've held off on these since they would likely require some refactoring and are not issues with the tests themselves.
Screen Shot 2022-04-21 at 2 09 04 pm
Screen Shot 2022-04-21 at 2 08 53 pm
Screen Shot 2022-04-21 at 2 08 46 pm

packages/date-picker/src/DatePicker/DateRangePicker.spec.tsx
NatalieLoyzaga: Interesting - Will look into a permeant fix later on.

NatalieLoyzaga (approved):
Looks great! Nice work :)

PR #2691, opened 2022-04-14
chore(deps): update design-tokens deps to include version 8

mcwinter07:
## Why
Required to add support for version 8 of design tokens following recent elm removal causing a major version bump

## What
- update design-tokens dependencies to include version 8


draft-packages/button/package.json
```diff
@@ -42,7 +42,7 @@
     "rimraf": "^3.0.2"
   },
   "peerDependencies": {
-    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0",
+    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0",
```
HeartSquared, line 45: I reckon we probably don't need to update the packages which have a blessed counterpart without breaking changes (since you're going to remove them as part of elm removal anyway).

HeartSquared (approved):

PR #2692, opened 2022-04-14
feat: add automationId and autoHide props to `draft-filter-menu-button`

Moskvyak:
## Why

filter-menu-button is using menu component, however does not provide the control over the autoHide behaviour like here 
https://cultureamp.design/storybook/?path=/story/components-menu--auto-hide-behaviours

https://cultureamp.atlassian.net/browse/KDS-402 

## What

- Added `autoHide` optional property. It is set to "on" by default
- Added `automationId` optional property for automated testing


draft-packages/filter-menu-button/docs/FilterMenuButton.stories.tsx
```diff
@@ -242,3 +242,163 @@ export const DefaultWithChildrenAdvancedFilter = () => {
...
+        metadata={checkedTraits}
+        isDropdownVisible={isDropdownVisible1}
+        toggleDropdown={toggleDropdown1}
+        hideDropdown={hideDropdown1}
+        autoHide="on"
+      >
+        <>
+          <div className={styles.content}>
+            <CheckboxGroup labelText="Traits">
+              {DROPDOWN_OPTIONS.map(trait => (
```
HeartSquared, line 300: When I select any option in any of the 3 dropdowns, the selected values apply to all 3 buttons as opposed to just the one I chose from.
Since the contents of the menu don't matter as long as there's at least one option, you could simplify and just hard code one option per each.

draft-packages/filter-menu-button/docs/FilterMenuButton.stories.tsx
```diff
@@ -242,3 +242,163 @@ export const DefaultWithChildrenAdvancedFilter = () => {
...
+            />
+          </div>
+        </>
+      </FilterMenuButton>
+    </>
+  )
+}
+
+AutoHideBehaviours.storyName = "Auto hide behaviours"
+AutoHideBehaviours.parameters = { chromatic: { disable: false } }
```
HeartSquared, line 404:
```suggestion
```

Since autohide is functional and not visually different, we should just not send it to Chromatic.

draft-packages/filter-menu-button/docs/FilterMenuButton.stories.tsx
```diff
@@ -242,3 +242,163 @@ export const DefaultWithChildrenAdvancedFilter = () => {
...
+    IS_INITIAL_DROPDOWN_VISIBLE
+  )
+  const [isDropdownVisible2, setIsDropdownVisible2] = useState<boolean>(
+    IS_INITIAL_DROPDOWN_VISIBLE
+  )
+  const [isDropdownVisible3, setIsDropdownVisible3] = useState<boolean>(
+    IS_INITIAL_DROPDOWN_VISIBLE
+  )
+  const [appliedFilters, setAppliedFilters] = useState<DropdownOption[]>(
+    IS_CHROMATIC ? DROPDOWN_OPTIONS_CHROMATIC_SELECTED : []
```
HeartSquared, line 257:
```suggestion
    []
```

We don't need to send this story to Chromatic (as mentioned in another comment)

draft-packages/filter-menu-button/KaizenDraft/FilterMenuButton/FilterMenuButton.tsx
```diff
@@ -49,17 +56,23 @@ export interface FilterMenuButtonProps {
...
   toggleDropdown,
   hideDropdown,
   onFilterClear,
 }: FilterMenuButtonProps) => {
   const dropdownId = `${id}-dropdown`
   return (
-    <div id={id} className={styles.filterMenuButton}>
+    <div
+      id={id}
+      data-automation-id={automationId}
```
BurntCaramel, line 73: Is there a reason that I‘d want the test id to be on the containing div and not the control itself? What are these test ids usually used for? I usually prefer getByRole so try to not use the getByTestId function

HeartSquared: Literally just for testing purposes. It's an old pattern to use `automationId` that we're deprecating (the complex components haven't had the fix yet due to the amount of work needed). Ideally we should avoid using test ids as you do, but it's just there as an escape hatch when you can't target with anything else (which generally signifies another problem, but that's a whole different discussion).

If you don't need to use the test id immediately, I reckon we just not add it.

In terms of why it's on the container, it's just a case of ease of use as drilling down is generally easier than drilling up.

Moskvyak: Good points,

We are going to rethink how we'd like to use the automated tests before implementing extra props, I removed it from this PR

HeartSquared (approved):

PR #2695, opened 2022-04-20
feat: DateInputPicker

NatalieLoyzaga:
## Why
This PR is quite a large refactor of the DatePicker as it was originally made as a button due to time constraints. We've now updated the button to be an input for a11y reasons and with that comes a lot more logic and UX decisions.
### UX
I'd suggest having a play with it to grasp the basic UX patterns. As a short list of our decisions and why:
- When you click into the input, you can type but it doesn't open the calendar yet.
  - This is for keyboard users that want to input the date manually or enter a date they know e.g. their birthday.
- Pressing the down arrow when in the input opens it.
  - This was a decision we made that we thought was a simple way of tying opening it to a key that wasn't enter (as that may submit a form).
- When a date is selected and you're not in the input, it formats to our fancy format e.g. March 14, 2022.
  - We don't want the user to be able to edit the fancy format, as that's a nightmare to parse. This is a simple way of keeping it in just a single format when editable.
- When a date is selected and you focus back into the input, it formats to the mm/dd/yyyy version so the user can update with the keyboard.
  - Note: it will only validate when you blur. Otherwise validation will be noisy.
- When selecting a date inside the calendar via keyboard, on enter, the focus will return back to the button.
  - This allows the user to open it back up on enter e.g. click of the button.

### API
We've made this as simple as possible so the consumer doesn't have to fiddle around with managing a lot of state. You just pass it your state and setter and the component will do the rest.

### What hasn't made this PR
- Due to @lloydstubber leaving in a week, we've left localisation out of this PR so we can start on converting the DateRangePicker over to an input as well. We had a look at how this would be done and it's pretty straight forward.
- RTL Support. The library has some basic weeks etc. baked in but for the sake of time this will come later.

## What

**Toggle between formats on focus and onBlur**

https://user-images.githubusercontent.com/48232362/164342866-5c1471a1-72f9-4454-ad03-f12933ecce3d.mov

**Typing date into input focusing out (validation). Opening calendar and date is visibly selected**

https://user-images.githubusercontent.com/48232362/164345123-7fce4563-44e4-4a4f-913f-5699468d8584.mov

**Selecting date, updating to invalid date via input, focusing out to trigger validation. Updating back to valid date and focusing out to remove error message**

https://user-images.githubusercontent.com/48232362/164349406-e20e39d6-8522-4f9d-8eb9-88c43c435fb8.mov

**Date Input Sticker Sheet**

image


jmitsios:
@NatalieLoyzaga is there a date range limit this won't format to by any chance?
I'm testing it out with my birthday and it seems to throw a validation error:

https://user-images.githubusercontent.com/95595231/164366788-70a51d7e-7ba4-428f-97ea-9188a153a1c0.mov

NatalieLoyzaga:
@jmitsios This is because that date is disabled so it throws a validation error.  A good reminder that we should of removed disabled dates from the demo story! Will update that now :)

packages/date-picker/src/DatePicker/components/DateInput/DateInput.scss
```diff
@@ -0,0 +1,96 @@
...
+
+  svg {
+    margin-top: 2px;
+  }
+}
+
+///////////////////////////////////////////////////
+// VALIDATION STYLES
+///////////////////////////////////////////////////
+
```
mcwinter07, line 69: We don't currently have any stories showcase the success or caution states for the datepicker - is this something we're able to simulate in the stickersheet or a separate story?

draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss
```diff
@@ -125,6 +125,11 @@ $input-with-icon-padding: $input-icon-size + $ca-grid * 0.75;
   .endIconAdornment {
     @include vertically-center-icon;
     @include ca-position($end: $ca-grid / 2);
+
```
mcwinter07, line 128: You need to add some rtl support to handle to handle the icon and text placement to ensure they don't overlap

Screen Shot 2022-04-22 at 12 19 13 pm

packages/date-picker/src/DatePicker/components/DateInput/DateInput.scss
```diff
@@ -0,0 +1,96 @@
...
+.caution {
+  @include input-icon($color-yellow-200, $color-purple-700);
+  @include ca-animation-fade(in);
+}
+
+.message {
+  &.disabled {
+    opacity: $input-disabled-opacity;
+  }
+}
```
mcwinter07, line 96: I think this is more DatePicker implementation but I realized that rtl changes the direction of the calendar month arrows, which should ideally remain in the same direction.

Screen Shot 2022-04-22 at 12 31 25 pm

mcwinter07 (commented):
Looking good so far, I've made a few comments and left some feedback. It's a pretty chunky commit so I'd be keen to go through the rest of it when we're in the office on Tuesday / Wednesday :)

draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss
NatalieLoyzaga: We aren't handling RTL in this iteration :)

packages/date-picker/src/DatePicker/components/DateInput/DateInput.scss
NatalieLoyzaga: We aren't handling RTL in this iteration :)

draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss
mcwinter07: gotcha 👍

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -123,7 +123,7 @@ export const DateInput: React.VFC<DateInputProps> = ({
       {...inputProps}
     />
 
-    {status && (
+    {status != "default" && (
```
mcwinter07, line 126: small nit, but I feel like we should use a strict equal `!==` here since it should always match to default as a string. It also help avoid type coercion unless there's a chance we want to have that there for safety?

packages/date-picker/docs/DatePicker.stories.tsx
```diff
@@ -22,115 +17,28 @@ export default {
...
-        selectedRange={selectedDateRange}
-        modifiers={modifiers}
-        {...props}
-      />
-    </div>
-  )
-}
+DefaultStory.storyName = "Date Picker"
 
 const CalendarTemplate: Story = props => {
```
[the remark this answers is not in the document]
NatalieLoyzaga: I have updated this so that the stories now use this like before, nice find

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
NatalieLoyzaga: Yep, my bad

dougmacknz (commented):
Couple of small things

packages/date-picker/docs/DatePicker.stories.tsx
```diff
@@ -22,115 +19,28 @@ export default {
...
-        initialMonth={new Date(2022, 2)}
-        range
-        selectedRange={selectedDateRange}
-        modifiers={modifiers}
-        {...props}
-      />
-    </div>
-  )
-}
+DefaultStory.storyName = "Date Picker"
```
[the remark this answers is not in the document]
NatalieLoyzaga: Thought we were slowly moving away from this?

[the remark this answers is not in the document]
dougmacknz: Oh, right. I've lost track of what we're doing with this. I'll defer to you as the Storybook master :)

lloydstubber:
@NatalieLoyzaga I've done the removal of the description, placeholder and validation. Then also the returning undefined to the consumer when date is invalid. 

Just have a look at the [last commit](https://github.com/cultureamp/kaizen-design-system/pull/2695/commits/d1a30bd372c781d3ce67cff92292ddf20a6c4ef4) when you want to add validation back in.

@dougmacknz We've left out the clicking -> opens the calendar for the time being as it was a big of a large rejig in terms of UX and how our focusing works. So that will need to come later.

I'll leave this for merging on Monday when you're back Nat. Hell of a job on this one!

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+          automationId={`${id}-field-description`}
+          message={description}
+          reversed={reversed}
+        />
+      </div>
+    )}
+  </FieldGroup>
+)
+
+DateInput.displayName = "DatePickerInput"
```
dougmacknz, line 138:
```suggestion
DateInput.displayName = "DateInput"
```

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+  calendarId,
+  isOpen,
+  onKeyDown,
+  handleOnBlur,
+  ...inputProps
+}) => (
+  <FieldGroup
+    id={`${id}-field-group`}
+    automationId={`${id}-field-group`}
+    inline={inline}
```
dougmacknz, line 65: You can just hardcode this to `true` and remove the `inline` prop.

This prop was used to remove built in margins, which we've moved away from.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+    className={classnames(styles.withLabel, {
+      [styles.withDisabled]: disabled,
```
dougmacknz, line 67: I can't see either of these classes in the .scss file.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+      }
+      isDatePicker
+      onKeyDown={onKeyDown}
+      {...inputProps}
+    />
+
+    {description && (
+      <div
+        className={classnames(styles.message, {
+          [styles.disabled]: disabled,
```
dougmacknz, line 124: Surprised there's no `disabled` prop for `FieldMessage` so that we don't have to do this, but that's work for another time.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
+import React from "react"
+import { Icon } from "@kaizen/component-library/components/Icon"
```
dougmacknz, line 2:
```suggestion
import { Icon } from "@kaizen/component-library"
```

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+import classnames from "classnames"
+import {
+  FieldGroup,
+  FieldMessage,
+  Input,
+  InputProps,
+  Label,
+} from "@kaizen/draft-form/KaizenDraft/Form"
```
dougmacknz, line 10:
```suggestion
} from "@kaizen/draft-form"
```

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -33,10 +29,10 @@ export interface DatePickerProps {
   initialMonth?: Date
 
   // The date passed in from the consumer that renders in the input and calendar.
-  value?: Date
+  valueDate: Date | undefined
 
-  // Event passed from consumer to handle the date on change.
-  onChange: (day: Date) => void
+  // Setter for updating the selected date used by the consumer.
+  setValueDate: React.Dispatch<React.SetStateAction<Date | undefined>>
```
dougmacknz, line 35: 🤔 This seems a bit odd to limit this to a react set state action.

The usual way we do this was how it was before, curious what drove the change here.

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -145,48 +158,126 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
-              : "Choose date"
+          handleOnBlur={() =>
+            valueString !== undefined && handleFormatChange(valueString)
           }
+          onFocus={() =>
+            valueString !== undefined && handleFormatChange(valueString)
+          }
+          labelText={labelText}
+          description="mm/dd/yyyy"
+          placeholder="mm/dd/yyyy"
```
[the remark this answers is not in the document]
NatalieLoyzaga: Yeah agreed

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+  onButtonClick,
+  calendarId,
+  isOpen,
+  onKeyDown,
+  handleOnBlur,
+  ...inputProps
+}) => (
+  <FieldGroup
+    id={`${id}-field-group`}
+    automationId={`${id}-field-group`}
```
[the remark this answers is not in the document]
NatalieLoyzaga: What should replace this -> `ariaDescribedBy={`${id}-field-message`}` that is on Input or do I remove?

Assuming I keep the `id` for Input and the `htmlFor={`${id}-field-input`}` and `id` for Label?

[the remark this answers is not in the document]
dougmacknz: Yep keep those ids. Sorry should have clarified—ids are still sometimes needed for that reason, but the automationIds can all go away.

draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss
```diff
@@ -125,6 +125,11 @@ $input-with-icon-padding: $input-icon-size + $ca-grid * 0.75;
...
+    &.datePicker {
```
HeartSquared, line 129: I was able to replicate the styles required without the change to the primitive.

In `DateInput.scss` `.iconButton`, add:
```
  right: -0.25rem;
  top: -45%;
```
and you'll be able to remove all the changes in this component :)

HeartSquared:
You'll want to run `yarn install` and commit the change for `yarn.lock`.

draft-packages/form/KaizenDraft/Form/Primitives/Input/styles.scss
NatalieLoyzaga: Awesome! I was about to also look into this because we would need to update this later for RTL so a lot better its contained

NatalieLoyzaga:
Comment reminder for myself - This is a breaking change.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+export interface validationMessagesProps {
+  error?: string | React.ReactNode
+}
+
+export interface DateInputProps extends Omit<InputProps, OmittedInputProps> {
+  id: string
+  calendarId?: string
+  buttonRef?: React.RefObject<HTMLButtonElement>
+  labelText: string | React.ReactNode
+  inline?: boolean
```
dougmacknz, line 29: This inline prop can go away as well, and just hardcode to true where it was being used

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -145,48 +161,125 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
           }
+          onFocus={() =>
+            valueString !== undefined && handleFormatChange(valueString)
+          }
+          labelText={labelText}
+          description="Format: mm/dd/yyyy"
+          icon={dateStart}
+          onButtonClick={handleOpenClose}
+          onChange={handleTextChange}
+          calendarId={"calendar-dialog"}
```
dougmacknz, line 279: This will need to be unique for cases where people put more than one datepicker on a single screen.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+import styles from "./DateInput.scss"
+
+type OmittedInputProps =
+  | "startIconAdornment"
+  | "endIconAdornment"
+  | "inputType"
+  | "inputValue"
+
+export interface validationMessagesProps {
+  error?: string | React.ReactNode
```
dougmacknz, line 21: /nit There's other `string | React.ReactNode` typings in this file that can just be `ReactNode`

HeartSquared:
Changes have been made in other PRs :)

PR #2696, opened 2022-04-21
refactor(loading-placeholder): deprecate loadingPlaceholder

NatalieLoyzaga:
## Why

loading placeholder is deprecated and @kaizen/loading-skeleton is to be used instead.

## What

- Update loading placeholder stories to have component deprecated message and deprecated sub category.
- Update Readme.mdx to deprecated = true, and deprecated message
- Remove stories from chromatic
- Update LoadingSkeleton storyId .mdx


draft-packages/loading-placeholder/docs/LoadingPlaceholder.stories.tsx
```diff
@@ -177,35 +150,6 @@ export const DefaultMultipleCombinedBlockAndInline = () => (
 )
 DefaultMultipleCombinedBlockAndInline.storyName =
   "Default, Multiple, Combined block and inline"
-DefaultMultipleCombinedBlockAndInline.parameters = {
-  chromatic: { disable: false },
-}
-
-export const DefaultWithoutBottomMargin = () => (
```
christian314159, line 184: Any specific reason you deleted these stories? They might still be helpful for someone using this package.

christian314159 (dismissed):

draft-packages/loading-placeholder/docs/LoadingPlaceholder.stories.tsx
NatalieLoyzaga: I initially removed it to reduce stories but not necessary since I can remove the chromatic parameter. I have added them back in :)

site/docs/components/loading-skeleton.mdx
```diff
@@ -6,7 +6,7 @@ tags: ["Skeleton", "Stencil", "Loading Placeholders", "Placeholder UI"]
 githubLabels: ["component:loading-skeleton"]
 needToKnow:
 - Give people an idea of what’s about to come and what’s happening (it's currently loading) when a page full of content/data takes more than 300ms to load.
-demoStoryId: components-loading-skeleton-loading-skeleton--sticker-sheet-default
+demoStoryId: components-loading-skeleton-loading-headings--sticker-sheet-reversed
```
HeartSquared, line 9: Not sure what would be the most optimal example, but perhaps the site docs should embed this story instead? 
https://dev.cultureamp.design/KDS-47/deprecate-loading-placeholder/storybook/?path=/story/components-loading-skeleton--sticker-sheet-default

HeartSquared (approved):

site/docs/components/loading-skeleton.mdx
NatalieLoyzaga: It unfortunately isn't responsive and breaks - we would want to make a separate one instead maybe?

HeartSquared: Fair enough. A separate story later which showcases the different loading skeletons might be nice just for the site (without the sample loaded content).

draft-packages/loading-placeholder/docs/LoadingPlaceholder.stories.tsx
christian314159: 👍

christian314159 (approved):

PR #2701, opened 2022-04-29
docs(inputeditmodal): improve stories to showcase sample code

HeartSquared:
## Why

https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-415

The existing story caused confusion as it gave the impression that the sample Select we put into `children` was thought to be part of the component.

## What

Surface the code sample (as it was only showing `<No Display Name />` which isn't helpful) to better showcase that the content is not part of the component.


christian314159 (approved):

christian314159 (approved):

dalebaldwin (approved):
Thanks, that makes a huge difference

PR #2702, opened 2022-04-29
chore(deps): remove react-intersection-observer from design-tokens

HeartSquared:
## Why

Questioning whether we should continue upgrading `react-intersection-observer` (https://github.com/cultureamp/kaizen-design-system/pull/2697) or just remove it since it appears to never be used.

## What

Remove `react-intersection-observer`


mcwinter07 (approved):
Nice, doesn't appear to be used anywhere so should be good to 🪓

PR #2709, opened 2022-05-02
fix: update index.ts to be on root level of package

NatalieLoyzaga:
## Why

Update the index.ts to be root of the package.


HeartSquared:
The shared styles are actually throwing me off 😅 

Based off our rules, we should be looking at a structure more like this:
image
as `LoadingSkeleton` itself isn't a component, but rather just the package name (similar to `form`), and the `components` directory is for subcomponents only, which these aren't.

Following that train of thought, does that mean shared styles should be named after the package name `loading-skeleton.scss` since they aren't tied into a component? 🤔 What are your thoughts?

NatalieLoyzaga:
@HeartSquared I've updated the file structure :)

HeartSquared (approved):
LGTM 🎉

PR #2711, opened 2022-05-03
fix: button focus styling

NatalieLoyzaga:
## Why

Button focus styles broken as the pagination style border-radius was over riding it.

## What

- Remove the styling that is overriding all focus-visible border-radius styles. 
- Add the pagination border-radius styles under the %caCircleButton class styles.
- Remove %caCircleButtonFocus as it's no longer needed


HeartSquared (approved):
Very hard to tell what's happening, but it looks to fix the styles

PR #2712, opened 2022-05-03
chore(deps): update jest and ts-jest to v27

HeartSquared:
## Why

Resolve https://github.com/cultureamp/kaizen-design-system/pull/2472 and https://github.com/cultureamp/kaizen-design-system/pull/2698 (this wants v28, but `jest` and `ts-jest` must match so I've gone with v27 since `ts-jest` doesn't have v28 yet).

## What

- Update `jest` and `ts-jest` to v27
	- Set jest `testEnvironment` to `jsdom` (default was changed to `node` in version update)
- Fix broken tests (GenericNotification and useResizeObserver)


mcwinter07 (approved):
Looks good to me - cheers for adding the describe to the Notifications test 👍 Merge on 🦖

PR #2720, opened 2022-05-06
docs: update component health

HeartSquared:
## Why

Fortnightly update

## What

- Update LoadingSkeleton to be responsive (confirmed with @NatalieLoyzaga)
- Remove `Link vs button` links in Avatar/AvatarGroup as they aren't relevant (confirmed with @jmitsios)
- Remove tags `Tag` and `Label` from Badge
- Remove tags `Badge` and `Label` from Tag
- Fix the Modal page story link


dougmacknz:
Few more I can think of:
* Filter menu should probably be implemented false
* Rich text editor can go to implemented true (leaving everything else as false for now)
* Likert scale can go to accessible true (that was just waiting on the border update which is done now)

dougmacknz (approved):

PR #2724, opened 2022-05-12
WIP: Add links to RTE

dougmacknz:
Known issues/things to improve after the first pass:
* The toolbar button should be disabled until there's a text selection, OR
* The modal should include the ability to add/edit the link text
* The Popover styling isn't quite right because of the built in padding on our Popover component
* The URL isn't checked to be a valid URL (only 'is empty' validation)


dougmacknz:
Superseded by https://github.com/cultureamp/kaizen-design-system/pull/2795

PR #2725, opened 2022-05-12
fix: Enable noUncheckedIndexedAccess for design tokens and fix TS errors

vietanhtran16:
## Why

We had an incident on unified-home recently and that was caused by a piece of code accessing item in an array via index like this `items[0]` but not checking whether that can be undefined before using it. Because of that, we are enabling [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) for TS so that it warns us of this potential issue ([trello ticket](https://trello.com/c/lj3dwRpy/341-add-nouncheckedindexedaccess-true-to-tsconfigts-and-fix-ts-errors)). While we enabling that config, we got this error from `design-tokens` so we created this PR to address it. This TS config is helpful and recommended to avoid potential issue
image

Another way to fix it could be to exclude `src` from the published package but would still recommend to switch this option to true

## What

The changes that I made basically handle scenario where `leafKey` can be `undefined` because of the below so I added a check so that if it is `undefined` just throw an error. From my understanding, this piece of code is being ran at build time before publish so we will get quick feedback.
```
const leafKey = leafPath[leafPath.length - 1]
```

I tried to turn on this TS option for all packages but there are a few other errors there

### DIFF
--- a/packages/design-tokens/src/lib/makeCssVariableTheme.ts
+++ b/packages/design-tokens/src/lib/makeCssVariableTheme.ts
@@ -47,6 +47,9 @@ export function makeCSSVariableTheme<
         (child[segment] || (child[segment] = {})) as Record<string, unknown>,
       augmentedTheme as Record<string, unknown>
     )
+    if (!leafKey) {
+      throw new Error("leafKey is undefined")
+    }
     const cssVariablesOfToken = addExtraThemeEntries(
       leafPath,
       leafKey,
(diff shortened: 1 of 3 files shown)


dougmacknz (approved):
Thanks! I'll put a card on our board to investigate turning this config on for the rest of our packages

vietanhtran16:
> Thanks! I'll put a card on our board to investigate turning this config on for the rest of our packages

Awesome 👍 . Thanks for looking at it

vietanhtran16:
> Thanks! I'll put a card on our board to investigate turning this config on for the rest of our packages

Also I forgot to mention, because the tsconfig in each package is called `tsconfig.dist.json`, VS Code does not seem to pick up that config and show the error in the IDE but it does fail when compile. VS Code would be able to pick up when they are just `tsconfig.json` I think

PR #2727, opened 2022-05-12
refactor: update RTE api to consume objects with group keys

mcwinter07:
## Why
This is an alternative to drop the multi dimensional array in favour of object notation


packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -65,132 +65,106 @@ function liftOrIndentList(action: "lift" | "indent"): Command {
...
 
   if (schema.marks.bold) {
     const type = schema.marks.bold
-    const groupIndex = controlsSections.get("bold")
-    toolbarControls[groupIndex]["bold"] = {
-      isActive: markIsActive(editorState, type),
-      action: toggleMarkCommand(type),
-      label: "Bold",
-      icon: boldIcon,
+    toolbarControls[controlGroups["bold"]] = {
```
dougmacknz, line 83: Might be worth putting `controlGroups["bold"]` into a variable to make it a little easier to read

PR #2729, opened 2022-05-13
feat: Add validation to DatePicker

NatalieLoyzaga:
## Why

- Handle validation for whether the date is valid and not disabled. 
- Still allow the consumer to override the `status` and the `validationMessage`.

## What

- Add `onValidate` callback function as a prop of DatePicker. 
- `onValidate` provides an `validationResponse` object back to the consumer.
- Update stories to include an error state example on the sticker sheet
- Add a Validation story that provides an example of how to add additional validation on top the `isValid` and `isDisabled`.
- Add extra doc comments to props.
- Remove deprecated Input Props

**An Example of Validation with the validationResponse printed out**

image

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -47,7 +47,7 @@ describe("<DatePicker />", () => {
   })
 
   it("should pre-fill the input when an initial date is provided", async () => {
-    render(<DatePickerWrapper selectedDay={new Date(2022, 2, 1)} />)
+    render(<DatePickerWrapper selectedDay={new Date("2022-03-1")} />)
 
     expect(screen.getByDisplayValue("Mar 1, 2022")).toBeInTheDocument()
   })
@@ -57,10 +57,8 @@ describe("<DatePicker />", () => {
 
     const button = screen.getByRole("button")
 
-    // Make sure calendar popup is not in the DOM
     expect(screen.queryByRole("dialog")).not.toBeInTheDocument()
 
-    // Click button and test calendar popup is showing
     await act(async () => button.click())
     expect(screen.getByRole("dialog")).toBeVisible()
   })
@@ -70,10 +68,8 @@ describe("<DatePicker />", () => {
 
     const input = screen.getByRole("combobox")
 
-    // Make sure calendar popup is not in the DOM
     expect(screen.queryByRole("dialog")).not.toBeInTheDocument()
 
-    // Click button and test calendar popup is showing
     await act(async () => {
       input.focus()
       userEvent.keyboard("{arrowdown}")
@@ -82,13 +78,12 @@ describe("<DatePicker />", () => {
   })
 
   it("is able to select date and shows in input", async () => {
-    render(<DatePickerWrapper />)
+    render(<DatePickerWrapper initialMonth={new Date("2022-03-01")} />)
 
     const button = screen.getByRole("button")
 
     await act(async () => button.click())
 
-    // Focus on date and select
     const selectedDate = screen.getByRole("gridcell", {
       name: "Sun Mar 06 2022",
     })
@@ -101,13 +96,12 @@ describe("<DatePicker />", () => {
   })
 
   it("returns focus to the button once date has been selected", async () => {
-    render(<DatePickerWrapper />)
+    render(<DatePickerWrapper initialMonth={new Date("2022-03-01")} />)
 
     const button = screen.getByRole("button")
 
     await act(async () => button.click())
 
-    // Focus on date and select
     const selectedDate = screen.getByRole("gridcell", {
       name: "Sun Mar 06 2022",
     })
@@ -119,62 +113,65 @@ describe("<DatePicker />", () => {
     expect(button).toHaveFocus()
   })
 
-  it("displays the message when status is error", async () => {
-    render(
-      <DatePickerWrapper status="error" validationMessage="Invalid Date." />
-    )
-
-    describe("Validation", () => {
-      expect(screen.getByText("Invalid Date.")).toBeInTheDocument()
+  describe("Validation", () => {
+    describe("Custom Validation", () => {
+      it("displays the message when status is error", async () => {
+        render(
+          <DatePickerWrapper status="error" validationMessage="Invalid Date." />
+        )
+        expect(screen.getByText("Invalid Date.")).toBeInTheDocument()
+      })
     })
 
-    it("displays error message when selected day is invalid", async () => {
-      render(<DatePickerWrapper selectedDay={new Date("potato")} />)
+    describe("Inbuilt Validation", () => {
+      it("displays error message when selected day is invalid", async () => {
+        render(<DatePickerWrapper selectedDay={new Date("potato")} />)
 
-      expect(screen.getByText("Date is invalid")).toBeInTheDocument()
-    })
-
-    it("displays error message when selected day is disabled", async () => {
-      render(
-        <DatePickerWrapper
-          disabledBefore={new Date(2022, 4, 15)}
-          selectedDay={new Date(2022, 4, 5)}
-        />
-      )
-
-      expect(
-        screen.getByText("05/05/2022 is not available, try another date")
-      ).toBeInTheDocument()
-    })
-
-    it("displays error message when input date is invalid", async () => {
-      render(<DatePickerWrapper />)
-
-      const input = screen.getByRole("combobox")
-      userEvent.type(input, "05/05/2022Blah")
-
-      await act(async () => {
-        userEvent.tab()
+        expect(screen.getByText("Date is invalid")).toBeInTheDocument()
       })
 
-      expect(
-        screen.getByText("05/05/2022Blah is an invalid date")
-      ).toBeInTheDocument()
-    })
+      it("displays error message when selected day is disabled", async () => {
+        render(
+          <DatePickerWrapper
+            disabledBefore={new Date("2022-05-15")}
+            selectedDay={new Date("2022-05-05")}
+          />
+        )
 
-    it("displays error message when input date is disabled", async () => {
-      render(<DatePickerWrapper disabledBefore={new Date(2022, 4, 15)} />)
-
-      const input = screen.getByRole("combobox")
-      userEvent.type(input, "05/05/2022")
-
-      await act(async () => {
-        userEvent.tab()
+        expect(
+          screen.getByText("05/05/2022 is not available, try another date")
+        ).toBeInTheDocument()
       })
 
-      expect(
-        screen.getByText("05/05/2022 is not available, try another date")
-      ).toBeInTheDocument()
+      it("displays error message when input date is invalid", async () => {
+        render(<DatePickerWrapper />)
+
+        const input = screen.getByRole("combobox")
+        userEvent.type(input, "05/05/2022Blah")
+
+        await act(async () => {
+          userEvent.tab()
+        })
+
+        expect(
+          screen.getByText("05/05/2022Blah is an invalid date")
+        ).toBeInTheDocument()
+      })
+
+      it("displays error message when input date is disabled", async () => {
+        render(<DatePickerWrapper disabledBefore={new Date("2022-05-15")} />)
+
+        const input = screen.getByRole("combobox")
+        userEvent.type(input, "05/05/2022")
+
+        await act(async () => {
+          userEvent.tab()
+        })
+
+        expect(
+          screen.getByText("05/05/2022 is not available, try another date")
+        ).toBeInTheDocument()
+      })
     })
   })
 })
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -1,8 +1,10 @@
-import { act, fireEvent, render, screen } from "@testing-library/react"
-import React from "react"
+import { act, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import React, { useState } from "react"
+import { FieldMessageStatus } from "@kaizen/draft-form"
 import userEvent from "@testing-library/user-event"
-import { DatePicker } from "./DatePicker"
+import { DatePicker, ValidationResponse } from "./DatePicker"
 import "@testing-library/jest-dom"
+import { DatePickerProps } from "."
 
 const defaultProps = {
   id: "date-picker",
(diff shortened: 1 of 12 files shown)


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
 const defaultProps = {
   id: "date-picker",
   labelText: "Choose date",
   selectedDay: undefined,
   initialMonth: new Date(2022, 2),
   onDayChange: jest.fn(),
+  onValidate: jest.fn(),
+}
+
+const MockDatePicker = ({
```
BurntCaramel, line 18: Maybe `DataPickerHarness` instead of `MockDatePicker`? There’s no mocking here?

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -146,25 +177,67 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
     }
-    onDayChange(day)
-    setIsOpen(false)
+
+    if (date === undefined) {
+      onValidate({
+        ...baseResponse,
+        isEmpty: true,
+      })
+      return onDayChange(undefined)
```
BurntCaramel, line 200: I wonder if instead of having two callbacks, if you considered having a second argument to `onDayChange` containing the validation information?

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -190,6 +263,36 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
     }
   }
 
+  useEffect(() => {
```
BurntCaramel, line 266: By validating after the first render has been committed, the validation error won’t appear in the server rendered output. Not sure if that’s something that matters?

HeartSquared: The initial render validation is more of an edge case since the user has yet to interact with the datepicker. It is for the situation when the api has given an invalid or disabled date, so we communicate to the user that there may be something they need to fix, but nothing more than.

Unless you can think of an example where you'd want us to handle it differently?

packages/date-picker/src/DatePicker/DatePicker.tsx
HeartSquared: Yes we considered it, however when we were building it out we found instances where we would want to run the validation and not update the day (currently in the initial render, but there were previous changes too).

In saying that, we discussed a next iteration where we might handle the inbuilt validation without the consumer needing to create state for validation should they not have any custom validation themselves. If we find that both functions will always be called together, then we could consider merging them into one (since there isn't much usage of the component yet).

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
HeartSquared: Will review the naming. It's more common practice to just call these wrappers though I believe (at least, in most examples of tests I've seen).

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -190,6 +263,36 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
+        status: "error",
+        validationMessage: "Date is invalid",
+        isInvalid: true,
+        isDisabled: false,
+        isEmpty: false,
+        isValidDate: false,
+      })
+      return
+    }
+    if (selectedDay && isDisabledDate(selectedDay, disabledDays)) {
```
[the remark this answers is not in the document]
HeartSquared: Interesting question. This useEffect has been put in only for the initial render, and we have validation elsewhere for user interaction.

My question to you would be, can you give me an example of when you imagine the disabled days may change without user interaction? You're certainly not wrong, but depending on the situation, we may choose to handle things differently (eg. perhaps we should handle the initial render validation in a different way).

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
+  selectedDay,
+  ...restProps
+}: Partial<DatePickerProps>) => {
+  const [status, setStatus] = useState<FieldMessageStatus>("default")
+  const [validationMessage, setValidationMessage] = useState<string>()
+  const [selectedDate, setValueDate] = useState<Date | undefined>(selectedDay)
```
HeartSquared, line 24:
```suggestion
  status: propsStatus,
  validationMessage: propsValidationMessage,
  ...restProps
}: Partial<DatePickerProps>) => {
  const [status, setStatus] = useState<FieldMessageStatus>(propsStatus || "default")
  const [validationMessage, setValidationMessage] = useState<React.ReactNode>(propsValidationMessage)
  const [selectedDate, setValueDate] = useState<Date | undefined>(selectedDay)
```

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
+      onDayChange={setValueDate}
+      status={status}
+      validationMessage={validationMessage}
+      selectedDay={selectedDate}
+    />
+  )
 }
 
 describe("<DatePicker />", () => {
   it("renders DatePicker with an empty input value", async () => {
```
HeartSquared, line 45:
```suggestion
  it("should have an empty input value when a date is not provided", async () => {
```

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
-    render(<DatePicker {...defaultProps} />)
+    render(<MockDatePicker {...defaultProps} />)
 
     expect(screen.getByRole("combobox")).toHaveValue("")
   })
 
   it("renders DatePicker and displays inital date within input", async () => {
```
HeartSquared, line 51:
```suggestion
  it("should pre-fill the input when an initial date is provided", async () => {
```

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
-    render(<DatePicker {...defaultProps} selectedDay={new Date(2022, 2, 1)} />)
+    render(
+      <MockDatePicker {...defaultProps} selectedDay={new Date(2022, 2, 1)} />
+    )
 
     // Make sure date renders in the button
     expect(screen.getByDisplayValue("Mar 1, 2022")).toBeInTheDocument()
   })
 
   it("renders DatePicker and shows/hides calendar on button press", async () => {
```
HeartSquared, line 60:
```suggestion
  it("shows/hides calendar on button press", async () => {
```

And similar for other descriptions. The outer `describe` already tells us it's about the DatePicker.

Also unsure which button is being referred to in these descriptions (and in the test for that matter) since there are so many buttons in a datepicker. Worth revisiting these tests later to make them a little clearer.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1,33 +1,64 @@
...
     expect(screen.getByRole("combobox")).toHaveValue("")
   })
 
   it("renders DatePicker and displays inital date within input", async () => {
-    render(<DatePicker {...defaultProps} selectedDay={new Date(2022, 2, 1)} />)
+    render(
+      <MockDatePicker {...defaultProps} selectedDay={new Date(2022, 2, 1)} />
+    )
 
     // Make sure date renders in the button
```
HeartSquared, line 56:
```suggestion
```

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -90,4 +123,67 @@ describe("<DatePicker />", () => {
 
     expect(button).toHaveFocus()
   })
+
+  describe("Validation", () => {
+    it("displays error message when status is error", async () => {
```
HeartSquared, line 128: This infers there should be tests for the other statuses too.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -90,4 +123,67 @@ describe("<DatePicker />", () => {
...
+          {...defaultProps}
+          status="error"
+          validationMessage="Invalid Date."
+        />
+      )
+
+      expect(screen.getByText("Invalid Date.")).toBeInTheDocument()
+    })
+
+    it("displays error message when selected day is invalid", async () => {
```
HeartSquared, line 140: This (and below) can go in their own `describe` (it can live within the Validation one) since it's tests about inbuilt validation, whereas the one above is about custom validation.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -90,4 +123,67 @@ describe("<DatePicker />", () => {
...
+      render(<MockDatePicker selectedDay={new Date("potato")} />)
+
+      expect(screen.getByText("Date is invalid")).toBeInTheDocument()
+    })
+
+    it("displays error message when selected day is disabled", async () => {
+      render(
+        <MockDatePicker
+          disabledBefore={new Date()}
```
HeartSquared, line 149:
```suggestion
          disabledBefore={new Date(2022, 4, 15)}
```
Best to be explicit with set values.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -90,4 +123,67 @@ describe("<DatePicker />", () => {
...
+          selectedDay={new Date(2022, 4, 5)}
+        />
+      )
+
+      expect(
+        screen.getByText("05/05/2022 is not available, try another date")
```
HeartSquared, line 155: It still makes me uncomfortable that the date and string don't match. I think this needs revisiting later.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -90,4 +123,67 @@ describe("<DatePicker />", () => {
...
+        userEvent.tab()
+      })
+
+      expect(
+        screen.getByText("05/05/2022Blah is an invalid date")
+      ).toBeInTheDocument()
+    })
+
+    it("displays error message when input date is disabled", async () => {
+      render(<MockDatePicker disabledBefore={new Date()} />)
```
HeartSquared, line 175:
```suggestion
      render(<MockDatePicker disabledBefore={new Date(2022, 4, 15)} />)
```

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -33,13 +35,18 @@ export interface DatePickerProps
   /**
    * Accepts a DayOfWeek value to start the week on that day. By default,
    * it's set to Monday.
+   *  @default 1
```
HeartSquared, line 38: Does this show up in the docs? If not, I think it's fine to leave it out.
But if it's useful, I believe it should be `DayOfWeek.Mon`

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -80,11 +91,31 @@ export interface DatePickerProps
    * e.g. disabledDaysOfWeek={[DayOfWeek.Mon, DayOfWeek.Tue]}
    */
   disabledDaysOfWeek?: DayOfWeek[]
+  /**
+   * Callback when a date is attempted to be selected.
+   * ValidationResponse object will be returned.
+   * @returns {ValidationResponse}
```
HeartSquared, line 97: Technically wrong since `ValidationResponse` is the param provided to the consumer, and not the returned value of the callback (which is `void`).

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -80,11 +91,31 @@ export interface DatePickerProps
...
+   */
+  onValidate: (validationResponse: ValidationResponse) => void
+  /**
+   * Updates the styling of the validation FieldMessage.
+   */
+  status: FieldMessageStatus
+  /**
+   * A descriptive message for the 'status' states.
+   */
+  validationMessage: string | React.ReactNode
```
HeartSquared, line 107:
```suggestion
  status: FieldMessageStatus | undefined
  /**
   * A descriptive message for the 'status' states.
   */
  validationMessage: string | React.ReactNode | undefined
```

Since we're making them required fields (for now), we need to define that is it valid for them to be `undefined`.

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -146,25 +177,67 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
     }
+  }
 
-    onDayChange(undefined)
+  const handleInputChange = (
+    date: Date | undefined,
+    inputValue: string
+  ): void => {
+    handleDayChange(date, inputValue)
   }
```
HeartSquared, line 241:
```suggestion
  ): void => handleDayChange(date, inputValue)
```

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -230,13 +333,11 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
...
+            onDayChange={handleOnCalendarDayChange}
             onKeyDown={handleKeyDown}
           />
         </FocusOn>
       )}
     </div>
   )
 }
-
-DatePicker.displayName = "DatePicker"
```
HeartSquared, line 242: Where did this go? :')

packages/date-picker/src/DatePicker/components/DateInput/DateInput.spec.tsx
```diff
@@ -17,6 +17,7 @@ const defaultProps = {
   onBlur: () => jest.fn(),
   calendarId: "calendar-dialog",
   valueDate: undefined,
+  onValidation: jest.fn(),
```
HeartSquared, line 20: I don't think this is needed.

```suggestion
```

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -24,6 +24,11 @@ type OmittedInputProps =
   | "reversed"
   | "onBlur"
   | "value"
+  | "type"
+  | "ariaLabel"
+  | "ariaDescribedBy"
+  | "defaultInputValue"
+  | "automationId"
```
HeartSquared, line 31:
```suggestion
  | "automationId"
  | "status"
```

Looks like `InputProps` also has a status of its own and is used to style the input itself. TypeScript didn't pick up on the conflict because the values just happen to be the same as the validation ones, but it's safer to omit it here so if there's any change in the future to either we don't get an error.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -36,21 +41,42 @@ export interface DateInputProps extends Omit<InputProps, OmittedInputProps> {
...
+   * The callback for then onBlur is triggered on the input
+   */
+  onBlur: (date: Date | undefined, inputValue: string) => void
   disabledDays?: Modifier | Modifier[]
+  /**
+   * A descriptive message for `status` states
+   */
+  validationMessage?: string | React.ReactNode
+  /**
+   * Updates the styling of the validation FieldMessage.
```
HeartSquared, line 63:
```suggestion
   * Updates the styling of the validation FieldMessage
```

I don't remember if we have `.` or not, but best to be consistent.

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -36,21 +41,42 @@ export interface DateInputProps extends Omit<InputProps, OmittedInputProps> {
...
+   */
+  status?: FieldMessageStatus
 }
 
 const formatDateAsText = (
   date: Date,
   disabledDays: Modifier | Modifier[],
   onValidDate: (newFormattedDate: string) => void
```
HeartSquared, line 71:
```suggestion
  onFormat: (newFormattedDate: string) => void
```

Since it's no longer just for valid dates

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -150,7 +181,21 @@ export const DateInput: React.VFC<DateInputProps> = ({
         onFocus={handleFocus}
         {...inputProps}
       />
-
+      {status && (
+        <div
+          className={classnames(styles.message, {
+            [styles.disabled]: disabled,
+          })}
+        >
```
HeartSquared, line 189: Interesting. Is this an existing pattern we have? I can't imagine us showing a validation message if the field is disabled (but correct me if I'm wrong).

packages/date-picker/src/DatePicker/components/DateInput/DateInput.tsx
```diff
@@ -150,7 +181,21 @@ export const DateInput: React.VFC<DateInputProps> = ({
...
+          <FieldMessage
+            id={`${id}-field-message`}
+            automationId={`${id}-field-validation-message`}
```
HeartSquared, line 192:
```suggestion
```

Unless someone needs them for good reason, we shouldn't be adding ids (definitely not automation ones too since they're deprecated).

HeartSquared (approved):

HeartSquared (approved):

PR #2733, opened 2022-05-18
refactor: fix dart-sass division errors

HeartSquared:
## Why

There is a lot of noise when building the site.

## What

Reduce as much dart-sass warnings as possible.

We could not fix the `strip-unit` function because it's not doing a recommended calculation. Also could not ignore it, so it's still there.


mcwinter07 (dismissed):
Looks good to me. If its all green in chromatic we're good to go 🦖

PR #2735, opened 2022-05-18
ci: Deploy storybook build to the root of branch deploys

zioroboco:
## Why
We're about to break site deploys in this repo (as part of moving the site into a new repo), but we don't want branch previews of storybook builds to stop being available.

## What
- Write storybook assets to the root of branch deploys
  - This allows us to remove the site deploy from the design system repo
without breaking storybook deploys in dev.
- Update PR annotations for branch previews

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -7,11 +7,6 @@ extract_artifact() {
 }
 
 main() {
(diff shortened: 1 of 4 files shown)


zioroboco:
Closing in favour of #2736

PR #2755, opened 2022-05-25
ci(chromatic): update GH action to trigger Chromatic build on label add

HeartSquared:
## What

Update GH action to trigger Chromatic build when label `run chromatic` is added.

Also added the missing `yarn.lock` changes.

## Why

Provides a manual trigger in PRs for Chromatic so we don't need to wait for an approval to see if we broke anything.

### DIFF
--- a/.github/workflows/chromatic.yaml
+++ b/.github/workflows/chromatic.yaml
@@ -3,6 +3,8 @@
 name: "chromatic"
 
 on:
+  pull_request:
+    types: [labeled]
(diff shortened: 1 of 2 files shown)


HeartSquared:
image

No idea why this is getting stuck, but it seems to get stuck even when I don't make any changes to the workflow. Going to just ignore it.

mcwinter07 (approved):
noice 🦖 merge in good health!

PR #2761, opened 2022-05-27
feat: Upgrade react-day-picker to v8

NatalieLoyzaga:
## Why

- Upgrade react-day-picker to the latest version v8.

## What

**Updated prop names:**
- `firstDayOfWeek` > `weekStartsOn`
- `initialMonth` > `defaultMonth`

**Updated types:**

- `Modifiers` > `Matcher`
- `RangeModifier` > `DateRange`

**Update styles:**
- New defaultClassNames- `DayPicker-Day-` >  `rdp-`


packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -12,14 +12,12 @@ $cell-border-radius: 3px;
   background-color: $color-white;
   z-index: 1010;
 
-  .wrapper {
-    padding-bottom: 0;
+  .container {
```
HeartSquared, line 15: Perhaps we should refactor this file later. The nesting isn't necessary since CSS modules ensures the classes are unique and don't clash with global ones.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.spec.tsx
```diff
@@ -1,14 +1,16 @@
-import { render, screen, cleanup } from "@testing-library/react"
+import { render, screen, act } from "@testing-library/react"
```
HeartSquared, line 1:
```suggestion
import { render, screen } from "@testing-library/react"
```

packages/date-picker/src/DatePicker/components/Calendar/Calendar.spec.tsx
```diff
@@ -1,14 +1,16 @@
...
 import React from "react"
+import userEvent from "@testing-library/user-event"
```
HeartSquared, line 3:
```suggestion
```

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -25,76 +28,68 @@ export type CalendarProps = {
...
-    day: Date,
-    modifiers: DayModifiers,
-    e: React.MouseEvent<HTMLDivElement, MouseEvent>
-  ) => void
-  range?: boolean
-  selectedRange?: RangeModifier
-  modifiers?: RangeModifier
-  onKeyDown?: (e: React.KeyboardEvent) => void
+  defaultMonth?: Date
+  weekStartsOn: 0 | 1 | 2 | 3 | 4 | 5 | 6 | undefined
```
HeartSquared, line 32: Ah, the culprit. We should have this accept the enum and do the type guard. Then the DatePicker and DateRangePicker can both just happily use the enum.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -25,76 +28,68 @@ export type CalendarProps = {
...
+  disabledDays?: Matcher[]
+  onDayChange: DayClickEventHandler
+  selectedRange?: DateRange
+  mode: DaySelectionMode
```
HeartSquared, line 36:
```suggestion
  mode: "single" | "range"
```

Since we don't want the whole set, we should just select the valid subset options.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -106,23 +101,43 @@ export const Calendar: React.VFC<CalendarProps> = ({
...
+            disabled={disabledDays}
+            onDayClick={onDayChange}
+            classNames={defaultCalendarClasses}
+            components={{
+              IconRight,
+              IconLeft,
+            }}
+          />
+        )}
+        {mode === "range" && (
```
HeartSquared, line 119: With restricting the possibilities of `mode`, we can remove the duplication again.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -106,23 +101,43 @@ export const Calendar: React.VFC<CalendarProps> = ({
...
+            components={{
+              IconRight,
+              IconLeft,
+            }}
+            modifiers={
+              {
+                [calendarStyles.from]: modifiers?.from,
+                [calendarStyles.to]: modifiers?.to,
+              } as DayModifiers
+            }
```
HeartSquared, line 138: I was going to suggest removing `as DayModifiers`, but then realised why you had it in there.

I question whether this is still required, or if there's another way to do this (`modifiers` should not be just a date range).
Either they go into DateRangePicker, or we find a different way of adding these styles.

I suggest we comment this out for the moment until we look at DateRangePicker again and properly fix it.

packages/date-picker/src/DatePicker/components/Calendar/CalendarClasses.ts
```diff
@@ -1,30 +1,32 @@
+/* eslint-disable camelcase */
 import calendarStyles from "./Calendar.scss"
 
 export const defaultCalendarClasses = {
```
HeartSquared, line 4:
```suggestion
import { ClassNames } from "react-day-picker"
import calendarStyles from "./Calendar.scss"

export const defaultCalendarClasses: ClassNames = {
```

Which actually surfaces that `interactionDisabled` is an invalid key

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -1,11 +1,8 @@
 import React, { RefObject, useEffect, useRef, useState } from "react"
 import dateStart from "@kaizen/component-library/icons/date-start.icon.svg"
-import "react-day-picker/lib/style.css"
+import "react-day-picker/dist/style.css"
```
HeartSquared, line 3: This should belong in `Calendar`.

Looking at the styles though, if we are overriding certain default styles with our own, we might as well just completely style it ourselves and not import this (we can look at this outside of this PR).

packages/date-picker/src/utils/calculateDisabledDays.spec.ts
```diff
@@ -1,53 +1,119 @@
...
+
+  it("calculates only an array of dates", () => {
+    const disabledDays = calculateDisabledDays({
+      disabledDates: [DISABLED_DATE__1, DISABLED_DATE__2],
+    })
+
+    const expectedResult = [DISABLED_DATE__1, DISABLED_DATE__2]
+
+    expect(disabledDays).toEqual(expectedResult)
+  })
```
HeartSquared, line 60: This appears to be the same test as L21-27

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -51,50 +46,52 @@ export const Calendar: React.VFC<CalendarProps> = ({
...
+      return defaultMonth
     }
   }
 
+  const IconRight = (props: StyledComponent) => (
+    <Icon icon={arrowRight} role="presentation" {...props} />
+  )
+  const IconLeft = (props: StyledComponent) => (
+    <Icon icon={arrowLeft} role="presentation" {...props} />
+  )
```
HeartSquared, line 93: You're not wrong, but the props provided by `StyledComponent` (className, style, children) actually aren't compatible with `Icon`, so we shouldn't be trying to spread them in.

```suggestion
  const IconRight: React.VFC = () => <Icon icon={arrowRight} role="presentation" />
  const IconLeft: React.VFC = () => <Icon icon={arrowLeft} role="presentation" />
```

Alternatively, we could destructure the props and pass them in in a compatible way (eg. `classNameOverride={className}`), but since we aren't even adding extra styles to these icons (unless you did want to?), it doesn't really seem necessary.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -51,50 +46,52 @@ export const Calendar: React.VFC<CalendarProps> = ({
...
+  const getdefaultMonth = () => {
     if (selectedRange?.from) {
       return selectedRange.from
     } else if (value) {
       return value
     } else {
-      return initialMonth
+      return defaultMonth
     }
   }
```
HeartSquared, line 86:
```suggestion
const getdefaultMonth = () => selectedRange?.from || value || defaultMonth
```

packages/date-picker/src/DatePicker/components/Calendar/CalendarClasses.ts
```diff
@@ -1,30 +1,27 @@
...
-  todayButton: "DayPicker-TodayButton",
-  today: `${calendarStyles.today} DayPicker-Day--today`,
-  selected: `${calendarStyles.selected} DayPicker-Day--selected`,
-  disabled: `${calendarStyles.disabled} DayPicker-Day--disabled`,
-  outside: `${calendarStyles.outside} DayPicker-Day--outside`,
-  from: `${calendarStyles.from} DayPicker-Day--from`,
-  to: `${calendarStyles.to} DayPicker-Day--to`,
-  range: `${calendarStyles.range} DayPicker-Day--range`,
+export const defaultCalendarClasses: ClassNames = {
+  root: `${calendarStyles.container}`,
```
HeartSquared, line 5:
```suggestion
  root: calendarStyles.container,
```

This can be applied to all of them :)

packages/date-picker/src/DatePicker/components/DateInput/DateInput.spec.tsx
```diff
@@ -12,9 +12,9 @@ const defaultProps = {
   description: "Description text",
   icon: dateStart,
   isOpen: false,
-  onButtonClick: () => jest.fn(),
-  onKeyDown: () => jest.fn(),
-  onBlur: () => jest.fn(),
+  onButtonClick: jest.fn(),
+  onKeyDown: jest.fn(),
+  onBlur: jest.fn(),
```
HeartSquared, line 17:
```suggestion
  onButtonClick: jest.fn<void, []>(),
  onKeyDown: jest.fn<void, [React.KeyboardEvent<HTMLInputElement>]>(),
  onBlur: jest.fn<void, [Date | undefined, string]>(),
```

HeartSquared (approved):

HeartSquared (approved):

PR #2766, opened 2022-06-01
chore(stylelint): add stylelint to lint:fix and run in ci

HeartSquared:
## What

- Add stylelint to `lint:fix`
	- I put `stylelint` at the front because we want `prettier` to run after it
- Fix existing linting errors
- Update GH actions to run linting as a single step (eslint, prettier, stylelint)

## Why

Turns out we were never enforcing stylelint in the CI, nor applying its autofixes when calling `lint:fix`


packages/split-button/src/SplitButton/components/BaseButton/BaseButton.scss
```diff
@@ -34,11 +34,6 @@
     &#{$polyfill--focus-visible},
     &#{$story-className__focus} {
       z-index: 3;
-    }
-
-    &:focus-visible,
-    &#{$polyfill--focus-visible},
-    &#{$story-className__focus} {
```
dougmacknz, line 41: This was definitely meant to happen? 🤨

dougmacknz (approved):

dougmacknz (approved):

PR #2769, opened 2022-06-02
feat: Add highlight mood (purple-100) for BrandMoment

kemblesong:
## Why

We are introducing a brand-moment for Develop that uses the purple-100 background colour. This is known as the 'highlight' mood in the UI kit.
https://www.figma.com/file/GDJ3KsznTE7Z7GwyJzANcB/%F0%9F%9A%A7-Develop-2021-Q3?node-id=6088%3A158270

## What

In line with the other moods, added as an option to the mood prop

### DIFF
--- a/packages/brand-moment/src/BrandMoment.scss
+++ b/packages/brand-moment/src/BrandMoment.scss
@@ -26,6 +26,10 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
   &.negative {
     background: $color-red-100;
   }
+
+  &.highlight {
+    background: $color-purple-100;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


dougmacknz:
Happy with the addition here but looking at that design, that is going beyond what was envisioned when the BrandMoment component when it was built. It was built on the expectation that it will always be full screen and the only thing on the screen. We should catch up about whether we should adjust the component to work for this case, or whether this should just be a custom build in your repo.

kemblesong:
Fair, but do you not agree that the case in point is full screen and the only thing on the screen?

kemblesong:
Going with a custom page layout, as it's not _exactly_ a brand moment

PR #2770, opened 2022-06-02
Ak/update titleblock subtitle prop type

arikark:
## Why
Allow the use of a component, such as a link, in the subtitle as in the below screenshot:
image

## What
Change the subtitle prop type in TitleBlockZen to ReactNode from string .


draft-packages/title-block-zen/docs/TitleBlockZen.stories.tsx
```diff
@@ -268,6 +268,11 @@ export const AdminVariant = () => (
       title="Page title"
       variant="admin"
       sectionTitle="Default questions"
+      subtitle={
```
HeartSquared, line 271: Normally I'd say it should have its own story, but these stories aren't in a good enough state to easily do that anyway (and the component is legacy) so I'll let it slide.

HeartSquared (approved):

PR #2771, opened 2022-06-03
style(title block zen): set max-width of subtitle to max-content

arikark:
…to stop wrapping of long subtitles

## Why
Prevent long subtitles from wrapping. See slack [conversation](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1654212985638349?thread_ts=1654143872.284079&cid=C0189KBPM4Y).

## What
Set max-width of the subtitle to max-content


HeartSquared:
As discussed with @jmitsios, we are opting not to make this change as it could open us up for other UI problems. The Title Block needs a design review in general, and this issue has been noted by the designers.

PR #2775, opened 2022-06-07
fix: Catch the video element error

sugendran:
The try..catch won't catch async errors unless the function is actually async.

## Why
We're getting a lot of errors in the logs when media is disabled

### DIFF
--- a/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
@@ -99,9 +99,7 @@ export const VideoPlayer = ({
     if (prefersReducedMotion) {
       videoElement.pause()
     } else if (autoplay && !prefersReducedMotion) {
-      try {
-        videoElement.play()
-      } catch (e) {
+      videoElement.play().catch(e => {
         /*
          * An DOMException _may_ be raised by some browsers if we
          * programatically interact with the video before the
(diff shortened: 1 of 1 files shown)


draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -99,17 +99,15 @@ export const VideoPlayer = ({
     if (prefersReducedMotion) {
       videoElement.pause()
     } else if (autoplay && !prefersReducedMotion) {
-      try {
-        videoElement.play()
-      } catch (e) {
+      videoElement.play().catch(e => {
```
jakubriedl, line 102: [Some older browsers](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#return_value) don't return promise from `play()`. I know we don't support them so just FYI

sugendran: People using those browsers are going to have way more problems than the empty state not rendering.

jakubriedl: I know 😉 , I just had ptsd flash from when I was trying to make autoplay work cross browser & iframe & embeds & chinese android forks web views. It took me several weeks to iron out all issues.

dougmacknz (dismissed):

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
sugendran: Careful.. you're going to trigger my ActiveX|Quicktime|Flash PTSD 😜

dougmacknz (approved):

PR #2776, opened 2022-06-07
fix(rte): rte hover state causing clipping

lokesh-everest:
## Why
Rte editor state causing clipping while hovering [JIRA](https://cultureamp.atlassian.net/browse/COMP-156)

BEFORE:
Screenshot 2022-06-07 at 12 06 14 PM

AFTER:
Screenshot 2022-06-07 at 12 04 00 PM

## What
Adding border radius and removing border radius when toolbar is present

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
@@ -51,16 +51,16 @@
   }
 }
 
+.editor.hasToolbar > :global(.ProseMirror) {
+  border-top-left-radius: 0px;
+  border-top-right-radius: 0px;
+}
+
 .editorWrapper {
   position: relative;
   border: $border-solid-border-width $border-solid-border-style $color-gray-500;
   background: $color-white;
   border-radius: $border-solid-border-radius;
-
-  .activeToolbar > :global(.ProseMirror) {
-    border-top-left-radius: 0px;
-    border-top-right-radius: 0px;
-  }
 }
 
 @for $i from 1 through 20 {
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
@@ -5,7 +5,7 @@
 @import "@kaizen/design-tokens/sass/typography";
 
 .editor > :global(.ProseMirror) {
-  border-radius: 0 0 $border-solid-border-radius $border-solid-border-radius;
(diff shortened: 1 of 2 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
```diff
@@ -56,6 +56,14 @@
   border: $border-solid-border-width $border-solid-border-style $color-gray-500;
   background: $color-white;
   border-radius: $border-solid-border-radius;
+
+  div + .editor :global(.ProseMirror) {
```
dougmacknz, line 60: We generally try to avoid targetting element types like this, because if we for some reason decide to change `div` to a `span`, or something like that, this breaks silently.

What I'd do instead is create a class based on whether there are any controls, and as a result if there will be a toolbar.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
```diff
@@ -56,6 +56,14 @@
...
+    &:hover,
+    &:focus {
```
dougmacknz, line 62: I think this can be applied regardless of hover or focus, like it is being done above?

lokesh-everest: Yes @dougmacknz

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.scss
lokesh-everest: Changed it. Let me know what you think

dougmacknz:
@lokesh-everest I made a couple of adjustments:
* Used the `classnames` object for the classname conditional which is now checking for nullish and length of `controls`
* Renamed `activeToolbar` to `hasToolbar` (somewhat of a nit pick)
* Removed the nesting on the `hasToolbar` styling, because the other border-radius is being set without the `editorWrapper` nesting.

LMK if that all looks good to you and we can merge.

lokesh-everest:
> @lokesh-everest I made a couple of adjustments:
> 
> * Used the `classnames` object for the classname conditional which is now checking for nullish and length of `controls`
> * Renamed `activeToolbar` to `hasToolbar` (somewhat of a nit pick)
> * Removed the nesting on the `hasToolbar` styling, because the other border-radius is being set without the `editorWrapper` nesting.
> 
> LMK if that all looks good to you and we can merge.

Yes @dougmacknz it makes sense. Thanks for taking time and explaining it to me

dougmacknz (approved):

PR #2777, opened 2022-06-07
fix: Move to camelCase and rename bold and italic to strong and em

dougmacknz:
## Why
https://cultureamp.atlassian.net/browse/KDS-498

## What
* Renames bold and italic in **schema** to `strong` and `em`
  * Note: prop names have stayed the same, so I've reworked the `createSchemaFromControls` to remove the tight coupling between prop names and schema names. It's also more consistent with how we've done things in the controlmap and keymap.
* Changes snake case names to camel case (both in props and in schema)
* I've also snuck in a little change to Storybook
  * I've gotten rid of the story without any controls, and made the controls one the default.


packages/rich-text-editor/src/RichTextEditor/schema.ts
```diff
@@ -1,100 +1,57 @@
...
+
+  if (allControlNames.includes("bulletList")) {
+    newNodes["bulletList"] = coreNodes.bulletList
+    newNodes["listItem"] = coreNodes.listItem
+  }
+
+  if (allControlNames.includes("orderedList")) {
+    newNodes["orderedList"] = coreNodes.orderedList
+    newNodes["listItem"] = coreNodes.listItem
+  }
```
mcwinter07, line 51: Nice, this is a lot easier to read 👍

mcwinter07 (approved):
Looks grand! I'm a big fan of the refactor in the schema.ts, much more consumable 🦖 Merge in good health

packages/rich-text-editor/src/RichTextEditor/schema.ts
mcwinter07: Meaning the `allControlNames` reducer

PR #2779, opened 2022-06-07
feat: update indentation logic to include disabled states

mcwinter07:
## Why
This updates the behavior of list indentation to allow for disabled states to be passed into buttons

## Still to do
update keyboard focus handler to exclude disabled buttons

## What
- adds disabled checks for list buttons and updates expected behaviors outlined [here](https://cultureamp.atlassian.net/browse/KDS-521)


dougmacknz (approved):

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -33,6 +33,10 @@
     right: calc(-1 * #{$focus-ring-offset});
     bottom: calc(-1 * #{$focus-ring-offset});
   }
+
+  &:disabled {
+    opacity: 0.4;
```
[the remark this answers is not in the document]
mcwinter07: Will update that one 👍

packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -37,36 +38,67 @@ function toggleMarkCommand(mark: MarkType): Command {
...
+
+    return sinkListItem(listItemNode)(state, dispatch)
   }
 }
 
+// If there is a valid list item its indent can be decrease or 'lifted'
+function liftListIsDisabled(state: EditorState): boolean {
+  const { $from } = state.selection
+  const listItemNode = $from.node($from.depth - 1)?.type
+  const isValidListItem = listItemNode?.name === "list_item" || false
```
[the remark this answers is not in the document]
mcwinter07: True, will have to pull those changes across from master :)

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
```diff
@@ -33,76 +33,120 @@
     right: calc(-1 * #{$focus-ring-offset});
     bottom: calc(-1 * #{$focus-ring-offset});
   }
+
+  &[aria-disabled="true"] {
```
dougmacknz, line 37: We should probably use a class here. Even though I don't think this property will change, there's an explicit relationship between classes and styling which we don't get from this.

Usually this is an issue when people use an element type as a selector, but the principle stands that if someone changes this property, they wouldn't expect that they would also be changing something related to styling.

packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -37,36 +38,67 @@ function toggleMarkCommand(mark: MarkType): Command {
...
+// If there is a valid list item its indent can be decrease or 'lifted'
+function liftListIsDisabled(state: EditorState): boolean {
+  const { $from } = state.selection
+  const listItemNode = $from.node($from.depth - 1)?.type
+  const isValidListItem = listItemNode?.name === "listItem" || false
+
+  return !isValidListItem
+}
+
+// If the is a valid list item and it is not the first in a list it can be indented
```
dougmacknz, line 84: `If there*` meant to be here I think?

dougmacknz (commented):
Strangely, the reduced opacity is working on the ToggleIconButton sticker sheet, but not the RTE itself
image

packages/rich-text-editor/src/RichTextEditor/controlmap.ts
mcwinter07: or is it thee? (yes it is meant to be `there `😭 )

packages/rich-text-editor/src/RichTextEditor/components/ToggleIconButton/ToggleIconButton.scss
mcwinter07: that's fair, happy to change that over :)

packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -176,13 +208,15 @@ export function buildControlMap(
 
     toolbarControls[groupIndex].push(
       {
-        action: liftOrIndentList("lift"),
+        action: liftList(),
```
dougmacknz, line 211: I should have picked this up earlier, but putting a function here with parentheses will invoke the function right?

mcwinter07: I believe what it returns is a `Command` that is used to invoke `liftListItem` and shouldn't actually execute

mcwinter07: although it might be better to do something like this and update the indentList function to:
```
const toggleIndentList: Command = (
  state: EditorState,
  dispatch: ((tr: Transaction) => void) | undefined
) => {
  const { $from } = state.selection
  const listItemNode = $from.node($from.depth - 1)?.type

  return sinkListItem(listItemNode)(state, dispatch)
}
```

dougmacknz: Oh so it's actually invoking this function here on purpose, in order to get the function that it returns. Mind blown. I guess the 2 without parameters (`liftList` and `indentList`) could just directly be the functions they are returning in that case, rather than a function that returns a function?

dougmacknz (approved):

PR #2781, opened 2022-06-07
feat: open calendar on click and focus on input

NatalieLoyzaga:
## Why

Fixes [card](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-480)
- Add onClick to input to open the calendar dialog.
- Remain focus within the input if the input is clicked.

## What

UI:
- Returns focus to input when a date is selected through the calendar and:
  -  it was opened via input click
  -  it was opened via input keydown down arrow

- Returns focus to the calendar button when a date is selected through the calendar and:
  - it was opened via calendar button click
  - it was opened via calendar button keydown enter
- Closes the calendar when user is within the input and a date is selected and keydown on Tab.

Technical:
- Add `onClick` to input to open calendar dialog
- Created a check for when the focus should go to the calendar once opened or remain within the input.
- Update `handleInputBlur` and `handleReturnFocus` to conditionally call depending on the `lastTrigger` element.
- Move `FocusOn` wrapper to be around both the `DateInput` and the `Calendar` to allow tabbing across.
- Add new `DatePicker` tests for each scenario of opening the calendar Dialog. 

**NOT IN**

- Allow user to press {enter} when focused within Input to confirm input 
- The calendar to change month as we type

### DIFF
--- a/packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
+++ b/packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
@@ -59,12 +59,11 @@ export const Calendar: React.VFC<CalendarProps> = ({
   modifiers,
   mode,
   // shouldFocusOnCalendar = false,
-  onMount
+  onMount,
 }) => {
   const calendarRef = useRef<CalendarElement>(null)
 
   useEffect(() => {
-    console.log("calendarRef",calendarRef)
     if (calendarRef.current) onMount && onMount(calendarRef.current)
   }, [calendarRef])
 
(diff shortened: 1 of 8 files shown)


packages/date-picker/src/DatePicker/components/Calendar/Calendar.tsx
```diff
@@ -138,5 +127,3 @@ export const Calendar: React.VFC<CalendarProps> = ({
     </div>
   )
 }
-
-Calendar.displayName = "Calendar"
```
NatalieLoyzaga, line 142: Assuming mistake - Will add this back in

PR #2783, opened 2022-06-08
fix: Fixed issue of modal background going full opacity in Chrome 102 by changing animation speed [KDS-523]

christian314159:
## Why
- See issue here: https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-523
- Modal backdrop would snap to full opacity at the end of the fade-in animation; a window resize would fix it

image

## What
- Changing the Modal package all the way back to v6.x didn't help, showed same problem, so likely no code regression
- Investigation showed that it is somehow correlated with the latest Blink engine 102.x; MS Edge 95 was fine, once updated to 102 demonstrated the bug; Firefox & Safari also did not exhibit the issue
- After some exploration it showed that changing the animation duration from `rapid` (200ms) to just something slightly off (201ms) made it work; fwiw 300ms also didn't work 🤷

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.scss
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.scss
@@ -59,11 +59,12 @@
   transition-duration: $ca-duration-fast;
 
   .backdropLayer {
-    @include ca-animation-fade(
-      $duration: $ca-duration-rapid,
-      $from: 0,
-      $to: 0.5
-    );
+    /*
+      There's a weird bug in Chrome/Blink 102.x that animates the opacity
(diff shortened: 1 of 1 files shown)


draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.scss
```diff
@@ -78,7 +79,11 @@
   transition-duration: $ca-duration-rapid;
 
   .backdropLayer {
-    @include ca-animation-fade($from: 0.5, $to: 0);
+    @include ca-animation-fade(
```
christian314159, line 82: I'm reverting a change here that I accidentally did in my first PR.

mcwinter07 (approved):

PR #2790, opened 2022-06-10
feat: Allow aria-labelledby as an alternative to labelText on RTE

lavanya-everest:
## Why
Using a discriminated union ideally, to make sure consumers are using one of the two. For more details [https://cultureamp.atlassian.net/browse/COMP-161](url)

## What

Added new property `labelledBy` to use it as `aria-labelledby` for Rich Text Editor.

In RTE, label will be created only if `labelledBy` is not passed.

### DIFF
--- a/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
+++ b/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
@@ -20,25 +20,28 @@ export default {
   },
 }
 
-export const Default: Story<RichTextEditorProps> = args => {
+export const Default = args => {
   const [rteData, setRTEData] = useState<EditorContentArray>([])
   return (
     <>
       <RichTextEditor
         value={rteData}
         onChange={data => setRTEData(data)}
-        labelText={"Label"}
-        rows={3}
-        controls={[
-          { name: "bold", group: "inline" },
-          { name: "italic", group: "inline" },
-          { name: "underline", group: "inline" },
-          { name: "orderedList", group: "list" },
-          { name: "bulletList", group: "list" },
-        ]}
+        {...args}
       />
     </>
   )
 }
 
 Default.storyName = "Default (Kaizen Demo)"
+Default.args = {
+  labelText: "Label",
+  rows: 3,
+  controls: [
+    { name: "bold", group: "inline" },
+    { name: "italic", group: "inline" },
+    { name: "underline", group: "inline" },
+    { name: "orderedList", group: "list" },
+    { name: "bulletList", group: "list" },
+  ],
+}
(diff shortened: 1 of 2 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -34,6 +34,7 @@ export interface RichTextEditorProps
   onChange: (content: EditorContentArray) => void
   value: EditorContentArray
   labelText: ReactNode
+  labelledBy: string
```
dougmacknz, line 37: @lavanya-everest I was hoping for a [discriminated union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) here, to ensure that engineers must pass either `labelText` or `labelledBy`.

As this currently stands, engineers will have to pass **both**, which is not quite what we're after. The easy solution is to make both of them optional, but given accessibility implications, I would like to enforce that one of 2 needs to be provided, and TypeScript is the best way to do that.

packages/rich-text-editor/docs/RichTextEditor.stories.tsx
```diff
@@ -14,28 +20,49 @@ export default {
...
-  rows: 3,
-  controls: [
-    { name: "bold", group: "inline" },
-    { name: "italic", group: "inline" },
-    { name: "underline", group: "inline" },
-    { name: "orderedList", group: "list" },
-    { name: "bulletList", group: "list" },
-  ],
+
+export const WithLabelledBy: Story<RichTextEditorProps> = args => {
```
dougmacknz, line 46: I think we can live without this story. We're trying to move away from having lots of stories, because our Chromatic budget isn't very large.

packages/rich-text-editor/docs/RichTextEditor.stories.tsx
```diff
@@ -14,28 +20,49 @@ export default {
...
+          { name: "orderedList", group: "list" },
+          { name: "bulletList", group: "list" },
+        ]}
       />
     </>
   )
 }
 
 Default.storyName = "Default (Kaizen Demo)"
-Default.args = {
```
dougmacknz, line 31: Please bring these back in as well, they are useful for allowing people to play around with props in storybook.

dougmacknz (commented):
Nice, just a couple minor things and this will be good to go!

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -42,6 +41,17 @@ export interface RichTextEditorProps
   rows?: EditorRows
 }
 
+interface RTEWithLabelText extends BaseRichTextEditorProps {
+  labelText: ReactNode
+  labelledBy?: never
```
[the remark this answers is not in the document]
lavanya-everest: Renamed `labelledBy` to `ariaLabelledBy`

dougmacknz (approved):
Awesome, thank you! 🚢

PR #2799, opened 2022-06-20
refactor: remove zen theme from kds

mcwinter07:
BREAKING CHANGE:
zenTheme will no longer be exported themes and all config on zen design-tokens will
be removed

## Why
Zen theming is no longer supported in kaizen and we are removing all references and exports for it
in our repository.

## What
- Zen theming is no longer supported in kaizen
- All references and config to zen theme removed from design tokens


dougmacknz (approved):
Make it be gone!

PR #2800, opened 2022-06-20
chore: Remove Heading and Paragraph components from component-library

dougmacknz:
**BREAKING CHANGE:** Heading and Paragraph have been moved to the `@kaizen/typography` package. See https://github.com/cultureamp/node-packages/tree/master/packages/jscodeshift for a codemod to bulk replace the imports in your repo.

## What
Removes the `Heading` and `Paragraph` components from the `component-library` package. 

## Why
These 2 components have been moved to a new `@kaizen/typography` package

## Notes
I haven't removed the Elm versions of these components, which means I've had to leave the CSS files in. Depending on which goes in first, either this PR or https://github.com/cultureamp/kaizen-design-system/pull/2633 will need to remove the CSS files.


site/src/components/Footer.tsx
```diff
@@ -1,4 +1,5 @@
-import { Icon, Paragraph } from "@kaizen/component-library"
+import { Icon } from "@kaizen/component-library"
+import { Paragraph } from "@kaizen/typography"
 import classnames from "classnames"
```
mcwinter07, line 3: I think this might be made redundant / cause a conflict if Cass merges in her site removal, but other than that all looks good to me 👍

mcwinter07 (approved):

mcwinter07 (approved):

PR #2802, opened 2022-06-20
chore: remove gatsby site

HeartSquared:
## What

Remove the gatsby site and associated code.

## Why

[KDS-535](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-535)
Site has been moved to `kaizen-site`.

## Changes

- Remove `site` directory
- Remove gatsby related packages
- Add carat to `svgo-load` (we had version locked it because of gatsby)
- Remove site references from gitignore and prettierignore
- Update `js-audit-allow-list` to remove references to gatsby
- Update README to remove mentions of gatsby
- Add `raw-loader` to root package.json
	- It is used for other packages, not just within the site
- Add simplified versions of site header and footer to storybook
	- The responsive header has been simplified to remove site links instead of the sidebar as it didn't actually work
	- The logo now links back to the homepage too :)

### Desktop

No change.

|OLD|NEW|
|-----|-----|
|image|image|

### Responsive

Left is when the responsive menu is open.

|OLD|NEW|
|-----|-----|
|image|image|


.gitignore
```diff
@@ -1,8 +1,6 @@
 node_modules
 yarn-*.log
 lerna-*.log
-/site/.cache/
-/site/public/
 /storybook/public/
```
mcwinter07, line 4: We should also call out in WOL just to make sure no one merges in these files after we merge this in 😉

tsconfig.json
```diff
@@ -15,7 +15,8 @@
   "include": [
     "./packages/**/*",
     "./draft-packages/**/*",
-    "./legacy-packages/**/*"
+    "./legacy-packages/**/*",
+    "./storybook/**/*"
```
mcwinter07, line 19: Was storybook never included in our original tsconfig? interesting 🤔

mcwinter07 (approved):
Left a couple of comments that are more observations, but it all looks hunky dory to me! Merge in good health my friend 🦖

tsconfig.json
HeartSquared: Yep :P Maybe it started not having any TS so was exempt. But then we started adding things, and I thought it was weird that I wasn't getting any help/errors.

.gitignore
HeartSquared: True! Good call 📞

PR #2803, opened 2022-06-21
chore(deps): address vulnerabilities

HeartSquared:
## What

Address vulnerability warnings and update js audit allow list.

- Upgraded Storybook packages to ^6.5.9
- Reviewed and updated js audit allow list
- Removed `axios` resolution
- Added comment for why we have a resolution for `@types/react`

## Why

We're vulnerable :(

## Notes

We still have quite a long list of warnings (which are in the allow list) which are caused by:
- `lerna` 
	- we couldn't upgrade it simply, and since we're looking to change to `changesets` later, no point investing too much time into it
- `commitizen` 
	- the package itself seems to have updated its dependencies automatically, but there hasn't been a release done. Since it's a dev tool, not too fussed
- `@storybook/*` 
	- the main vulnerability seems to be `ansi-regex` in `node-logger`, which is quite embedded.
	- `trim-newlines` is a regex DOS which is not relevant (checked with Foundations)
- `svgo` 
	- seems that this requires extra config changes to get working, but there are new versions (same with `svgo-loader`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,8 +114,7 @@
     "webpack": "^5.68.0"
   },
   "resolutions": {
-    "@types/react": "^18.0.9",
-    "axios": "0.21.2"
+    "@types/react": "^18.0.9"
   },
   "config": {
     "commitizen": {
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-circus": "^27.4.2",
     "jest-static-stubs": "^0.0.1",
-    "lerna": "^3.22.1",
+    "lerna": "3.22.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-circus": "^27.4.2",
     "jest-static-stubs": "^0.0.1",
-    "lerna": "^3.22.1",
+    "lerna": "3.22.1",
```
mcwinter07, line 97: The second last final nail 🔨

mcwinter07 (commented):
Looks hunky dory 🦖 Merge on!

mcwinter07 (approved):

PR #2821, opened 2022-06-21
chore: Delete legacy title-block package

dougmacknz:
## What
Delete the legacy title-block package.

## Why
With Elm now moved out, and Murmur updated to import from the Kaizen Elm report, there's no repo consuming this package anymore.


HeartSquared (approved):
Is this technically a breaking change because it's a deletion? In case someone somewhere on the internet is using it 🙃

dougmacknz:
The whole package is being deleted so doing this as a breaking change wouldn't do anything I think?

PR #2828, opened 2022-06-23
refactor: remove deprecated styles and helpers from EmptyState

mcwinter07:
## Why
Empty state was relying on deprecated-component-library-helpers and using kz- typography styles,
which was importing old sizes, weights and line-heights. This has been removed and the Paragraph and
Heading components have used to align to the UI toolkit

## What
- replaces divs with `Paragraph` and `Heading` components
- replaces deprecates typographic styles with design tokens
- aligns sizes and responsiveness to ui toolkit

### DIFF
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
@@ -9,6 +9,7 @@ import {
   EmptyStatesPositive,
   AnimatedSceneProps,
 } from "@kaizen/draft-illustration"
+import { Paragraph, Heading } from "@kaizen/typography"
 import styles from "./styles.scss"
 
 const ILLUSTRATIONS: { [k: string]: React.VFC<AnimatedSceneProps> } = {
@@ -94,8 +95,16 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       </div>
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
-          <div className={styles.heading}>{headingText}</div>
-          <div className={styles.description}>{bodyText}</div>
+          <Heading
+            variant="heading-3"
+            classNameOverride={styles.heading}
+            tag="div"
+          >
+            {headingText}
+          </Heading>
+          <Paragraph variant="body" classNameOverride={styles.description}>
+            {bodyText}
+          </Paragraph>
           {children}
         </div>
       </div>
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
@@ -1,34 +1,44 @@
-@import "~@kaizen/deprecated-component-library-helpers/styles/layout";
 @import "~@kaizen/design-tokens/sass/layout";
 
 $small: 300px;
 $medium: $layout-breakpoints-medium;
 $large: $layout-breakpoints-large;
 
+// comes from ca-layout-padding x2 which is no longer used
+$side-padding: 48px;
 $sidebarWidth: 240px;
-$sidebarWithPadding: $sidebarWidth + (2 * $ca-layout-side-padding);
-$contentPadding: 2 * $ca-layout-side-padding;
+$sidebarWithPadding: $sidebarWidth + $side-padding;
+
+$typography-heading-3-font-size-small: 1.25rem;
+$typography-heading-3-font-weight-small: 700;
+$typography-heading-3-line-height-small: 1.5rem;
 
 @mixin large-sidebar-and-content {
-  @media (min-width: $large + $sidebarWithPadding + $contentPadding) {
+  @media (min-width: ($large + $sidebarWithPadding + $side-padding)) {
     @content;
   }
 }
 
 @mixin large-content-only {
-  @media (min-width: $large + $contentPadding) {
+  @media (min-width: ($large + $side-padding)) {
     @content;
   }
 }
 
 @mixin medium {
-  @media (min-width: $medium + $contentPadding) {
+  @media (min-width: ($medium + $side-padding)) {
     @content;
   }
 }
 
 @mixin small {
-  @media (max-width: $medium + $contentPadding - 1px) {
+  @media (max-width: ($medium + $side-padding - 1px)) {
     @content;
   }
+}
+
+@mixin typography-heading-3-sm {
+  font-size: $typography-heading-3-font-size-small;
+  line-height: $typography-heading-3-line-height-small;
+  font-weight: $typography-heading-3-font-weight-small;
 }
--- a/draft-packages/empty-state/package.json
+++ b/draft-packages/empty-state/package.json
@@ -35,7 +35,7 @@
     "@kaizen/component-base": "^1.1.0",
     "@kaizen/component-library": "^15.0.3",
     "@kaizen/draft-illustration": "^5.1.15",
-    "@types/classnames": "^2.3.0",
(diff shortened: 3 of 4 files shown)


draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -94,8 +96,20 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       </div>
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
-          <div className={styles.heading}>{headingText}</div>
-          <div className={styles.description}>{bodyText}</div>
+          <Box mb={1}>
+            <Heading
+              variant="heading-3"
+              classNameOverride={styles.heading}
+              tag="div"
```
mcwinter07, line 103: I've left this as a non semantic tag to ensure that the heading level structure of pages that use this component will not be broken

mcwinter07: not ideal but we could potentially add a prop to make this more accessible by giving them the option to set the tag level on the component - something like `headingTag`?

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -94,8 +96,20 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
...
+            >
+              {headingText}
+            </Heading>
+          </Box>
+          <Box mb={1}>
+            <Paragraph variant="body" classNameOverride={styles.description}>
```
christian314159, line 109: Ha, love how you use `classNameOverride` here within a system component. It really shows the need for more flexibility.

christian314159: `styles.description` includes `margin-bottom` as far as I can tell, and you wrap it into a `Box` to add even more margin (though they might collapse). I would add the `margins` directly to the `Heading` and `Paragraph` elements with the classes that already exist tbh.

draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
```diff
@@ -1,34 +1,43 @@
-@import "~@kaizen/deprecated-component-library-helpers/styles/layout";
 @import "~@kaizen/design-tokens/sass/layout";
 
 $small: 300px;
 $medium: $layout-breakpoints-medium;
 $large: $layout-breakpoints-large;
 
+$side-padding: 48px;
```
christian314159, line 7: There are already some existing `px` values in here, but I feel we should only add `rem` values, if we can - plus change the old `px` ones if we can easily test it causes no regressions.
This will help when users zoom in/out.

draft-packages/empty-state/package.json
```diff
@@ -35,11 +35,12 @@
...
     "@kaizen/component-library": "^15.0.3",
     "@kaizen/draft-illustration": "^5.1.13",
-    "@types/classnames": "^2.3.0",
+    "@kaizen/typography": "^2.2.3",
     "classnames": "^2.3.1"
   },
   "devDependencies": {
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "@types/classnames": "^2.3.0"
```
christian314159, line 43: Nice catch moving that to `devDependencies` 👏

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
christian314159: Potentially, but I'd call that out for now. It behaves as it always did using the `div`.

draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
mcwinter07: I was on the fence whether I should change this to a rem value, it could be supplanted by a spacing-xl but given it was only used in the media query below which uses a pixel value from design tokens I opted to leave it as is so as to not worry about calculating pixels and rems together.

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
mcwinter07: hahah yeah, originally it did have that margin but I removed it to use the box components - I could leave it in there and not have the DOM as populated with Boxes but given we have that it felt like I should have used it. Honestly happy to add that margin back in and keep it all in the styles to reduce the DOM nesting

christian314159: To me it would be much cleaner to not use the `Box` just for `margin` but have styles in the class. I think we should do it.

I think if the only use of `Box` is for `margins` then something is wrong. It really sounds like the good ol' "spacer gif" at that point ;) But that's a discussion outside of this PR.

draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
christian314159: I don't have strong feelings about either way here at this point. We as a team should probably align on what to use (and I'd say it should be `rems`). Don't have the link handy but you might remember the post I linked about which value is best for accessibility and how it behaves.

draft-packages/empty-state/package.json
HeartSquared: I actually think it can probably be removed since it's in the root package.json 😬 Unless something explodes which proves that it must be within every package

draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
sentience: It's worth looking at how the component behaves when you change the browser text size and using that to make a judgement call on whether the padding should grow/shrink with font size or not.

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
HeartSquared: Random unverified idea: We make an abstract component (which doesn't insert into the dom, but instead modifies the child) to deal with margins and we add it to all our components and expose it as a prop to consumers 🙃

draft-packages/empty-state/package.json
mcwinter07: Good point 🤔 will try it out and see if anything explodes

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
christian314159: @HeartSquared I think I've seen exactly that somewhere using Tailwind. The "abstract/wrapper" component doesn't add a DOM element but adds the Tailwind margin class to all its children.

HeartSquared: Very tempted to POC this idea for my next learning day :)

draft-packages/empty-state/KaizenDraft/EmptyState/responsive.scss
mcwinter07: @christian314159 That's fair, I think in this instance it since its replacing the a pixel value calculated off $ca-layout-side-padding we should leave it as is. On Kevins recommendation I tried out it out with resized text and it didn't have any unfavourable impacts on the layout. I've added a small comment in the scss just to mark where the value comes from.

christian314159 (approved):

christian314159 (approved):

PR #2829, opened 2022-06-24
chore: Remove legacy-packages folder

dougmacknz:
## What
Delete the legacy-packages folder and any config related to it.

## Why
There are no packages remaining in the `legacy-packages` folder, so this can be removed


HeartSquared (approved):
🔥

PR #2830, opened 2022-06-24
fix: Remove tab on scene illustrations [KDS-564]

christian314159:
## Why
- See: [KDS-564](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-564)
- Scene illustrations get tab focus (at least in Firefox), probably because they're `video` elements
image

## What
- Added `tabindex="-1"` to the `video` tag
- Refactor: Removed some duplication in the current component

### DIFF
--- a/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
@@ -76,6 +76,7 @@ export const VideoPlayer = ({
 
   useEffect(() => {
     if (!window) return
+
     const reducedMotionQuery = window.matchMedia(
       "(prefers-reduced-motion: reduce)"
     )
@@ -96,6 +97,7 @@ export const VideoPlayer = ({
   useEffect(() => {
     const { current: videoElement } = videoRef
     if (!videoElement) return
+
     if (prefersReducedMotion) {
       videoElement.pause()
     } else if (autoplay && !prefersReducedMotion) {
@@ -120,6 +122,7 @@ export const VideoPlayer = ({
     // Add event listeners for the video element
     const { current: videoElement } = videoRef
     if (!videoElement || !onEnded) return
+
     videoElement.addEventListener("ended", onEnded)
 
     return function cleanup() {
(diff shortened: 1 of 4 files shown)


draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -120,69 +122,48 @@ export const VideoPlayer = ({
...
+        <source src={assetUrl(`${source}.webm`)} type="video/webm" />
       )}
-    </>
+      <source src={assetUrl(`${source}.mp4`)} type="video/mp4" />
+    </video>
   )
+
+  if (aspectRatio) {
+    return (
+      <figure className={styles[aspectRatio] + " " + styles.aspectRatioWrapper}>
```
HeartSquared, line 162:
```suggestion
      <figure className={classNames([
      	styles[aspectRatio],
      	styles.aspectRatioWrapper
    ])}>
```

or

```suggestion
      <figure className={`${styles[aspectRatio]} ${styles.aspectRatioWrapper}`}>
```

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -120,69 +122,48 @@ export const VideoPlayer = ({
...
-           * This seems counter-intuitive, but webm support is codec specific.
-           * Only offer webm if we are positive the browser supports it.
-           * Reference: https://bugs.webkit.org/show_bug.cgi?id=216652#c1
-           */}
-          {canPlayWebm() && (
-            <source src={assetUrl(`${source}.webm`)} type="video/webm" />
-          )}
-          <source src={assetUrl(`${source}.mp4`)} type="video/mp4" />
-        </video>
+  const videoPlayer = (
```
HeartSquared, line 133:
```suggestion
  const VideoPlayer = (
```

Since it's a JSX Element, we should use PascalCase (might need to be renamed though of course, since it conflicts with the component name)

christian314159: I contemplated using PascalCase but decided against, because then it might look too much like a component and I'm not doing `...<VideoPlayer />...` but `...{videoPlayer}...`.

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
christian314159: Good call, I too lazily copied what was there ;)

draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
HeartSquared: Yeah it's a funny one. But it probably is moreso a subcomponent than a variable. I'd make it a component, but eh, not a big deal. Up to you whether you change it or not.

HeartSquared (approved):

PR #2833, opened 2022-06-27
test: improve DatePicker tests

BurntCaramel:
## Why

[KDS-584](https://cultureamp.atlassian.net/browse/KDS-584?atlOrigin=eyJpIjoiYzE2NTIwODc4OWI5NDliNzk1N2IyNWZkNzZhOTM4ZmIiLCJwIjoiaiJ9)
There’s some minor niceties we can use to make our tests more reflect the intended behaviour.

## What
- Use `getByRole` instead of using `selector` argument, or traversing DOM element’s parent
- Skip some assertions that have been done in previous tests
- Use `toBeVisible` instead of `toBeInTheDocument`, which makes sure the element is in the document _and_ the accessibility tree.
- Use tabbing instead of focusing on specific elements.

@HeartSquared changes:
- Remove unnecessary usages of `async`
- Clean up tests to use consistent patterns
- Add tests for `setFocusInCalendar`

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -165,7 +165,6 @@ describe("<DatePicker /> - Focus element", () => {
       const dateToSelect = screen.getByRole("button", {
         name: "6th March (Sunday)",
       })
-      expect(dateToSelect).toBeInstanceOf(HTMLElement)
       userEvent.click(dateToSelect)
 
       expect(input).toHaveFocus()
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from "react"
 import { render, screen, waitFor } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
-import format from "date-fns/format"
 import { FieldMessageStatus } from "@kaizen/draft-form"
 import { DatePicker, ValidationResponse } from "./DatePicker"
 import { DatePickerProps } from "."
@@ -83,27 +82,6 @@ describe("<DatePicker />", () => {
       name: "Go to previous month",
     })
     expect(arrowButton).toHaveFocus()
-  })
-})
-
-// @todo: Move this to be tests for utils/setFocusInCalendar
-describe("<DatePicker /> - Focus within calendar", () => {
-  it("shows focus on today when no date is selected", () => {
-    const today = new Date()
-    const todayFormatted = format(today, "do MMMM (eeee)") // e.g 6th June (Monday)
-
-    render(<DatePickerWrapper />)
-
-    const calendarButton = screen.getByRole("button", {
-      name: "Choose date",
-    })
-    userEvent.click(calendarButton)
-    waitFor(() => {
-      expect(screen.queryByRole("dialog")).toBeVisible()
-    })
-
-    const dateToSelect = screen.getByRole("button", { name: todayFormatted })
-    expect(dateToSelect).toHaveFocus()
   })
 })
 
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -49,7 +49,7 @@ describe("<DatePicker />", () => {
 
   it("should have an empty input value when a date is not provided", () => {
     render(<DatePickerWrapper />)
-    const input = screen.getByRole("combobox", { name: "Input label" })
+    const input = screen.getByLabelText("Input label")
     expect(input).toHaveValue("")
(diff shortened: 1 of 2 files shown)


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -141,31 +142,28 @@ describe("<DatePicker /> - Focus element", () => {
...
       userEvent.click(input)
 
       waitFor(() => {
-        expect(screen.queryByRole("dialog")).toBeInTheDocument()
+        expect(screen.queryByRole("dialog")).toBeVisible()
       })
 
-      const dateToSelect = screen.getByText("6th March (Sunday)").parentElement
-      if (dateToSelect) userEvent.click(dateToSelect)
+      const dateToSelect = screen.getByText("6th March (Sunday)").parentElement!
```
HeartSquared, line 164: You're being quite inconsistent in your changes?

If `const dateToSelect = screen.getByRole("button", { name: todayFormatted })` is the pattern you recommend, then it's best to use it consistently.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -283,25 +276,21 @@ describe("<DatePicker /> - Input format", () => {
...
     render(<DatePickerWrapper selectedDay={new Date("2022-03-01")} />)
 
-    const input = screen.getByLabelText("Input label", { selector: "input" })
+    const input = screen.getByRole("combobox", { name: "Input label" })
     expect(input).toHaveValue("Mar 1, 2022")
 
-    await waitFor(() => {
-      input.focus()
-    })
+    userEvent.tab()
```
HeartSquared, line 282: If there is any change in the dom structure, this will fail. The idea was that because `focus` on the input could be triggered through click or keyboard navigation, calling `.focus` would cater for both.

Nonetheless, you're right that calling `.focus` probably isn't the best way to replicate user intention, so I'd say it would be better to click on the input to be more explicit. Or, if we are using `tab`, then to first assert that the input has the focus.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -283,25 +276,21 @@ describe("<DatePicker /> - Input format", () => {
...
```
HeartSquared, line 293: Same as the above

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
BurntCaramel: The intent here is that “the user is real life will hit the tab key” but I understand if it looks a bit loose. But I believe that is the ‘contract’ that we are having with the user?

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -57,29 +57,29 @@ describe("<DatePicker />", () => {
     expect(screen.getByDisplayValue("Mar 1, 2022")).toBeInTheDocument()
   })
 
-  it("allows you to tab through input, button and calendar", () => {
+  it("allows you to tab through input, button and calendar", async () => {
```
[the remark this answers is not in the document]
BurntCaramel: Oh I experimented with using

```js
expect(await screen.findByRole("dialog")).toBeVisible()
```

instead of:

```js
waitFor(() => {
  expect(screen.getByRole("dialog")).toBeVisible()
})
```

What do you think? Better to always use `waitFor` rather than the more subtle `findByRole`?

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
BurntCaramel: One idea I had was to have one test that verifies that tabbing focuses the input we want, and then always just tab in other tests.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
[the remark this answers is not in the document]
HeartSquared: Hmm... I think it reads slightly clearer when we use `waitFor`. That being said, I don't really have an issue with either. I think it depends on what might be deemed the better practice? Perhaps one is more performant than the other? Might be worth a quick google to see. I definitely know when I was looking up things recently that I tended to see suggestions using `waitFor` and not many that suggest `await findBy`.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -48,7 +48,7 @@ describe("<DatePicker />", () => {
 
   it("should have an empty input value when a date is not provided", () => {
     render(<DatePickerWrapper />)
-    const input = screen.getByLabelText("Input label", { selector: "input" })
```
BurntCaramel, line 51: Let me know if you want this brought back

NatalieLoyzaga (approved):
Nice!!

PR #2834, opened 2022-06-27
refactor: update RTE nested list styles

mcwinter07:
## Why
The default lists in the rte did not have any nested styles.

## What
- adds recurring nested list styles for `ul` and `ol` tags up to a dept of 6
- I also used this PR to update the focus ring to match design since this is a very small scss change we wanted to fix

| Prev lists | Updated lists |
|---|---|
| old_list_styles | new_list_styles | 

| Prev border | Updated border |
|---|---|
|  old_focus   |   new_focus  |


christian314159 (approved):

dougmacknz (approved):
nice

PR #2837, opened 2022-06-27
feat: Add chevron up to Select when dropdown is open

JakePitman:
## Why
To indicate that the dropdown can be closed by clicking the Select component while open.

## What
Changes the Select dropdown indicator to an upwards chevron when the dropdown is open.


HeartSquared (dismissed):
This works, but worth adding an example to the sticker sheet of an open Select so Chromatic can snapshot it :)

Looking at it, the sticker sheet could be improved, so we can look at revising it if you'd like.
Also the non-sticker sheet stories don't need to be sent to Chromatic.

HeartSquared (approved):

HeartSquared (dismissed):

HeartSquared (approved):

PR #2839, opened 2022-06-28
feat: Add en-AU and en-US localisation

NatalieLoyzaga:
## Why

- Add localisation for en-AU and en-US

## What

image


packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -360,14 +385,15 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
             inputRef={inputRef}
             buttonRef={buttonRef}
             id={id}
-            calendarId={`${id}-calendar-dialog`}
```
NatalieLoyzaga, line 363: Here @HeartSquared

PR #2840, opened 2022-06-29
fix: Update styles to use import instead of require

abaayd01:
## Why

- when upgrading performance-ui to @cultureamp/frontend-build v2, we discovered style modules referenced using require were not being applied
  - see this slack thread for more context: https://cultureamp.slack.com/archives/C016N8Y1E10/p1656464394609989
- the suggested solution from FE foundations was to update kaizen packages to use import instead of require
- it looks like this change was made for a bunch of packages a while ago, but these couple components got missed: https://github.com/cultureamp/kaizen-design-system/pull/782

## What

- no functional change, using `import` should result in the same behaviour


HeartSquared (approved):

HeartSquared:
You'll need to update your commit message to use a conventional commit (`yarn commit` can help), but the changes look good :)

HeartSquared (approved):

PR #2842, opened 2022-06-30
feat: Add RTL styles to Calendar and DateInput

NatalieLoyzaga:
## Why

Add missing RTL styles triggered when global dir="rtl".

## What

- Update arrows on calendar
- Update calendar button spacing

### DIFF
--- a/packages/date-picker/CHANGELOG.md
+++ b/packages/date-picker/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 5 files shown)


HeartSquared:
I think we have a few questions we need answered for RTL behaviour.

Here is some strangeness we've encountered, which may be valid behaviour that just looks strange in English:
|||
|---|---|
|image|image|

Also, clicking on the Previous and Next buttons in the calendar, as they are now swapped, the right one goes backwards, and the left one goes forwards. Unsure if this is actually the expected behaviour or if people who read RTL still expect the left to be Prev and right to be Next.

HeartSquared (approved):

PR #2843, opened 2022-06-30
fix: lexical error on sass variable

NatalieLoyzaga:
## Why

Lexical error on sass variable

## What

Wrap variable with #{}


HeartSquared (approved):
🤞🏻

PR #2846, opened 2022-07-04
docs: Update Datepicker stories into separate stickersheets

NatalieLoyzaga:
## Why

Clearer sticker sheets and allows for more component examples

## What

Create a new sticker sheet story for `Date Input` and one for `Calendar`.


HeartSquared:
I would actually expect the Storybook menu structure to look like this:
```
DatePicker
- Primitives (maybe _Primitives to surface it to the top?)
-- Calendar
--- Sticket Sheet (Default)
--- Sticker Sheet (Reversed)
-- DateInput
--- Sticket Sheet (Default)
--- Sticker Sheet (Reversed)
- Date Picker
-- Date Picker (although it seems weird that we're now at 3 layers lol, but that's another discussion)
-- Validation
- Date Range Picker
-- Date Range Picker
```

Or, if DateInput will be exclusive to DatePicker, the sticker sheets for DateInput could sit within DatePicker and DateInput will not be exposed as a primitive (this is dependent on how the DateRangePicker design evolves - or we could just do it).

One other thing I realised was that we should actually embed the icon into DateInput since it shouldn't be anything else - splitting out DateInput highlights this unnecessary flexibility.

NatalieLoyzaga:
@HeartSquared Also noticing we do not have reversed styling... Should I remove the reversed stickersheets till we do? Assuming we will need a design :/

HeartSquared:
>Should I remove the reversed stickersheets till we do?

Yeah I suppose if we don't have a design, then it makes sense not to have it yet. Good thing it just uses a template :)
We need to bring it up with @jmitsios though.

NatalieLoyzaga:
@HeartSquared Okay I removed the reversed sticker sheets and I'll create a card in design for jason.

NatalieLoyzaga:
@HeartSquared  Anymore comments for this one?

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const SUPPORTED_LOCALES = ["en-US", "en-AU"]
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.datePicker}/Primitives/Calendar`,
+  component: Calendar,
+  parameters: {
+    docs: {
+      description: {
+        component: 'import { Calendar, DateInput } from "@kaizen/date-picker"',
```
HeartSquared, line 18:
```suggestion
        component: 'This is a subcomponent - use DatePicker or DateRangePicker.',
```

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+  argTypes: {
+    locale: {
+      options: SUPPORTED_LOCALES,
+      control: { type: "radio" },
+    },
+  },
+}
+
+const CalendarExample = (props: Partial<CalendarProps>): JSX.Element => {
+  const [popperElement, setPopperElement] = useState<HTMLDivElement | null>(
```
HeartSquared, line 31: Not to be fixed here, but this shows that the Calendar component shouldn't know about popper (it should be the parent's responsibility).

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+    modifiers: [
+      {
+        name: "offset",
+        options: {
+          offset: [0, 15],
+        },
+      },
+    ],
+    placement: "bottom-start",
+  })
```
HeartSquared, line 48: I suspect these options aren't required since we aren't actually using the popper properly? But test it just in case.

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <div ref={setReferenceElement}>
+      <Calendar
+        mode="single"
+        id="calendar-dialog"
```
HeartSquared, line 54: ~~You'll end up with 2 elements with the same id. This needs to be dynamic.~~

Seems you override it. A little unclear at a glance though, so ideally if we could remove the assignment here it would be better.

packages/date-picker/docs/DatePicker.stories.tsx
```diff
@@ -290,7 +249,7 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
             onValidate={() => undefined}
             isReversed={isReversed}
             status="error"
-            validationMessage="Invalid Date"
+            validationMessage="Invalid Date."
```
HeartSquared, line 252: You probably didn't want this :)

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+      },
+    ],
+    placement: "bottom-start",
+  })
+
+  return (
+    <div ref={setReferenceElement}>
+      <Calendar
+        mode="single"
+        id={props.id ? props.id : "calendar-dialog"}
```
HeartSquared, line 55:
```suggestion
        id={props.id}
```
I take it this doesn't work? 😅

HeartSquared (approved):

packages/date-picker/docs/Calendar.stories.tsx
NatalieLoyzaga: Yeah it doesn't I tried - ts error

PR #2847, opened 2022-07-04
docs: add pseudo states for calendar sticker sheet

NatalieLoyzaga:
## Why

Add pseudo states for Calendar to be sent to Chromatic

## What

- Use storybook .play() to get elements and add className focus-visible and hover story class.
- Checked chromatic to ensure the popovers aren't overlapping

image

### DIFF
--- a/packages/date-picker/docs/Calendar.stories.tsx
+++ b/packages/date-picker/docs/Calendar.stories.tsx
@@ -46,7 +46,7 @@ const CalendarExample = (
 const StickerSheetCalendarTemplate: Story<{ isReversed: boolean }> = ({
   isReversed,
 }) => (
-  <div style={{ height: "1600px" }}>
+  <>
     <StoryWrapper isReversed={isReversed}>
       <StoryWrapper.RowHeader
         headings={["Hover", "Focus", "Disabled", "Disabled Focus"]}
@@ -85,13 +85,13 @@ const StickerSheetCalendarTemplate: Story<{ isReversed: boolean }> = ({
         <CalendarExample id="navigation-focus" />
       </StoryWrapper.Row>
     </StoryWrapper>
-  </div>
+  </>
 )
 
 export const StickerSheetCalendar = StickerSheetCalendarTemplate.bind({})
 StickerSheetCalendar.storyName = "Sticker Sheet (Default)"
 StickerSheetCalendar.parameters = {
-  chromatic: { disable: false, viewports: [1800] },
+  chromatic: { disable: false },
   controls: { disable: true },
 }
 
@@ -106,16 +106,16 @@ StickerSheetCalendar.play = ({ canvasElement }) => {
   }
 
   const calendars = [
-    { calendar: "default", name: "5th September (Monday)" },
-    { calendar: "selected", name: "5th September (Monday)" },
-    { calendar: "navigation", name: "Go to previous month" },
+    { row: "default", name: "5th September (Monday)" },
+    { row: "selected", name: "5th September (Monday)" },
+    { row: "navigation", name: "Go to previous month" },
   ]
 
-  calendars.forEach(cal => {
-    getElementWithinCalendar(`${cal.calendar}-hover`, cal.name).classList.add(
-      "story_hover"
+  calendars.forEach(({ row, name }) => {
+    getElementWithinCalendar(`${row}-hover`, name).classList.add(
+      "story__datepicker__calendar--hover"
     )
-    getElementWithinCalendar(`${cal.calendar}-focus`, cal.name).classList.add(
+    getElementWithinCalendar(`${row}-focus`, name).classList.add(
       "focus-visible"
     )
   })
--- a/packages/date-picker/docs/Calendar.stories.tsx
+++ b/packages/date-picker/docs/Calendar.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { Story } from "@storybook/react"
 import { enAU } from "date-fns/locale"
+import { within } from "@storybook/testing-library"
 import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
 import { DayOfWeek } from "../src/enums"
 import { Calendar, CalendarProps } from "../src/_primitives/Calendar"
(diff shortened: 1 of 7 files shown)


packages/date-picker/package.json
```diff
@@ -43,7 +43,8 @@
   },
   "devDependencies": {
     "@kaizen/design-tokens": "^9.0.3",
-    "@types/classnames": "^2.3.1"
+    "@types/classnames": "^2.3.1",
+    "@storybook/testing-library": "^0.0.13"
```
HeartSquared, line 47: I'd put this in the root `package.json`

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -251,3 +251,18 @@ $rdp-cell-size: 40px;
   background-color: $color-blue-100;
   color: $color-blue-500;
 }
+
+// Add Storybook psuedo classes
```
HeartSquared, line 255:
```suggestion
// Add Storybook pseudo classes
```

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -251,3 +251,18 @@ $rdp-cell-size: 40px;
...
+:global(.story__day-hover) {
```
HeartSquared, line 256: Ideally, we'd be able to put this class just wherever we've got the styles already defined. This way, we wouldn't need to change multiple places to ensure they align.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
```diff
@@ -251,3 +251,18 @@ $rdp-cell-size: 40px;
...
+  &:not([aria-disabled="true"]) {
+    &:not(.outside) {
```
HeartSquared, line 258: We should probably revisit this `.outside` class and find out exactly what it's doing.

HeartSquared (commented):
Leaving a few comments first. Haven't looked at the stories yet.

packages/date-picker/src/DatePicker/components/Calendar/Calendar.scss
NatalieLoyzaga: i hate this word ahahh

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -43,24 +44,82 @@ const CalendarExample = (
...
+    { calendar: "default", name: "5th September (Monday)" },
+    { calendar: "selected", name: "5th September (Monday)" },
+    { calendar: "navigation", name: "Go to previous month" },
+  ]
+
+  calendars.forEach(cal => {
+    getElementWithinCalendar(`${cal.calendar}-hover`, cal.name).classList.add(
+      "story_hover"
+    )
+    getElementWithinCalendar(`${cal.calendar}-focus`, cal.name).classList.add(
```
HeartSquared, line 118:
```suggestion
  calendars.forEach(({ calendar, name }) => {
    getElementWithinCalendar(`${calendar}-hover`, name).classList.add(
      "story_hover"
    )
    getElementWithinCalendar(`${calendar}-focus`, name).classList.add(
```

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -43,24 +44,82 @@ const CalendarExample = (
...
+
+  const getElementWithinCalendar = (id: string, name: string) => {
+    const calendar = canvas.getByTestId(id)
+    return within(calendar).getByRole("button", {
+      name,
+    })
+  }
+
+  const calendars = [
+    { calendar: "default", name: "5th September (Monday)" },
```
HeartSquared, line 109:
```suggestion
    { row: "default", name: "5th September (Monday)" },
```

Moreso a nitpick, but I got a little confused thinking `calendar` was referencing a single Calendar, but it seems to be referencing a row of calendars

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -43,24 +44,82 @@ const CalendarExample = (
...
+      </StoryWrapper.Row>
+    </StoryWrapper>
+  </div>
 )
 
 export const StickerSheetCalendar = StickerSheetCalendarTemplate.bind({})
 StickerSheetCalendar.storyName = "Sticker Sheet (Default)"
 StickerSheetCalendar.parameters = {
-  chromatic: { disable: false },
+  chromatic: { disable: false, viewports: [1800] },
```
HeartSquared, line 94: No longer needed since the popper isn't there to cause weirdness anymore :)

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -43,24 +44,82 @@ const CalendarExample = (
 const StickerSheetCalendarTemplate: Story<{ isReversed: boolean }> = ({
   isReversed,
 }) => (
-  <StoryWrapper isReversed={isReversed}>
-    <StoryWrapper.RowHeader headings={["Selected Date", "Disabled Dates"]} />
-    <StoryWrapper.Row rowTitle="Calendar">
-      <CalendarExample
-        id="calendar-dialog-selected"
-        value={new Date(2022, 1, 5)}
+  <div style={{ height: "1600px" }}>
```
HeartSquared, line 49: No longer needed since the popper isn't there to cause weirdness anymore :)

packages/date-picker/src/_primitives/Calendar/Calendar.scss
```diff
@@ -98,7 +98,8 @@ $rdp-cell-size: 40px;
     border: $border-solid-border-width $border-solid-border-style transparent;
     border-radius: $cell-border-radius;
 
-    &:hover {
+    &:hover,
+    &:global(.story_hover) {
```
HeartSquared, line 102: Since it's a global class we should be more specific ([bem is helpful here](http://getbem.com/naming/)), otherwise we might end up with global styles from different component sticker sheets affecting others.

```suggestion
    &:global(.story__datepicker__calendar--hover) {
```

packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -30,37 +31,97 @@ export default {
...
 ): JSX.Element => (
-  <Calendar
-    mode="single"
-    weekStartsOn={DayOfWeek.Sun}
-    onDayChange={() => undefined}
-    defaultMonth={new Date(2022, 1, 5)}
-    locale={enAU}
-    {...props}
-  />
+  <div data-testid={props.id}>
```
HeartSquared, line 34: I was wondering why we couldn't just put this on `Calendar` itself (it's what I would expect) and it turns out we don't accept further props.

While we could certainly put it in now, the question is whether it should go on the container or the DayPicker. 
Ideally I'd expect to extend DayPicker because it's the main element, but it appears we can't actually remove the container because DayPicker doesn't accept `ref`, `id`, `role`, of which case it means `data-testid` should therefore go on the container.

Strange predicament here, but I suppose as a subcomponent it can be less flexible 🤷🏻 

TLDR; you can either leave it, or add `...restProps` to Calendar and put it in the container div (remember to use `OverrideClassName<HTMLAttributes<HTMLDivElement>>` if doing so).

HeartSquared (approved):
Love your work! 💯

PR #2848, opened 2022-07-05
test: Add list tests to RTE

mcwinter07:
## Why
Increase test coverage of the Kaizen RTE component

## What
- add list tests


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+      const decreaseIndentBtn = screen.getByRole("button", {
+        name: "Decrease indent",
+      })
+      const increaseIndentBtn = screen.getByRole("button", {
+        name: "Increase indent",
+      })
+
+      expect(decreaseIndentBtn).toHaveAttribute("aria-disabled", "true")
+      expect(increaseIndentBtn).toHaveAttribute("aria-disabled", "true")
```
dougmacknz, line 151: Is there a `toBeDisabled` assertion that can be used here?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+      fireEvent.focus(screen.getByRole("textbox", { name: "List RTE" }), {
+        target: { textContent: "this will be a ul" },
+      })
+
+      await waitFor(() => {
+        userEvent.click(screen.getByRole("button", { name: "Bullet List" }))
+        screen.getByRole("list")
+
+        expect(screen.getByRole("list")).toBeInTheDocument()
+        expect(document.querySelectorAll("ul").length).toBeGreaterThan(0)
```
dougmacknz, line 83: Is the second assertion here redundant? Wouldn't the first one fail if there's no ul?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      },
+    ]
+
+    it("indent buttons are 'disabled' by default", () => {
```
dougmacknz, line 140: /nit these test cases don't read nicely like your others "it indent buttons..."

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
mcwinter07: That is true, hang up from when I was tryign to be thorough to verify it was working, the less assertions the better hey

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
mcwinter07: there is, but sadly it doesn't capture aria-disabled, only disabled={true}

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.spec.tsx
mcwinter07: nit away :) could definitely update this!

dougmacknz (approved):

PR #2852, opened 2022-07-06
Exclude branches created to resolve Snyk findings

mechanicalpete:
## Why
Ignore branches, created by Snyk, to reduce the cost of running Chromatic


HeartSquared:
I actually don't think this will work since we have protection set up requiring `UI Tests` (an action triggered from a response received from Chromatic) to pass before we can merge.

That being said, perhaps it would make sense to not need to run Chromatic? Or maybe not in case the package upgrade somehow breaks a component. We'll probably need to give it a little more thought within our team.

PR #2853, opened 2022-07-07
refactor: move popper usage from Calendar to Date(Range)Picker

HeartSquared:
## What

Removed the requirement of setting popper attributes from Calendar.
Adds `CalendarWrapper` which handles popper functionality for both DatePicker and DateRangePicker.

## Why

While reviewing https://github.com/cultureamp/kaizen-design-system/pull/2847 I realised there was an unnecessary dependency for popper attributes in the Calendar component.

### DIFF
--- a/packages/date-picker/src/_primitives/CalendarWrapper/CalendarWrapper.tsx
+++ b/packages/date-picker/src/_primitives/CalendarWrapper/CalendarWrapper.tsx
@@ -1,9 +1,12 @@
 import React, { HTMLAttributes, useState } from "react"
+import { OverrideClassName } from "@kaizen/component-base"
+import classnames from "classnames"
 import { usePopper } from "react-popper"
 import { Calendar } from "../Calendar"
 import calendarWrapperStyles from "./CalendarWrapper.scss"
 
-interface CalendarWrapperProps extends HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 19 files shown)


packages/date-picker/src/_primitives/CalendarWrapper/CalendarWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Calendar } from "../Calendar"
+import calendarWrapperStyles from "./CalendarWrapper.scss"
+
+interface CalendarWrapperProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
+  children: React.ReactElement<typeof Calendar>
+  referenceElement: HTMLElement | null
+}
+
+export const CalendarWrapper: React.VFC<CalendarWrapperProps> = ({
```
HeartSquared, line 14: Not 100% sure if this is a good name. Was going to name it `FloatingCalendarWrapper` to give it more context, but it sounded a little weird too. Open to suggestions.

NatalieLoyzaga: Haha I think it's fair and still gives people a good idea of what it is - Should it be in a components folder under Calendar?

PR #2854, opened 2022-07-07
feat: Create RichTextContent and EditableRichTextContent components

dougmacknz:
## What
Creates two new components in the `@kaizen/rich-text-editor` package.

**`RichTextContent`**
Renders content created from the `RichTextEditor`. Creates a schema that includes all of our possible features/controls.

**`EditableRichTextContent`**
Renders the content with some styling and accessibility built in. Adds hover and focus styling, and a visually hidden button.

Smaller changes:
* Created some demo rich text data for our stories
* Moved `Toolbar` and `ToggleIconButton` into a `Subcomponents` subfolder on Storybook
* Moved all of the common styling rules for our content (things like typography and list styling) into a single mixin that is applied to both the editor and content.
* Created a constant array with our possible control names, to make it easy to generate a schema with *everything* for the content component.
* Moved some stuff out of the `RichTextEditor` folder into the `src` folder so that it can used across multiple components.

### Known issues
* I have done a temporary solution for the focus styling of the editable content area. Currently, when you're focused on a link inside the content area, you will see the focus indicator for the content area as well. I'll be following up with an adjustment for this.
* When clicking a link inside of the editable content wrapper, the event bubbles up and the onClick of the editable wrapper is triggered.

Both of these things will be fixed in a follow up.

## Why
Our teams need a way of rendering the content that's created with our RTE component, and this gives them an easy way to do that.

We've also seen the requirement for the pattern built here where the content is a clickable area, which will allow a user to switch to editing the content.

### DIFF
--- a/packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
+++ b/packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
@@ -1,0 +1,73 @@
+import React, { useState } from "react"
+import {
+  EditableRichTextContent,
+  RichTextEditor,
+  EditorContentArray,
+} from "@kaizen/rich-text-editor"
+import { Button } from "@kaizen/button"
+import { Box } from "@kaizen/component-library"
+import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
+import dummyContent from "./dummyContent.json"
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.richTextEditor}/Editable Rich Text Content`,
+  component: EditableRichTextContent,
+  parameters: {
+    docs: {
+      description: {
+        component:
+          'import { EditableRichTextContent } from "@kaizen/rich-text-editor"',
+      },
+    },
+  },
+}
+
+export const EditableRichTextContentStory = args => <InlineEditor {...args} />
+EditableRichTextContentStory.storyName = "Default"
+EditableRichTextContentStory.args = {
+  content: dummyContent,
+  labelText: "Shared notes",
+  isLabelHidden: false,
+}
+
+function InlineEditor(props: {
+  content: EditorContentArray
+  labelText: string
+}) {
+  const [editMode, setEditMode] = useState<boolean>(false)
+  const [rteData, setRTEData] = useState<EditorContentArray>(dummyContent)
+  const handleContentClick = () => setEditMode(true)
+  const handleCancel = () => setEditMode(false)
+
+  if (editMode) {
+    return (
+      <>
+        <RichTextEditor
+          labelText={props.labelText}
+          controls={[
+            { name: "bold", group: "inline" },
+            { name: "italic", group: "inline" },
+            { name: "underline", group: "inline" },
+            { name: "orderedList", group: "list" },
+            { name: "bulletList", group: "list" },
+            { name: "link", group: "link" },
+          ]}
+          value={rteData}
+          onChange={data => setRTEData(data)}
+        />
+        <Box mt={0.5} style={{ display: "flex", justifyContent: "end" }}>
+          <Button label="Cancel" secondary onClick={handleCancel} />
+          <Button label="Save" primary onClick={handleCancel} />
+        </Box>
+      </>
+    )
+  }
+
+  return (
+    <EditableRichTextContent
+      onClick={handleContentClick}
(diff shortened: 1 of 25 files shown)


packages/rich-text-editor/src/_mixins.scss
```diff
@@ -0,0 +1,41 @@
...
+    margin-bottom: 0;
+  }
+
+  ol {
+    list-style-type: decimal;
+    ol {
+      list-style-type: lower-alpha;
+      ol {
+        list-style-type: lower-roman;
+        @content;
```
mcwinter07, line 27: One thing I just wanted to call out here is if they go into more than 3 nested lists it'll just repeat the last list-style, ie: lower-roman or square. I wasn't sure whether more nesting would be required or if it was overkill, thus the:
  ```
  @include nestedListStyles {
    @include nestedListStyles;
  }
  ```
  If we don't need it, we can probably get rid of the @content 😄

packages/rich-text-editor/src/RichTextEditor/schema.ts
```diff
@@ -3,7 +3,8 @@ import {
   nodes as coreNodes,
   marks as coreMarks,
 } from "@cultureamp/rich-text-toolkit"
-import { ToolbarItems } from "./RichTextEditor"
+import { TOOLBAR_CONTROLS } from "../constants"
```
mcwinter07, line 6: Nice, some good refactoring in here 👍

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.editableContainer :global(.ProseMirror) {
+  padding: $spacing-sm;
+  position: relative;
+  border-radius: $border-solid-border-radius;
+  border: $border-focus-ring-border-width $border-focus-ring-border-style
+    transparent;
+  transition: background-color $animation-duration-immediate,
+    border-color $animation-duration-immediate;
+}
```
mcwinter07, line 14: I was thinking we should have a border when its inactive just to make it clear its an input? That is unless a design says otherwise, in which case feel free to ignore 😆

packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+            { name: "italic", group: "inline" },
+            { name: "underline", group: "inline" },
+            { name: "orderedList", group: "list" },
+            { name: "bulletList", group: "list" },
+            { name: "link", group: "link" },
+          ]}
+          value={rteData}
+          onChange={data => setRTEData(data)}
+        />
+        <Box mt={0.5} style={{ display: "flex", justifyContent: "end" }}>
```
mcwinter07, line 58: Don't know if this is required since its just a story example but thought I'd just call out this minor spacing nit
Screen Shot 2022-07-13 at 10 33 18 am

mcwinter07 (commented):
Aside from what we've already talked about re: the click event bubbling I've left a couple of comments and observations around some styles. Other than that it all looks hunky dory to me 👍

packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
dougmacknz: Yeah got a little lazy here

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.scss
mcwinter07: Nice 👍

packages/rich-text-editor/src/_mixins.scss
dougmacknz: Oh oops, sounds like I misunderstood how this was working

mcwinter07: That said having 6 potential nested lists seem very unlikely -  the `@include nestedListStyles` was just saved on writing really verbose scss - its really a judgment call on whether we want to support anything more the 3 levels deep in a list

dougmacknz: I've restored it :)

mcwinter07 (approved):

PR #2855, opened 2022-07-07
feat: update focus and hover styles for Calendar component

NatalieLoyzaga:
## Why

Update styles within the UI kit.

## What

- Update the focus ring on days and arrows.
- Add the hover styling when focused.


HeartSquared (approved):
Looks great!

PR #2856, opened 2022-07-11
feat: update secondary select styles

NatalieLoyzaga:
## Why

Secondary placeholder was missing padding and reversed styles.

## What

- Add padding to placeholder
- Add reversed colour to placeholder and label when reversed.


mcwinter07 (approved):
Nice work, the fix looks much better 🦖 I might suggest checking in with @jmitsios before you merge it in because the design kit has two different spaces between the select placeholder text and the arrow (figma snaps below). Also if you've already done this ignore and merge on 👍 

### Matches your spacing
Screen Shot 2022-07-11 at 5 09 35 pm

### Slightly narrower spacing

Screen Shot 2022-07-11 at 5 06 48 pm

PR #2857, opened 2022-07-11
refactor: update list behaviours

mcwinter07:
## Why
- The RTE was unable to use any of the buttons if the rich text editor content was empty.
- This adds a function for the Mark and Node toggle in the control map to create an initial <p> tag if there is no content in the editor. An initial tag is required by prosemirror before and node is able to be wrapped.

## Before
https://user-images.githubusercontent.com/36558508/178204989-56c45675-18bd-4ae3-a1ff-8ef2a112b9ef.mp4

## After
https://user-images.githubusercontent.com/36558508/178205011-ed35d15c-b9f2-48c9-b673-a4e6e1dae820.mp4

## What
- Adds `chainTransaction` helper  to chain multiple commands and dispatch each transitions in sequential order
  - chainCommand exist on recieving true value so this needed to be created
- Adds `createInitialParagraph` which adds a `<p>` tag to the initial Selection (which should be 0 if no content is in the editor)
- Adds conditional `docIsEmpty` check to ensure that the chaining is only run if required
- Adds descriptions

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/controlmap.ts
+++ b/packages/rich-text-editor/src/RichTextEditor/controlmap.ts
@@ -7,7 +7,7 @@ import decreaseIndentIcon from "@kaizen/component-library/icons/decrease-indent.
 import increaseIndentIcon from "@kaizen/component-library/icons/increase-indent.icon.svg"
 import linkIcon from "@kaizen/component-library/icons/add-link.icon.svg"
 
-import { EditorState, Transaction, TextSelection } from "prosemirror-state"
+import { EditorState, Transaction } from "prosemirror-state"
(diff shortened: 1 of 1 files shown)


packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -32,18 +32,73 @@ type ControlGroupTypes = {
   [key in ToolbarControlTypes]?: string
 }
 
+/**
+ * Chains multiple commands to dispatch each transitions in sequential order
+ */
+function chainTransactions(...commands: Command[]): Command {
```
mcwinter07, line 38: Depending on whether this is helpful or not, we could potentially export this from the toolkit

dougmacknz (approved):

PR #2860, opened 2022-07-13
test: add playwright user interaction tests

mcwinter07:
## Why
To extend the functionality of testing in Kaizen and to make it easier to simulate E2E tests with components like the Rich Text Editor. Relates to this [ticket](https://cultureamp.atlassian.net/browse/KDS-668)

## What
- Adds playwright step to Buildkite CI
- Adds playwright dev and default config
- Adds new 'with data' story for RTE to make the default more testable


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.playwright.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const iframePath = "iframe.html?args=&id="
+const storyId = "components-rich-text-editor-rich-text-editor--default"
+
+test.beforeEach(async ({ page }) => {
+  // Go to the starting url before each test.
+  await page.goto(`./${iframePath}${storyId}`)
+})
+
+test.describe("Test list keyboard shortcuts", async () => {
```
christian314159, line 11: > `Test list keyboard shortcuts...`

Don't want to be nitpicking here, but I think these being the first Playwright tests we should set the right example: The word "Test" feels redundant, as well as in the other `describes` below. It could just be `List keyboard shortcuts`... though that can read like `list` is a verb.... 😓 So maybe `Keyboard shortcuts for Lists`?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.playwright.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const editor = await page.locator('div[role="textbox"]')
+
+    await editor.click()
+    await editor.type("-")
+    await editor.press("Space")
+    await editor.type("By the pricking of my thumbs")
+    await editor.press("Enter")
+    await editor.type("Something wicked this way comes")
+
+    expect(await editor.locator("ul li").count()).toBe(2)
```
christian314159, line 22: Again, super nitpicky, but technically `expect(await editor.locator("ul li").count()).toBe(2)` could have been `true` even before you interacted with the editor.
If you want to be super thorough you'd start the test with:
`expect(await editor.locator("ul li").count()).toBe(0)`

christian314159 (commented):
👍 Nice work adding Playwright, just left some comments on how we could potentially improve the tests themselves.

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.playwright.spec.tsx
mcwinter07: That's fair, test is a little redundant giving this is a test 😆 I'll update this to make it more contextual

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.playwright.spec.tsx
mcwinter07: Good tip, will amend that test

christian314159 (approved):
👍

PR #2862, opened 2022-07-20
feat: remove all deprecated styles

NatalieLoyzaga:
## Why

Remove all deprecated styles and mixins.

## What

Remove all instances of the old design style sheets and replace with new design tokens. 
Ran Chromatic to ensure no changes.


christian314159 (approved):
I compared all the notifications in Storybook and they all still look identical 👍

PR #2865, opened 2022-07-21
refactor: remove deprecated styles from Tag component

mcwinter07:
## Why
To remove dependency on old deprecate library helpers - [see ticket ](https://cultureamp.atlassian.net/browse/KDS-570) and [original discovery](https://cultureamp.atlassian.net/browse/KDS-548)

## What
- replaces style dependent on `deprecated-component-library-helpers` with design tokens and modern CSS
- remove redundant @types/classname from package.json dependency

### DIFF
--- a/draft-packages/tag/KaizenDraft/Tag/Tag.scss
+++ b/draft-packages/tag/KaizenDraft/Tag/Tag.scss
@@ -57,6 +57,11 @@ $small: $spacing-md;
 
 .textContent {
   margin-top: -1px;
+  font-family: $typography-paragraph-small-font-family;
+  font-weight: $typography-paragraph-small-font-weight;
+  font-size: $typography-paragraph-small-font-size;
+  line-height: $typography-paragraph-small-line-height;
+  letter-spacing: $typography-paragraph-small-letter-spacing;
   white-space: nowrap;
 }
 
@@ -67,14 +72,10 @@ $small: $spacing-md;
   height: 100%;
   align-items: center;
   padding: 0 $spacing-xs;
-  margin-right: -0.6625rem;
-  margin-left: -0.225rem;
+  margin-inline-end: -0.6625rem;
+  margin-inline-start: -0.225rem;
   color: rgba($color-purple-800-rgb, 0.7);
   cursor: pointer;
-  [dir="rtl"] & {
-    margin-left: -0.6625rem;
-    margin-right: -0.225rem;
-  }
 
   &:hover {
     color: $color-purple-800;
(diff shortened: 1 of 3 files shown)


draft-packages/tag/KaizenDraft/Tag/Tag.scss
```diff
@@ -43,20 +42,20 @@ $small: $spacing-md;
...
-  @include ca-margin($start: calc(#{$spacing-md} * -0.15), $end: $spacing-xs);
+  margin-inline-start: calc(#{$spacing-md} * -0.15);
+  margin-inline-end: $spacing-xs;
   display: flex;
   align-items: center;
 }
 
 .textContent {
-  @include kz-typography-paragraph-small;
-  @include ca-inherit-baseline;
```
HeartSquared, line 59: This definitely isn't needed, and nothing needs to be put in as a replacement?

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
```diff
@@ -133,16 +134,18 @@ const Tag = (props: TagProps) => {
                 return
             }
           })()}
-        <span
-          className={classNames(styles.textContent, {
+        <Paragraph
```
HeartSquared, line 137: Mmm...I'm a little iffy about this change. If we make any change to Paragraph later, it shouldn't be changing the styles of Tag, and this opens us up for that.

draft-packages/tag/KaizenDraft/Tag/Tag.scss
mcwinter07: Not if we're using <Paraghraph> - especially not ca-inherit-baseline - that's a super old pattern that should have long been deprecated

draft-packages/tag/KaizenDraft/Tag/Tag.tsx
mcwinter07: That's fair, if we're using the design tokens in the scss then that should be fine. My thinking that brought to this solution was to ensure continuity with text styles with little maintenance - mainly to avoid things being out of sync like here

draft-packages/tag/KaizenDraft/Tag/Tag.scss
```diff
@@ -68,14 +72,12 @@ $small: $spacing-md;
   height: 100%;
   align-items: center;
   padding: 0 $spacing-xs;
+  margin-inline-end: -0.6625rem;
+  margin-inline-start: -0.225rem;
   margin-right: -0.6625rem;
   margin-left: -0.225rem;
```
HeartSquared, line 78: If `inline` is compatible, then we probably wouldn't need these lines either?

mcwinter07: this is correct! I should've removed them 👍

HeartSquared (approved):

PR #2870, opened 2022-07-25
fix: Rename ref to forwardRef on custom button element render

dougmacknz:
BREAKING CHANGE: Any usages of the 'component' prop on Button/IconButton that also made use of ref will need to adjust to the new 'forwardRef' name

## What
Renames the `ref` prop to `forwardRef` on `CustomButtonProps`.

## Why
Because this prop was titled `ref`, React would throw these warnings.

image


packages/button/src/Button/components/GenericButton.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+import GenericButton, { CustomButtonProps } from "./GenericButton"
+
+it("renders a custom component when the 'component' prop is provided", () => {
+  render(<ExampleComponent />)
+  expect(
+    screen.getByRole("button", { name: "A custom button" })
+  ).toBeInTheDocument()
+})
+
+function ExampleComponent() {
```
HeartSquared, line 12: I think declarations should be made before usage. It's a lot clearer, as opposed to needing to potentially search throughout the file to find it.

I'm also a little unsure about putting the ref into the test and then not actually testing it (I know we shouldn't be testing internal behaviour, so no issues there). It gives a slight impression that it must be declared when using the component.

packages/button/src/Button/components/GenericButton.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const someRef = React.useRef()
+
+  return (
+    <GenericButton
+      label="A custom button"
+      ref={someRef}
+      href="https://cultureamp.design"
```
HeartSquared, line 19: This is a little confusing. Seeing `href` being declared makes me expect an anchor element, but the CustomComponent makes it a button. It's not really something anyone should be doing, so I don't think we should be making an example with.

packages/button/src/Button/components/GenericButton.tsx
```diff
@@ -17,7 +17,7 @@ import styles from "./GenericButton.scss"
 export type CustomButtonProps = {
   id?: string
   className: string
-  ref: Ref<any>
+  forwardRef: Ref<any>
```
HeartSquared, line 20: Hmm... I would expect the fix to actually be using React's `forwardRef` then usage using `ref` as opposed to having a custom prop named `forwardRef`. Is this really the right direction? 😬

packages/button/src/Button/components/GenericButton.spec.tsx
dougmacknz: Gotta say I disagree, I think we should be doing this much more often. IMO the main purpose/function of the file should be at the top, and any supporting private functions should be a secondary focus below that.

packages/button/src/Button/components/GenericButton.spec.tsx
dougmacknz: That's kinda the point here, the custom component's purpose is to override what you expect to be rendered. I think we probably disagree on something here: I don't think tests should be a place for best practice examples, that's what stories and tech docs are for.

Another factor is that the main purpose of this prop is for use with a router library component, which we're never going to pull in just for this. So we're never going to be able to make an example here that makes real world sense.

packages/button/src/Button/components/GenericButton.spec.tsx
HeartSquared: There's actually linting rules that exist for it, as it is a more common practice for reducing confusion (I'm one of them who finds it confusing). It would be much easier to enforce moving forward too
https://eslint.org/docs/latest/rules/no-use-before-define

>In JavaScript, prior to ES6, variable and function declarations are hoisted to the top of a scope, so it’s possible to use identifiers before their formal declarations in code. **This can be confusing and some believe it is best to always declare variables and functions before using them.**

packages/button/src/Button/components/GenericButton.tsx
OliBridgman: It's a similar solve to the problem discussed in the react docs here for Higher-order-components: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components

The Button is a forward ref-able component, but to forward the ref _again_ to a React component, it needs to pass forward in this manner.

packages/button/src/Button/components/GenericButton.spec.tsx
HeartSquared: I think if that's the case, then it would be better to go even more obscure and have the custom component be a whole different element? It took a bit of thinking to figure out why passing in a `href` resulted in a `button` (which is the main element the component renders) without looking at all the other props. As it is, if the `component` prop didn't work and the href check failed, wouldn't this still pass?

packages/button/src/Button/components/GenericButton.tsx
OliBridgman: I'd suggest we change the prop to `forwardedRef` tho :P

OliBridgman:
BUMP

dougmacknz:
Opting to go for a much more simple solution: https://github.com/cultureamp/kaizen-design-system/pull/2997 (solve the problem by deleting it)

PR #2871, opened 2022-07-25
refactor: Kds 433/rename packages scss files

JakePitman:
## Why
- Needed for the Next.js transition

## What
- Renames all `<FILENAME>.scss` files to `<FILENAME>.module.scss` within the packages folder
- Also changes any `styles.scss` format to `ComponentName.module.scss`


packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -8,7 +8,7 @@ import {
   DropdownButton,
   DropdownButtonProps,
 } from "./components"
-import styles from "./SplitButton.scss"
+import styles from "./SplitButton.module.scss"
```
HeartSquared, line 11: You missed the subcomponents 🙃

packages/tabs/src/Tab.tsx
```diff
@@ -2,7 +2,7 @@ import { Tab as ReachTab } from "@reach/tabs"
 import React, { ReactNode, useState, SyntheticEvent } from "react"
 import classnames from "classnames"
 import { Badge } from "@kaizen/draft-badge"
-import styles from "./styles.scss"
+import styles from "./Tabs.module.scss"
```
HeartSquared, line 5: `Tabs.tsx` doesn't actually pull in any styles itself, but the stylesheet infers that the styles are for that component.

See if it's possible to break out the styles into files relating to the components using them.

packages/date-picker/src/_primitives/Calendar/Calendar.module.scss
```diff
@@ -263,7 +263,7 @@ $rdp-cell-size: 40px;
   -moz-appearance: none;
   -webkit-appearance: none;
   appearance: none;
-  position: absolute !important;
+  position: absolute;
```
JakePitman, line 266: This breaks the linter after taking it out of global styles. I removed `!important` and checked it out in storybook locally, and seemed to behave the same without it. `position: absolute` still ends up in the class rule & doesn't seem to get overridden by anything

HeartSquared (approved):

PR #2874, opened 2022-07-26
fix(rte): fix bug where long strings overflow editor

JakePitman:
Fixes an issue where a long, unbroken string will overflow the editor. This fix makes it wrap to a new line.


mcwinter07 (approved):
Looks good to me 🦖 merge in good health

PR #2877, opened 2022-07-27
fix: update control map to filter empty control groups

mcwinter07:
## Why
An empty section wrapper was being created when ever no “ungrouped” controls were provided. This create an unnecessary DOM node and some minor visual regression when resizing the window. 

I've also use this a chance to clean up the code comments and typing for additional clarity.

## What
- Adds a filter function
- Cleaned up code comments and typings


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -89,7 +89,7 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
       <div className={styles.editorWrapper}>
         {controls && (
           <Toolbar aria-controls={editorId} aria-label="Text formatting">
-            {Object.values(controlMap).map((section, index) => (
+            {controlMap.map((section, index) => (
```
mcwinter07, line 92: I've moved this into the controlMap filtering function for a less verbose .map

JakePitman (approved):

dougmacknz (approved):

PR #2878, opened 2022-07-27
chore: address vulnerabilities

HeartSquared:
## What

- Update commitizen
	- adds resolution for commitizen as cz-conventional-changelog still has a dependency on 4.2.4
- Update js audit allow list

## Why

Vulnerabilities should be addressed often. Although @christian314159 will be checking on whether Snyk will replace the need to check these.


mcwinter07 (approved):
Nice work 👍 always happy to see any change that stops the console screaming at us

PR #2879, opened 2022-07-27
feat: Kds 607 avoid fatal error when rte data doesnt match schema

JakePitman:
## Why
Current error state is jarring and confusing to the user.
Currently no ability for consumers to log errors.

## What
- Adds an optional `onDataError` prop that gets fired off when RichTextEditor receives bad data
- Adds an optional `dataError` prop (React Element) that gets displayed instead of the editor, when it receives bad data

### DIFF
--- a/packages/rich-text-editor/docs/dummyMalformedContent.json
+++ b/packages/rich-text-editor/docs/dummyMalformedContent.json
@@ -1,0 +1,12 @@
+[
+  {
+    "type": "plaragraph",
+    "content": [
+      {
+        "type": "text",
+        "marks": [{ "type": "strong" }],
+        "text": "Some notes I'd like to share in bold text:"
+      }
+    ]
+  }
+]
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
@@ -12,6 +12,7 @@ import {
   createLinkManager,
 } from "@cultureamp/rich-text-toolkit"
 import { OverrideClassName } from "@kaizen/component-base"
+import { InlineNotification } from "@kaizen/notification"
 import { ToolbarItems, EditorContentArray, EditorRows } from "../types"
 import { createSchemaFromControls } from "./schema"
 import { buildKeymap } from "./keymap"
@@ -58,7 +59,7 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
     classNameOverride,
     controls,
     rows = 3,
-    dataError = <p>Something went wrong</p>,
+    dataError = "Something went wrong",
     onDataError,
     ...restProps
   } = props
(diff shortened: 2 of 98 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -56,25 +58,41 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
     classNameOverride,
     controls,
     rows = 3,
+    dataError = <p>Something went wrong</p>,
```
JakePitman, line 61: I'm wondering if it makes sense to have a default, and if so - what would be best?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -30,6 +30,8 @@ export interface BaseRichTextEditorProps
    * The editable area will autogrow, so this only affects the component when the content doesn't exceed this height.
    */
   rows?: EditorRows
+  dataError?: React.ReactElement
```
JakePitman, line 33: Would ReactElement be best here? Or ReactNode? Do we want consumers to pass in raw strings, or should we restrict it to elements only?

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -56,25 +58,41 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
...
+              })
+            : null,
+          schema,
+          plugins: getPlugins(controls, schema),
+        }),
+        { "aria-labelledby": labelId, role: "textbox" }
+      )
+    } catch {
+      return new Error("Bad data error")
+    }
```
mcwinter07, line 86: Looks good 👍

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -56,25 +58,41 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
...
+  })()
+
+  if (useRichTextEditorResult instanceof Error) {
+    onDataError && onDataError()
+    return dataError
```
mcwinter07, line 91: We could render the inline notification here and pass in the dataError in as a child

mcwinter07:
Just hand a thought, we might also want to create a story in the RichTextEditor.stories.tsx

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -56,25 +59,45 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
...
+      )
+    } catch {
+      return new Error("Bad data error")
+    }
+  })()
+
+  if (useRichTextEditorResult instanceof Error) {
+    onDataError && onDataError()
+    return (
+      <InlineNotification title="Error" type="negative">
```
mcwinter07, line 93: Wondering if we should throw a `persistent` prop on the notification so it's not dismissible?

packages/rich-text-editor/docs/dummyMalformedContent.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "type": "plaragraph",
```
mcwinter07, line 3: so happy you went with plaragraph in the end 😄

mcwinter07 (approved):
Awesome stuff, the changes look hunky dory!  🦖 merge in good health

PR #2881, opened 2022-07-28
feat: Add custom component render prop to navigation tab of title block

OliBridgman:
## Why
We want to render `Link` components from our routing library for tabs.

## What

Similar to the GenericButton `component` prop, we've added a custom render prop to the  TitleBlock.NavigationTab. Most props are forwarded through the render prop, however, className is calculated by the internals and then forwarded through.

## Notes

We removed the inner div of the navigation tab link. This allowed the classNames to be passed as one string to the resulting custom component. All styles of the div could be safely moved to the anchor tag.

Unlike the GenericButton, the NavigationTab was not a forwardRef component. We did not add that functionality here, but it may be useful to add in the future.

### DIFF
--- a/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/NavigationTabs.tsx
+++ b/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/NavigationTabs.tsx
@@ -18,7 +18,12 @@ export type NavigationTabProps = {
   variant?: Variant
   id?: string
   automationId?: string
-  component?: (props: CustomNavigationTabProps) => JSX.Element
+  /**
+   * Custom render for the tab. Commonly used to replace the link with a router link component.
+   * Props given to the NavigationTab component will be passed back, along with a decorated className.
+   * It is up to you to reapply them to your custom component.
+   */
+  renderTab?: (props: CustomNavigationTabProps) => JSX.Element
 }
 
 const isLight = (variant: Variant | undefined): boolean =>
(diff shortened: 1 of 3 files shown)


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/NavigationTabs.tsx
```diff
@@ -3,6 +3,13 @@ import classnames from "classnames"
 import { NON_REVERSED_VARIANTS, Variant } from "./TitleBlockZen"
 import styles from "./NavigationTabs.module.scss"
 
+export type CustomNavigationTabProps = Omit<
+  NavigationTabProps,
+  "component" | "variant"
+> & {
+  className: string
+}
```
OliBridgman, line 11: What if we ditched the href!!

dougmacknz: Don't you want it for the other side so that you can do `<Link to={props.href} />`?

OliBridgman: Once we're in userland, the user can do what they want. I guess if they're passing the component by reference like in our test they might want it but personally I'd do this:
```
component: (props) => <Link to={IControlMyOwnDestiny}>{AndMyChildrensDestiny}</Link>
```

`className` is almost the _only_ thing we want passed. It's the only thing a custom component shares with a default component.

OliBridgman: Maybe we should use a discriminated union with the component prop and everything else? If you are going to use the renderProp API, you _cannot_ pass the other props.

dougmacknz (approved):

PR #2882, opened 2022-07-28
refactor: add override and change target of css focus

mcwinter07:
## Why
There is a double focus ring on the editable rich text content for keyboard navigation when tabing to inner anchor tags in the rte content.

## What
- useses the  visually hidden classNameOverride to target the sibling element for focus

## Side by side

https://user-images.githubusercontent.com/36558508/181419532-1413da79-86e4-4c2c-820c-2f9a892408b3.mp4

### DIFF
--- a/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
+++ b/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
@@ -19,10 +19,7 @@
   border-color: $color-blue-500;
 }
 
-// Temporary solution for focus styling the content area
-// (the problem with this is that it shows this styling when focusing a link within the content
-.editableContainer:focus-within :global(.ProseMirror) {
-  background-color: $color-gray-200;
+.editableContainer .hiddenButton:focus-within + * > :global(.ProseMirror) {
   border: $border-focus-ring-border-width $border-focus-ring-border-style
     $color-blue-500;
 }
--- a/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
+++ b/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
@@ -46,7 +46,7 @@ export const EditableRichTextContent: React.VFC<
         className={classnames(styles.editableContainer, classNameOverride)}
         {...restProps}
       >
-        <VisuallyHidden>
+        <VisuallyHidden classNameOverride={styles.hiddenButton}>
           <button
             type="button"
             onClick={onClick}


packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
```diff
@@ -46,13 +46,12 @@ export const EditableRichTextContent: React.VFC<
         className={classnames(styles.editableContainer, classNameOverride)}
         {...restProps}
       >
-        <VisuallyHidden>
```
mcwinter07, line 49: I'll also need to remove the import of this if we go this route but holding off until I can get feedback :)

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
```diff
@@ -19,10 +19,17 @@
   border-color: $color-blue-500;
 }
 
-// Temporary solution for focus styling the content area
-// (the problem with this is that it shows this styling when focusing a link within the content
-.editableContainer:focus-within :global(.ProseMirror) {
-  background-color: $color-gray-200;
+.editableContainer .hiddenButton {
```
mcwinter07, line 22: this duplicated the VisuallyHidden component styles

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
```diff
@@ -19,10 +19,17 @@
...
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
+
+.editableContainer .hiddenButton:focus + * > :global(.ProseMirror) {
```
mcwinter07, line 32: This work but keen to hear opinions of this going too far into the route of scss hackery (or sackery for short)

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
dougmacknz: VisuallyHidden doesn't have `classNameOverride` yet but you could add it and then you could still use it?

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
dougmacknz: Oh nice! I forgot that it was the sibling that needed the focus styling rather than the parent.

I'm happy with this.

packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.module.scss
mcwinter07: that's true, would save on the duplication of the hidden styles!

mcwinter07: Do you think it should be in a seperate PR or just bundle this one in?

dougmacknz: You can bundle if you want, but it does mean you're tied to a single release type (patch or minor) for both things

mcwinter07: hmmm does feel like I should break it up into `feat` for the VisuallyHidden and `refactor` for this

dougmacknz (approved):

PR #2886, opened 2022-08-01
feat: add maximize and minimize icons to comp lib

mcwinter07:
## Why
There’s a need by designer Aru in team Apollo for a Maximize icon, as currently there is only a Minimize icon.  This will be used to expand and contract a data table they’re working on, with the impetus to go fullscreen if the user needs.

## What
- adds max and minimize icons


HeartSquared (approved):
The code for the svgs don't have a new line at the end of the file, but maybe it's fine? 🤷🏻

PR #2887, opened 2022-08-01
fix(rte): darken the rte border on hover

JakePitman:
Darkens the RTE's border from gray-500 to gray-600 on hover, including the border between the toolbar and text area

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.module.scss
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.module.scss
@@ -52,6 +52,13 @@
   background: $color-white;
   border-radius: $border-solid-border-radius;
   word-wrap: break-word;
+
+  &:hover {
+    border-color: $color-gray-600;
+    .toolbar {
+        border-color: $color-gray-600
+    }
+  }
 }
 
 @for $i from 1 through 20 {
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.module.scss
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.module.scss
@@ -53,7 +53,9 @@
   border-radius: $border-solid-border-radius;
   word-wrap: break-word;
 
-  &:hover {
+  &:hover,
+  &:focus-within
+  {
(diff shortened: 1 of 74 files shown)


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.module.scss
```diff
@@ -52,6 +52,13 @@
   background: $color-white;
   border-radius: $border-solid-border-radius;
   word-wrap: break-word;
+
+  &:hover {
```
mcwinter07, line 56: Do we want to also add a focus-within on that one too? Depends whether we want to remove the darker for keyboard users or not? @jmitsios might need a quick consult on this one 🙏 
Screen Shot 2022-08-01 at 5 52 00 pm
Screen Shot 2022-08-01 at 5 51 24 pm

jmitsios: @mcwinter07 — Just to make sure I'm following along here, is the question about whether we want all the borders to still darken to Gray-600 for keyboard users, when placing focus on the text input area?  (I'm not too sure what focus-within is, sorry 😞)

mcwinter07: @jmitsios no worries, to clarify - if the user is focused on the input (1st screenshot) should the entire border be Gray-600 like on hover (2nd screenshot). Currently it has the focus ring but not the darkened border

jmitsios: Ahh! I see what you mean now.  I believe that most of our components have their hover states triggered upon focusing, as it's seen as the 'keyboard equivalent' of hovering over it.   So in this instance, it feels like it makes sense to trigger the borders when the focus ring is active.

mcwinter07: Cool beans, thanks for confirming! @JakePitman so we should be able to chuck a `&:focus-within` to action that :)

mcwinter07 (approved):
nice work 🦖 merge in good health

PR #2889, opened 2022-08-02
feat(tooltip): allow consumer to override animation duration through …

JakePitman:
## Why
The RichTextEditor's Toolbar component uses tooltips, but the animation delay causes an issue where, when hovering over multiple icons consecutively, the tooltips pile up and look buggy.

## What
Adds optional `animationDuration` and `animationBuffer` props to Tooltip so that the consumer can control/ remove animation durations

### DIFF
--- a/draft-packages/tooltip/KaizenDraft/Tooltip/AppearanceAnim.tsx
+++ b/draft-packages/tooltip/KaizenDraft/Tooltip/AppearanceAnim.tsx
@@ -6,7 +6,6 @@ type AnimationProps = {
   isAnimOut?: boolean
   prevIsOpen?: boolean
   animationDuration?: number
-  animationBuffer?: number
   isVisible: boolean
 }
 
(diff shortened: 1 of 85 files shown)


dougmacknz:
There's a comment on this `AppearanceAnim.tsx` file about how the duration value needs to stay in sync with the sass variable in `AppearanceAnim.module.scss`, so I don't think this is going to do everything we need it to.

We might not be able to have this completely customisable, you might need to come up with a few presets and use that to adjust the sass side.

draft-packages/tooltip/KaizenDraft/Tooltip/AppearanceAnim.tsx
```diff
@@ -23,7 +25,12 @@ export const AnimationContext = React.createContext<AnimationProps>({
  * When the component is no longer needed, it will no longer be rendered to the
  * dom.
  */
-export const AnimationProvider = ({ isVisible, ...otherProps }) => {
+export const AnimationProvider = ({
+  isVisible,
+  animationDuration = ANIM_DURATION_MS,
+  animationBuffer = ANIM_BUFFER,
+  ...otherProps
+}) => {
```
christian314159, line 33: I saw @dougmacknz had some comments about how this would need to be solved differently, but I'd also expect a test added for this change as this is a functional change to the component (it takes more props now).

dougmacknz (approved):

PR #2890, opened 2022-08-02
fix(date-picker): remove unnecessary :not(.outside) selectors in Calendar styles

HeartSquared:
## Why

When "today" is disabled in the Calendar, the disabled styles (grey font) were not coming through; although they did when the date was focused.

|OLD|NEW|
|---|---|
|image|image|

## What

Removed the `:not(.outside)` selectors which don't appear to do anything except make things overspecific.

### DIFF
--- a/packages/date-picker/docs/Calendar.stories.tsx
+++ b/packages/date-picker/docs/Calendar.stories.tsx
@@ -97,6 +97,12 @@ const StickerSheetCalendarTemplate: Story<{ isReversed: boolean }> = ({
           id="today-selected"
           value={new Date("2022-05-01")}
         />
+        <CalendarExample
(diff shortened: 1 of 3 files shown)


packages/date-picker/docs/Calendar.stories.tsx
```diff
@@ -97,6 +97,12 @@ const StickerSheetCalendarTemplate: Story<{ isReversed: boolean }> = ({
           id="today-selected"
           value={new Date("2022-05-01")}
         />
+        <CalendarExample
```
HeartSquared, line 100: Should we actually add all the combination hover+focus states to the sticker sheet too to capture any possible unintentional overridden styles? Or would it be overkill?

NatalieLoyzaga (approved):

PR #2891, opened 2022-08-03
fix: Styling tweaks to Select component

christian314159:
## Why
I noticed some styling issues with our Select component, see tweaks below.

### 1) Reversed mode: input text is wrong color & caret clashes with text

Before:

image

image

After:

image

image

### 2) Input text in secondary small is too big

Before:

image

After:

image

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.module.scss
+++ b/draft-packages/select/KaizenDraft/Select/Select.module.scss
@@ -141,7 +141,8 @@ $focus-border-color: $color-blue-500;
 
   .input {
     color: $color-purple-800;
-    padding-right: 20px;
+    padding: 0 20px 0 0;
+    margin: 0;
   }
 
   &.secondarySmall {
@@ -338,10 +339,10 @@ $focus-border-color: $color-blue-500;
       top: 0px;
       transform: none;
       max-width: 100%;
-      padding-right: 10px;
+      padding-right: 14px;
     }
     .placeholderOverrides {
-      padding-right: 10px;
+      padding-right: 14px;
     }
     .singleValue {
       max-width: 100%;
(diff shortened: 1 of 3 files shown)


HeartSquared (dismissed):
It's not exactly UI parity (still differs from Figma), but certainly fixes worth having in the meantime I think :)

jmitsios:
@christian314159 Nice pickups and fixes CK.  I had no idea there was a 'secondary small' too - this is the first time learning about it!  We should add it to the UI Kit. Not sure how granular you wanted me to get though, but here's some small issues I found:

**Secondary small**
- Reversed theme doesn't seem to have a hover state:
image
- Default theme has a hover state - but it's showing a lack of proper padding to the left of the text:
image

**Secondary**
- Clicking inside the input field when a selection is made places the text indicator far from the input text:
image

christian314159:
@jasononeil [My search](https://github.com/search?q=org%3Acultureamp+%22secondary-small%22&type=code) within the Culture Amp org only showed this variant referenced and used within our very own Kaizen repo - what about I removed the stories in this PR and then create another card to deprecate it completely?

jmitsios:
> @jmitsios [My search](https://github.com/search?q=org%3Acultureamp+%22secondary-small%22&type=code) within the Culture Amp org only showed this variant referenced and used within our very own Kaizen repo - what about I removed the stories in this PR and then create another card to deprecate it completely?

If it only appears within our Kaizen repo and nowhere else, then it's probably best to create its own card about it and recognise if it needs to be deprecated (once we explore if it's 100% certain that it doesn't exist anywhere else in the product)

christian314159:
fwiw I found out `secondary small` can sometimes be used in the mobile version of the TitleBlock, thus can't be removed.

PR #2892, opened 2022-08-03
feat: Add secondary action prop

reetaa:
## Why
In `InputEditModal` modal - we currently have 3 buttons.
1) A primary button to Submit action.
2) A secondary button that is attached to Dismiss action.
3) A close button that is attached to Dismiss action.

We in Polaris have a product need to separate these 3 buttons with 3 distinct actions. 

## What
* When two props `onSecondaryAction` and `secondaryLabel` are supplied, secondary action feature is available in the secondary button.
* Otherwise, the modal works as usual.

## Image of the modal with Secondary and Primary buttons
Screen Shot 2022-08-04 at 3 33 47 pm

## Notes
* Gearing towards minimal change and only `minor` version upgrade.
* So, not touching other prop names.
* I'd like to avoid major package upgrade.
* Have done design review with designer in #community_amplify_1on1

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.spec.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.spec.tsx
@@ -85,7 +85,7 @@ describe("<InputEditModal />", () => {
         onSubmit={handleSubmit}
         onDismiss={handleDismiss}
         onSecondaryAction={handleSecondaryAction}
-        dismissLabel="Secondary button"
+        secondaryLabel="Secondary button"
       >
         Example modal body
       </InputEditModalWrapper>
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.spec.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.spec.tsx
@@ -75,7 +75,7 @@ describe("<InputEditModal />", () => {
     expect(handleDismiss).toHaveBeenCalledTimes(0)
   })
 
-  it("supports a secondary action when secondary button is pressed", () => {
+  it("supports a secondary action when secondary-action and secondary-label both props are provided", () => {
     const handleSubmit = jest.fn()
     const handleDismiss = jest.fn()
     const handleSecondaryAction = jest.fn()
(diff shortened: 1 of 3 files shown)


mcwinter07 (commented):
From a perspective of increasing this component utility I think this additional prop could be pretty handy. Not sure if this is tied into the same work but we did also have a question recently about this in our #prod_design_system channel.

I'm kind of curious about the naming of the secondary action as `onConfirm`. It does seem a like a very specific event name, which granted we may want to be, but just wanted to hear any underlying reason / make sure that there is enough delineation between that and onSubmit event - it could potentially benefit from something a little more generic like onSecondaryAction? 

Also just wanted to call out the [docs on InputModal](https://cultureamp.design/components/modal/#inputedit-modal) describes the secondary action as "destructive" so might be worth just checking that this is either a pattern it aligns to, or that we are ok with the new prop slightly changing the intention of this type of modal (might want to rope in our designer @jmitsios to confer).

mcwinter07:
> 

@reetaa sorry I just re-read [the thread](https://cultureamp.slack.com/archives/CRWGPPFR9/p1659514212490839) we were tagged in 😆 and have got a little more context on on the `onConfirm` naming convention. I'm still inclined to lean toward `onSecondaryAction` even though it match as nicely with `onSubmit`, but agree that we should try to release this as a minor (maybe change it to onPrimaryAction after as a major?) @dougmacknz keen to hear your thoughts too :)

reetaa:
Yeah, the question was asked from #community_amplify_1on1 [in this thread here](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1658300726619619) which is why I tagged you in this PR @mcwinter07. 
Added the UI in the description as you suggested.

I read the following from the design doc.
> All modals use a single Primary Action Button. Never two primary action buttons. If a second action is an option, use a Secondary Action button (e.g. Cancel).

In this usage for 1-1s - they are not two primary actions but primary and secondary actions. Hence, I believe we are still within the realm of design system guidelines.

About `Destructive action` - I read in the design doc that it's referred as `Primary/Destructive action`. So, the wording could be clearer since we leave the decision to the consumers what's destructive. As design system - our guidance should be sufficient from Primary and Secondary wording, in my opinion. Would love to hear DST thoughts.

draft-packages/modal/KaizenDraft/Modal/Presets/InputEditModal.spec.tsx
```diff
@@ -74,4 +74,25 @@ describe("<InputEditModal />", () => {
...
+    const handleSubmit = jest.fn()
+    const handleDismiss = jest.fn()
+    const handleSecondaryAction = jest.fn()
+
+    const { getByText } = render(
+      <InputEditModalWrapper
+        onSubmit={handleSubmit}
+        onDismiss={handleDismiss}
+        onSecondaryAction={handleSecondaryAction}
+        dismissLabel="Secondary button"
```
dougmacknz, line 88: I think it would be a good idea to create a new label here specific to the secondary action, 

I'm kinda surprised to see that the cross button in the top right isn't using the `dismissLabel` passed in, I think it probably should be.

Either way, having to use the `dismissLabel` prop for the secondary button label is going to be quite confusing.

reetaa: Yeah, makes sense, changes made. This will need a cleanup done afterwards. That'd be a major release, otherwise we'll end up with too many props doing similar things.

draft-packages/modal/docs/InputEditModal.stories.tsx
```diff
@@ -98,7 +108,8 @@ InputEditModalExample.storyName = "Input Edit Modal"
 InputEditModalExample.args = {
   mood: "positive",
   title: "Input-edit modal title",
-  submitLabel: "Submit label",
+  submitLabel: "Primary label",
+  secondaryLabel: "Secondary label",
```
dougmacknz, line 112: Given this is more of an edge case and we only have one story for this modal, I'm thinking this story shouldn't have the secondary action by default.

dougmacknz (approved):

dougmacknz (approved):

PR #2893, opened 2022-08-04
docs(date-picker): fix validation story

HeartSquared:
## What

Removes the spread props for the validation story because the controls have been removed.

## Why

Accidentally broke the story when updating DatePicker stories.


NatalieLoyzaga (approved):
The storybook looks good!

PR #2896, opened 2022-08-04
fix: Kds 651/remove tooltip animation delay in rte

JakePitman:
## Why
Lingering tooltips obstructs view of toolbar buttons, and appears buggy to users

## What
- Bump @kaizen/draft-tooltip
- Set `animationDuration` to `0` to make tooltips disappear immediately upon mouse-off event


christian314159 (approved):
The 🐍 is gone!

PR #2897, opened 2022-08-08
feat(date-picker): validate input on pressing Enter key

HeartSquared:
## What

Validate input when the user presses the Enter key.
Updated the Validation story to use a `<form>`.

## Why

Pressing Enter submits the form (if it exists), but we did not trigger the validation when it happens, which lead to an inaccurate error message (lagging one step behind).


NatalieLoyzaga (approved):
Looks good!

PR #2898, opened 2022-08-08
refactor(DatePicker): split out validation and use validateDate util

NatalieLoyzaga:
## Why

Refactor the handle functions to utilise new `validateDate` util.

## What

- Create new `validateDate` util with tests
- Use new util when can
- Update `formatNumeral` to return empty string rather than "Invalid Date".


packages/date-picker/src/utils/validateDate.ts
```diff
@@ -0,0 +1,69 @@
...
+      validationResponse: {
+        ...baseResponse,
+        status: "error",
+        validationMessage: `${newInputValue} is not available, try another date`,
+        isDisabled: true,
+      },
+      newDate: undefined,
+    }
+  }
+  return {
```
HeartSquared, line 62: Nitpick: Empty line above here would be less confusing.

HeartSquared (approved):
Looks great 💃🏻 Only one comment, but it's just an empty line 🤷🏻

PR #2899, opened 2022-08-08
fix(multi-filter-select): add padding to control buttons

JakePitman:
Adds padding to control buttons, as per [the figma designs](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%9A%A7-UI-Kit%3A-Beta-Components?node-id=24636%3A129157)

### DIFF
--- a/packages/select/src/FilterMultiSelect/components/SelectionControlButton/SelectionControlButton.scss
+++ b/packages/select/src/FilterMultiSelect/components/SelectionControlButton/SelectionControlButton.scss
@@ -22,7 +22,8 @@
   border-color: $border-borderless-border-color;
   border-radius: $border-solid-border-radius;
 
-  padding: 7px 11px;
+  //       7px       11px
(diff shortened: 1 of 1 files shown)


packages/select/src/FilterMultiSelect/components/SelectionControlButton/SelectionControlButton.scss
```diff
@@ -22,7 +22,7 @@
   border-color: $border-borderless-border-color;
   border-radius: $border-solid-border-radius;
 
-  padding: 0;
+  padding: 7px 11px;
```
HeartSquared, line 25: Is there a way to use our design tokens? If not (which is a different problem), this should probably still be in `rem`.

HeartSquared (approved):

HeartSquared (approved):

HeartSquared (approved):

PR #2902, opened 2022-08-08
test(SelectionProvider): fix several problematic waitFor

louiseliang:
## Why

[KDS-699](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-699)

Some strange behaviours with tests in SelectionProvider is reported.
1. `console.error` for overlapping act
2. `expect()` under `waitFor` clause without `await`  passes regardless it is truth or false

## What

Reasons:

- In the test, `selectedKeys = new Set(["option-2"])` is passed as a props to `SelectionProvider`, which make it controlled. Clear Button on its own won't change the passed static value `new Set(["option-2"])` so hence the test should fail.
- everything under `waitFor` without `await` will pass, which masks the failed tests.

Solution:

- hook up whole controlled system with `selectedKeys` and `onSelectionChange` with external state to make it work
- remove all `waitFor`


HeartSquared (dismissed):
Nice fixes!

Although there are a few other things I think may need some updates (happy for it to be done here or separately, up to you):
- Many of the `expects` do not actually have an assertion behind it (eg. L70)
	- While I know that the `getByRole` will fail if it does not find the element, for readability, we should always be writing a full assertion
- Unused variable `listBox` in L296
- Unused variable `option1` in L370

HeartSquared (approved):

PR #2903, opened 2022-08-09
feat(date-picker): render inbuilt validation when props not set

NatalieLoyzaga:
## Why

- Improve developer experience

## What

- Allow `status`, `validationMessage`, and `onValidate` to be optional and use inbuilt validation.

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -264,18 +264,18 @@ describe("<DatePicker /> - Input format", () => {
 })
 
 describe("<DatePicker /> - Validation", () => {
-  it("displays the message when status is error", () => {
-    render(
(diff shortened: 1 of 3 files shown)


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -282,22 +264,96 @@ describe("<DatePicker /> - Input format", () => {
 })
 
 describe("<DatePicker /> - Validation", () => {
+  it("displays the message when status is error", () => {
```
HeartSquared, line 267:
```suggestion
  it("displays custom validation message when provided (overrides inbuilt validation)", () => {
```

Hmm... actually we may need to adjust the props to make it meet an inbuilt validation error, which we ensure isn't shown.

HeartSquared (approved):
Love it :)

PR #2904, opened 2022-08-09
docs(filter-menu-button): deprecate filter menu button

JakePitman:
Deprecate FilterMenuButton in favour of FilterMultiSelect


NatalieLoyzaga (commented):
You can also remove this parameter so that the stories aren't being run in chromatic anymore `{ chromatic: { disable: false } }`

NatalieLoyzaga (approved):

PR #2906, opened 2022-08-10
feat(date-picker): update DatePicker description from "Format" to "Input format"

HeartSquared:
## What

Update embedded description from "Format" to "Input format"

## Why

Design request for clarity (as the format is only accurate when focused in the input)


dougmacknz (approved):
LGTM

PR #2908, opened 2022-08-11
fix: use max-width on label

aaronknol:
## Why
https://cultureamp.atlassian.net/browse/KDS-716

The `label` inside a `RadioGroup` is given a width of 400px, which could cause problems on small screens and increased font-size.

## What
As suggested by @christian314159 using a `max-width` instead of `width` is the best solution to prevent any scenarios that do require the width


mcwinter07 (approved):
Looks good to me 👍 Thanks for the contribution

PR #2912, opened 2022-08-15
fix(filter-multi-select): use spacing-xs between children and chevron

JakePitman:
## What
On the FilterMultiSelect's trigger button, changes spacing between children and the chevron from 12px to $spacing-xs

## Why
Spacing was previously 12px, which is a larger value than the figma design

### Before
image

### After
image

### DIFF
--- a/packages/select/src/FilterMultiSelect/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
+++ b/packages/select/src/FilterMultiSelect/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
@@ -24,10 +24,12 @@ export const TriggerButtonBase: React.VFC<TriggerButtonBaseProps> = ({
       ref={buttonRef}
       className={classNames(styles.button, classNameOverride)}
     >
-      <span>{children}</span>
+      {children}
+
       <Icon
         icon={menuTriggerState.isOpen ? chevronUp : chevronDown}
         role="presentation"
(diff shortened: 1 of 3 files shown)


jmitsios:
@JakePitman Looks excellent to me!

jmitsios (dismissed):

packages/select/src/FilterMultiSelect/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
```diff
@@ -24,11 +24,14 @@ export const TriggerButtonBase: React.VFC<TriggerButtonBaseProps> = ({
       ref={buttonRef}
       className={classNames(styles.button, classNameOverride)}
     >
-      <span>{children}</span>
-      <Icon
-        icon={menuTriggerState.isOpen ? chevronUp : chevronDown}
-        role="presentation"
-      />
+      <span className={styles.innerContentContainer}>
```
HeartSquared, line 27: Could we get away with not adding another container? Eg. Putting the styles on the button element instead.

JakePitman: Actually yes - turns out it was just the `gap` rule. Thanks

HeartSquared (approved):
Looks fine to me :)

PR #2915, opened 2022-08-15
fix: FilterMultiSelect stories - FilterBar 

louiseliang:
## Why
A bug was raised that when having multiple filter category selected, removing the first category will clear the second one as well.

## What
The proposed solution:

- remove the internal state within the [DemographicValueSelect.tsx](https://github.com/cultureamp/kaizen-design-system/compare/ll/fix-filter-multi-select-stories?expand=1#diff-fb7011fd06c523f4db5afb33622454f35bbb27efdd224b70c16c68b94f9a800d) 
- allow its value to be controlled externally

## Result

https://user-images.githubusercontent.com/8596435/184593993-f5d23529-06d2-455b-9312-f3dad54b8fef.mov

### DIFF
--- a/packages/select/docs/FilterMultiSelect.stories.tsx
+++ b/packages/select/docs/FilterMultiSelect.stories.tsx
@@ -1,4 +1,4 @@
-import React, { Key, useState } from "react"
+import React, { useState } from "react"
 import { ComponentMeta } from "@storybook/react"
 import { withDesign } from "storybook-addon-designs"
 import { Selection } from "@react-types/shared"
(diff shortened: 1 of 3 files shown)


packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { Key, useState } from "react"
```
louiseliang, line 1:
```suggestion
import React, { useState } from "react"
```

dougmacknz (approved):

PR #2917, opened 2022-08-16
feat: Preload fonts in ThemeProvider & handle default theme

hqta:
## Why
In `unified-home` and `unified-navigation` we preload the font resources and create a Theme Manager using the default theme. This is repeated across both these repos. This PR takes in these commonalities and incorporates it within Kaizen `ThemeProvider`

Thread: https://cultureamp.slack.com/archives/C0189KBPM4Y/p1660535756937689

### DIFF
--- a/packages/design-tokens/react/ThemeProvider.tsx
+++ b/packages/design-tokens/react/ThemeProvider.tsx
@@ -68,13 +68,6 @@ export const ThemeProvider = ({
       />
       <link
         rel="preload"
-        href="https://d1e7r7b0lb8p4d.cloudfront.net/fonts/inter/inter-light.woff2"
-        as="font"
-        type="font/woff2"
-        crossOrigin="crossorigin"
-      />
-      <link
-        rel="preload"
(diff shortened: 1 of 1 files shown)


packages/design-tokens/react/ThemeProvider.tsx
```diff
@@ -11,28 +11,69 @@ export const ThemeProvider = ({
...
+      <link
+        rel="preload"
+        href="https://d1e7r7b0lb8p4d.cloudfront.net/fonts/inter/inter-bold.woff2"
+        as="font"
+        type="font/woff2"
+        crossOrigin="crossorigin"
+      />
+      <link
+        rel="preload"
+        href="https://d1e7r7b0lb8p4d.cloudfront.net/fonts/inter/inter-demi-bold.woff2"
```
dougmacknz, line 50: Wonder if this was meant to be `semi-bold` instead of `demi-bold` when it was uploaded. Ah well 🤷 lol

dougmacknz (approved):

PR #2918, opened 2022-08-16
fix: Kds 720 filter multi select padding issues

JakePitman:
## Why
Styling issues were found in the FilterMultiSelect, as detailed with snapshots in [this jira ticket](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-720)

## What
- Center aligns multi select options vertically
- Makes badge background colour white when option is in hover state
- Removes the margin that separates multi select options from the scrollbar when there is no scrollbar
- Fixes issue where selection control buttons can be both focused and disabled at the same time. Disabled buttons will no longer recieve blue backgrounds
- Replaces `border` focus styles with focus ring (adds a 1px gap between button and ring)


jmitsios:
@JakePitman Great updates!  Only a couple minor things:

- The **number badge** doesn't look vertically centered within the hover container
image
- When there's longer labels within the item name, the checkbox and badge need to remain at the top:
**Storybook:**
image
**UI Kit design:**
image
- Button-link's padding are still not consistent with the design:
**Storybook:**
image
**UI Kit design:**
image
- We'll need to update the visuals of the Disabled 'Focus' state to match a pattern we've been using recently _(We can chat about this on the side!)_
**Storybook:**
image
**UI Kit Example from our Datepicker**
image

JakePitman:
> @JakePitman Great updates! Only a couple minor things:
> 
>     * The **number badge** doesn't look vertically centered within the hover container
>       image
> 
>     * When there's longer labels within the item name, the checkbox and badge need to remain at the top:
>       **Storybook:**
>       image
>       **UI Kit design:**
>       image
> 
>     * Button-link's padding are still not consistent with the design:
>       **Storybook:**
>       image
>       **UI Kit design:**
>       image
> 
>     * We'll need to update the visuals of the Disabled 'Focus' state to match a pattern we've been using recently _(We can chat about this on the side!)_
>       **Storybook:**
>       image
>       **UI Kit Example from our Datepicker**
>       image

👍

christian314159 (approved):

PR #2920, opened 2022-08-17
fix: Kds 618 select component missing focus states

JakePitman:
## Why
Select control and option components don't currently receive focus rings
image

## What
Adds focus rings to control and option components on focus/hover
image


draft-packages/select/KaizenDraft/Select/Select.module.scss
```diff
@@ -101,11 +107,16 @@ $focus-border-color: $color-blue-500;
     @include base-font-style();
     border-radius: 4px;
```
dougmacknz, line 108: There's a border radius property here you can get rid of

NatalieLoyzaga (approved):
Looks great!

PR #2921, opened 2022-08-18
fix: Splitbutton flaky test on chromatic

NatalieLoyzaga:
## Why

The split button's menu causes flaky chromatic tests when the position is a few pixels off.

## What

- Wrap the chromatic rows in a chromatic ignore

### DIFF
--- a/packages/split-button/docs/SplitButton.stories.tsx
+++ b/packages/split-button/docs/SplitButton.stories.tsx
@@ -28,25 +28,17 @@ const WrapperChromaticIgnore: React.VFC<{ children: React.ReactNode }> = ({
 }) => <div data-chromatic="ignore">{children}</div>
 
 const DROPDOWN_CONTENT__ENABLED = (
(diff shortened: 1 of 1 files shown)


packages/split-button/docs/SplitButton.stories.tsx
```diff
@@ -193,7 +181,7 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
       </StoryWrapper>
 
       {IS_CHROMATIC && (
```
HeartSquared, line 183: You know what, since it looks like we only wanted it in Chromatic, and now we're saying to ignore it, you might as well just plain remove it.

I thought it was still going to show in Storybook, but I guess it didn't have that benefit.

NatalieLoyzaga: I was thinking this too - I can remove it

HeartSquared (approved):
Thanks Nat :)

PR #2923, opened 2022-08-18
build: update tsconfig and clean up package.json files for component packages

HeartSquared:
## What

Update tsconfig and clean up package.json `files` for all component packages. Packages with pre-existing custom config were not touched (eg. `design-tokens`).

Update `!docs` in `package.json` to `!docs/**/*` to actually exclude the folder (I found it often did not work prior).

## Why

[KDS-627](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-627)

Optimise the release process by updating the tsconfig to only compile files within its own package. Previously it would compile all packages (including draft-packages) each time it wanted to compile one.

Removed unnecessary globs in the package.json `files` attribute (eg. `stories`, `draft-templates`).

## Note

I did realise that some of the globs within the package.jsons did not actually do what was expected. In most cases, `"!**/*.snap"` (which is intended to exclude snapshot files) does not work. Many also do not exclude the `.ts(x)` files as they should. But we will be revisiting what we want our distributions should look like, so it has been omitted here.


draft-packages/well/tsconfig.dist.json
```diff
@@ -1,10 +1,5 @@
 {
   "extends": "../../tsconfig.dist.json",
-  "compilerOptions": {
-    "allowJs": false,
-    "declaration": true,
-    "noEmit": false,
-    "sourceMap": true
-  },
-  "exclude": ["docs"]
+  "include": ["index.ts", "*.tsx"],
```
HeartSquared, line 3: This one is different because the component files are put in the root and not within a folder like every other package 😬

packages/component-base/package.json
```diff
@@ -21,8 +21,6 @@
     "!**/*.ts",
     "!**/*.tsx",
     "**/*.d.ts",
-    "!**/*.spec.*",
-    "!docs",
```
HeartSquared, line 25: I didn't leave omitting the `docs` folder from this because `component-base` is not really a component. But I can add it back in if anyone sees it might happen.

NatalieLoyzaga (approved):
Nice work!

PR #2924, opened 2022-08-18
feat: remove deprecated styles from checkbox

NatalieLoyzaga:
## Why

- Remove deprecated styles and mixins

## What

- Remove deprecated styles and mixins from checkboxField, Checkbox and checkboxGroup
- Update stories to use stickersheets


christian314159 (approved):
👍

PR #2926, opened 2022-08-19
chore(lerna): update to ignore stories/test files/readmes on release

HeartSquared:
## What

Tell lerna not to publish a package when we only update stories/test files/README.

## Why

Not necessary to bump versions for these changes.


christian314159 (approved):
Excited to see less unintended version bumps!

PR #2929, opened 2022-08-23
fix: add margin fix for list children selector

mcwinter07:
## Why
A single list in the RTE content has too much spacing because the first and last children do not inherit the margin resets:

image

## Solution walk through
https://user-images.githubusercontent.com/36558508/186082370-4c95c11b-adea-4572-ae16-2ed979b8c395.mp4

## What
- updates the content styles mixin to use wild cards instead of the p tag selector
- adds selector traverse into the list elements


packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
```diff
@@ -8,6 +8,7 @@ import { Button } from "@kaizen/button"
 import { Box } from "@kaizen/component-library"
 import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
 import dummyContent from "./dummyContent.json"
+import dummyLists from "./dummyLists.json"
```
mcwinter07, line 11: I'll remove these changes, but just wanted to give a preview of the problem

packages/rich-text-editor/docs/dummyLists.json
```diff
@@ -0,0 +1,44 @@
+[
```
mcwinter07, line 1: This will get removed

packages/rich-text-editor/src/_mixins.scss
```diff
@@ -14,8 +14,22 @@
...
+  }
+
+  > *:first-child {
+    margin-top: 0;
+    > li:first-of-type {
+      p {
+        margin-top: 0;
+      }
+    }
   }
```
mcwinter07, line 33: for context to use the same selector without the wildcard it would look approximately like this:
```css
  > ul:last-child,
  > ol:last-child {
    margin-bottom: 0;
    > li:last-of-type {
      p {
        margin-bottom: 0;
      }
    }
  }

  > ul:first-child,
  > ol:first-child {
    margin-top: 0;
    > li:first-of-type {
      p {
        margin-top: 0;
      }
    }
  }
  
  ```

dougmacknz (approved):
Nice!

dougmacknz (approved):

PR #2930, opened 2022-08-24
fix: KDS-737 Only apply 'select all' and 'clear' to filtered items when using the search

JakePitman:
## Why
"Select all" and "Clear" buttons are currently scoped to affect all options, instead of just those that are filtered.

## What
Scopes "Select all" and "Clear" button click behaviour to affect only filtered options.

- When options have been filtered, "Select all" will
  - Select all filtered options
  - Not select unselected, un-filtered options
  - Not clear selected un-filtered options
  
- When options have been filtered, "Clear will
  - Deselect all filtered options
  - Not deselect selected unfiltered options


dougmacknz (approved):
Just have a peak at Slack for feedback about the PR title, otherwise looking good 👍

PR #2934, opened 2022-08-30
docs: Kds 755 - Fix outdated Paragraph stories issue

JakePitman:
## Why
Current docs suggest that padding props are available, when they aren't

## What
- Removes invalid Paragraph stories
- Adds a default Paragraph story with controls
- Adds a sticker sheet (+ reversed) displaying different variants of Paragraph


NatalieLoyzaga (approved):
Nice

NatalieLoyzaga (approved):

PR #2935, opened 2022-08-30
fix: Remove unused %caIconBlockLayout scss var

JakePitman:
## Why
This class rule is unused, and may lead to confusion around whether or not a prop should be available to turn the Icon into a block layout. It is expected that this should be achieved through classNameOverride instead, so removing this should make that less confusing.

## What
Removes an unused css rule that applies `display: block` to the Icon component.

## Breaking change
If importing this variable in your own repo, please replace all instances of `%caIconBlockLayout` with your own `display: block` css rule.

### DIFF
--- a/packages/component-library/components/Icon/Icon.module.scss
+++ b/packages/component-library/components/Icon/Icon.module.scss
@@ -25,10 +25,6 @@
   @extend %caIconInheritSize;
 }
 
-.blockLayout {
-  @extend %caIconBlockLayout;
-}
-
 .interactiveIconWrapper {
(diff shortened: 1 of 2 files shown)


dougmacknz:
Just be aware that Murmur has somewhere that is using this. Really easy fix for them though to just replace that with a `display: block` property.
image

JakePitman:
Yeah good point - it looks like it's used in [other repos](too). Do all of these need to be updated to `display: block` before this can get merged?

dougmacknz (approved):

PR #2939, opened 2022-09-01
(feat): Add control props to FlterMultiSelect Menu Trigger

OliBridgman:
## Why
Allow a consumer to control the menu state of a FilterMultiSelect.

## What
`useMenuTriggerState` allows for two more props, `isOpen` and
`onOpenChange`. Add these props to the FilterMultiSelect and pass
them down to the MenuTriggerProvider. This allows a consumer to
control the menu display via their own state and callbacks.


packages/select/src/FilterMultiSelect/components/Root/Root.tsx
```diff
@@ -22,7 +22,9 @@ export interface RootProps
 }
 
 interface MenuTriggerProps {
+  isOpen?: boolean
   defaultOpen?: boolean
+  onOpenChange?: (isOpen: boolean) => void
```
HeartSquared, line 27: Suggestion: This interface could be exported and extended by `MenuTriggerProviderProps` for consistency (eg. if one of the prop's type changes)

packages/select/src/FilterMultiSelect/provider/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
```diff
@@ -10,9 +10,15 @@ import {
...
 const MenuTriggerProviderWrapper = ({
+  isOpen,
   defaultOpen,
+  onOpenChange,
 }: Partial<MenuTriggerProviderProps>) => (
-  <MenuTriggerProvider defaultOpen={defaultOpen}>
+  <MenuTriggerProvider
+    defaultOpen={defaultOpen}
+    isOpen={isOpen}
+    onOpenChange={onOpenChange}
```
HeartSquared, line 20: If you want to save some lines

```suggestion
(props: Partial<MenuTriggerProviderProps>) => (
  <MenuTriggerProvider {...props} />
```

packages/select/src/FilterMultiSelect/provider/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
```diff
@@ -30,24 +36,51 @@ describe("<MenuTriggerProvider /> - Visual content", () => {
...
+
+    it("shows the menu when defaultOpen is set to true", () => {
+      render(<MenuTriggerProviderWrapper defaultOpen />)
+      const menu = screen.queryByText("menu-content-mock")
+      expect(menu).toBeVisible()
+    })
+  })
+  describe("when controlled", () => {
+    it("shows the menu based on the isOpen prop", () => {
+      const { rerender } = render(<MenuTriggerProviderWrapper isOpen={true} />)
```
HeartSquared, line 61:
```suggestion
      const { rerender } = render(<MenuTriggerProviderWrapper isOpen />)
```

packages/select/src/FilterMultiSelect/provider/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
```diff
@@ -30,24 +36,51 @@ describe("<MenuTriggerProvider /> - Visual content", () => {
...
+
+      const menu = screen.queryByText("menu-content-mock")
+      expect(menu).toBeVisible()
+
+      rerender(<MenuTriggerProviderWrapper isOpen={false} />)
+      expect(screen.queryByText("menu-content-mock")).not.toBeInTheDocument()
+    })
+
+    it("fires the onOpenChange callback when the trigger is interacted", () => {
+      const onOpenChange = jest.fn()
```
HeartSquared, line 71:
```suggestion
      const onOpenChange = jest.fn<void, [boolean]>()
```

HeartSquared (commented):
Implementation looks fine. Just left a bunch of suggestions for the code, but they can be considered nitpicking.

It appears that the controls on Storybook do not work though, so that would be good to fix.

OliBridgman:
I had some strange issues with this after rebasing onto master, a lotta build files around and unrelated tests failing. Anyway, I deleted the whole repo and re-installed seemed to fix it.

OliBridgman:
@HeartSquared Have updated based on your feedback. A few more potential weird things:

* A lotta interfaces over types, which means if we wanna share the MenuTriggerProps _I think_ I needed to create an empty interface which is not allowed by a random eslint rule somewhere.
* Deep down in the depths we end up relying on `useControlledState` from react-stately. I think because `defaultOpen` is always set to false (rather than undefined) in the Root props, we need to be specific when passing `isOpen` boolean props (so preferring false over not passing). I'm not sure if you want to make this a more robust API on the `Root` component, forcing the consumer to make the call upfront about controlled/uncontrolled, and setting `defaultOpen` _after_ we know that decision.

HeartSquared:
>A lotta interfaces over types, which means if we wanna share the MenuTriggerProps I think I needed to create an empty interface which is not allowed by a random eslint rule somewhere.

Yeah I recall encountering that eslint rule. Just let it do what it wants. We tend to use interface because the design system should be extendable (there's only minor differences between interface/type that do not matter most of the time).

>Deep down in the depths we end up relying on useControlledState from react-stately. I think because defaultOpen is always set to false (rather than undefined) in the Root props, we need to be specific when passing isOpen boolean props (so preferring false over not passing). I'm not sure if you want to make this a more robust API on the Root component, forcing the consumer to make the call upfront about controlled/uncontrolled, and setting defaultOpen after we know that decision.

I'm not too familiar with the api, so I'm not 100% sure what you mean about this. Does it conflict when one is `false` while the other is `undefined`? What are the changes you suggest to make it more robust?

packages/select/src/FilterMultiSelect/components/Root/Root.tsx
```diff
@@ -48,9 +47,9 @@ export type FilterMultiSelectProps = RootProps
 export const Root: React.VFC<RootProps> = ({
   trigger,
   children,
-
-  defaultOpen = false,
-
+  isOpen,
+  defaultOpen,
```
HeartSquared, line 51: Had a discussion with @OliBridgman and now that we allow the controlled option, we should remove the default value for `defaultOpen` since the internals in react-aria treats `false` and `undefined` differently (`defaultOpen` should only be defined if it's uncontrolled, but having it always defined could potentially override `isOpen`'s `false` if it were defined even though it should be controlled).

HeartSquared (approved):

PR #2940, opened 2022-09-01
feat: Kds 656 implement smarts to avoid really long select buttons

NatalieLoyzaga:
## Why

- truncate label options based on character limit rather than options length.

## What

- Add optional prop to the Trigger Button to accept char limit
- Update truncation function to recursively check against char limit
- Add story to show the new functionality


packages/select/src/FilterMultiSelect/utils/getTruncatedLabel/getTruncatedLabel.ts
```diff
@@ -1,13 +1,22 @@
...
+  labelsTruncatedCount = 0
+): string => {
+  const moreIndicator =
+    labelsTruncatedCount > 0 ? ` + ${labelsTruncatedCount} more` : ""
+  if (labels.length === 1) {
+    return `${labels.join(CONNECTOR)}${moreIndicator}`
+  }
+  if (labels.join("").length < limit) {
+    return `${labels.join(CONNECTOR)}${moreIndicator}`
   }
```
HeartSquared, line 15:
```suggestion
  if (labels.length === 1 || labels.join("").length < limit) {
    const moreIndicator =
      labelsTruncatedCount > 0 ? ` + ${labelsTruncatedCount} more` : ""
    return `${labels.join(CONNECTOR)}${moreIndicator}`
  }
```

packages/select/src/FilterMultiSelect/utils/getTruncatedLabel/getTruncatedLabel.ts
```diff
@@ -1,13 +1,22 @@
...
+  labels: string[],
+  limit: number,
+  labelsTruncatedCount = 0
+): string => {
+  const moreIndicator =
+    labelsTruncatedCount > 0 ? ` + ${labelsTruncatedCount} more` : ""
+  if (labels.length === 1) {
+    return `${labels.join(CONNECTOR)}${moreIndicator}`
+  }
+  if (labels.join("").length < limit) {
```
HeartSquared, line 13: We definitely don't want to count the `, ` as part of the length?

packages/select/src/FilterMultiSelect/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
```diff
@@ -7,12 +7,14 @@ export type FilterTriggerButtonProps = {
   label: string
   selectedOptionLabels: string[]
   classNameOverride?: string // TODO: migrate it to use OverrideClassName<T> and omit the props controlled by React-Aria
+  characterLimit?: number
```
HeartSquared, line 10: Hmm...given this prop name, I would expect it to cap the label length, however the functionality would actually still display a full length long string if it were the first option to display.

Perhaps clarify with @jmitsios if we want to actually truncate the first label if it exceeds the character limit? If not, we should try and think of a better name. 

It could benefit from a description regardless too :)

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -69,6 +70,63 @@ export const DefaultKaizenSiteDemo = args => {
 
 DefaultKaizenSiteDemo.storyName = "Default (Kaizen Site Demo)"
 
+export const TruncatedLabels = args => {
```
HeartSquared, line 73: Do you want this story to have controls? If so, you would want to spread the `args` into the story somewhere. Otherwise, disable the controls and change `args` to `()`.

Bonus points if you type the story too with either `Story` or `ComponentStory` :)

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -69,6 +70,63 @@ export const DefaultKaizenSiteDemo = args => {
...
+  const [selectedKeys, setSelectedKeys] = useState<Selection>(
+    new Set(["id-fe"])
+  )
+  const [characterLimit, setCharacterLimit] = useState<number>(100)
+
+  const handleSelectionChange = (keys: Selection) => {
+    keys && setSelectedKeys(keys)
+  }
```
HeartSquared, line 81:
```suggestion
  const handleSelectionChange = (keys: Selection) => setSelectedKeys(keys)
```

Based on your params, `keys` will always be defined

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -69,6 +70,63 @@ export const DefaultKaizenSiteDemo = args => {
...
+
+  const handleCharacterLimitChange = e => {
```
HeartSquared, line 83: Type please :)

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -69,6 +70,63 @@ export const DefaultKaizenSiteDemo = args => {
...
+      <div style={{ marginBottom: "3rem" }}>
+        <Label
+          labelText="Update character limit: "
+          style={{ marginRight: "1rem" }}
+        />
+        <input
+          type="number"
+          onChange={handleCharacterLimitChange}
+          value={characterLimit}
+        ></input>
```
HeartSquared, line 98:
```suggestion
        />
```

packages/select/src/FilterMultiSelect/utils/getTruncatedLabel/getTruncatedLabel.ts
```diff
@@ -1,13 +1,22 @@
...
   }
 
-  return `${[...labels].splice(0, LABELS_TRUNCATED_LIMIT).join(CONNECTOR)} +${
-    labels.length - LABELS_TRUNCATED_LIMIT
-  } more`
+  const newLabels = labels.slice(0, labels.length - 1)
+  return truncateByCharacterLimit(newLabels, limit, labelsTruncatedCount + 1)
 }
+
+export const getTruncatedLabels = (labels: string[], limit: number): string =>
```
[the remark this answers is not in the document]
JakePitman: Yeah - added this to remove the possibility of setting the `labelsTruncatedProp` manually from the outside. It's purely there for recursion. So `getTruncatedLabels` is the same thing, minus the third param. I _did_ forget to stop exporting the recursive function though, which defeats the whole purpose... I'll fix that

packages/select/src/FilterMultiSelect/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
NatalieLoyzaga: We discussed this in the kickoff and the first label regardless of its char length should always display. We also discussed there isn't an obvious name to convey that obscurity. I will add a description which will defs help.

packages/select/src/FilterMultiSelect/utils/getTruncatedLabel/getTruncatedLabel.ts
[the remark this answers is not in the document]
HeartSquared: If the third param is optional, then it should be to use anyway :)

packages/select/src/FilterMultiSelect/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
HeartSquared: Hmm yeah that's hard 🤔 I can't think of anything better currently either. A description will do for now :)

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -29,25 +30,25 @@ export default {
...
-      label="Engineer"
+      {...args}
       onSelectionChange={handleSelectionChange}
       selectedKeys={selectedKeys}
       items={items}
       trigger={() => (
         <FilterMultiSelect.TriggerButton
           selectedOptionLabels={getSelectedOptionLabels(selectedKeys, items)}
-          label="Engineer"
+          {...args}
```
HeartSquared, line 51: Hmm... This probably doesn't make sense, because not all props for `FilterMultiSelect` would apply to `TriggerButton`.

This could probably work though?

```suggestion
          label={args.label}
```

packages/select/src/FilterMultiSelect/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
dougmacknz: > The first label regardless of its char length should always display

Definitely agree but now I feel like we should probably account for how long the label is in this calculation (if that's not happening already?) 

Is this too wordy: `labelCharacterLimitBeforeTruncate`?

HeartSquared: If it adds clarity, there's nothing wrong with a long variable imo. Shippit

HeartSquared (approved):
Code looks fine. Give some consideration to Doug's suggestion for the prop name though (since it would be a breaking change if you wanted to change it later).

JakePitman (approved):

JakePitman (approved):

PR #2942, opened 2022-09-02
feat(filter-multi-select): Allow disabled options

NatalieLoyzaga:
## Why

- Allow disabled options

## What

- Allow for a disabledKeys prop which takes a set.
- Add a tests to check for a disabled item


dougmacknz:
Nice. The other thing I noticed is that there's still hover styling for disabled options.

dougmacknz:
Something's not quite right
https://user-images.githubusercontent.com/1811583/188338876-6933abc1-3051-440c-98cb-6aedc8181c6c.mov

JakePitman (approved):

PR #2949, opened 2022-09-06
fix(select): fix selected option styling

JakePitman:
Removes blue background unless focused, and fixes font styling

## Why
Implementation is out of sync with the UI kit - screenshots [here](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-396&text=select)

## What
- Removes blue-100 background for selected item (unless focused. This is to differentiate between what is focused, and what is selected)
- Adds bold font weight, and color blue-500 to selected item, to distinguish from non-selected items

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.module.scss
+++ b/draft-packages/select/KaizenDraft/Select/Select.module.scss
@@ -117,6 +117,7 @@ $focus-border-color: $color-blue-500;
     color: $color-blue-500;
     border-color: $focus-border-color;
     &.disabledOption {
+      color: $color-purple-800;
       border-style: $border-dashed-border-style;
       border-width: $border-focus-ring-border-width;
       border-color: $color-gray-500;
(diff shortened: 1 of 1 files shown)


jmitsios:
@JakePitman Nice work mate!  Only things I've picked up on is:

- When you click the input field to open the menu after an option has been selected, the focus jumps to the selected item.   I'm unsure if the expected behavior is to default to the **first** item in the list, so the user navigating with a keyboard, or screen reader, can go through the options from the top (rather than skipping over a few to the selected option.)
image

- When navigating with the keyboard over a disabled option, can we align the style of this with the pattern we've used in other components?
image
image

jmitsios:
@JakePitman 
- The disabled focus state still shows a blue highlight on the text when focused on. Could we please keep the text as is, and just display the dashed focus ring?
image

NatalieLoyzaga (approved):

PR #2951, opened 2022-09-08
fix: be able to specify root element applying theme

Tommy-Chen-cultureamp:
## Why
In order to let nextjs app set default styles on server-side. We have to set defaultTheme at html component before this fix.

## What
This is an easy fix that, in the nextjs app, we can specify the component with id `kaizen-root` as a root element and it can render default styles on the server side.

### DIFF
--- a/packages/design-tokens/src/ThemeManager.ts
+++ b/packages/design-tokens/src/ThemeManager.ts
@@ -50,7 +50,9 @@ export class ThemeManager<Theme extends BaseTheme = BaseTheme> {
   }
   public applyCurrentTheme = () => {
     if (typeof window !== "undefined") {
-      this.setRootElement(document.documentElement)
+      this.setRootElement(
+        document.querySelector("#kaizen-root") ?? document.documentElement
+      )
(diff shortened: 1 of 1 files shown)


packages/design-tokens/src/ThemeManager.ts
```diff
@@ -50,7 +50,9 @@ export class ThemeManager<Theme extends BaseTheme = BaseTheme> {
   }
   public applyCurrentTheme = () => {
     if (typeof window !== "undefined") {
-      this.setRootElement(document.documentElement)
+      this.setRootElement(
+        document.querySelector("#kaizen-root") ?? document.documentElement
```
dougmacknz, line 54: This is a little magical for my liking. What about if `ThemeManager` took an optional second argument: `rootElementId`, rather than having a hardcoded `#kaizen-root`?

Tommy-Chen-cultureamp: ah that sounds better! I will do that

dougmacknz (approved):

PR #2953, opened 2022-09-09
fix: Extra catch around illustration video player

sugendran:
## Why

Some older (unsupported) browsers did not have a promise as the result of the `.play` function so they are spamming the error logs. Although we don't support those browsers, we do need to add the guard against it as they are spamming our error logs.

## What
Have wrapped the `.play` in a try..catch and also check that the result of `.play` returns a callable `.catch` before adding the catch.

### DIFF
--- a/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
@@ -101,15 +101,23 @@ export const VideoPlayer = ({
     if (prefersReducedMotion) {
       videoElement.pause()
     } else if (autoplay && !prefersReducedMotion) {
-      videoElement.play().catch(e => {
-        /*
-         * An DOMException _may_ be raised by some browsers if we
-         * programatically interact with the video before the
-         * user has interacted with the page. This is okay - so
-         * we're going to catch this error without handling it. See:
-         * https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide#autoplay_availability
+      try {
+        // Older browsers may not return a promise, so .play could return undefined
+        videoElement.play()?.catch(e => {
+          /*
+           * An DOMException _may_ be raised by some browsers if we
+           * programatically interact with the video before the
+           * user has interacted with the page. This is okay - so
+           * we're going to catch this error without handling it. See:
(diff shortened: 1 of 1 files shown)


draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.tsx
```diff
@@ -101,15 +101,25 @@ export const VideoPlayer = ({
     if (prefersReducedMotion) {
       videoElement.pause()
     } else if (autoplay && !prefersReducedMotion) {
-      videoElement.play().catch(e => {
```
HeartSquared, line 104:
```suggestion
videoElement.play()?.catch(e => {
```

Would optional chaining possibly help? Although of course this means that if `catch` doesn't exist, we get no error logs. How much do we care about them?

sugendran: I always forget about optional chaining. We don't care about the error as it fallsback to not playing the animation. Will update.

sugendran: I ended up typing it so that it's explicit to people why we have optional chaining there. The `dom.ts` that we reference doesn't show that it could be undefined.

HeartSquared: I don't think you need the outer try/catch anymore with the optional chaining.

I'm not a big fan of type-casting though, because it opens us up to missing errors if there are any changes in the future. Certainly worth documenting though, but I think a comment is enough in this situation.

sugendran: The `.play` can fail as well so you'd still need the outer try catch

sugendran: removed the typecasting and left the comment

HeartSquared: Ah okay. I'll trust you tested it then :)

HeartSquared (dismissed):

HeartSquared (approved):

PR #2954, opened 2022-09-09
feat(Collapsible): Use an IconButton for header toggle (a11y improvement)

dougmacknz:
## What
Rather than the whole Collapsible header being a `button` element, turn the chevron that sits in the header into an `IconButton` which will act as the toggle button for keyboard & screen reader users. Then add add an `onClick` to the header div as an additional layer so that mouse users can still click the header to open and close. 

## Why
Two reasons:

1: There are times where we want the header to contain other actions, e.g.
image

In these cases, we're forced to have _buttons inside buttons_, which is invalid HTML and causes many issues.

2: Elements inside of buttons have their semantics stripped and simplified down to a string. If you were to put an `h3` into this heading (and it's wrapped in a button), the document will no longer see the `h3` as an `h3`.

### DIFF
--- a/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
+++ b/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
@@ -67,10 +67,19 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
   background-color: white;
   border: none;
   transition: background-color $ca-duration-rapid;
-  padding: $spacing-sm $spacing-sm $spacing-sm $spacing-md;
+  padding-top: $spacing-sm;
+  padding-bottom: $spacing-sm;
+  padding-left: $spacing-md;
+  padding-right: $spacing-sm;
 
   &:hover {
     background-color: $heading-active-color;
+  }
+
+  [dir="rtl"] & {
+    text-align: right;
+    padding-left: $spacing-sm;
+    padding-right: $spacing-md;
   }
 }
 
--- a/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
+++ b/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
@@ -1,5 +1,6 @@
 @import "~@kaizen/design-tokens/sass/border";
 @import "~@kaizen/design-tokens/sass/shadow";
+@import "~@kaizen/design-tokens/sass/spacing";
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/deprecated-component-library-helpers/styles/layout";
 @import "~@kaizen/component-library/styles/border";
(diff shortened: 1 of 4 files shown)


draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
```diff
@@ -61,25 +62,23 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
...
   border: none;
   transition: background-color $ca-duration-rapid;
-
-  @include ca-padding(
-    $top: $ca-grid,
-    $end: $ca-grid,
-    $bottom: $ca-grid,
-    $start: $ca-grid
-  );
+  padding: $spacing-sm $spacing-md;
```
HeartSquared, line 70: How come these values aren't matching anymore?

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.tsx
```diff
@@ -97,20 +97,19 @@ export class Collapsible extends React.Component<CollapsibleProps, State> {
...
             [styles.clearVariant]: open && variant === "clear",
             [styles.sticky]: sticky,
             [styles.open]: open,
           })}
           style={sticky && { top: sticky.top }}
-          onClick={this.handleClick}
-          aria-expanded={open}
-          aria-controls={sectionId}
-          data-automation-id={`collapsible-button-${id}`}
+          onClick={this.handleSectionToggle}
```
HeartSquared, line 111: Just for curiosity, should this div now need a focus state because it's clickable? Or does that have the same effect as having a button in a button?

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
dougmacknz: It's because the IconButton has padding (for the hover/focus styling), whereas the Icon didn't have any.

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.tsx
dougmacknz: In terms of a keyboard only experience we now have this:
https://user-images.githubusercontent.com/1811583/189813421-dcc542ef-fa93-489b-a6ce-25b554b91503.mov

For the mouse experience, we don't have any focus styling on either the header or the IconButton.

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
HeartSquared: Hmm I've noticed that the chevron has moved a lot further to the left than before now though. Do we need to check with @jmitsios?

HeartSquared (dismissed):
I'll just approve it for now. The only thing that really stood out was the chevron moving a bit of a distance to the left compared to before. If Jason is okay with it, then you can go ahead :)

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
jmitsios: @HeartSquared - You're right, it looks like it has moved to the left _(unless it's always been that way and gone unnoticed til now?)_ 

Ideally I'd prefer consistent padding around it, if we could reduce the right margin to `12px`?
image

draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.module.scss
```diff
@@ -61,23 +62,30 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
 
 .button {
```
HeartSquared, line 63:
```suggestion
.header {
```

Nitpick, but it's not a button anymore, so this was a little confusing.

draft-packages/collapsible/package.json
```diff
@@ -32,6 +32,7 @@
     "@kaizen/component-base": "^1.1.1",
     "@kaizen/component-library": "^16.1.1",
     "@kaizen/typography": "^2.3.2",
+    "@kaizen/button": "^1.3.3",
```
HeartSquared, line 35: You may want to move this up so it's in the correct alphabetical order, otherwise things will move around when anyone uses a command to add a package.

HeartSquared (dismissed):

HeartSquared (approved):

PR #2960, opened 2022-09-12
chore(title-block-zen): Rename renderTab prop to render

dougmacknz:
## What
Renames the `renderTab` prop to just `render` on Title Block's `NavigationTab` component.

Also adds a prop description to the `render` prop from the last PR 'cause I forgot.

## Why
To create a consistent name for this prop, rather than having something unique each time.

This should technically be a breaking change, but we just added this prop a couple weeks ago and no one is using it yet so I'm looking to get away without a breaking change :)


NatalieLoyzaga (approved):
Nice! Sneakily ship it

PR #2961, opened 2022-09-12
docs: update menu component stories

mcwinter07:
## Why
Confusion around Menu has popped up a couple of time and the components stories are written in the most digestible way. This is an effort to split up the stories into more readable chunks and provide better guidance on how to use this component.

## What
- adds subcomponent stories for MenuItems and MenuList
- move and rename the <MenuInstance> helper component to MenuContentExample in the docs components folder and provide a typed description
- Move more complicated examples of Menu into its own story folder (probably still need refactoring)
- simplify initial story and used MenuItem and MenuList as children to provide better guidance


draft-packages/menu/docs/Menu.stories.tsx
```diff
@@ -97,328 +32,143 @@ export default {
...
-        <MenuInstance />
-      </Menu>
-    </Box>
-  </>
-)
-
-AutoHideBehaviours.storyName = "Auto hide behaviours"
-
-export const LabelAndIconReversed = () => (
+export const DefaultStory = args => (
```
christian314159, line 35: I'd assume many people would look at the default story for a code example, but its menu content is very simplistic - e.g. missing links, it just has 3 buttons.
Not necessary for this PR, but I think it could be an improvement to make this example more elaborate.

draft-packages/menu/docs/MenuExamples.stories.tsx
```diff
@@ -0,0 +1,342 @@
...
+          <IconButton secondary label={"Edit menu"} icon={meatballsIcon} />
+        }
+      >
+        <MenuContentExample />
+      </Menu>
+    </StoryWrapper.Row>
+  </StoryWrapper>
+)
+
+IconExample.storyName = "Icon button menus"
```
christian314159, line 99: This might just be me, but I had to read "Icon button menus" 3 times before I could make sense of it.
To me "Menu using Icon Buttons" would be much clearer.

draft-packages/menu/docs/components/MenuContentExample.tsx
```diff
@@ -0,0 +1,81 @@
+import React from "react"
```
christian314159, line 1: Nice refactoring to move this out.

The one thing I don't like about this though (and I guess that was how it was before, too) is that it makes the code of the stories "un-copy & paste-able": You can't just look at the story, click the "Code view" button and then copy the code to your code base because you have to find `<MenuContentExample />` first - which you'd have to have the Kaizen repo for.

I wish we could `spread/unwrap` the component somehow.

(Also nothing you need to solve in this PR IMO)

christian314159 (approved):
Some comments, but nothing that would block merging 👍

draft-packages/menu/docs/components/MenuContentExample.tsx
mcwinter07: Oh yeah, this isn't great from that perspective, kinda hate how storybook's code export only shows a shallow render of components like this, but I opted to go this route instead of producing a lot of duplicate code.

Originally this was in the actual stories as <MenuInstance />, which wasn't really clear that it's not an actual component, so hopefully the name will help clear up that confusion. I've also tried to clamp it down to use in examples where the inner content isn't as relevant to understanding the story. 🤞 that doesn't get lost

draft-packages/menu/docs/MenuExamples.stories.tsx
mcwinter07: Yeah that sounds much better 👍 Will update that

draft-packages/menu/docs/Menu.stories.tsx
```diff
@@ -97,328 +32,143 @@ export default {
...
+        }}
+        icon={trashIcon}
+        destructive
+        label="Delete item"
+      />
+      <MenuItem
+        href="https://cultureamp.design/components/menu/"
+        label="Learn more about Menu"
+      />
+      <MenuList heading="Deprecated props">
```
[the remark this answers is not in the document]
mcwinter07: Nice catch, that is not meant to be a list in list 😆  will fix that

draft-packages/menu/docs/Menu.stories.tsx
mcwinter07: I'm gonna fix up a could of other things so will add that one in as well 👍

PR #2963, opened 2022-09-13
feat: remove deprecated styles from form primitives

NatalieLoyzaga:
## Why

Remove deprecated styles from primitive form components

## What

- Remove and replace ca- or kz- tokens

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.module.scss
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.module.scss
@@ -1,19 +1,20 @@
 @import "~@kaizen/design-tokens/sass/border";
 @import "~@kaizen/design-tokens/sass/color";
-@import "~@kaizen/deprecated-component-library-helpers/styles/layout";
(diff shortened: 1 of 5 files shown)


draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.module.scss
```diff
@@ -95,17 +103,28 @@ $input-with-icon-padding: calc(#{$input-icon-size} + calc(#{$ca-grid} * 0.75));
   }
 
   .input {
-    @include ca-padding(
-      $start: $input-with-icon-padding,
-      $end: $input-base-padding-horizontal
-    );
+    padding-right: $input-base-padding-horizontal;
```
christian314159, line 106: Just wondering, is this something where we should use `padding-inline-start/end` instead now that we don't have to support IE11 anymore?

christian314159 (approved):
Looking good, but maybe ask in our `#wol` channel about using `padding-inline-...` instead, might be a nice improvement.

PR #2967, opened 2022-09-15
chore(tsconfig): add noUncheckedIndexedAccess as true to config

NatalieLoyzaga:
## Why

Fixes https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?label=health_week&selectedIssue=KDS-446

## What

- Add `"noUncheckedIndexedAccess": true` to tsconfig

### DIFF
--- a/packages/stylelint-plugin/tsconfig.dist.json
+++ b/packages/stylelint-plugin/tsconfig.dist.json
@@ -9,8 +9,7 @@
     "declarationMap": true,
     "declaration": true,
     "target": "es5",
-    "module": "CommonJS",
-    "noUncheckedIndexedAccess": true
+    "module": "CommonJS"
   },
   "include": ["src", "types"]
 }
--- a/packages/stylelint-plugin/tsconfig.dist.json
+++ b/packages/stylelint-plugin/tsconfig.dist.json
@@ -9,7 +9,8 @@
     "declarationMap": true,
     "declaration": true,
     "target": "es5",
-    "module": "CommonJS"
+    "module": "CommonJS",
+    "noUncheckedIndexedAccess": true
   },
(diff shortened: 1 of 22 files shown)


packages/stylelint-plugin/tsconfig.dist.json
```diff
@@ -9,7 +9,8 @@
     "declarationMap": true,
     "declaration": true,
     "target": "es5",
-    "module": "CommonJS"
+    "module": "CommonJS",
+    "noUncheckedIndexedAccess": true
```
HeartSquared, line 13: This (and the other packages) shouldn't require it since they all extend the root dist file

NatalieLoyzaga: ahhh I knew you did some magic recently - I'll update now

NatalieLoyzaga: I'll remove from design-tokens too as that was previously added before this update.

HeartSquared (dismissed):

packages/stylelint-plugin/tsconfig.dist.json
NatalieLoyzaga: So did you want me to put it back? @HeartSquared

HeartSquared: Hmm... it will probably be better for the releases since I suspect it would release an unnecessary patch. But I'm not too fussed either.

NatalieLoyzaga: @HeartSquared Reverted and ready for another tick :)

HeartSquared (approved):

PR #2971, opened 2022-09-16
fix(design-tokens): update dist structure

HeartSquared:
## What

Fix the module structure for design-tokens.

- Moved `react` dir into `src`
- Flattened `dist` dir to mirror `src`
- Moved module root to `src` (TS only - was causing errors)
- Added a `index.ts` just for local development

## Why

Different repos were treating different files as the module root.

As you can see, the distributed package has an `index.ts` in the root, while also having a `index.js` in the `dist` directory.
image
Depending on the repo config, it may find the one in `index.ts` even though it should be defining the package root as `dist/index.js`.


packages/design-tokens/tokens/animation.json.d.ts
```diff
@@ -1,5 +1,5 @@
 import { DeepMapObjectLeafs, Theme } from "../src/types"
-import { useTheme, ThemeProvider } from "../react"
+import { useTheme, ThemeProvider } from "../src/react"
```
HeartSquared, line 2: These `.d.ts` files should be pointing to the `dist` files (generally they shouldn't even be committed), but we opted not to fix it in this PR.

zioroboco (approved):

mcwinter07 (approved):
Looks hunky dory! merge in good health 🦖

PR #2976, opened 2022-09-19
feat(filterMultiSelect): Allow grouped items within sections

NatalieLoyzaga:
## Why

- allow consumers to specify grouped items

## What

- Create new `Node<ItemType>` and `ItemType` variations of `MutliSelectOption` - Update all usages of `FilterMultiSelect.Option` to use `FilterMultiSelect.OptionNode`
- Allow `children` as an optional prop of `ItemType`
- Create a grouped items story
- Created a `Section` component that displays the section label and children as options.
- Update the `SelectionProvider` to allow for items with children.
- Update `getSelectedOptionsLabels` function to flatten the selected items.

image
image


NatalieLoyzaga:
Question for @jmitsios - What to do when there are 2 groups with sub-items and then the last item is not within any groups? Currently, `Engineering-type-6` is actually not a part of `Group 2` in terms of the data structure but visually you would not be able to tell.

image

jmitsios:
> Question for @jmitsios - What to do when there are 2 groups with sub-items and then the last item is not within any groups? Currently, `Engineering-type-6` is actually not a part of `Group 2` in terms of the data structure but visually you would not be able to tell.
> 
@NatalieLoyzaga -
Is this because they're automatically sorted into groups from a database? I believe we'll have to be quite prescriptive to consumers who use it, and mention they have to put their items into groups so we're not left with stragglers like this.    In any case, if they don't do that, can they be placed into a group called 'Other'?

In the design review of it I found a few other things to make mention:

- Could we remove this gap being created at the base of the list and scrollbar (highlighted in red)? So that overflow content sits flush with the divider line?
image

- Disabled states still seem to have a style change when hovered over.  The checkbox goes darker, and the badge numbers background turns white.  Disabled states should really have no visual change, other than the bespoke focus ring styling.
image

- I noticed when the first item in the menu has focus and I press the Down arrow key to navigate, it jumps 3 items down the list (See frame 1 and 2.)  Then pressing it again skips down over 2 items (See frame 3)
image

- Some minor alignment and spacing issues with the group header. In the UI Kit the group headers are left-aligned with the row containers (shown in the hover state), whereas in the Storybook build they're aligned with the checkboxes. Also a slightly more amount of spacing is needed above and below the headers.
image

- Padding around the checkbox looks to be larger in the hover states in the Storybook build (Image shown UI Kit versus Build)
image

JakePitman:
Updated most of the style changes, but I'm not sure about the padding difference in the checkbox.
They definitely look different in that screenshot, but the [UI Kit](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%9A%A7-UI-Kit%3A-Beta-Components?node-id=24783%3A132249) seems to specify `6px`, and storybook is using a padding of `0.375rem`, which ends up being `6px`

PR #2977, opened 2022-09-19
feat(KDS-754): create time picker component

jshli:
## Why
Perf-UI needs a time picker component that is localised, accessible allows custom user input, and displays a dropdown of time increments to help users quickly choose a time

## What
A time picker built using a number of `React-Spectrum` libraries.


packages/time-picker/src/TimePicker/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,22 @@
...
+import styles from "./Menu.module.scss"
+
+export const Menu = (props: AriaMenuOptions<TIME_OPTION> & TreeProps<any>) => {
+  const state = useTreeState(props)
+  const ref = React.useRef(null)
+  const { menuProps } = useMenu(props, state, ref)
+
+  return (
+    <ul {...menuProps} ref={ref} className={styles.menuContainer}>
+      {/* @ts-ignore */}
```
HeartSquared, line 15: Whatever this is, it's probably not something we'd want to ignore? If there's a good reason though, make sure to put a comment to explain it :)

packages/time-picker/src/TimePicker/components/MenuItem/MenuItem.module.scss
```diff
@@ -0,0 +1,33 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/typography";
+
+.menuItem {
+  color: #2f2438;
+  background: transparent;
+  padding: 6px calc(#{$spacing-sm} - 2px);
```
HeartSquared, line 9:
```suggestion
  padding: $spacing-xs calc(#{$spacing-sm} - 2px);
```

Is there something else we could do instead the `-2px` on a spacing? Although I assume we'd probably be wanting a review from Jason at some point for design, so we can probably address it a little later once we're happy with functionality.

packages/time-picker/src/TimePicker/components/MenuItem/types.ts
```diff
@@ -0,0 +1,34 @@
+export interface Node<T> {
```
HeartSquared, line 1: You mentioned this was copied from elsewhere right? We'd ideally not want to need to redefine this ourselves (especially not with the name `Node`). Let's try and find it to import :)

packages/time-picker/src/TimePicker/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react"
+import {
+  useOverlay,
+  AriaOverlayProps,
+  DismissButton,
+} from "@react-aria/overlays"
+import { FocusScope } from "@react-aria/focus"
+import styles from "./Popover.module.scss"
+
+export const Popover: React.FunctionComponent<
```
HeartSquared, line 10:
```suggestion
export const Popover: React.VFC<
```

As with all the other components (I know it's deprecated in React 18, but since we still support 16 and 17, we haven't changed it yet).

packages/time-picker/src/TimePicker/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,40 @@
...
+      onClose,
+      shouldCloseOnBlur: true,
+      isDismissable: true,
+      shouldCloseOnInteractOutside,
+    },
+    ref
+  )
+  // Add a hidden <DismissButton> component at the end of the popover
+  // to allow screen reader users to dismiss the popup easily.
+  if (!isOpen) {
```
HeartSquared, line 26: If the hook takes in `isOpen`, does it not mean they have handled this for us? If not, would it make more sense (or not) for the consuming component to render this component conditionally?

jshli: for some reason, the `isOpen` prop didn't actually seem to do anything, therefore this conditional render. But I'll come back to look at this

packages/time-picker/src/TimePicker/components/MenuItem/MenuItem.module.scss
jshli: I have a feeling I might have copied these styles from somewhere. I'll have a look at where I found it from

packages/time-picker/src/TimePicker/components/Menu/Menu.tsx
jshli: Ooh yeah sorry there's going to be a few of these scattered around. Haven't gotten around to cleaning it up yet

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,169 @@
...
+export type StatusType = "default" | "error"
+
+export interface TimePickerProps
+  extends Omit<
+    TimeFieldStateOptions,
+    "errorMessage" | "validationState" | "value" | "onChange" | "label"
+  > {
+  id: string
+  label: string
+  // if Value takes a date, it returns a date. Else returns time string?
```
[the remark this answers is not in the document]
jshli: Sorry this is now outdated!

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,169 @@
...
+    expect(
+      screen.getByTestId("timepicker-menu").firstChild?.textContent
+    ).toEqual("12:00 am")
+  })
+  it("shows time in a h:mm A format for en-US", () => {
+    render(<TimePickerWrapper locale="en-US" />)
+    expect(screen.getAllByRole("spinbutton")).toHaveLength(3)
+    fireEvent.click(screen.getByTestId("timepicker-input"))
+    expect(
+      screen.getByTestId("timepicker-menu").firstChild?.textContent
```
BurntCaramel, line 70: Why not `.toHaveText("12:00 AM")` or `.toHaveText(/12:00 AM/)` ?

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,169 @@
...
+    ).toEqual("00:00")
+  })
+})
+
+it("changes value when selecting a menu item", () => {
+  render(<TimePickerWrapper />)
+  fireEvent.click(screen.getByTestId("timepicker-input"))
+  fireEvent.click(screen.getByText("01:30"))
+  // FIXME: There should actually be a leading 0, i.e. 01
+  expect(screen.getAllByRole("spinbutton")[0].textContent).toEqual("1")
```
BurntCaramel, line 89: Again why not `toHaveText`? Traversing the DOM feels fragile to me.

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { useState } from "react"
+import { fireEvent, render, screen } from "@testing-library/react"
+import { TimePicker, TimePickerProps } from "./TimePicker"
+
+const mockSetValue = jest.fn()
+const UTC_ZERO_TIMEZONE = "Africa/Abidjan"
+
+const pressArrowKey =
+  (direction: "ArrowUp" | "ArrowDown") => (element: HTMLElement) =>
```
BurntCaramel, line 9: It doesn’t really matter but I’m curious why this is a curried function? Why not have two arguments?

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  const state = useTimeFieldState({
+    ...restProps,
+    // @ts-expect-error controlled values should not be undefined and library throws warning if such is supplied for value,
+    // however library does not allow for null values despite handling properly
+    // https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/utils/src/useControlledState.ts#L23
+    // https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/datepicker/src/useTimeFieldState.ts
+    value: value ? parseAbsolute(value.toISOString(), timeZone) : null,
```
BurntCaramel, line 95: Why about falling back to an empty string instead of `null`?

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const options = useMemo(
+    () =>
+      getAllTimeOptions({
+        locale,
+        timeZone,
+        date: value,
+        increments: dropdownIncrements,
+      }),
+    [locale, timeZone, dropdownIncrements]
```
BurntCaramel, line 128: How come `value` is not a dependency?

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,202 @@
...
+  )
+
+  return (
+    <>
+      <Label data-testid="timepicker-label">{`${label} ${
+        restProps.hideTimeZone && timeZoneLabel ? "" : `(${timeZoneLabel})`
+      }`}</Label>
+      <div className={styles.wrapper}>
+        {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */}
+        <div
```
BurntCaramel, line 138: Is there a reason this isn’t a `<button>`?

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,202 @@
...
+            [styles.error]: state.validationState === "invalid",
+          })}
+        >
+          {state.segments.map((segment, i) => (
+            <DateSegment key={i} segment={segment} state={state} />
+          ))}
+          <div className={styles.focusRing} />
+          <Button
+            {...menuTriggerProps}
+            aria-label={`${id} dropdown`}
```
BurntCaramel, line 156: Why is the `id` shown to the user? Why not the label?

packages/time-picker/src/TimePicker/components/Button/Button.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Button } from "./Button"
+
+const ButtonWrapper = (customProps?): JSX.Element => (
+  <Button
+    exampleRequiredString="Hello!" /** @todo: Add default values for your required props (override them with customProps if needed) */
+    {...customProps}
+  />
+)
+
+describe.skip("<Button />", () => {
```
BurntCaramel, line 12: Pointless test, should this just be deleted?

packages/time-picker/src/TimePicker/components/DateSegment/DateSegment.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { DateSegment } from "./DateSegment"
+
+const DateSegmentWrapper = (customProps): JSX.Element => (
+  <DateSegment
+    exampleRequiredString="Hello!" /** @todo: Add default values for your required props (override them with customProps if needed) */
+    {...customProps}
+  />
+)
+
+describe.skip("<DateSegment />", () => {
```
BurntCaramel, line 12: Another test that does nothing

packages/time-picker/src/TimePicker/components/DateSegment/DateSegment.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <div
+      {...segmentProps}
+      ref={ref}
+      className={`${styles.segment} ${
+        segment.isPlaceholder ? styles.placeholder : ""
+      }`}
+    >
+      {segment.text}
+      {/* {renderText()} */}
```
BurntCaramel, line 37: WIP stuff?

packages/time-picker/src/TimePicker/components/Menu/Menu.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Menu } from "./Menu"
+
+const MenuWrapper = (customProps): JSX.Element => (
+  <Menu
+    exampleRequiredString="Hello!" /** @todo: Add default values for your required props (override them with customProps if needed) */
+    {...customProps}
+  />
+)
+
+describe.skip("<Menu />", () => {
```
BurntCaramel, line 12: Another test that does nothing

packages/time-picker/src/TimePicker/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const Menu = (
+  props: AriaMenuOptions<TIME_OPTION> & TreeProps<TIME_OPTION>
+) => {
+  const state = useTreeState(props)
+  const ref = React.useRef(null)
+  const { menuProps } = useMenu(props, state, ref)
+
+  return (
+    <ul {...menuProps} ref={ref} className={styles.menuContainer}>
+      {[...Array.from(state.collection)].map(item => (
```
BurntCaramel, line 17: You can pass the mapping function as the second argument to `Array.from()`: http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

Also doing `Array.from()` and then splatting that into a new array creates a needless copy.

BurntCaramel, line 17:
```suggestion
      {Array.from(state.collection, item => (
```

packages/time-picker/src/TimePicker/components/MenuItem/MenuItem.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { MenuItem } from "./MenuItem"
+
+const MenuItemWrapper = (customProps): JSX.Element => (
+  <MenuItem
+    exampleRequiredString="Hello!" /** @todo: Add default values for your required props (override them with customProps if needed) */
+    {...customProps}
+  />
+)
+
+describe.skip("<MenuItem />", () => {
```
BurntCaramel, line 12: Another test that does nothing

packages/time-picker/src/TimePicker/components/Popover/Popover.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Popover } from "./Popover"
+
+const PopoverWrapper = (customProps): JSX.Element => (
+  <Popover
+    exampleRequiredString="Hello!" /** @todo: Add default values for your required props (override them with customProps if needed) */
+    {...customProps}
+  />
+)
+
+describe.skip("<Popover />", () => {
```
BurntCaramel, line 12: Another test that does nada

packages/time-picker/src/TimePicker/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,40 @@
...
+    {
+      isOpen,
+      onClose,
+      shouldCloseOnBlur: true,
+      isDismissable: true,
+      shouldCloseOnInteractOutside,
+    },
+    ref
+  )
+  // Add a hidden <DismissButton> component at the end of the popover
```
BurntCaramel, line 24: What’s this comment — a todo?

packages/time-picker/src/TimePicker/utils/convertTimeToZonedDateTime.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      day,
+      TIME_ZONE,
+      now(TIME_ZONE).offset,
+      4,
+      56
+    )
+
+    expect(result).toStrictEqual(expectedResult)
+    expect(result.toString()).toStrictEqual(
+      `${year}-0${month}-${day}T04:56:00+10:00[Australia/Melbourne]`
```
BurntCaramel: Melbourne just changed to `AEDT` which is +11 so having this test use the current time will be flakey I imagine.

packages/time-picker/src/TimePicker/utils/convertTimeToZonedDateTime.ts
```diff
@@ -0,0 +1,26 @@
...
+}): ZonedDateTime => {
+  // using this instead of new Date().getTimezoneOffset as ZonedDateTime constructor measures offset different
+  const today = now(timeZone)
+  const todayDate = today.toDate()
+  return new ZonedDateTime(
+    (date ?? todayDate).getFullYear(),
+    (date ?? todayDate).getMonth() + 1,
+    (date ?? todayDate).getDate(),
+    timeZone,
+    today.offset,
```
BurntCaramel, line 22: Won’t the offset be dependent on the passed date, not on the offset of today?

packages/time-picker/src/TimePicker/components/Button/Button.spec.tsx
jshli: Ah sorry these tests are WIP, I never got around to writing them. 

Not sure these particularly need unit tests, so may end up just deleting them

packages/time-picker/src/TimePicker/utils/convertTimeToZonedDateTime.ts
jshli: We do generate the offset from the passed in timezone, but maybe I can assign a variable name to this so it's less confusing.

packages/time-picker/src/TimePicker/TimePicker.tsx
jshli: There's been a few conversations about this, and I think there were a few reasons why this is a `div` - the primary reason I believe is because we wanted the dropdown chevron to also be a button

packages/time-picker/src/TimePicker/TimePicker.tsx
jshli: We wanted the behaviour of having an empty time picker with no value set if there was no default value

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
jshli: I thought i was going to be clever and have a bunch of different functions for different elements and keystrokes. But didn't really work out that way.

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
jshli: I wanted strict equality checking (without having to write regex). 
I believe `toHaveTextContent` uses `node.textContent` more or less under the hood, so it should be the same behaviour, but I've changed it regardless.

packages/time-picker/src/TimePicker/TimePicker.tsx
jshli: Returning back to this - @HeartSquared talked to dug about this, and confirmed we want to keep this a div, as turning it into a button may remove all the accessibility information from inside it.

packages/time-picker/src/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,215 @@
...
+    fireEvent.click(screen.getByRole("group", { name: LABEL }))
+    expect(screen.getByText("00:00")).toBeInTheDocument()
+  })
+})
+
+it("changes value when selecting a menu item", () => {
+  render(<TimePickerWrapper />)
+  fireEvent.click(screen.getByRole("group", { name: LABEL }))
+  fireEvent.click(screen.getByText("01:30"))
+  // FIXME: There should actually be a leading 0, i.e. 01
```
BurntCaramel, line 84: Is this fix easy to be done before merging?

packages/time-picker/src/TimePicker/utils/generateRegexFromInput.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const generateRegexFromInput = (segments: DateSegment[]): RegExp => {
+  const regexString = segments
+    .map(segment => {
+      if (segment.isPlaceholder || segment.type === "literal") {
+        if (segment.type === "hour") {
+          return ".{1,2}"
+        }
+        // replaces all placeholder text characters with regex match any characters
+        return Array.from(Array(segment.text.length).keys())
```
BurntCaramel, line 11: Could this be achieve with a regex like `.{,${segment.text.length}}`

Or something a bit clearer?

packages/time-picker/src/TimePicker/utils/filterTimeOptions.ts
```diff
@@ -0,0 +1,14 @@
+import { DateSegment } from "@react-stately/datepicker"
+import { TimeOption } from "../types"
+import { generateRegexFromInput } from "./generateRegexFromInput"
+
+export const generateFilteredTimeOptions = (
+  allOptions: Record<string, TimeOption>,
+  segments: DateSegment[]
+) =>
+  Object.keys(allOptions).reduce((filteredOptions, optionKey) => {
+    if (generateRegexFromInput(segments).test(optionKey)) {
```
BurntCaramel, line 10: `segments` isn’t changing here but we are throwing away the result of `generateRegexFromInput` in each iteration of this loop. Perhaps perform `generateRegexFromInput` once outside the loop?

packages/time-picker/src/TimePicker/utils/generateRegexFromInput.ts
```diff
@@ -0,0 +1,24 @@
+import { DateSegment } from "@react-stately/datepicker"
+
+export const generateRegexFromInput = (segments: DateSegment[]): RegExp => {
```
BurntCaramel, line 3: Why do we need this function? Do we actually need to generate a regex? Perhaps we could encapsulate the Regex inside, and return a function that just returns `true` or `false`. For now it could keep the regex and use its `.test` method — but encapsulate that.

packages/time-picker/src/TimePicker/utils/getAllTimeOptions.ts
```diff
@@ -0,0 +1,26 @@
...
+type GetAllTimeOptionsConfig = {
+  locale: string
+  increments?: IncrementValues
+}
+
+export const getAllTimeOptions = ({
+  locale,
+  increments = 30,
+}: GetAllTimeOptionsConfig): Record<string, TimeOption> =>
+  Array.from(Array(24).keys()).reduce((options, hour) => {
```
BurntCaramel, line 14: It doesn’t really matter, but it does feel a bit strange to use this "create an array of length N and throw it away" when this looks like a good use case for a `for (let hour = 0; hour <= 24; hour++)` style loop?

gyfchong:
Can we close this PR?

PR #2979, opened 2022-09-19
refactor: upgrade RTE Toolkit to 2.0.0 and update ProseMirror imports

mcwinter07:
## Why
- Misalignment between prosemirror dependencies (specifically prosemirror-model and prosemirror-state) in the Toolkit and RTE was causing the component to break on render
- As these repo are not connected by the same dependancy bot there is no guarantee that these issue would not regularly arise
- This was surfaced on a lock file [maintenance PR](https://github.com/cultureamp/kaizen-design-system/pull/2868/files), which caused a cascade of issues

## What
- Update import statements
- Fixes type errors
- remove prose-mirror dependencies
- This will rely on this [PR](https://github.com/cultureamp/rich-text-toolkit/actions/runs/3080111238/jobs/4977082827) to be merged in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ lerna-*.log
 .DS_Store
 dist
 .cache
+.yalc
+yalc.loc
--- a/packages/rich-text-editor/package.json
+++ b/packages/rich-text-editor/package.json
@@ -29,7 +29,7 @@
   "private": false,
   "license": "MIT",
   "dependencies": {
-    "@cultureamp/rich-text-toolkit": "^1.11.0",
+    "@cultureamp/rich-text-toolkit": "^2.0.0",
     "@kaizen/a11y": "^1.6.4",
     "@kaizen/component-base": "^1.1.1",
     "@kaizen/component-library": "^16.1.1",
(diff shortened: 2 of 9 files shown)


packages/rich-text-editor/src/RichTextEditor/controlmap.ts
```diff
@@ -215,7 +213,7 @@ export function buildControlMap(
     const type = schema.marks.strong
     const groupIndex = getGroupIndex(controlGroupIndex, "bold")
     toolbarControls[groupIndex].push({
-      isActive: markIsActive(editorState, type),
+      isActive: false,
```
dougmacknz, line 216: Left over testing stuff

.gitignore
```diff
@@ -12,3 +12,5 @@ lerna-*.log
 .DS_Store
 dist
 .cache
+.yalc
```
mcwinter07, line 15: Adding these here to make sure that any yalc files from local development doesn't get accidentally committed in the future

packages/rich-text-editor/package.json
```diff
@@ -29,32 +29,18 @@
...
-    "prosemirror-inputrules": "^1.1.3",
-    "prosemirror-keymap": "^1.1.5",
-    "prosemirror-model": "^1.16.1",
-    "prosemirror-schema-list": "^1.1.6",
-    "prosemirror-state": "^1.3.4",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@types/classnames": "^2.3.0",
-    "@types/prosemirror-commands": "^1.0.4",
```
mcwinter07, line 51: All the @type have been removed because they no longer are required in the latest ProseMirror update, which transitioned to Typescript

HeartSquared (approved):

PR #2980, opened 2022-09-19
fix(draft-form): Label margin should be 6px and not 10px

arikark:
[KDS-640](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-640)

## Why
The UI Kit components for the [Checkbox Field](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=14462%3A196) use different spacing between the checkbox and the label than what is shown in [Storybook](https://dev.cultureamp.design/develop/filter-select/?path=/story/components-form-checkbox-field--interactive-kaizen-site-demo).

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/Label/Label.module.scss
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/Label/Label.module.scss
@@ -3,7 +3,7 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "../../variables";
 
-$label-start-margin: 10px;
+$label-start-margin: $spacing-xs;
 
 .label,
 :global(.ideal-sans) .label {
--- a/draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.module.scss
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/ToggleSwitch/ToggleSwitch.module.scss
@@ -7,7 +7,7 @@
 
 $animation-timing: $animation-duration-immediate
   $animation-easing-function-linear;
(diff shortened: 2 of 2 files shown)


draft-packages/form/KaizenDraft/Form/Primitives/Label/Label.module.scss
```diff
@@ -3,7 +3,7 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "../../variables";
 
-$label-start-margin: 10px;
+$label-start-margin: $spacing-xs;
```
HeartSquared, line 6: One thing we did notice was that this affects not only Checkbox, but also RadioField and ToggleSwitchField. Also if any consumers are using this component directly in their repos.

HeartSquared:
@jmitsios has mentioned that the spacing for the focus ring on the Toggle Switch should be 1px as opposed to the 2px it is. 
I took a glance at Figma, and it seems that it may be applicable to Checkbox and Radio too, but since Jason isn't in today I'll check in with him tomorrow.

Technically we could release the fixes separately, but since it visually throws off the spacing with this PR, I think it makes more sense to release them in the same patch.

HeartSquared:
Caught up with @jmitsios. No idea why the code doesn't accurately reflect the visuals, but apparently the 2px offset in the code for Checkbox and Radio actually render with only a 1px offset, so the only one we'd need to fix is the one in Toggle.

HeartSquared (approved):

PR #2982, opened 2022-09-20
feat: heading tag for EmptyState component

caitlinmwillington:
## Why

**Use heading tag and add `headingVarient` prop in EmptyState component**
Currently the "Empty State" component heading renders as a `<div>`, but is styled as `heading-3` which doesn't follow best a11y semantic practices (and has created a violation on the survey configure questions edit page - [here is ticket](https://cultureamp.atlassian.net/browse/KD-1721))  

## What

1. Use `headingVarient` to control heading level
2. Use `tag="div"` when no `headingVarient` provided
3. Add optional `headingVarient` prop so that correct heading can be applied for the page (in cases where no preceding `<h2>`)
4. use `heading-3` styling by default when no `headingVarient` provided

> Note: an alternative to this PR could be adding `role="heading"` to the div, but seemed best to make a heading a heading

### DIFF
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
@@ -61,11 +61,7 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
   illustrationType = "informative",
   layoutContext = "sidebarAndContent",
   headingText,
-  headingProps = {
-    variant: "heading-3",
-    tag: "div",
-    color: "dark",
-  },
+  headingProps,
   bodyText,
   straightCorners,
   isAnimated = true,
@@ -105,14 +101,17 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       </div>
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
-          <Heading
-            variant={headingProps.variant}
-            classNameOverride={styles.heading}
-            tag={headingProps.tag}
-            color={headingProps.color}
-          >
-            {headingProps.children || headingText}
-          </Heading>
+          {headingProps ? (
+            <Heading classNameOverride={styles.heading} {...headingProps} />
+          ) : (
+            <Heading
+              variant="heading-3"
+              classNameOverride={styles.heading}
+              tag="div"
+            >
+              {headingText}
+            </Heading>
+          )}
           <Paragraph variant="body" classNameOverride={styles.description}>
             {bodyText}
           </Paragraph>
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
@@ -63,7 +63,7 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
   loop = false,
   automationId,
   classNameOverride,
-  headingVariant: headingVariant = "heading-3",
+  headingVariant,
   ...props
 }) => {
   const IllustrationComponent = ILLUSTRATIONS[illustrationType]
@@ -97,7 +97,11 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       </div>
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
-          <Heading variant={headingVariant} classNameOverride={styles.heading}>
+          <Heading
+            variant={headingVariant ? headingVariant : "heading-3"}
+            tag={headingVariant ? undefined : "div"}
+            classNameOverride={styles.heading}
+          >
             {headingText}
           </Heading>
           <Paragraph variant="body" classNameOverride={styles.description}>
--- a/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
+++ b/draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
@@ -9,7 +9,7 @@ import {
   EmptyStatesPositive,
   AnimatedSceneProps,
 } from "@kaizen/draft-illustration"
-import { Paragraph, Heading, HeadingVariants } from "@kaizen/typography"
+import { Paragraph, Heading, HeadingProps } from "@kaizen/typography"
 import styles from "./EmptyState.module.scss"
 
 const ILLUSTRATIONS: { [k: string]: React.VFC<AnimatedSceneProps> } = {
@@ -36,10 +36,14 @@ export interface EmptyStateProps
   id?: string
   illustrationType?: IllustrationType
   layoutContext?: LayoutContextType
-  headingText: string | React.ReactNode
   bodyText: string | React.ReactNode
   straightCorners?: boolean
-  headingVariant?: HeadingVariants
+  headingProps?: HeadingProps
+  /**
+   * **Deprecated:** Use headingProps
+   * @deprecated
+   */
+  headingText?: string | React.ReactNode
   /**
    * **Deprecated:** Use test id compatible with your testing library (eg. `data-testid`).
    * @deprecated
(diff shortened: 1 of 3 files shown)


draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -95,11 +97,7 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       </div>
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
-          <Heading
-            variant="heading-3"
-            classNameOverride={styles.heading}
-            tag="div"
-          >
+          <Heading variant={headingVariant} classNameOverride={styles.heading}>
```
mcwinter07, line 100: Love the A11Y considerations in this PR and introducing more flexibility intro the component. One important call out in `<Heading />` is that `tag` controls the semantic tag type, whereas the `variant` is the styling. If `variant` is the only prop that is provided the component will attempt to infer the tag type.

The issue that could arise from this is that if we are not on a page with a h1 or h2 tag - we would be forced into a situation where we could not use h3 styling without breaking the heading level hierarchy. We could instead have another prop for the tag type that uses `AllowedHeadingTags` so that the user can specify the tag without impacting the style.

mcwinter07: Also a consideration about the h3 being default tag type - In this instance I'd err on the side of keeping the default tag to a `div` as it was originally. While it may be correct on most pages to default to h3, on some pages and h2 or even h1 may be more appropriate dependent on the content of the page and where its used in the product. 

If released as a non-breaking change that could also cause a number of accessibility bugs to arise.

caitlinmwillington: Ah great point!! I have just pushed a change so that the fall back is a div with the "heading-3" styling 👍

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -96,9 +98,9 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
       <div className={styles.textSide}>
         <div className={styles.textSideInner}>
           <Heading
-            variant="heading-3"
+            variant={headingVariant ? headingVariant : "heading-3"}
+            tag={headingVariant ? undefined : "div"}
```
dougmacknz, line 102: I think we should probably create a `headingTag` prop as well, so that semantics and styling aren't tightly coupled. From my experience it's more rare that the styling and semantics actually align.

dougmacknz: Actually on second thought, that will mean you end up with 3 separate props for headings things, which is a bit gross.

Couple of options:
1. We expose the heading props directly in a single prop. Something like
```
<EmptyState
  ...
  heading={{
    children: "My heading",
    variant: "heading-3",
    tag: "h4"
  }}
```
We'd need to deprecate the `headingText` prop, and also set defaults for the `variant` and `tag` parts, which might be a bit tricky.

2. We add a render prop instead, which gives people ultimate flexibility to add whatever they want here.

I'm leaning towards option 2 because it's simple and gives people more flexibility. Interested to get @HeartSquared's take on it as the interim tech lead. If we were doing this from scratch I think it would be a different API completely, but these 2 options might be the most realistic given what we're already working with.

mcwinter07: I'd also suggest that either way we go with this, we should also have a couple of tests to test the new prop/s. If we're going the route of render we'd probably want a story example too 🤔

HeartSquared: I've personally been going down the pattern from option 1 for stylistically opinionated subcomponents. It is easier for the consumer to create consistency without the need for an extra import.

That being said, I'm not against option 2 if people think `heading={<Heading />}` is nicer to consume.

Both options are a breaking change regardless. As @dougmacknz said, if we were to rebuild the component, we would likely choose a different option altogether.

dougmacknz: The render prop could just be an additional prop, rather than replacing `headingText`, so that there's a simple option and a flexible option. That's what we've generally done when adding render props.

caitlinmwillington: This conversation reminded me of something I was going to take a look at next - the [Inline Notification component ](https://cultureamp.design/storybook/?path=/story/components-notification-inline-notification--default-kaizen-demo)heading is a H6, which causes a violation when used on our [Confidentiality config page](https://frontend-assets.usw2.dev-us.cultureamp-cdn.com/dev/survey-configuration/storybook/krypton/caitlin/accessibility-fixes--confidentiality-page/index.html?path=/story/confidentiality-confidentialitypage--default-view) as the heading levels skip from H3 to H6. It would be good if we could also have control over the heading level in that component, or use a Div to avoid the violation.

mcwinter07: Hey @caitlinmwillington we talked about this as team at standup and think the best course of action is to go with something closer to option 1 dug recommended originally, which should give granular enough control of the semantic without you needing to pull in the Heading as in as an import into your page. Super rough example but something along the lines of this:

```ts
export interface EmptyStateProps
  extends OverrideClassName<HTMLAttributes<HTMLDivElement>>,
    Pick<AnimatedSceneProps, "isAnimated" | "loop"> {
  ...
  // we could make this optional and give a deprecation warning on the prop
  /**
   * **Deprecated:** Use headingProps
   * @deprecated
   */
  headingText?: string | React.ReactNode
  // We could import Heading Props from "@kaizen/typography"
  headingProps?: HeadingProps
  ...
}
```

and then providing some defaults in the component restructuring that can be used in the component

```tsx
  export const EmptyState: React.VFC<EmptyStateProps> = ({
    ...
    headingText,
    headingProps = {
      variant: "heading-3",
      tag: "div",
      color: "dark"
    },
    ...
    
    return (
    ...
    <Heading
      variant={headingProps.variant}
      classNameOverride={styles.heading}
      tag={headingProps.tag}
      color={headingProps.color}
    >
      {headingProps.children || headingText}
    </Heading>
  ...
  )
```

mcwinter07: It might be good to pair on this one if you've got some time free this week to shorten the feedback loop? :)

caitlinmwillington: Thanks @mcwinter07 , sounds like a great idea! I have booked some time in for Thurs but feel free to move it if the time doesn't suit

caitlinmwillington: @mcwinter07 I have pushed those changes, added a test and updated the storybook to use the new heading props over the deprecated one

mcwinter07: Brill, just reviewing it now

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
```diff
@@ -96,11 +106,12 @@ export const EmptyState: React.VFC<EmptyStateProps> = ({
...
           <Heading
-            variant="heading-3"
+            variant={headingProps.variant}
             classNameOverride={styles.heading}
-            tag="div"
+            tag={headingProps.tag}
+            color={headingProps.color}
           >
-            {headingText}
+            {headingProps.children || headingText}
```
mcwinter07, line 114: Just caught up with Cass and she made a good recommendation around passing the props directly into a Heading component like below:

```
          {headingProps ? (
            <Heading classNameOverride={styles.heading} {...headingProps} />
          ) : (
            <Heading
              variant="heading-3"
              classNameOverride={styles.heading}
              tag="div"
            >
              {headingText}
            </Heading>
          )}
```

This way we don't need to define default props on ln 64 and it makes it easier to untangle the deprecated `headingText` from the new `headingProps`.

mcwinter07 (commented):
@caitlinmwillington I've left a comment about changing the way we render the Heading component to a ternary, but after that update we'll be good to merge 👍

draft-packages/empty-state/KaizenDraft/EmptyState/EmptyState.tsx
caitlinmwillington: Kewl, thanks. Just pushed that change 👍

mcwinter07 (approved):
Looks hunky dory to me 🦖  I'll update the branch for you and get that merged in. Awesome work and thanks again for the contribution

PR #2983, opened 2022-09-21
fix: correct ID on validation message used for describedBy on text field

caitlinmwillington:
## Why

When using a validation message, the text input field would receive this violation: "ARIA attribute element ID does not exist on the page" . You can see in storybook, the `aria-describedby` on the input field is looking for `kaizen-demo-text-field-field-validation-message` but the ID of the validation message is `kaizen-demo-text-field-field-message`

## What

Correct ID used for validation text so `aria-describedby` works as intended

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
+++ b/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
@@ -114,8 +114,8 @@ export const TextField: React.VFC<TextFieldProps> = ({
           })}
         >
           <FieldMessage
-            id={`${id}-field-validation-message`}
-            automationId={`${id}-field-validation-message`}
+            id={validationMessageAria}
+            automationId={validationMessageAria}
             message={validationMessage}
             status={status}
             reversed={reversed}
--- a/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
+++ b/draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
@@ -130,8 +130,8 @@ export const TextField: React.VFC<TextFieldProps> = ({
           })}
         >
           <FieldMessage
-            id={`${id}-field-message`}
-            automationId={`${id}-field-description`}
+            id={descriptionAria}
+            automationId={descriptionAria}
             message={description}
             reversed={reversed}
           />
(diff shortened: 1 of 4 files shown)


draft-packages/form/KaizenDraft/Form/TextField/TextField.tsx
```diff
@@ -114,7 +114,7 @@ export const TextField: React.VFC<TextFieldProps> = ({
           })}
         >
           <FieldMessage
-            id={`${id}-field-message`}
+            id={`${id}-field-validation-message`}
```
dougmacknz, line 117: I think there's a variable that we can use further up in this file, that way we make sure these never go out of sync again.

dougmacknz:
Looking good, but @mcwinter07 has reminded me on the other PR you have, we should add some tests here to make sure this is all working as expected with all of its incarnations (no description or validation message = no aria-describedby, description and no validation message = aria-describedby with the description ID, etc)

Are you happy to take that on @caitlinmwillington? Otherwise we can talk about doing it within our team.

caitlinmwillington:
> Are you happy to take that on @caitlinmwillington? Otherwise we can talk about doing it within our team.
@dougmacknz  Of course, I have pushed some tests :)

mcwinter07 (approved):
This all looks really awesome @caitlinmwillington. I did have a general wondering on the order of the aria-describedby output for the accessible description - ie: should it be error message first then description or description first then error. I can't say I know the answer so I'd prompted a question in the #prod_accessibility [channel](https://cultureamp.slack.com/archives/C0405G8EJ48/p1664167868675539).

Feel free to chime in there if you have any thoughts, otherwise I'm happy for us to merge this in now and update the order with a quick fix later if we need to later. We can always confirm in our next Intopia session :)

caitlinmwillington:
> This all looks really awesome @caitlinmwillington. I did have a general wondering on the order of the aria-describedby output for the accessible description - ie: should it be error message first then description or description first then error. I can't say I know the answer so I'd prompted a question in the #prod_accessibility [channel](https://cultureamp.slack.com/archives/C0405G8EJ48/p1664167868675539).

Ooh interesting thought... Will be a good question for Russ! I am inclined to merge this in now so I can resolve the bug on page asap but happy to wait if you think we should 🙂

mcwinter07:
> @mcwinter07 Ooh interesting thought... Will be a good question for Russ! I am inclined to merge this in now so I can resolve the bug on page asap but happy to wait if you think we should 🙂

@caitlinmwillington I reckon merge it in now and we can add a fix after if needed, especially if its causing a bug. All we'd need to do is swap the position of `validationMessageAria` and `descriptionAria` in the reducer so it'd be a quick one anyhow 👍

PR #2984, opened 2022-09-21
fix(filterMultiSelect): Kds 825 refactor disabled keys to be part of the data structure

NatalieLoyzaga:
## Why

- More intuitive developer experience.
- We do not expect consumer to need to control disabled keys at run time.

## What

- Remove disabledKeys prop
- Built disabled keys based on items and pass to SelectionProvider


JakePitman (approved):

NatalieLoyzaga:
BREAKING CHANGE: **Remove disabledKeys prop and update itemType type**

- `disabledKeys` prop has been removed.
- `itemType` type now includes the `isDisabled` property to specify whether a single item is disabled.
- `item` prop on FilterMultiSelect is now required.

PR #2985, opened 2022-09-26
chore: Skipping playwright tests on master branch

christian314159:
## Why
- Playwright currently runs on our `master` builds, but it's pretty useless then because it only runs after things have been released - and it already ran before the merge

## What
- added branch filtering rule to the Playwright step in our Buildkite config

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -74,7 +74,7 @@ steps:
     key: "playwright-step"
     command: ".buildkite/scripts/run-playwright.sh"
     timeout_in_minutes: 15
-    branches: "!master"
+    branches: "!master !canary"
     depends_on: "publish-step"
     <<: *defaults
     agent_query_rules: ["queue=build-unrestricted-large"]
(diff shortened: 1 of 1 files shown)


mcwinter07 (dismissed):
Looks hunky dory to me 👍

.buildkite/pipeline.yaml
```diff
@@ -74,6 +74,7 @@ steps:
     key: "playwright-step"
     command: ".buildkite/scripts/run-playwright.sh"
     timeout_in_minutes: 15
+    branches: "!master"
```
HeartSquared, line 77:
```suggestion
    branches: "!(master|canary)"
```

Canary releases probably don't need to run Playwright either :)

mcwinter07: True! chuck that canary in a cage ||🐦||

.buildkite/pipeline.yaml
```diff
@@ -74,6 +74,7 @@ steps:
     key: "playwright-step"
     command: ".buildkite/scripts/run-playwright.sh"
     timeout_in_minutes: 15
+    branches: "!master !canary"
```
mcwinter07, line 77: Nice, didn't know you could do this without a logical operator but seems like you can add as many filters as you like #theMoreYouKnow 👍 
Screen Shot 2022-09-27 at 4 44 46 pm

mcwinter07 (approved):

PR #2986, opened 2022-09-27
feat: Kds 657/float selected items

JakePitman:
## Why
We want consumers to have a way to float selected items to the top. We also want to give consumers more flexibility in how they organise dropdown options.

## What
- Exposes `selectedItems`, `unselectedItems` and `disabledItems` to the consumer, as arguments in `FilterMultiSelect`'s `children` function.
- Adds `ListBoxSection`, allowing consumers to group options


packages/select/src/FilterMultiSelect/components/ListBox/ListBox.tsx
```diff
@@ -19,6 +23,27 @@ export const ListBox: React.VFC<ListBoxProps> = ({ children }) => {
     }
     setIsOverflown(listElement.scrollHeight > listElement.clientHeight)
   }, [listRef])
+  // TODO Move to util functions
```
HeartSquared, line 26:
```suggestion
```
Not sure what you're wanting to move, but do that too :)

packages/select/src/FilterMultiSelect/components/ListBox/ListBox.tsx
```diff
@@ -19,6 +23,27 @@ export const ListBox: React.VFC<ListBoxProps> = ({ children }) => {
...
+  const items = selectionState.collection
+  const disabledKeys = selectionState.disabledKeys
+  const selectedKeys = selectionState.selectionManager.selectedKeys
```
HeartSquared, line 29:
```suggestion
  const { collection, disabledKeys, selectionManager: { selectedKeys } } = selectionState
```

packages/select/src/FilterMultiSelect/components/ListBox/ListBox.tsx
```diff
@@ -29,10 +54,11 @@ export const ListBox: React.VFC<ListBoxProps> = ({ children }) => {
...
     >
-      {Array.from(selectionState.collection).map(item =>
-        // pass item to render children
-        children(item)
-      )}
+      {children({
+        selectedItems: itemsState.selectedItems,
+        unselectedItems: itemsState.unselectedItems,
+        disabledItems: itemsState.disabledItems,
+      })}
```
HeartSquared, line 61:
```suggestion
      {children(itemsState)}
```

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -58,7 +59,36 @@ export const DefaultKaizenSiteDemo: ComponentStory<
           <>
             <FilterMultiSelect.SearchInput />
             <FilterMultiSelect.ListBox>
-              {item => <FilterMultiSelect.Option key={item.key} item={item} />}
+              {({ selectedItems, unselectedItems, disabledItems }) => (
```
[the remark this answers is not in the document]
JakePitman: Not like this. Atm the `children` callback only exposes `selectedItems`, `unselectedItems` and `disabledItems`, so the user could just combine all of these into one array themselves and then loop over that.
Alternatively, we expose `allItems` as well, and they can just use that.
I'll go ahead and expose that, make these stories loop over all items like before, and add a new one that shows off the ability to access `selectedItems` etc

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -67,10 +97,6 @@ export const DefaultKaizenSiteDemo: ComponentStory<
           </>
         )}
       </FilterMultiSelect>
-      <Paragraph variant={"body"}>
-        Items:{" "}
-        <CodeBlock language="json" code={JSON.stringify(items, null, "\t")} />
-      </Paragraph>
```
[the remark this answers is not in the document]
JakePitman: Ah, deleted this because "Show code" was displaying the full list of items at the time. But I've changed it now, so I'll add this back in.

NatalieLoyzaga (approved):
Wooooo

jmitsios:
@JakePitman — Looks really great! Nicely done 👍 

I noticed if you have a couple of items selected and run a search; if the results aren't split between the two divided groups (selected and non-selected) the divider line remains visible, giving this double-up of dividers at the bottom. 

How easy is it to hide that divider line if the results aren't showing a division of groups?
image

JakePitman:
Ah good pickup @jmitsios 

This behaviour is currently up to the consumer. I had a pretty in-depth chat with @HeartSquared about how we want to handle dividers, and decided that there isn't a way for us to bake them in without restricting the consumer, or adding complexity to the implementation, so we left divider logic up to them.

That being said, I can't think of an easy way to avoid this doubling up problem as a consumer... We may need to make a proper card for this

PR #2987, opened 2022-09-29
refactor: Added factory function for spot illus so they fit on one line

christian314159:
## Why
I was looking into adding a new Spot illustration and noticed that the file was ~1300 lines long and consisted of a lot of repetition as almost all components were identical but for the `url` they used.
The length of the file, plus the noise from the repetition, made it really hard to get an overview of what's going on and how/where to add new illustrations.

## What
I created a simple factory function that creates the illustration component using a given `url`.
Unfortunately, Prettier would still wrap these much simpler statements due to its 80 char max default, which made it still quite "noisy" to look at IMO. Thus Ito add a `// prettier-ignore` to each line (group statements are not supported :( ), which isn't the best - but still better than the alternative, I think.

So previously an illustration would take up 10 lines:
```
export const LeadingThroughCrisis = ({
  enableAspectRatio,
  ...props
}: SpotProps) => (
  <Base
    aspectRatio={enableAspectRatio ? "square" : undefined}
    {...props}
    name="illustrations/heart/spot/template-library-leading-through-crisis.svg"
  />
)
```
but in this PR it's just 1:
```
/* prettier-ignore */ export const LeadingThroughCrisis = createSpotIllustration("illustrations/heart/spot/template-library-leading-through-crisis.svg")
```

Given we use large screens, I find this so much easier to maintain now:
image

### DIFF
--- a/draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx
+++ b/draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx
@@ -199,1106 +199,175 @@ export const Assertive = ({
   return StaticIllustration
 }
 
+function createSpotIllustration(url: string) {
+  return ({ enableAspectRatio, ...props }: SpotProps) => (
+    <Base
+      aspectRatio={enableAspectRatio ? "square" : undefined}
+      {...props}
+      name={url}
+    />
+  )
+}
+
 /**
  * Template Library / Engagement
  */
-export const BenefitsSurvey = ({ enableAspectRatio, ...props }: SpotProps) => (
-  <Base
-    aspectRatio={enableAspectRatio ? "square" : undefined}
-    {...props}
-    name="illustrations/heart/spot/template-library-benefits-survey.svg"
-  />
-)
-
-export const CustomSurvey = ({ enableAspectRatio, ...props }: SpotProps) => (
-  <Base
-    aspectRatio={enableAspectRatio ? "square" : undefined}
-    {...props}
-    name="illustrations/heart/spot/template-library-custom-survey.svg"
-  />
-)
-
-export const CustomUnattributedSurvey = ({
-  enableAspectRatio,
-  ...props
-}: SpotProps) => (
-  <Base
-    aspectRatio={enableAspectRatio ? "square" : undefined}
-    {...props}
-    name="illustrations/heart/spot/template-library-custom-unattributed-survey.svg"
-  />
-)
-
-export const EngagementSurvey = ({
-  enableAspectRatio,
-  ...props
-}: SpotProps) => (
-  <Base
-    aspectRatio={enableAspectRatio ? "square" : undefined}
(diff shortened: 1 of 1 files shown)


christian314159:
Looking at the above screenshot, maybe it's possible to squeeze it to 80 chars, if I moved the base path into a `const`?

draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx
```diff
@@ -199,1106 +199,175 @@ export const Assertive = ({
...
 
+function createSpotIllustration(url: string) {
+  return ({ enableAspectRatio, ...props }: SpotProps) => (
+    <Base
+      aspectRatio={enableAspectRatio ? "square" : undefined}
+      {...props}
+      name={url}
+    />
+  )
+}
```
HeartSquared, line 210:
```suggestion
const createSpotIllustration = (url: string): React.VFC<SpotProps> =>
  ({ enableAspectRatio, ...restProps }) => (
    <Base
      aspectRatio={enableAspectRatio ? "square" : undefined}
      {...restProps}
      name={url}
    />
  )
```

I think that should work?

draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx
```diff
@@ -199,1106 +199,175 @@ export const Assertive = ({
...
-  enableAspectRatio,
-  ...props
-}: SpotProps) => (
-  <Base
-    aspectRatio={enableAspectRatio ? "square" : undefined}
-    {...props}
-    name="illustrations/heart/spot/template-library-emergency-response.svg"
-  />
-)
+/* prettier-ignore */ export const BenefitsSurvey = createSpotIllustration("illustrations/heart/spot/template-library-benefits-survey.svg")
```
HeartSquared, line 215: Would be nice if we could specify a rule or ignore a range, but as you said, apparently it's not easy to do.

I wonder if prettier-eslint would make things better. But that can be for a future investigation.

draft-packages/illustration/KaizenDraft/Illustration/Spot.tsx
```diff
@@ -199,1106 +203,181 @@ export const Assertive = ({
...
 }
 
+// ------------------------------------------------------------------------- //
+// STATIC ILLUSTRATIONS                                                      //
+// ------------------------------------------------------------------------- //
+
+const SPOT_ILLUSTRATION_BASE_PATH = "illustrations/heart/spot/"
+const createSpotIllustration =
+  (fileName: string): React.VFC<SpotProps> =>
+  ({ enableAspectRatio, ...props }: SpotProps) =>
```
HeartSquared, line 213: Is `: SpotProps` not inferred? Interesting :o

HeartSquared (approved):

PR #2989, opened 2022-09-29
feat: Added new “Calendar Sync” spot illustration

christian314159:
## Why
We have a new illustration for "Calendar Sync".

## What
The SVG has already been added to our assets repo, so all that needed to be done here was adding the new component.
image


draft-packages/illustration/KaizenDraft/Illustration/__snapshots__/illustration.spec.tsx.snap
```diff
@@ -600,6 +600,16 @@ exports[`<Illustration /> Spot BlankSurvey should exist and render an alt tag 1`
 </div>
 `;
 
+exports[`<Illustration /> Spot CalendarSync should exist and render an alt tag 1`] = `
```
HeartSquared, line 603: What a useless snapshot 😂 Can we get rid of this file?

HeartSquared (approved):

PR #2991, opened 2022-09-30
feat: use headingProps in notification to control heading level

caitlinmwillington:
## Why

Applying the same thinking addressed here: https://github.com/cultureamp/kaizen-design-system/pull/2982 for Inline notifications. Currently the default is to use a `h6` which causes accessibility violations on some pages. 

## What

1. Add `HeadingProps` to `InlineNotification` to control the heading level used
2. Use `Headings` component in `GenericNotification` to render headings
3. Use div as the default heading wrapper when none supplied, same as https://github.com/cultureamp/kaizen-design-system/pull/2982
4. Add test to ensure correct heading is rendering when supplied
5. Add `autohide: false` to toast notification story to follow convention for error notifications


packages/notification/src/components/GenericNotification.tsx
```diff
@@ -192,4 +197,32 @@ const CancelButton = ({ onClick }: CancelButtonProps) => (
...
+        classNameOverride={styles.notificationTitle}
+      >
+        {headingProps.children}
+      </Heading>
+    )
+  } else if (titleProp) {
+    return (
+      <Heading
+        variant="heading-6"
+        tag={style === "toast" ? "h6" : "div"}
```
caitlinmwillington, line 216: This may be a bit controversial, changing the default from a `h6` to a `div` when no `headingProps` provided, but follows the discussion in https://github.com/cultureamp/kaizen-design-system/pull/2982 not to cause violations

dougmacknz: Yeah I don't think there was any good reason for this to be an `h6`, any hardcoded heading level like this in a reusable component seems like a bad idea.

Can we get rid of it for `style === toast` as well? Any particular reason you've opted to continue doing that here?

caitlinmwillington: @dougmacknz I was nervous to change the toast in case it messed up with any announcements for screen readers-  but if thats not an issue I can apply the same!

dougmacknz: Yeah I don't think this is helping with that, we really need a live region for toasts. For now unfortunately they're not screen reader accessible.

packages/notification/src/components/GenericNotification.module.scss
```diff
@@ -33,6 +33,14 @@
 .cancelLabel {
   @extend %ca-notification__cancel__label;
 }
+.toastTitle {
+  top: 0.38em;
+  position: relative;
+}
+.notificationTitle {
+  padding-right: 6px;
+  align-self: center;
```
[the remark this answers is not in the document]
caitlinmwillington: Ah ok fair! I've just reverted it to use the old way but adjusted a little as the value wasn't exactly centred

packages/notification/src/components/GenericNotification.tsx
caitlinmwillington: Just set div as the default 👍 
@mcwinter07 as the ternary isn't there anymore I wont add the extra test

packages/notification/src/components/GenericNotification.module.scss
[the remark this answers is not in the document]
mcwinter07: Nice, thanks for the fix 👍

mcwinter07 (approved):
Looks all hunky dory to me ✅ Thanks again for all the fixes @caitlinmwillington and awesome work. Assuming nothing comes up on the chromatic tests I'll update this to the latest and merge this in for you

PR #2993, opened 2022-10-03
feat: add default reverse, not full width support and rm ie11 styles

mcwinter07:
## Why
There is a UI bug reported by Apollo whereby if using `fullWidth:false` with the `isClearable` prop, the component will overlap the selects input. This was due to absolute positioning being used without a margin to pad out the overlap between the two containers.

Screen Shot 2022-10-03 at 1 38 50 pm

Rather than add hardcoded values into the CSS I have opted to remove it in favour of display: inline-flex since ie11 support is no longer required.

Note: the clearable button's obscure focus behaviour is not in scope of this ticket and will not be fixed here.

## Also
This prompted some other fixes for the Select missing variants and A11Y contrast issues, which we bundled into this update since it relates to this style fixes.

## What
- Remove absolute positioning on the Selects clearable parent container
- Add default reverse variant
- Remove Error thrown for the notFullWidth and default & reverse class now that the styles are supported
- Add input wrapper and class to allow for input text colour on reverse to be white as it would fail colour contrast


dougmacknz (dismissed):
Nice, those comments really came in handy

draft-packages/select/KaizenDraft/Select/Select.module.scss
```diff
@@ -317,25 +316,15 @@ $focus-border-color: $color-blue-500;
       }
     }
 
-    .indicatorsContainer {
```
dougmacknz, line 320: Should the reference to this classname be removed from the component now as well?

dougmacknz:
Just tried to test this with the `fullWidth` control and ran into this (not specific to your branch), if you wanna have a quick go at fixing this as well
image

mcwinter07:
> Just tried to test this with the `fullWidth` control and ran into this (not specific to your branch), if you wanna have a quick go at fixing this as well image

Took a little bit of extra finagling but I've updated it be able to remove the error handlers and add the supported styles that are in Figma - wasn't too much extra effort but we're now adding functionality do you think it should be a `feat:` instead of a `fix`?  @dougmacknz

dougmacknz:
> do you think it should be a feat: instead of a fix?

Sorry yeah I didn't read the error well enough and didn't realise it was something we were throwing! Scope creeped ya. Yeah either switch this to a feat, or break out the feature stuff into a new PR.

dougmacknz (approved):
Looking good, you might just want to adjust the PR title to mention the feature part as well.

mcwinter07:
> Looking good, you might just want to adjust the PR title to mention the feature part as well.

Good point, updated the title and the PR description to better capture those details 👍

PR #2994, opened 2022-10-04
chore: Release @kaizen/eslint-plugin v1

dougmacknz:
## What
Manually bump the eslint-plugin package to 1.0.0

## Why
Breaking changes on sub v1 versions (0.x.x) don't trigger a major version release, it has to be manually updated to v1 before that happens

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kaizen/eslint-plugin",
-  "version": "0.1.3",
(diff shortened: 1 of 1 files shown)


HeartSquared (approved):

HeartSquared:
It's not actually being used anywhere though, so maybe we should just delete it instead?

dougmacknz:
Deleting this package instead https://github.com/cultureamp/kaizen-design-system/pull/2995

PR #2995, opened 2022-10-04
chore: Delete eslint-plugin package

dougmacknz:
## What
Deletes the `@kaizen/eslint-plugin` package.

## Why
The one rule that was created in this package was back when we were thinking we were going to create a 'blessed' equivalent to every draft package as a 1:1. That is no longer the case, and this was never implemented anywhere, so I'm just removing this to cut down on the code/packages that we maintain.


HeartSquared (approved):
👋🏻

I suspect we'll need to deprecate it from npm after too as a follow up step.

PR #2997, opened 2022-10-04
fix(button): Remove ref from custom component render to avoid forwardRef console warnings

dougmacknz:
BREAKING CHANGE: Ref is no longer passed through when using the 'component' render prop. I don't think anyone is using this. If your TS build passes then there's nothing to worry about.

## What
Removes the passing through of ref to the custom component render.

## Why
This is causing console warnings (especially in testing environments), even after attempting to implement `forwardRef`. Passing ref through the button and into the custom component render makes life more complicated than it needs to be. Consumers can give a ref to their custom component directly instead, rather than putting it on `Button` and then pulling it from the `CustomButtonProps` on the custom render.


mcwinter07 (approved):
Nice 👍 Nearly caught myself out the console warning for the draft IconButton ref, but given we want peeps to upgrade to the "blessed" button package that should be hunky dory. Might want to call that out in the updates_design_system post just in case though

PR #3004, opened 2022-10-05
Kds 754/time picker filter

nadeemaajihil:
Filtering

### DIFF
--- a/packages/time-picker/src/TimePicker/TimePicker.tsx
+++ b/packages/time-picker/src/TimePicker/TimePicker.tsx
@@ -10,7 +10,7 @@ import {
 import { I18nProvider } from "@react-aria/i18n"
 import { useTimeField } from "@react-aria/datepicker"
 import { useMenuTrigger } from "@react-aria/menu"
-
+import { useKeyboard } from "@react-aria/interactions"
 import { Item } from "@react-stately/collections"
 import { useMenuTriggerState } from "@react-stately/menu"
 import {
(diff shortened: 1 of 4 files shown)


packages/time-picker/src/TimePicker/utils/time.ts
```diff
@@ -25,19 +25,42 @@ export const formatDateToTime = (date: Date, locale, timeZone): string =>
...
+  if (isNaN(parseInt(aNum, 10))) {
+    return false
+  }
+  return true
+}
+
+const getDateRegEx = (segments: DateSegment[] | undefined) => {
+  let regex = ""
+  if (segments) {
+    for (const [, { type, text, isPlaceholder }] of Object.entries(segments)) {
```
jshli, line 38: Why apply `Object.entries` on an array? 
Is there a way we write this without using a `for...of` loop as loops aren't a common practice within the codebase? 

Not super tested, but either:
```suggestion
     const regex = segments
    .filter(
      segment =>
        ((segment.type === "literal" || segment.type === "dayPeriod") &&
          segment.isPlaceholder !== undefined &&
          !segment.isPlaceholder) ||
        isNumber(segment.text)
    )
    .map(segment => segment.text)
    .join("")
```

or if you don't like multiple loops:
```suggestion
   const regex = segments.reduce((regexString, segment) => {
    if (
      ((segment.type === "literal" || segment.type === "dayPeriod") &&
        segment.isPlaceholder !== undefined &&
        !segment.isPlaceholder) ||
      isNumber(segment.text)
    ) {
      regexString += segment.text
    }
    return regexString
  }, "")

```

nadeemaajihil: Oh, I initially wanted the index here but I can just for..of the object. I think simple javascript is well understood by everyone.

packages/time-picker/src/TimePicker/TimePicker.tsx
```diff
@@ -117,17 +117,33 @@ export const TimePicker: React.VFC<TimePickerProps> = ({
...
-  const options = useMemo(
-    () =>
-      getAllTimeOptions({
-        locale,
-        timeZone,
-        date: value,
-        increments: dropdownIncrements,
-      }),
-    [locale, timeZone, dropdownIncrements, value]
+  const [filteredOptions, setFilteredOptions] = React.useState(
```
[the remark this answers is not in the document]
nadeemaajihil: My initial assumption was the component value can be both controlled/uncontrolled so I was treating this filtering separate. Are we sure there shouldn't be an uncontrolled version of this component?

HeartSquared:
Filtering functionality has been revised by designers.
Code for old filtering was also adapted in https://github.com/cultureamp/kaizen-design-system/pull/2977

PR #3011, opened 2022-10-07
test: KDS-837 Add Playwright tests for FilterMultiSelect

JakePitman:
## Why
- Better test coverage. Adds interaction testing.

## What
- Adds Playwright tests to FilterMultiSelect


mcwinter07 (approved):
Nice work, that all look honky dory to me 👍 I think the new selectors should make this a lot less brittle

PR #3014, opened 2022-10-11
chore: Deprecate draft-tabs

dougmacknz:
## What
Add deprecation doc block to `draft-tabs` Tabs.

## Why
See the deprecation message :)


HeartSquared (dismissed):
The props put `@deprecated` at the end of the comment because it won't show up in Storybook if it was put at the top. I think usually it's the first thing in deprecation comments though. But not too fussed, since it still seems to work.

Might be worth adding a link to the W3C docs if we're telling people to look at them though.

HeartSquared (approved):

PR #3015, opened 2022-10-11
fix(filter-multi-select): strip items of undefined values

JakePitman:
## Why
disabledItems currently returns an array that includes filtered-out items as `undefined`. This causes `disabledItems.length` to return unexpected results, causing confusion for the consumer.

## What
Strips `disabledItems` of all undefined values, leaving only items that are filtered in.


NatalieLoyzaga (approved):
Nice! LGTM 🥳

PR #3017, opened 2022-10-12
feat(timepicker): add timepicker component

jshli:
Why
Perf-UI needs a time picker component that is localised, accessible and allows custom user input

What
A time picker built using a number of React-Spectrum libraries. 

image
image
image


packages/time-picker/src/TimePicker/components/DateSegment/DateSegment.tsx
```diff
@@ -0,0 +1,40 @@
...
+      className={classNames(
+        styles.segment,
+        segment.type === "literal" && styles.literal,
+        segment.isPlaceholder && styles.placeholder,
+        segment.type === "dayPeriod" && styles.dayPeriod
+      )}
+    >
+      {segment.isPlaceholder && segment.type !== "dayPeriod"
+        ? "--"
+        : segment.text}
```
HeartSquared, line 35: This could use a test. Perhaps abstract it into a util.

HeartSquared (dismissed):

HeartSquared (approved):

PR #3018, opened 2022-10-13
fix: remove incorrect design tokens type location

ckychris:
## Why
`design-tokens@10.0.3` has exported types in different locations, `package.json` has not been updated.

## What
Updated `package.json` to contain no types key, limiting one file pattern. Instead, allow typescript to find types file automatically.

## Referenece 
https://cultureamp.slack.com/archives/C02NUQ27G56/p1664939298139269

### DIFF
--- a/packages/design-tokens/package.json
+++ b/packages/design-tokens/package.json
@@ -19,7 +19,6 @@
     "!**/*.spec.*",
     "!docs"
   ],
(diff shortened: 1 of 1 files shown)


ckychris:
Failing because of branch name length, reopened with short branch name 😢

PR #3020, opened 2022-10-13
chore(yarn.lock): updated the yarn.lock to reflect new package versions

gyfchong:
## Why
Yarn lock fell out of sync with packages.

## What
updates the versions of installed packages in yarn lock


HeartSquared (approved):
Thanks Geoff! 🙌🏻

PR #3030, opened 2022-10-18
feat: Add 'control' to FilterMultiSelect search input

OliBridgman:
## Why
We want to use the FilterMultiSelect with asynchronous data. To do so, we need a way to either control or observe the search input of the dropdown. The easier option was to observe the search input. This is some brittle code, and if you really want control I think you'd need to re-imagine how this component thinks about the world.

We also need a way to get more results. I have added a load more button. This is presentational, and could do _whatever you want it to do!_. I imagine people will use it to... Load more.

## What

* I added react-query to the dev dependencies! 😲 
* I added react-querys queryClient to the "top level" storybook preview, rather than the selects story. Adding it to the select story was never instantiating the queryClient in the tree and causing the page to crash. I suspect this has something to do with stories being lazily loaded.
* Added a LoadMoreButton. Nothing in this package seemed to use a `@kaizen/Button`, so I didn't either.
* Added an optional `onSearchInputChange` prop to the `MultiSelectFilter`.
    * This prop turns off filtering for the component
    * This prop calls back to the consumer when the searchInput is changed
* Added some async "docs"
    * Controlled query via opening the dropdown, searching and pagination


packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -334,3 +335,92 @@ export const DefaultKaizenSiteDemoWithoutScrollbar = () => {
...
+              people
+            )}
+            label={"People"}
+          />
+        )}
+        onSearchInputChange={searchInput => setSearchState(searchInput)}
+        onOpenChange={isOpen => setOpen(isOpen)}
+        onSelectionChange={keys => {
+          if (keys === "all") {
+            return
```
OliBridgman, line 395: I don't know why keys can be `all`. It sends back all the keys when you select all. It doesn't send this string.

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -334,3 +335,92 @@ export const DefaultKaizenSiteDemoWithoutScrollbar = () => {
...
+          }
+          setSelectedPeople(Array.from(keys) as string[])
+        }}
+        isOpen={open}
+        selectedKeys={new Set(selectedPeople)}
```
OliBridgman, line 400: Sets everywhere. Why does this thing need Sets.

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -334,3 +335,92 @@ export const DefaultKaizenSiteDemoWithoutScrollbar = () => {
...
+      ({ pageParam = 1 }) =>
+        fetch(
+          `https://swapi.dev/api/people/?page=${pageParam}&search=${searchState}`
+        ).then(res => res.json()) as Promise<{
+          results: Array<{ name: string; url: string }>
+          next: string
+        }>,
+      {
+        enabled: open,
+        keepPreviousData: true,
```
OliBridgman, line 355: Keeping previous data is important here. Currently the `loadingSkeleton` prop blocks the entire component from being interacted with, including the search input. However, if you try and render something more controlled:
```
            {isLoading ? ( <Skeleton />) : (
            <>
              <FilterMultiSelect.ListBox>
                {({ allItems }) =>
                  allItems.map(item => (
                    <FilterMultiSelect.Option key={item.key} item={item} />
                  ))
                }
              </FilterMultiSelect.ListBox>
              {hasNextPage && (
                <FilterMultiSelect.LoadMoreButton onClick={() => fetchNextPage()}>
                  {isFetchingNextPage ? "Fetching..." : "Load more"}
                </FilterMultiSelect.LoadMoreButton>
              )}
              <FilterMultiSelect.MenuFooter>
                <FilterMultiSelect.SelectAllButton />
                <FilterMultiSelect.ClearButton />
              </FilterMultiSelect.MenuFooter>
              <>
              )}
            </>
```
The whole component blows up.

OliBridgman: Also `FilterMultiSelect.MenuLoadingSkeleton` isn't a component. I do not know why.

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -334,3 +335,92 @@ export const DefaultKaizenSiteDemoWithoutScrollbar = () => {
...
+            <FilterMultiSelect.ListBox>
+              {({ allItems }) =>
+                allItems.map(item => (
+                  <FilterMultiSelect.Option key={item.key} item={item} />
+                ))
+              }
+            </FilterMultiSelect.ListBox>
+            {hasNextPage && (
+              <FilterMultiSelect.LoadMoreButton onClick={() => fetchNextPage()}>
+                {isFetchingNextPage ? "Fetching..." : "Load more"}
```
dougmacknz, line 414: Some copy updates here to be in line with our guidance https://cultureamp.design/guidelines/infinite-scroll/

```suggestion
                {isFetchingNextPage ? "Loading…" : "View more"}
```

packages/select/package.json
```diff
@@ -56,7 +56,8 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "@tanstack/react-query": "^4.12.0"
```
dougmacknz, line 60: Given you've added the react-query stuff to the top level storybook preview, this would make make more sense in the root `package.json` now

packages/select/docs/FilterMultiSelect.stories.tsx
dougmacknz: > Also `FilterMultiSelect.MenuLoadingSkeleton` isn't a component. I do not know why.

Hm yeah, that might have been a mistake.

packages/select/src/FilterMultiSelect/components/LoadMoreButton/LoadMoreButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+import cx from "classnames"
+import styles from "./LoadMoreButton.module.scss"
+
+export const LoadMoreButton = React.forwardRef<
+  HTMLButtonElement,
+  JSX.IntrinsicElements["button"]
+>((props, ref) => {
+  const { children, className, ...otherProps } = props
+  return (
+    <button
```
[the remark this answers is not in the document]
OliBridgman: Updated.

packages/select/docs/FilterMultiSelect.stories.tsx
OliBridgman: Updated

packages/select/package.json
OliBridgman: Updated.

dougmacknz (approved):

PR #3033, opened 2022-10-19
fix(renamed timepicker to timefield): using patch as it only has 1 consuming team

gyfchong:
## Why
Design has noted that this input is no longer a picker but rather a field

## What
Re-names the TimePicker to TimeField.
Despite this normally being a breaking change, only 1 team uses it at the moment so it is submitted as a patch.


packages/date-picker/docs/TimePicker.stories.tsx
```diff
@@ -54,22 +54,22 @@ const StickerSheetTemplate: Story = () => {
     <StoryWrapper>
       <StoryWrapper.RowHeader headings={["Default", "Disabled", "Error"]} />
       <StoryWrapper.Row rowTitle="Input">
-        <TimePicker
+        <TimeField
           id="timepicker-default"
```
HeartSquared, line 58: You'll probably want to rename these ids too :)

gyfchong: Curiosity: why do we have id's here?

HeartSquared: We require ids for our form fields because they're used to tie the fields to their label for a11y :)

gyfchong: Ah right, that makes sense haha 😅

packages/date-picker/docs/TimePicker.stories.tsx
```diff
@@ -3,16 +3,16 @@ import { ComponentMeta, ComponentStory, Story } from "@storybook/react"
 import { withDesign } from "storybook-addon-designs"
 import { CATEGORIES } from "../../../storybook/constants"
 import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
-import { TimePicker } from "../index"
-import { ValueType } from "../src/TimePicker/types"
+import { TimeField } from "../index"
```
HeartSquared, line 6: You'll want to rename this file :)

NatalieLoyzaga (approved):
🥳

PR #3039, opened 2022-10-21
chore: Added XSS warning to RTE default story [KDS-526]

christian314159:
## Why
- We wanted to call out that dealing with user input from our Rich Text Editor can be a possible security concern as this content can't be trusted. There is a potential for it to include XSS or injection attacks.

## What
- Added a warning to the default story
image

### DIFF
--- a/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
+++ b/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
@@ -30,12 +30,13 @@ export const Default = args => {
       <Box pt={1}>
         <InlineNotification
           type="cautionary"
-          title="Security Hint"
+          title="Security hint"
           hideCloseIcon
         >
           Our Rich Text Editor deals with user input which can potentially
-          include malicious content (e.g. XSS, injection attacks). It should
-          always be treated with great care when processing or displaying again.
+          include malicious content (e.g. XSS, injection attacks). The data
+          returned by the RTE should hence always be treated with appropriate
+          caution when being processed or rendered by your application.
         </InlineNotification>
       </Box>
     </>
--- a/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
+++ b/packages/rich-text-editor/docs/RichTextEditor.stories.tsx
@@ -1,5 +1,7 @@
 import React, { useState } from "react"
 import { RichTextEditor, EditorContentArray } from "@kaizen/rich-text-editor"
+import { Box } from "@kaizen/component-library"
+import { InlineNotification } from "@kaizen/notification"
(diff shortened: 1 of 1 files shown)


packages/rich-text-editor/docs/RichTextEditor.stories.tsx
```diff
@@ -19,11 +21,24 @@ export default {
...
+        <InlineNotification
+          type="cautionary"
+          title="Security Hint"
+          hideCloseIcon
+        >
+          Our Rich Text Editor deals with user input which can potentially
+          include malicious content (e.g. XSS, injection attacks). It should
+          always be treated with great care when processing or displaying again.
+        </InlineNotification>
+      </Box>
```
gyfchong, line 40: Should this be part of the story? Concerned some may take this to mean they need to include it when they use the RTE for build.

christian314159: You mean like those birthday cakes where the instruction is "Please write something funny on it" and they literally write "Please write something funny on it" on the cake? ;)
I'm not too worried about this, also can't really think of a better place in our current docs.

jmitsios:
@christian314159
- Going by our [language style and punctuation](https://cultureamp.design/language/grammar/) conventions, let's change 'Hint' to use a lowercase h, 'hint'.
- Also I just wanted to question (and call out) the final word in the message how it says "again".  Is this intentional, or am I maybe not interpreting the message correctly?
The message reads a bit confusingly to me when I reach that word. 
- I'm wondering if these should be treated similarly to our validation messages, how they sit closer to the text field.   And will this current message impact a validation message with a potential double-up?

christian314159:
@jmitsios 

> Going by our [language style and punctuation](https://cultureamp.design/language/grammar/) conventions, let's change 'Hint' to use a lowercase h, 'hint'.

Will make that change.

> Also I just wanted to question (and call out) the final word in the message how it says "again". Is this intentional, or am I maybe not interpreting the message correctly? The message reads a bit confusingly to me when I reach that word.

Good call, "again" is intentional, but I think I can improve it by saying "... when processing or rendering elsewhere".

> I'm wondering if these should be treated similarly to our validation messages, how they sit closer to the text field. And will this current message impact a validation message with a potential double-up?

Ha, now I'm wondering if you fell into the trap that Geoff pointed out above (and I wasn't worried about) - this warning message is purely for an engineer looking at the component in Storybook, it's not meant to be part of an RTE implementation in an application - and hence there's no worry about it clashing with other messages.

gyfchong:
Decided not to add this message any more. Probably table it until we do some docs re-working

PR #3041, opened 2022-10-21
fix(splitbutton): explicitly setting the text color and text-decoration on hover

jared-luke:
## Why

Adding strict CSS hover rules for the `SplitButton` component to prevent the murmur global styles effecting the appearance of the component.

## What

I have explicitly set the text `color` on hover for both the default and reversed themes, plus set the `text-decoration` to be `unset`


packages/split-button/src/SplitButton/components/BaseButton/BaseButton.module.scss
```diff
@@ -48,9 +49,10 @@
 }
 
 .themeDefault {
+  $default-text-color: $color-purple-800
```
HeartSquared, line 52:
```suggestion
  $default-text-color: $color-purple-800;
```

packages/split-button/src/SplitButton/components/BaseButton/BaseButton.module.scss
```diff
@@ -73,9 +76,10 @@
 }
 
 .themeReversed {
+  $reversed-text-color: $color-white
```
HeartSquared, line 79:
```suggestion
  $reversed-text-color: $color-white;
```

HeartSquared (approved):

PR #3043, opened 2022-10-23
fix(searchfield): update onClear to return focus to input

NatalieLoyzaga:
## Why

- When user selects the clear button the focus does not return to the input
https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-611

## What

- Add a ref to input and force focus back onto the input when the onClear has been called.

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
@@ -1,4 +1,4 @@
-import React, { InputHTMLAttributes } from "react"
+import React, { InputHTMLAttributes, useRef } from "react"
 import classnames from "classnames"
 import { OverrideClassName } from "@kaizen/component-base"
 import { Icon } from "@kaizen/component-library"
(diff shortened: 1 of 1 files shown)


draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
```diff
@@ -32,7 +32,12 @@ export const InputSearch: React.VFC<InputSearchProps> = (
     secondary = false,
     ...restProps
   } = props
+  const inputRef = useRef<HTMLInputElement>(null)
 
+  const handleOnClear = (): void => {
+    inputRef.current?.focus()
+    onClear && onClear()
```
gyfchong, line 39:
```suggestion
    onClear?.()
```

gyfchong (approved):

draft-packages/form/KaizenDraft/Form/Primitives/InputSearch/InputSearch.tsx
NatalieLoyzaga: oo awesome nice to know - what do you call this?

gyfchong: It's another way to use Optional Chaining:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

PR #3045, opened 2022-10-24
fix: styles for link for RTE and RTC

louiseliang:
## Why

In our product design review for [career paths](https://github.com/cultureamp/career-pathways), we found that the style of links in RichTextContent is browser default and also found there is [figma design ](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%9A%A7-UI-Kit%3A-Beta-Components?node-id=23929%3A115344)for it. 

This PR is to simply include the style.

## What

RTC: 
image
RTE:
image

Please note that it won't affect any style added from RTE: 
image


packages/rich-text-editor/src/_mixins.scss
```diff
@@ -32,6 +33,11 @@
     }
   }
 
+  a {
+    color: $color-blue-500;
+    text-decoration: none;
```
mcwinter07, line 38: We may want to provide a hover state - I believe our default is just to add an underline on hover based off the Figma example [here](https://cultureamp.design/guidelines/link-vs-button/) - just confirming with @jmitsios

louiseliang: Updated according Jason's feedback. Could I please have another reviews? cc @mcwinter07 Thanks!

dougmacknz (approved):

louiseliang:
Thanks @dougmacknz!

PR #3049, opened 2022-10-25
docs(icons): add story to display all icons

NatalieLoyzaga:
## Why

- To display an up to date library of our icons

## What

- created a icons.ts file with all icons imported
- created a new all icons story


packages/component-library/stories/Icons.ts
```diff
@@ -0,0 +1,379 @@
...
+import trash from "@kaizen/component-library/icons/trash.icon.svg"
+import refresh from "@kaizen/component-library/icons/refresh.icon.svg"
+import sync from "@kaizen/component-library/icons/sync.icon.svg"
+import duplicate from "@kaizen/component-library/icons/duplicate.icon.svg"
+import undo from "@kaizen/component-library/icons/undo.icon.svg"
+import redo from "@kaizen/component-library/icons/redo.icon.svg"
+import send from "@kaizen/component-library/icons/send.icon.svg"
+import print from "@kaizen/component-library/icons/print.icon.svg"
+import configure from "@kaizen/component-library/icons/configure.icon.svg"
+import flagOn from "@kaizen/component-library/icons/flag-on.icon.svg"
```
christian314159, line 16: It's going to be great to have the icons in Storybook, but I think we can make it even easier for an engineer to import the icon that they need.

Look at the `flagOff` icon. How would the engineer know to import it like so?
```
import flagOffIcon from "@kaizen/component-library/icons/flag-off.icon.svg"
```
image

As an idea, could we put some info at the top of the stickersheet saying something like:
```
To import an icon use this import statement and replace the file name with the name under the icon:
import myIcon from "@kaizen/component-library/icons/{FILE_NAME}.icon.svg"
```

packages/component-library/stories/Icons.ts
```diff
@@ -0,0 +1,379 @@
...
+// DIRECTIONAL
+import arrowLeft from "@kaizen/component-library/icons/arrow-left.icon.svg"
+import arrowRight from "@kaizen/component-library/icons/arrow-right.icon.svg"
+import arrowDown from "@kaizen/component-library/icons/arrow-down.icon.svg"
+import arrowUp from "@kaizen/component-library/icons/arrow-up.icon.svg"
+import chevronDown from "@kaizen/component-library/icons/chevron-down.icon.svg"
+import chevronUp from "@kaizen/component-library/icons/chevron-up.icon.svg"
+import sortDescending from "@kaizen/component-library/icons/sort-descending.icon.svg"
+import sortAscending from "@kaizen/component-library/icons/sort-ascending.icon.svg"
+// import reverse from "@kaizen/component-library/icons/reverse.icon.svg"
```
christian314159, line 146: Why is this one (and some other ones) commented out?

packages/component-library/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@kaizen/component-base": "^1.1.1",
+    "@kaizen/component-library": "16.1.4",
```
christian314159, line 35: 🤔 Why does it have a dependency on itself?

NatalieLoyzaga: silly me was importing through the package rather than the file paths. I have updated now

packages/component-library/stories/Icons.ts
NatalieLoyzaga: They were ones that were missing from the component lib but are in figma - i have removed and made note.

packages/component-library/stories/Icons.ts
gyfchong: Perhaps another thing we could do is [copy the icon name to their clipboard](https://www.npmjs.com/package/copy-to-clipboard) when clicked

packages/component-library/stories/Icons.ts
christian314159: Great, is that "note" as in "card in Jira"? ;) If not, could you please put one up for those, thanks!

packages/component-library/stories/Icons.ts
christian314159: @gyfchong That's a great idea. I've create [a card](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-916) to track it, as I think it'd be helpful for all docs we have.

christian314159 (approved):

PR #3050, opened 2022-10-26
fix(FilterMultiSelect): Fix loading skeleton spacing

NatalieLoyzaga:
## Why

- spacing was broken

## What

- add loading story
- updated styling
- use loadingInput for buttons

image


dougmacknz (approved):
Nice!

PR #3051, opened 2022-10-26
feat(TitleBlockZen): update collapseNavigationArea styles

louiseliang:
## Why

[KDS-913](https://cultureamp.atlassian.net/browse/KDS-913)

Feature request from #team_act that allow TitleBlockZen to have a fully collapsed navigation area when:
- no navigation tabs
- no secondary items

[figma](https://www.figma.com/file/On69lVBEBbz87HxVw6ynCY/%F0%9F%9A%A7-Develop-2022-H2-%C2%B7%C2%A0Career-paths?node-id=10243%3A107087)
image

## What

changes:
- update the collapsed height to be 0
- remove border bottom when no sectionTitle and sectionTitleDescription and collapseNavigationArea = true 

image


HeartSquared (commented):
Since this is somewhat a new variant, it's best to add it to Storybook and also send it to Chromatic :)

Edit: Just realised there technically is as `Default (collapsed navigation)`

HeartSquared:
This seems to break usages when combined with Skirt. I think we need to check with @jmitsios if this is okay or not.
https://www.chromatic.com/test?appId=60a1e4a102f0cb003b5d19d6&id=6358a13c2f657f23d79edc62

HeartSquared:
Checked with @jmitsios and this was his response:
> That’s much better, and something I’ve been hoping to see fixed for a while.
From the example in Chromatic it looks like it’s only pulling the skirt up the same amount of space that’s being removed above.  That isn’t an issue to me.

HeartSquared (approved):

PR #3053, opened 2022-10-26
fix(deps): import brand as package

harrisrachel:
## Why
When creating the Expert Advice Collapsible, the Brand was imported incorrectly, making it not usable for consumers.

## What
Correct the way Brand is being imported into the component.


mcwinter07 (approved):
Nice catch! Thanks for the fix 👍

PR #3054, opened 2022-10-28
feat: Tailwind design tokens canary release

mcwinter07:
## Why
Canary release for current iteration of Kaizen tokens to test the end to end use of the design tokens update


packages/design-tokens/src/tailwind/tailwind-presets.ts
```diff
@@ -0,0 +1,126 @@
+// ts-check
+// import {Config} from "tailwindcss"
+import { defaultTheme } from "../"
+
+// TODO: create a color mapper since you can't just pass in destructure colour into here
```
mcwinter07, line 5: This is no longer true ^ will remove this and all other comments in this file later

packages/design-tokens/src/tailwind/tailwind-presets.ts
```diff
@@ -0,0 +1,126 @@
+// ts-check
+// import {Config} from "tailwindcss"
```
mcwinter07, line 2: Currently using this to incrementally check the updated presets, before ready for production I think it would be best to create our own type or interface. If we were to use the native config object we would also have to make tailwind a dependency of our design-tokens, which I don't know if we'd want that

.vscode/settings.json
```diff
@@ -22,5 +22,11 @@ Happy linting! 💖
   "playwright.env": {
     "IS_DEV": true
   },
-  "playwright.reuseBrowser": true
+  "playwright.reuseBrowser": true,
+  "tailwindCSS.classAttributes": [
```
mcwinter07, line 26: This allows the Tailwind Plugins intellisense pick up in the following string values

storybook/tailwind.scss
```diff
@@ -0,0 +1 @@
+@tailwind utilities;
```
mcwinter07, line 1: We could also add `@tailwind base;` here - tailwinds reset for native browser styles but I've held off for the moment in case we want to advise differently for our consumers

mcwinter07: You can see the issues with not providing a reset here:
Screen Shot 2022-11-03 at 12 08 56 pm

JakePitman (approved):

PR #3055, opened 2022-10-28
refactor(date-picker): abstract DatePicker subcomponents

HeartSquared:
## What

Refactor the subcomponents for DatePicker.
- Splits `Calendar` into `CalendarSingle` and `CalendarRange`
- Style `CalendarRange` as per UI Kit (double calendar, with potential divider)
- Rename `CalendarWrapper` to `FloatingCalendarWrapper` to give more usage context
- Abstract `DateInput` into `DateInput`, `DateInputWithButton`, `DateInputField`
- Adds tests and stories for abstracted components

## Why

To prepare for FilterDateRangePicker work (split from [POC here](https://github.com/cultureamp/kaizen-design-system/compare/KDS-750/filter-date-range-picker-base)).


packages/date-picker/src/_subcomponents/FloatingCalendarWrapper/FloatingCalendarWrapper.tsx
```diff
@@ -2,18 +2,22 @@ import React, { HTMLAttributes, useState } from "react"
...
+export interface FloatingCalendarWrapperProps
   extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
-  children: React.ReactElement<typeof Calendar>
+  children: React.ReactNode
   referenceElement: HTMLElement | null
+  popperOptions?: Partial<Options>
 }
 
-export const CalendarWrapper: React.VFC<CalendarWrapperProps> = ({
+export const FloatingCalendarWrapper: React.VFC<
```
gyfchong, line 15: Since there's no other `CalendarWrapper` it seems odd to call this out as a `Floating` one, perhaps just `CalendarPopper` would suffice since it implements the `usePopper` hook, or `Popover` to be clearer of it's use.

packages/date-picker/src/_subcomponents/Calendar/LegacyCalendarRange.tsx
```diff
@@ -0,0 +1,61 @@
...
+  classNameOverride?: string
+  defaultMonth?: Date
+  weekStartsOn?: DayOfWeek
+  disabledDays?: Matcher[]
+  selectedRange?: DateRange
+  locale: Locale
+  onDayChange: DayClickEventHandler
+}
+
+export const LegacyCalendarRange: React.VFC<LegacyCalendarRangeProps> = ({
```
gyfchong, line 25: This looks largely the same as your new `CalendarRange` why did it get set as `Legacy*`?

packages/date-picker/src/DatePicker/components/DateInputField/DateInputField.tsx
```diff
@@ -0,0 +1,79 @@
...
+   */
+  status?: FieldMessageStatus
+  /**
+   * A descriptive message for `status` states
+   */
+  validationMessage?: React.ReactNode
+  locale: Locale
+}
+
+export const DateInputField = React.forwardRef<
```
gyfchong, line 28: Why not `DateField` to be consistent with `TimeField` etc?

packages/date-picker/src/DatePicker/DatePicker.tsx
```diff
@@ -211,7 +177,7 @@ export const DatePicker: React.VFC<DatePickerProps> = ({
 
   const handleInputClick: React.MouseEventHandler<HTMLInputElement> = e => {
     setIsOpen(true)
-    onInputClick && onInputClick(e)
+    onInputClick?.(e)
```
gyfchong, line 180: Nice one 👍

packages/date-picker/src/utils/validateDate.ts
```diff
@@ -53,12 +53,13 @@ export const validateDate = (
       validationResponse: {
         ...baseResponse,
         status: "error",
-        validationMessage: `${newInputValue} is not available, try another date`,
+        validationMessage: `${inputValue} is not available, try another date`,
```
gyfchong, line 56: Should this also check if `inputValue` is undefined?

HeartSquared: Nope because a valid disabled date is not empty :)

gyfchong: Ah right, didn't check the rest of the file 😅

gyfchong (dismissed):

packages/date-picker/docs/FloatingCalendarWrapper.stories.tsx
```diff
@@ -49,7 +46,7 @@ const FloatingCalenderWrapperExample = ({
     <>
       <div
         ref={setReferenceElement}
-        style={{ paddingTop: "24px", marginBottom: `${rowHeight}px` }}
+        style={{ paddingBottom: "24px", marginTop: `${rowHeight}px` }}
```
gyfchong, line 49: So simplistic haha

gyfchong:
Probably something we don't have to be concerned with as it's a subcomponent that no one is supposed to really reach for but this is the docs behaviour 😅 

https://www.loom.com/share/2d6b29d470b44459a6ab91342563b090

HeartSquared:
> Probably something we don't have to be concerned with as it's a subcomponent that no one is supposed to really reach for but this is the docs behaviour 😅

Yeah I think people tend to look at sticker sheets in Canvas? Or maybe that's just me 😂 Either way, this silly thing is just for Chromatic. I actually overrode the popper styles for the sticker sheet just so I can display them spaced out against each other.

gyfchong:
> I actually overrode the popper styles for the sticker sheet just so I can display them spaced out against each other.

Yeah initially I was concerned that the override will show up in the docs giving people a false example to follow, but seems like it doesn't, so all good!

gyfchong (approved):

PR #3056, opened 2022-10-28
fix: margin issue when avatr group has exact two avatars

Tommy-Chen-cultureamp:
## Why
https://cultureamp.atlassian.net/browse/KDS-915
In `rtl` direction, there is another margin on the right and it should have a margin on the left instead
image

## What
1. Make the margin on the left work in `rtl` direction by using `margin-left` instead of `margin-inline-start`
2. Reduce the extra margin on the left when there are 3 and more avatars

Screen Shot 2022-10-28 at 15 59 05
Screen Shot 2022-10-28 at 15 59 24


christian314159:
Hey @Tommy-Chen-cultureamp Thanks for this contribution! Would you be able to add a "RTL" section to the AvatarGroup stickersheet story as well (also one using only 2 avatars as that was an issue you found & fixed)?
This way it'll be included in our Chromatic visual regression testing in the future.

HeartSquared:
I tried to see if we could use positioning to replicate the required styles (didn't work), but I've got this branch with a slightly different approach that removes some excess styles :)
https://github.com/cultureamp/kaizen-design-system/compare/KDS-915/fix-avatar-group-rtl-margin

Tommy-Chen-cultureamp:
@HeartSquared That looks absolutely better! Thanks for your suggestion!

Tommy-Chen-cultureamp:
Hey @christian314159 I am not sure how to add `dir='rtl'` to `AvatarGroup`because `dir` is a native prop.

christian314159:
@Tommy-Chen-cultureamp I actually just wrapped the story examples in a `<div dir="rtl">...</div>` and that worked.

Tommy-Chen-cultureamp:
Morning @christian314159 @HeartSquared just ping you the PR is ready to be reviewed again

draft-packages/avatar/docs/AvatarGroup.stories.tsx
```diff
@@ -114,3 +115,21 @@ StickerSheetReversed.parameters = {
...
+)
+StickerSheetRtl.storyName = "Sticker Sheet (RTL)"
+
+const twoAvatars: AvatarList = [EXAMPLE_USER_1, EXAMPLE_USER_2]
+export const StickerSheetRtlWithTwoAvatars = () => (
+  <div dir="rtl">
+    <StickerSheetTemplateComponent isReversed={false} avatars={twoAvatars} />
+  </div>
+)
+StickerSheetRtlWithTwoAvatars.storyName = "Sticker Sheet (Two avatars, RTL)"
```
HeartSquared, line 135: Rather than adding extra sticker sheets, I'd suggest adding extra rows into StickerSheetTemplate instead. Visual regression testing can get costly as each snapshot adds cost, so we specifically try and keep them in a single story :)

HeartSquared (approved):

PR #3058, opened 2022-10-31
fix: Change MenuLoadingSkeleton to component

JakePitman:
## Why
To have the `FilterMultiSelect.MenuLoadingSkeleton` consistent with other components.

## What
Makes `FilterMultiSelect.MenuLoadingSkeleton` a component instead of just a `const` with JSX.


packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
```diff
@@ -49,12 +49,12 @@ export const DemographicValueSelect = ({
   const handleSelectionChange = (keys: Selection) => {
     onSelectionChange(getSelectedOptionKeys(keys, demographicValueItems))
   }
-
+  FilterMultiSelect.MenuLoadingSkeleton
```
gyfchong, line 52: What is this used for?

JakePitman: Nothing - forgot to remove

packages/select/src/FilterMultiSelect/components/MenuLayout/MenuLoadingSkeleton/MenuLoadingSkeleton.tsx
```diff
@@ -4,7 +4,7 @@ import classNames from "classnames"
 import { MenuFooter } from "../MenuFooter"
 import styles from "./MenuLoadingSkeleton.module.scss"
 
-export const MenuLoadingSkeleton = (
+export const MenuLoadingSkeleton = () => (
```
NatalieLoyzaga, line 7: `export const MenuLoadingSkeleton: React.VFC = () => (`

Just to keep consistent with our other components?

JakePitman: Yeah sure - I noticed that `.VFC` is deprecated in favour of `.FC` though - do we want to use that instead?

NatalieLoyzaga: Yeah we need to still use .VFC because of compatibility with react 16 _(something like that)_

JakePitman: Ah cool, thanks Nat

NatalieLoyzaga (approved):
💯

PR #3063, opened 2022-11-02
fix(scene): remove HumanityAtWork illustration

JakePitman:
BREAKING CHANGE: HumanityAtWork and DataCatching scenes are no longer exported from draft-illustration

## Why
These illustrations were never intended for production use.

## What
Removes HumanityAtWork and DataCatching scenes.


draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
```diff
@@ -89,7 +89,7 @@ DefaultStory.argTypes = {
     options: ["spot", "scene"],
     mapping: {
       spot: <Informative alt="" />,
-      scene: <HumanityAtWork alt="" />,
+      scene: <BrandMomentPositiveOutro alt="" />,
```
gyfchong, line 92: If `BrandMomentPositiveOutro` is a replacement for `HumanityAtWork`, should we leave a deprecation notice for those who've already used `HumanityAtWork`?

NatalieLoyzaga (approved):

PR #3067, opened 2022-11-07
fix: Ensure legacy safari and legacy edge users don't call missing methods

OliBridgman:
Some of our users are using older versions of safari which do not have the `addEventListener` method on `matchMedia`.

There is a deprecated `addListener` function, but I'm following what was established in `useMediaQueries` and stopping the effect from running, rather than supporting old browsers.


NatalieLoyzaga (approved):

NatalieLoyzaga (approved):

gyfchong:
Thanks for this Oli!

PR #3068, opened 2022-11-07
feat(Select): Kds 925 basic single select

NatalieLoyzaga:
## Why

- Current Select is a shim over react-select and does not meet accessibility requirements as well as the up to date design
- Make the new Select composable

## What

Basic Single Click Select
- Accessible
- `placeholder` is required and will display if `selectedKey` is null
- `isFullWidth`: Sets the Button and ListBox to fullWidth (To be used when wanting a custom width)
- Button width by default is calculated by the longest `Item`
- User can select ONE `Item`
- Uses `SingleSelection` types from react aria
- Added interaction states to stickersheet for both select button and dropdown

**Updated Design:**
- tick icon to signify selected item
- Dropdown menu width is now 248px
- Trigger Button min-width is 180px or full width if set


packages/select/src/Select/_mixins.scss
```diff
@@ -0,0 +1,9 @@
+// reset user agent styles for button element type
+@mixin button-reset {
```
gyfchong, line 2: Is this not shared with the original button component?

packages/select/src/Select/components/MenuPopup/MenuPopup.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useOverlay, DismissButton } from "@react-aria/overlays"
+import { FocusScope } from "@react-aria/focus"
+import { mergeProps } from "@react-aria/utils"
+import { useMenuTriggerContext } from "../../provider"
+import styles from "./MenuPopup.module.scss"
+export interface MenuPopupProps {
+  children: React.ReactNode
+}
+
+export const MenuPopup: React.VFC<MenuPopupProps> = ({ children }) => {
```
gyfchong, line 12: So we can easily find them for a holistic approach later it would be good to rename this to `FloatingSelectWrapper` to match @HeartSquared's naming for Calendar: https://github.com/cultureamp/kaizen-design-system/tree/master/packages/date-picker/src/_subcomponents/FloatingCalendarWrapper

packages/select/src/FilterMultiSelect/components/ListBox/ListBox.module.scss
```diff
@@ -7,7 +7,7 @@
   padding-right: $spacing-sm;
   margin: 0 $spacing-sm 0 0;
   display: grid;
-  max-height: 210px;
+  max-height: 344px;
```
gyfchong, line 10: Actually, would it make sense to have this scale with zoom? I'm assuming the text inside will scale and then eventually get crushed inside the max height... Thoughts @dougmacknz?

dougmacknz: Ah the age old question. Yeah I think I'd go for rems here. Some thoughts on this topic here https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#widths-and-heights

gyfchong: Yeah great article

packages/select/src/Select/components/FloatingSelectWrapper/FloatingSelectWrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { FocusScope } from "@react-aria/focus"
+import { useOverlay, DismissButton } from "@react-aria/overlays"
+import { mergeProps } from "@react-aria/utils"
+import { useMenuTriggerContext } from "../../provider"
+import styles from "./FloatingSelectWrapper.module.scss"
+export interface FloatingSelectWrapperProps {
+  children: React.ReactNode
+}
+
+export const FloatingSelectWrapper: React.VFC<FloatingSelectWrapperProps> = ({
```
MrBaptastic, line 12: if you use React.FC instead of VFC, you wont need FloatingSelectWrapperProps as FC comes with the children prop

HeartSquared: We haven't changed to `FC` yet because of backwards compatibility. The old `FC` includes `children` which we don't want. When we no longer support the old version (I think 16), we can port it over.

packages/select/src/Select/components/Option/Option.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export interface OptionProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
+  item: Node<ItemType>
+}
+
+export const Option: React.VFC<OptionProps> = ({ item, classNameOverride }) => {
+  const { selectionState: state } = useSelectionContext()
+  // Get props for the option element
+  const ref = React.createRef<HTMLLIElement>()
```
MrBaptastic, line 22: My understanding is that this would create a new ref on re-render, it might be more advisable to use useRef() instead

packages/select/src/Select/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { OverrideClassName } from "@kaizen/component-base"
+import { Icon } from "@kaizen/component-library"
+import chevronDown from "@kaizen/component-library/icons/chevron-down.icon.svg"
+import chevronUp from "@kaizen/component-library/icons/chevron-up.icon.svg"
+
+import { useMenuTriggerContext } from "../../../provider/MenuTriggerProvider"
+import styles from "./TriggerButtonBase.module.scss"
+
+export interface TriggerButtonBaseProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
```
MrBaptastic, line 14: should it be button element type ?
```suggestion
  extends OverrideClassName<HTMLAttributes<HTMLButtonElement>> {
```

packages/select/src/Select/provider/MenuTriggerProvider/MenuTriggerProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onOpenChange,
+  children,
+  isDisabled = false,
+  isFullWidth = false,
+}: MenuTriggerProviderProps) {
+  // Create state based on the incoming props to manage the open/close
+  const state = useMenuTriggerState({ isOpen, defaultOpen, onOpenChange })
+
+  // Get A11y attributes and events for the menu trigger and menu elements
+  const ref = React.createRef<HTMLButtonElement>()
```
MrBaptastic, line 45: Might need to use useRef here to prevent it from changing on re-render

packages/select/src/Select/components/Trigger/SingleTriggerButton/SingleTriggerButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+      selectedOptionLabel === null && styles.placeholder,
+      classNameOverride,
+    ])}
+  >
+    {
+      <span>
+        {selectedOptionLabel != null
+          ? `${selectedOptionLabel}`
+          : `${placeholder}`}
+      </span>
```
MrBaptastic, line 29:
```suggestion
<span>{`${selectedOptionLabel ?? placeholder}`}</span>
```
Nice little trick https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

HeartSquared: You might not need the `${}` too :)

packages/select/src/Select/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
```diff
@@ -0,0 +1,50 @@
...
+  children: React.ReactNode
+}
+
+export const TriggerButtonBase: React.VFC<TriggerButtonBaseProps> = ({
```
HeartSquared, line 18: Question: Is this a Filter button? Or something else? 
Because there's one inside the `FilterMultiSelect`, so it's very confusing when they share the same name if they are not the same thing (I understand we are not applying DRY yet because our Filter pattern needs more thought).

packages/select/src/Select/_mixins.scss
HeartSquared: Yes, we don't want to share styles between packages :)

gyfchong: I guess another bonus of the All-in-One lol

packages/select/src/Select/components/Option/Option.tsx
gyfchong: I'm thinking that we can't have a null Ref passed to the component, and that's why we need to create one?

packages/select/src/Select/_mixins.scss
HeartSquared: Yessiree. Alt, we have some other way of applying reset styles. Perhaps we just export a stylesheet of resets.

packages/select/src/Select/components/Option/Option.tsx
MrBaptastic: https://beta.reactjs.org/apis/react/createRef#createref new doc is recommending against using createRef for function components, you can still create a ref with useRef()

`const ref = React.useRef< HTMLLIElement >(null);` should work here

packages/select/src/Select/components/FloatingSelectWrapper/FloatingSelectWrapper.tsx
MrBaptastic: @HeartSquared could we not use the React type ? then there is no need to change it later ?
```suggestion
export const FloatingSelectWrapper = ({ children }: FloatingSelectWrapperProps) => {
```

HeartSquared: Technically yes, however you'll also need to put in the return type of `JSX.Element` (which `FC/VFC` do for you). Currently it allows you to infer the return type, but I do want to make our linting stricter in the future. Typing the function removes the need to put in the return type (since it's part of the function signature).

I'm not actually opposed to either method as long as we are consistent. Currently the `VFC` pattern is what we've opted for, so we should keep to it until we revise our patterns.

packages/select/src/Select/components/Root/Root.tsx
```diff
@@ -0,0 +1,100 @@
...
+        classNameOverride={classnames([
+          !isFullWidth && rootStyles.notFullWidth,
+        ])}
+      >
+        {label}
+      </Label>
+      <MenuTriggerProvider {...menuTriggerProps}>
+        <div className={rootStyles.container}>
+          <HiddenSelectWrapper items={items} label={label} name={id} />
+          <MenuTriggerConsumer>{trigger}</MenuTriggerConsumer>
```
MrBaptastic, line 85: trigger (TriggerButtonBase) use the hook `useMenuTriggerContext` inside the component, you don't need to `MenuTriggerConsumer` hoc around it to make it work.

if you remove `MenuTriggerConsumer` you need to change the type `trigger: (value?: MenuTriggerProviderContextType) => React.ReactNode` to `trigger: (value?: MenuTriggerProviderContextType) => React.ReactNode`

the same can be said about `SelectionConsumer`

if you need a be more context around this we can huddle on it :)

packages/select/src/Select/components/Trigger/SingleTriggerButton/SingleTriggerButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { TriggerButtonBase } from "../TriggerButtonBase"
+import styles from "./SingleTriggerButton.module.scss"
+
+export type SingleTriggerButtonProps = {
+  placeholder: string
+  selectedOptionLabel: RootProps["selectedKey"]
+  classNameOverride?: string // TODO: migrate it to use OverrideClassName<T> and omit the props controlled by React-Aria
+}
+
+export const SingleTriggerButton: React.VFC<SingleTriggerButtonProps> = ({
```
MrBaptastic, line 13: I understand this a pattern copied from FilterMultiSelect which a bit more complicated that just the span with text, you could move this logic to TriggerButtonBase and have one less component to maintain

packages/select/src/Select/components/FloatingSelectWrapper/FloatingSelectWrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+      style={{
+        width: isFullWidth ? "100%" : undefined,
+        minWidth: !isFullWidth ? "248px" : undefined,
+      }}
+    >
+      <FocusScope contain autoFocus restoreFocus>
+        <DismissButton onDismiss={onClose} />
+
+        {children}
+        <DismissButton onDismiss={onClose} />
```
MrBaptastic, line 49: there are 2 `<DismissButton onDismiss={onClose} />` is that expected ?

packages/select/src/Select/components/ListBox/ListBox.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const ListBox: React.VFC<ListBoxProps> = ({ children }) => {
+  const { listBoxProps, listRef, selectionState } = useSelectionContext()
+  const [isOverflown, setIsOverflown] = useState(false)
+  useEffect(() => {
+    const listElement = listRef.current
+    if (!listElement) {
+      return
+    }
+    setIsOverflown(listElement.scrollHeight > listElement.clientHeight)
```
gyfchong, line 24: This seems excessive to remove the margin, why does the margin exist in the first place?

packages/select/src/Select/components/Root/Root.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <Label
+        htmlFor={id}
+        aria-label={label}
+        classNameOverride={classnames([
+          !isFullWidth && rootStyles.notFullWidth,
+        ])}
+      >
+        {label}
+      </Label>
+      <MenuTriggerProvider {...menuTriggerProps}>
```
gyfchong, line 78: Why does the MenuTriggerProvider wrap the entire select?

packages/select/src/Select/components/ListBox/ListBox.tsx
MrBaptastic: I just checked, the overflow class which add `margin-right: 0;` has no effect on the component, all that logic can be removed

packages/select/src/Select/components/Root/Root.tsx
MrBaptastic: aside from the div right underneath, all the other components use this context

gyfchong: Ah okay I see now. I think we need to re-name this provider to something like `SelectProvider` since it's actually storing most of the data about the entire select, not just the menu trigger.

And by extension the `SelectionProvider` is a bit confusing as well, why not have it all in one provider?

MrBaptastic: This provider only has context around the when the menu is open or not, SelectionProvider store the state of the menu items and what is item is selected, It could be possible to remove the contexts and use render props to pass the relevant informations but that would require some refactoring

gyfchong: I think the context is okay to keep, since multiple parts of the internal components need to more or less be sync'd with the props provided, but I don't see the need for two of them.

NatalieLoyzaga: @gyfchong We can refactor this to be using one Select Provider but involves quite a bit of refactor. We were thinking this could be a next iteration card and then merge this in? Unless you think this work should be done on this PR

cc: @MrBaptastic

gyfchong (approved):

gyfchong:
There's a curly brace on the loose!
image

NatalieLoyzaga:
@gyfchong Updated now :)

gyfchong (approved):

gyfchong:
https://github.com/cultureamp/kaizen-discourse/issues/24

PR #3087, opened 2022-11-15
chore(deps): update node.js to v16

gyfchong:
## Why
Upgrading to v18 caused [Playwright to fail](https://buildkite.com/culture-amp/kaizen-design-system/builds/35915#01847926-d7b8-478b-9bbc-cf7dbb1bc685) and because v18 is not a requirement right now, the decision is to stick to 16 for now.

## What
- Node versions are changed from 14 to 16
- Audit JS removed due to incompatibility with node upgrade + superseded by Snyk


mcwinter07 (approved):
Looks 🦖 🐜 ! Merge in good health

PR #3088, opened 2022-11-15
refactor: remove deprecated compoonent lib helpers and redundant deps

mcwinter07:
## Why
 - Component used deprecated component library helpers which blocks us from removing this package entirely
 - was using older typography mixins which do not have the right spacing and can cause inconsistency with designs

## What
- Replace deprecated kz- and ca- with modern tokens
- fix broken story but updated calc function in scss
- minor visual clean up


draft-packages/hero-card/docs/HeroCard.stories.tsx
```diff
@@ -14,7 +14,7 @@ const renderContent = () => (
       height: "300px",
     }}
   >
-    <p style={{ margin: "0" }}>
+    <p style={{ margin: "0 0 0.75rem 0" }}>
```
gyfchong, line 17: Interesting, why did the margin change here?

gyfchong (approved):

draft-packages/hero-card/docs/HeroCard.stories.tsx
mcwinter07: I added a minor change... couldn't abide by the story not actually having any spacing between a typographical element and a button 😆

mcwinter07: For context a before and after:
Screen Shot 2022-11-15 at 2 44 13 pm

PR #3096, opened 2022-11-15
docs: update storybook to deploy docs

gyfchong:
## Why
We want to make deprecation documentation more discoverable

## What
In a first step, storybook will be deployed in 'docs' mode for consumers


HeartSquared (commented):
The intro page is nice, but I'm still not a fan of when I change from Docs to Canvas then move to another story, it changes back to Docs instead of staying on Canvas :(

PR #3097, opened 2022-11-16
chore(deps): remove gatsby dep from renovate ignore + add node to renovate ignore

gyfchong:
## Why
We aren't moving to 18, so I added it to the ignore list.
#3087

We also don't use gatsby anymore.

## What
Renovate ignore deps are updated


HeartSquared (approved):
Thanks boss ☕

PR #3098, opened 2022-11-16
feat: update indicatorsContainer select styles

NatalieLoyzaga:
## Why

- add styles to the globally

## What

- add styles to the globally


gyfchong (approved):
Actually checked it this time 😆

PR #3100, opened 2022-11-16
fix(KDS-988): remove grid from confirmation modal

jshli:
## Why
The confirmation modal had a gap to the left of the body, causing incorrect text alignment. 

## What
Remove the grid styling from the body of the modal. 

## Context
UI Kit:
Screen Shot 2022-11-16 at 2 29 51 pm

Kaizen:
Screen Shot 2022-11-16 at 2 30 57 pm


HeartSquared:
Just a note that `style:` is for insignificant code changes like removing white space, and not for styles :P This is more of a `feat`

mcwinter07:
> Just a note that `style:` is for insignificant code changes like removing white space, and not for styles :P This is more of a `feat`

How about `refactor:` @HeartSquared ? Don't know if this would qualify as a `feat`

gyfchong:
Technically a `fix:` I think, since the design was incorrect, and now aligns.

mcwinter07 (approved):

PR #3104, opened 2022-11-16
feat: update toolkit location + snapshots

gyfchong:
## Why
Moving the @cultureamp/rich-text-toolkit into kaizen-design-system to remove deps on @cutureamp private packages, and simplify the stack. Going forward this will help us communicate to consumers the optimal path for rich text usage all in one place.

## What
Copy of https://github.com/cultureamp/rich-text-toolkit made in kaizen


packages/rich-text-editor/src/toolkit/plugins/LinkManager/types.d.ts
```diff
@@ -0,0 +1,17 @@
+import { Transaction, Command } from "prosemirror-state"
+import { EditorView } from "prosemirror-view"
+
+export type MaybeCommand = Command | Transaction
+
+export type Dispatcher = (tx: Transaction) => void
+
+export class CAEditorView extends EditorView {
+  dispatch(maybeCommand: MaybeCommand): void
```
gyfchong, line 9: This is a weird one, where dispatch throws this error if the file isn't a type definition file, so I've had to not ignore this file on https://github.com/cultureamp/kaizen-design-system/pull/3104/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R11

```
Dispatch a transaction. Will call dispatchTransaction when given, and otherwise defaults to applying the transaction to the current state and calling updateState with the result. This method is bound to the view instance, so that it can be easily passed around.

Function implementation is missing or not immediately following the declaration.
```

Any ideas @HeartSquared ?

HeartSquared: These files likely should not exist locally in dev. `*.d.ts` files are usually created during compiling for distribution.

@mcwinter07 how much do you know about this file?

gyfchong: Yeah seems odd, cause the original toolkit commits this too

gyfchong: Also the file importing this doesn't work without it

HeartSquared: Try renaming the file to `types.ts` instead

gyfchong: The error is from when it's a regular `types.ts` 😞

mcwinter07: @HeartSquared Honestly don't remember too much about the LinkManager, that was more a Doug baby 👶  The error that's getting thrown looks to be Prosemirror rather than us. As far as the .d.ts I feel like that's to support its use in the prosemirror plugin system, so could be required to be consumed by the larger component?

gyfchong:
Closed, reattempt another time

PR #3112, opened 2022-11-18
feat: add isLoading to SearchInput and story

NatalieLoyzaga:
## Why

- Add the ability to pass the isLoading state to the searchInput

## What

- Add isLoading prop to `SearchInput` which uses the Kaizen `InputSearch` which has the loading spinner inbuilt
- Update the story to pull out the `isRefetching` boolean from `useInfiniteQuery` and pass to the `SearchInput`
- Check that the searchState is not `"" `and `isRefetching` to avoid displaying the loading spinner when the search is empty initially


gyfchong (approved):
This didn't seem urgent? 😅

NatalieLoyzaga:
I was intrigued :P

PR #3117, opened 2022-11-24
refactor(storybook): update sticker sheet component

HeartSquared:
## What

Add new StickerSheet component which uses `<table>`.

## Why

[KDS-1025](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1025)

The existing `grid`-based component forces children to be full width even when they shouldn't be, and also forces RTL examples to float right as opposed to being aligned to their LTR counterparts.

Also allows the flexibility of not having a row title, but instead a section heading.

#### SplitButton
|EXISTING|NEW|
|---|---|
|image|image|

#### FilterTriggerButton (from FilterDRP branch)
|EXISTING|NEW|NEW v2|
|---|---|---|
|image|image|image|


NatalieLoyzaga:
So is the plan to have both and then slowly go through the previous ones to convert over?

HeartSquared:
> So is the plan to have both and then slowly go through the previous ones to convert over?

Yep, because I don't think I can keep up with all the new ones to make all the changes here 😅 Also then this PR would be HUGE

NatalieLoyzaga:
Yeah no i didn't think we would do it in this PR - Well looks good 👍

NatalieLoyzaga (approved):

PR #3118, opened 2022-11-24
feat(date-picker): add FilterDateRangePicker

HeartSquared:
## What

Add new component FilterDateRangePicker.

- Removes version lock for `component-library` and updates to latest
- Export `Matcher` from types.ts as it is used often
- Add `FilterTriggerButton` (adapted from FilterMultiSelect)

image

## Why

Iterative release phase 1.

### DIFF
--- a/packages/date-picker/docs/FilterDateRangePicker.stories.tsx
+++ b/packages/date-picker/docs/FilterDateRangePicker.stories.tsx
@@ -2,6 +2,7 @@ import React, { useState } from "react"
 import { ComponentStory, Story } from "@storybook/react"
 import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
+import { figmaEmbed } from "../../../storybook/helpers"
 import { FilterDateRangePicker } from "../src/FilterDateRangePicker"
 import { DateRange } from "../src/types"
 import { defaultMonthControls } from "./controls/defaultMonthControls"
@@ -21,6 +22,9 @@ export default {
           'import { FilterDateRangePicker } from "@kaizen/date-picker"',
       },
     },
+    ...figmaEmbed(
+      "https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=45526%3A98760&t=67y0ZhpSLd5K7Rpn-4"
+    ),
   },
   args: {
     label: "Dates",
--- a/packages/date-picker/docs/FilterDateRangePicker.stories.tsx
+++ b/packages/date-picker/docs/FilterDateRangePicker.stories.tsx
@@ -1,0 +1,99 @@
+import React, { useState } from "react"
+import { ComponentStory, Story } from "@storybook/react"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
+import { CATEGORIES, SUB_CATEGORIES } from "../../../storybook/constants"
+import { FilterDateRangePicker } from "../src/FilterDateRangePicker"
+import { DateRange } from "../src/types"
+import { defaultMonthControls } from "./controls/defaultMonthControls"
+import { disabledDayMatchersControls } from "./controls/disabledDayMatchersControls"
+import { dateRangePickerLocaleControls } from "./controls/localeControls"
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.datePicker}/Filter Date Range Picker`,
+  component: FilterDateRangePicker,
+  parameters: {
+    actions: {
+      argTypesRegex: "^on.*",
+    },
+    docs: {
+      description: {
+        component:
+          'import { FilterDateRangePicker } from "@kaizen/date-picker"',
+      },
+    },
+  },
+  args: {
+    label: "Dates",
+    locale: "en-AU",
+  },
+  argTypes: {
+    ...dateRangePickerLocaleControls,
+    ...defaultMonthControls,
+    ...disabledDayMatchersControls,
+    description: {
+      control: "text",
+    },
+  },
+}
+
+export const DefaultStory: ComponentStory<
+  typeof FilterDateRangePicker
+> = props => {
+  const [range, setRange] = useState<DateRange | undefined>()
+
+  return (
+    <FilterDateRangePicker
+      {...props}
+      selectedRange={range}
+      onRangeChange={setRange}
+    />
+  )
+}
+DefaultStory.storyName = "Filter Date Range Picker"
+DefaultStory.parameters = {
+  docs: { source: { type: "code" } },
+}
+DefaultStory.args = {
+  id: "filter-drp--default",
+}
+
+const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
+  isReversed,
+}) => {
+  const [rangeDefault, setRangeDefault] = useState<DateRange | undefined>()
+  const [rangeExisting, setRangeExisting] = useState<DateRange | undefined>({
+    from: new Date("2022-05-15"),
+    to: new Date("2022-06-22"),
+  })
+
+  return (
+    <StoryWrapper isReversed={isReversed}>
+      <StoryWrapper.Row rowTitle="Default">
+        <FilterDateRangePicker
+          id="stickersheet--filter-drp--default"
+          label="Dates"
+          locale="en-AU"
+          selectedRange={rangeDefault}
+          onRangeChange={setRangeDefault}
+        />
+      </StoryWrapper.Row>
+
+      <StoryWrapper.Row rowTitle="Existing value">
+        <FilterDateRangePicker
+          id="stickersheet--filter-drp--existing"
+          label="Dates"
+          locale="en-AU"
+          selectedRange={rangeExisting}
+          onRangeChange={setRangeExisting}
+        />
+      </StoryWrapper.Row>
+    </StoryWrapper>
+  )
+}
+
+export const StickerSheetDefault = StickerSheetTemplate.bind({})
+StickerSheetDefault.storyName = "Sticker Sheet (Default)"
+StickerSheetDefault.parameters = {
+  chromatic: { disable: false },
+  controls: { disable: true },
+}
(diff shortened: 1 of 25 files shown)


packages/date-picker/docs/FilterDateRangePicker.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const [rangeDefault, setRangeDefault] = useState<DateRange | undefined>()
+  const [rangeExisting, setRangeExisting] = useState<DateRange | undefined>({
+    from: new Date("2022-05-15"),
+    to: new Date("2022-06-22"),
+  })
+
+  return (
+    <StoryWrapper isReversed={isReversed}>
+      <StoryWrapper.Row rowTitle="Default">
+        <div>
```
HeartSquared, line 72: Our `StoryWrapper` component forces the contents into full width even if they aren't meant to. Will remove these `div`s when it is fixed ([draft in progress](https://github.com/cultureamp/kaizen-design-system/pull/3117)).

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -0,0 +1,104 @@
...
+    onRangeChange(dateRange)
+  }
+
+  const handleCalendarSelectRange: CalendarRangeProps["onSelect"] = range => {
+    handleDateRangeChange(range)
+  }
+
+  return (
+    <>
+      <FilterTriggerButton
```
HeartSquared, line 70: Will likely break this and the wrapper out later into some sort of Filter components so that they can have variants (eg. removable button)

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+const TriggerButtonBaseWrapper = (props: Partial<TriggerButtonBaseProps>) => (
+  <TriggerButtonBase isOpen={false} {...props}>
+    Pancake
+  </TriggerButtonBase>
+)
+
+describe("<TriggerButtonBase />", () => {
+  it("has the required attributes when not expanded", () => {
+    render(<TriggerButtonBaseWrapper />)
+    const button = screen.getByRole("button", { name: "Open filter - Pancake" })
```
HeartSquared, line 14: @dougmacknz I had noticed that the button in FilterMultiSelect did not indicate whether clicking on the button was going to open or close the filter, so I added in something. I'm not sure if it's a good label or not though (prepending with `Open/Close filter`), so any advice?

In the actual case of the FilterDRP with a value, the label looks something like this: `Open filter - Dates: 1 May 2022 - 25 Nov 2022`, where `Dates` is the label and `1 May 2022 - 25 Nov 2022` is the selected value (which may or may not exist).

packages/date-picker/src/utils/formatDateAsText.ts
```diff
@@ -1,6 +1,6 @@
 import { format } from "date-fns"
```
gyfchong, line 1: Can we replace this lib with [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat)?

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.spec.tsx
dougmacknz: `aria-expanded=true|false` should be communicating the open/closed state of the popup, so we shouldn't need to add it to the label.

https://user-images.githubusercontent.com/1811583/203704108-c17d2ac4-59fa-4703-94c2-088f0179d862.mov

HeartSquared: I see! Makes sense. Do we want to do anything about the label and the selected values being mushed together in the name (especially since it changes)? Or is that fine?

dougmacknz: image

Just gave this a quick test on VoiceOver and ran into something interesting: VoiceOver must have detected that it looks like a date range and tried to read out something more nice, except it screwed it up. It said "The 25th of November to 2030 November 2022"

So yeah it's probably a good idea for us to override the contents label with an aria-label that reads nicer.

packages/date-picker/src/FilterDateRangePicker/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { TriggerButtonBase, TriggerButtonBaseProps } from "../TriggerButtonBase"
+import styles from "./FilterTriggerButton.module.scss"
+
+export interface FilterTriggerButtonProps
+  extends Omit<TriggerButtonBaseProps, "children"> {
+  label: string
+  selectedValue?: string
+}
+
+export const FilterTriggerButton = forwardRef<
```
gyfchong, line 11: Could we simplify the need for this abstraction?

Would something like this inside `FilterDateRangePicker.tsx` be enough:
```
<TriggerButton>
     {hasSelectedValue ? (
             <>
                 <span className={styles.hasSelectedValues}>{label}:&nbsp;</span>
                 <span>{selectedValue}</span>
             </>
      ) : (
        <span>{label}</span>
      )}
</TriggerButton>
```

So instead of the future state being about implementing a `XTriggerButton`, we just pass in different formatted labels depending on the filter's needs

HeartSquared: Yes that's in my next PR :) I copy pasted-ish from the FilterMultiSelect to just get it working first.

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
```diff
@@ -0,0 +1,35 @@
...
+    className={classNames(styles.button, classNameOverride)}
+    aria-haspopup="true"
+    aria-expanded={isOpen}
+    {...restProps}
+  >
+    {children}
+    <Icon
+      icon={isOpen ? chevronUp : chevronDown}
+      role="presentation"
+      classNameOverride={styles.icon}
```
gyfchong, line 30: Lol using our own `classNameOverride`, could we use a flex on button instead?

HeartSquared: Nah, we still need some consistent gap in it. Nothing wrong with using our own override though :) That's how styling and extension works :D

gyfchong: Adding `gap` would achieve that too, right?

> Nothing wrong with using our own override though :) That's how styling and extension works :D
Sure, but I'd argue that we shouldn't need to since most of the use cases for it can be solved with CSS on containers which is much more scalable.

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.spec.tsx
HeartSquared: :(
Any suggestions?

Just tried out a few attempts with @mcwinter07 and any usage of `-` (including unicode) as it is causes the problem. It worked nicely when I gave it an aria-label of `"to"`, but then it means we'd need to translate it, and that may not translate well in other languages.
We also tried wrapping in `time` and `data` tags, but the same problem happens.
When we used a different separator, the dates were read out correctly, but the separator was not announced.

Would it still make sense if we just read the two dates? Or does that now cause confusion and no longer seems like a date range?

dougmacknz: Oh right, I didn't think about internationalisation. Maybe I'll test out the current label in NVDA (known for being fairly similar to JAWS) and we'll see how this goes. If it's just a VoiceOver bug, we may want to just roll with this for now, given the much lower number of people using VoiceOver.

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.tsx
HeartSquared: Shh, stop being smart :P

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.spec.tsx
HeartSquared: @mcwinter07 just tested on Airbnb and the same problem persists. It might be a VoiceOver bug after all?

gyfchong: How about "Between {date} and {date}"

gyfchong (approved):

packages/date-picker/src/FilterDateRangePicker/components/Trigger/TriggerButtonBase/TriggerButtonBase.spec.tsx
HeartSquared: Requires translation though 😞

gyfchong: That can be handled in the solution for translating, as I understand a few other places aren't translated yet either?

gyfchong (approved):

PR #3121, opened 2022-11-28
fix: Hoist stories and fix breaking ts build issue

mcwinter07:
## Why
Increase ease of discoverability by reducing number of nested folders in storybook

## What
- update title names and story names to hoist the stories
- minor updates to to titles to CSS property like names
- add utilities function to easily consume tailwind preset in stories
- add documentation around local dev gotchas

### DIFF
--- a/packages/tailwind/docs/LeadingHeight.stories.tsx
+++ b/packages/tailwind/docs/LeadingHeight.stories.tsx
@@ -1,13 +1,12 @@
 import React from "react"
 import { Story } from "@storybook/react"
 import { Divider } from "@kaizen/draft-divider"
+import { kaizenTailwindTheme } from "@kaizen/tailwind"
 import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 import styles from "./styles.module.scss"
 
-import { kaizenTailwindTheme } from "@kaizen/tailwind"
-
 export default {
-  title: "Tailwind/Typography/Leading Height",
+  title: "Tailwind/Typography/Line Height",
   component: <div>Hello</div>,
   parameters: {
     docs: {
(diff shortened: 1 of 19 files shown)


packages/tailwind/docs/Shadow.stories.tsx
```diff
@@ -1,13 +1,12 @@
...
 import { Divider } from "@kaizen/draft-divider"
+import { kaizenTailwindTheme } from "@kaizen/tailwind"
 import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 import styles from "./styles.module.scss"
 
-import { kaizenTailwindTheme } from "@kaizen/tailwind"
-
 export default {
-  title: "Tailwind/Effects/Shadow",
+  title: "Tailwind/Effects/Box Shadow",
```
mcwinter07, line 9: Just changing this to box shadow to make it easier to find it a dev is searching for its equivalent

packages/tailwind/docs/LeadingHeight.stories.tsx
```diff
@@ -1,13 +1,12 @@
...
 import { Divider } from "@kaizen/draft-divider"
+import { kaizenTailwindTheme } from "@kaizen/tailwind"
 import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
 import styles from "./styles.module.scss"
 
-import { kaizenTailwindTheme } from "@kaizen/tailwind"
-
 export default {
-  title: "Tailwind/Typography/Leading Height",
+  title: "Tailwind/Typography/Line Height",
```
mcwinter07, line 9: Leading is a pretty design centric term, so while tailwind uses it  I think we should document it as line height

PR #3124, opened 2022-11-28
fix(draft-button): remove draft-button package

JakePitman:
BREAKING CHANGE: Removes the draft-button package

## Why
This package is deprecated

## What
Removes the draft-button package


gyfchong:
Oh, I think your build is failing cause it can no longer find `draft-button`?

PR #3129, opened 2022-11-29
feat(date-picker): allow FilterDateRangePicker to be removable

HeartSquared:
## What

Allow FilterDateRangePicker to be removable from a filter bar (add button to the right of the trigger button).

image

- Refactor filter buttons
- Add filter button with a tooltip
	- Add `@kaizen/draft-tooltip` to deps
- Add filter button group
- Add new prop `onRemoveFilter` to FilterDateRangePicker

## Why

[KDS-1010](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1010)
Iterative release for FilterDRP


gyfchong:
Troll: Should we also have an alert on click that asks them if they're sure? 😛

packages/date-picker/src/FilterDateRangePicker/components/Trigger/FilterButtonGroup/FilterButtonGroup.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const FilterButtonGroup: React.VFC<FilterButtonGroupProps> = ({
+  children,
+  classNameOverride,
+  ...restProps
+}) => {
+  const childCount = React.Children.count(children)
+
+  return (
+    <div
```
gyfchong, line 22: Add a `role=group`? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role

packages/date-picker/src/FilterDateRangePicker/components/Trigger/FilterButtonGroup/FilterButtonGroup.module.scss
```diff
@@ -0,0 +1,84 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/color";
+@import "../variables";
+
```
gyfchong, line 4: This whole file is like witchcraft haha

HeartSquared: That's essentially what scss is. Precisely why our product teams don't like styling :P

packages/date-picker/src/FilterDateRangePicker/components/Trigger/FilterButtonGroup/FilterButtonGroup.tsx
```diff
@@ -0,0 +1,44 @@
...
+      role="group"
+      className={classnames(styles.filterButtonGroup, classNameOverride)}
+      {...restProps}
+    >
+      {childCount === 1
+        ? children
```
gyfchong, line 28: Would it be a safe assumption that if someone wanted to use `FilterButtonGroup` that it would be more than one child? Otherwise they wouldn't use it?

gyfchong: Ah right, so this works if you do something like this:
```
<FilterButtonGroup>
  <Button>
  { condition && <Button> }
</FilterButtonGroup>
```

HeartSquared: Nope 😅 It's going to sound strange, but because you'd need Fragments for your example, it would actually fail because I didn't put in the magic recursive check to check within Fragments (I found a solution online, but didn't feel it was necessary to look into and implement yet).

```
// This has a single child, which is the fragment
<FilterButtonGroup>
  <>
    <Button>
    { condition && <Button> }
  </>
</FilterButtonGroup>
```

Which now when I think about it (after typing it out, then deleting it), it means that it doesn't work after all! Okay, no conditional buttons for anyone yet 😂 Let's put it in later when someone says they want it.

gyfchong: Haha, sounds good 😄

packages/date-picker/src/FilterDateRangePicker/components/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
```diff
@@ -1,33 +1,53 @@
...
   HTMLButtonElement,
   FilterTriggerButtonProps
->(({ label, selectedValue, ...restProps }, ref) => {
-  const hasSelectedValue = selectedValue && selectedValue !== ""
+>(
+  (
+    { label, selectedValue, isOpen = false, classNameOverride, ...restProps },
+    ref
+  ) => {
+    const hasSelectedValue = selectedValue && selectedValue !== ""
```
gyfchong, line 24: Could you apply the empty string type guard to prevent this from being an issue?

HeartSquared: This is essentially a type guard lol. Is there a util out there??

gyfchong: I think Nat wrote one, but probably not worth digging it up for this reason lol

gyfchong (approved):

PR #3130, opened 2022-11-29
reafactor(Select): Use one Select Provider and useSelect

NatalieLoyzaga:
## Why

- use the useSelect hook with a main select provider rather than two separate providers
- Simplifies the code 
- Type select functionality
- Refactor Trigger Button


packages/select/src/Select/provider/SelectProvider/SelectProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+        listBoxProps,
+        labelProps,
+        valueProps,
+        state,
+      }}
+    >
+      <div
+        className={classnames([
+          rootStyles.container,
+          !otherProps.isFullWidth && rootStyles.notFullWidth,
```
gyfchong, line 106: I know it's not entirely relevant to the PR, but `notFullWidth` doesn't feel very descriptive, can you find a better name?

PR #3132, opened 2022-11-30
fix(draft-button): remove draft-button package

JakePitman:
BREAKING CHANGE: Removes the draft-button package

## Why
This package is deprecated.

## What
Deletes the draft-button package.


christian314159 (approved):
🚢

PR #3136, opened 2022-12-01
feat: add tailwind beta release

mcwinter07:
## Why
We're splitting tailwind from the design-tokens package to ensure that versioning this will likely undergo as we continue to develop in its early stages does not impact our consumers of design tokens that are not using tailwind.

## What
- will release @kaizen/tailwind as 0.0.1
- will add kaizen presets, story tables and documentation on local development


tailwind.config.js
```diff
@@ -0,0 +1,15 @@
+// ts-check
+/** @type {import('tailwindcss').Config} */
+
+const kzPresets = require("@kaizen/tailwind").TailwindPreset
```
gyfchong, line 4: I wonder if naming this `TailwindPreset` is now redundant? Thoughts on calling it `Preset`?

mcwinter07: True, now it's in its own package probs doesn't need to be explicitly called out.

packages/tailwind/README.md
```diff
@@ -0,0 +1,5 @@
+# Install
```
gyfchong, line 1: Should include examples of tailwind config and VSCode recommendations

gyfchong: And a huge disclaimer that API is subject to change lol

mcwinter07: True that, there's some guidance in the FE Template I should link back too as well

JakePitman (approved):

gyfchong (approved):

PR #3137, opened 2022-12-01
feat: Add component for 'No results found' message

dougmacknz:
## What
Creates a new `FilterMultiSelect.NoResults` component for showing a "No results found for [search term]" message inside the dropdown.

## Why
So that users get a confirmation that there's no results, rather than just getting an empty list and wondering if it's broken, or still loading, etc.

### DIFF
--- a/packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
+++ b/packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
@@ -1,5 +1,6 @@
 import React, { Key, useEffect, useState } from "react"
 import { Selection } from "@react-types/shared"
+import { Paragraph } from "@kaizen/typography"
 import {
   FilterMultiSelect,
   getSelectedOptionKeys,
@@ -74,8 +75,18 @@ export const DemographicValueSelect = ({
         <>
           <FilterMultiSelect.SearchInput />
           <FilterMultiSelect.ListBox>
-            {({ selectedItems, unselectedItems, disabledItems }) => (
+            {({
+              selectedItems,
+              unselectedItems,
+              disabledItems,
+              hasNoItems,
+            }) => (
               <>
+                {hasNoItems && (
+                  <FilterMultiSelect.NoResults>
+                    No results found.
+                  </FilterMultiSelect.NoResults>
+                )}
                 <FilterMultiSelect.ListBoxSection
                   items={selectedItems}
                   sectionName="Selected items"
--- a/packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
+++ b/packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
@@ -1,6 +1,5 @@
 import React, { Key, useEffect, useState } from "react"
 import { Selection } from "@react-types/shared"
-import { Paragraph } from "@kaizen/typography"
 import {
   FilterMultiSelect,
   getSelectedOptionKeys,
--- a/packages/select/docs/FilterMultiSelect.stories.tsx
+++ b/packages/select/docs/FilterMultiSelect.stories.tsx
@@ -466,8 +466,17 @@ export const Async: ComponentStory<typeof FilterMultiSelect> = args => {
               isLoading={isRefetching && searchState !== ""}
             />
             <FilterMultiSelect.ListBox>
-              {({ allItems }) => (
+              {({ allItems, hasNoItems }) => (
                 <>
+                  {hasNoItems && (
+                    <FilterMultiSelect.NoResults>
+                      No results found for{" "}
+                      <Paragraph variant="extra-small" tag="span" color="dark">
+                        {searchState}
+                      </Paragraph>
+                      .
+                    </FilterMultiSelect.NoResults>
+                  )}
                   {allItems.map(item => (
                     <FilterMultiSelect.Option key={item.key} item={item} />
                   ))}
--- a/packages/select/docs/FilterMultiSelect.stories.tsx
+++ b/packages/select/docs/FilterMultiSelect.stories.tsx
@@ -63,11 +63,19 @@ export const DefaultKaizenSiteDemo: ComponentStory<
           <>
             <FilterMultiSelect.SearchInput />
             <FilterMultiSelect.ListBox>
-              {({ allItems }) =>
-                allItems.map(item => (
+              {({ allItems, hasNoItems }) => {
+                if (hasNoItems) {
+                  return (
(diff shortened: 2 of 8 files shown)


packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -60,11 +60,19 @@ export const DefaultKaizenSiteDemo: ComponentStory<
           <>
             <FilterMultiSelect.SearchInput />
             <FilterMultiSelect.ListBox>
-              {({ allItems }) =>
-                allItems.map(item => (
+              {({ allItems, hasNoItems }) => {
+                if (hasNoItems) {
+                  return (
+                    <FilterMultiSelect.NoResults>
+                      No results found.
```
dougmacknz, line 67: Unfortunately I wasn't able to add the search term to this message for the non-async examples because the state of the search field is all uncontrolled and we don't have access to it here.

If we want to tackle this, it's going to be a separate piece of work.

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
```diff
@@ -74,8 +75,18 @@ export const DemographicValueSelect = ({
...
+              unselectedItems,
+              disabledItems,
+              hasNoItems,
+            }) => (
               <>
+                {hasNoItems && (
+                  <FilterMultiSelect.NoResults>
+                    No results found.
+                  </FilterMultiSelect.NoResults>
+                )}
```
HeartSquared, line 89: Kinda annoying as a consumer to need to put this in 😞
Would it be possible to bake it into the `ListBox` or are there scenarios which would prevent it from working?

packages/select/src/FilterMultiSelect/components/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react"
```
HeartSquared, line 1:
```suggestion
import React from "react"
```

packages/select/src/FilterMultiSelect/components/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Paragraph } from "@kaizen/typography"
+import styles from "./NoResults.module.scss"
+
+interface NoResultsProps {
+  children: React.ReactNode
+}
+
+export const NoResults: React.VFC<NoResultsProps> = (props: NoResultsProps) => (
```
HeartSquared, line 9:
```suggestion
export const NoResults: React.VFC<NoResultsProps> = props => (
```
or
```suggestion
export const NoResults: React.VFC<NoResultsProps> = ({ children }) => (
```

packages/select/src/FilterMultiSelect/components/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react"
+import { Paragraph } from "@kaizen/typography"
+import styles from "./NoResults.module.scss"
+
+interface NoResultsProps {
```
HeartSquared, line 5:
```suggestion
interface NoResultsProps extends HTMLAttributes<HTMLDivElement> {
```

Then spread the rest of the props into the container :)

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
```diff
@@ -74,8 +74,18 @@ export const DemographicValueSelect = ({
...
+            {({
+              selectedItems,
+              unselectedItems,
+              disabledItems,
+              hasNoItems,
+            }) => (
               <>
+                {hasNoItems && (
+                  <FilterMultiSelect.NoResults>
+                    No results found.
```
gyfchong, line 86: I think this would be nicer as a default message, so the consumer wouldn't need to add this all the time, but give them the option to override it with a custom one if the designer wants to provide more context to the items.

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
dougmacknz: It would be possible to put a generic message like this in there if there's no items, but...
* It wouldn't be possible to do the thing I did with the async one where it says "No results found for [search term].", because `ListBox` doesn't know anything about `SearchInput`
* It would go against the API pattern already established for this component, and in general it would feel weird for `ListBox` to show a message that more relates to `SearchInput` when they're not tightly coupled.
* Makes it less likely that consumers will implement a translated string for this. If there's something built in that looks like it works fine, they're gonna be more likely to just leave it.

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
dougmacknz: Cass commented the same thing just below 👇 check out that thread

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
HeartSquared: I guess in the case where the consumer can't get the input text to put into the message, I'd have expected Kaizen to translate it on their behalf.

Nonetheless, this is fine since the other points make enough sense for it the way it has been built :)

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
HeartSquared: Since it's not baked into the component, and Doug has raised some other points about why it can't seem to live there, we can't really do it sadly :')

packages/select/src/FilterMultiSelect/components/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { HTMLAttributes } from "react"
+import { Paragraph } from "@kaizen/typography"
+import styles from "./NoResults.module.scss"
+
+interface NoResultsProps extends HTMLAttributes<HTMLDivElement> {
```
HeartSquared, line 5:
```suggestion
export interface NoResultsProps extends HTMLAttributes<HTMLDivElement> {
```

HeartSquared (approved):

packages/select/docs/FilterBarExample/DemographicValueSelect.tsx
gyfchong: Sadness

HeartSquared (approved):

PR #3138, opened 2022-12-01
feat: add classNameOverride to TableHeaderRowCell and TableRowCell.

kaydbwithers:
## Why
I have Table component shown that has a lot of content inside of it. The default 1.5rem side padding will not work. I've added classNameOverride so that I can use custom padding/no-padding in the Header and Body cells.
[KDS-1017](https://cultureamp.atlassian.net/browse/KDS-1017)

image


draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -84,36 +85,38 @@ const ratioToPercent = (width?: number) =>
...
-  /**
-   * Shows an up or down arrow, to show that the column is sorted.
-   */
-  sorting?: "ascending" | "descending"
-  wrapping?: "nowrap" | "wrap"
-  align?: "start" | "center" | "end"
-  tooltipInfo?: string
-  sortingArrowsOnHover?: "ascending" | "descending" | undefined
-}>
+type TableHeaderRowCell = OverrideClassName<React.FunctionComponent> &
```
HeartSquared, line 88: `OverrideClassName` is a generic which removes `className` from props and adds `classNameOverride` in its place. It is mostly used for when we extend HTML attributes, which don't seem to be the case here.

Options:
- Use `OverrideClassName` on the props within `React.FunctionComponent` (which we should use `React.VFC` instead)
- Just add the prop and not use the generic

I'd probably recommend you pull out the props into their own `interface` too and export it.

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -377,7 +387,8 @@ export const TableRow: React.VFC<TableRowProps> = ({
  *        shrink, and basis, due to IE11 compatibility. eg. use "1 1 auto"
  *        instead of just "1".
  */
-type TableRowCellProps = {
+interface TableRowCellProps
+  extends OverrideClassName<HTMLAttributes<HTMLSpanElement>> {
```
HeartSquared, line 391:
```suggestion
  extends OverrideClassName<HTMLAttributes<HTMLElement>> {
```

The props don't appear to be spread into a `span`, but either an `a` or `div`. `HTMLElement` is the most generic :)

draft-packages/table/KaizenDraft/Table/Table.tsx
JakePitman: I've removed the `OverrideClassName` generic - it's just using the prop now

kaydbwithers: @JakePitman really sorry, but I think our changes clashed and I overwrote yours. I think we had the same idea though. I've added a BaseTableProps interface to match the other files.

draft-packages/table/KaizenDraft/Table/Table.tsx
kaydbwithers: Hey @HeartSquared, thanks for your feedback. I've made those changes if you'd like to re-review.

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -112,9 +117,13 @@ type TableHeaderRowCell = React.FunctionComponent<{
   wrapping?: "nowrap" | "wrap"
   align?: "start" | "center" | "end"
   tooltipInfo?: string
+  classNameOverride?: string
```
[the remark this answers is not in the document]
kaydbwithers: Thanks @JakePitman, this has now been amended.

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -9,6 +10,13 @@ import { Tooltip } from "@kaizen/draft-tooltip"
 import { Heading } from "@kaizen/typography"
 import styles from "./Table.module.scss"
 
+export interface BaseTableProps
+  extends OverrideClassName<Omit<HTMLAttributes<HTMLElement>, "onChange">> {
```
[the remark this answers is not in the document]
kaydbwithers: Thanks for reviewing @JakePitman , no sorry, this shouldn't be here. I have removed it.

JakePitman (approved):

PR #3140, opened 2022-12-01
fix: Resolving clearing issue of legacy multi-select [KDS-1030]

christian314159:
## Why
- When trying to click the "x" to remove one of the tags from the legacy multi-select, it does not remove it but opens the dropdown instead
- It does however work fine using the keyboard

https://user-images.githubusercontent.com/763385/205181669-3fc6513c-95c7-4849-89c9-ac89a385a003.mov

## What
- I noticed that when trying to look at the `Tags` using the DevTools, it wouldn't select them, but the wrapping `div` they're in instead; I couldn't figure out exactly why, but somehow the rendering order is off, so I added a `z-index` to the `Tags` to get them to be rendered on the top level, which now allows the "x" remove button to be clicked

https://user-images.githubusercontent.com/763385/205181938-dfd59ef3-d65c-4165-838a-aa763f5b4b27.mov

### DIFF
--- a/draft-packages/select/KaizenDraft/Select/Select.module.scss
+++ b/draft-packages/select/KaizenDraft/Select/Select.module.scss
@@ -36,6 +36,7 @@ $focus-border-color: $color-blue-500;
 .specificityIncreaser {
   .multiValue {
     margin: 0.2375rem;
+    z-index: 1;
   }
 
   .dropdownIndicator {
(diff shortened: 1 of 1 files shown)


draft-packages/select/KaizenDraft/Select/Select.module.scss
```diff
@@ -98,6 +99,7 @@ $focus-border-color: $color-blue-500;
     border: 6px $border-solid-border-style transparent;
     border-radius: $border-solid-border-radius;
     box-shadow: $shadow-large-box-shadow;
+    z-index: 3;
```
christian314159, line 102: This is the menu dropdown. Because it had no previous `z-index` the elements with a `z-index` were "shining" through (you can see it in the first video with the chevrons).

HeartSquared, line 102: Why not 2 instead of jumping straight to 3? 😄

christian314159: Wanted to have some wiggle room, just in case :)

HeartSquared (approved):

PR #3143, opened 2022-12-06
chore(deps): Remove unused typescript-plugin-css-modules dep

dougmacknz:
This isn't being used, and looking at [this PR where it was removed from our fops tooling](https://github.com/cultureamp/frontend-ops/pull/1138), it's not something that works for us.


HeartSquared (approved):
❤️

PR #3153, opened 2022-12-08
chore(docs): Adjust Collapsible custom header example to have IconButtons

dougmacknz:
This comes up every now and then. People want to put extra actions into the Collapsible heading. We have a story showing how to do a custom header with Collapsible, but it just has an Icon. I've adjusted that to have IconButtons showing how to do actions, most notably using `event.stopPropagation()` to avoid the additional action triggering the collapse behaviour.

### DIFF
--- a/draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
+++ b/draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
@@ -1,9 +1,10 @@
 import React from "react"
 import { Story } from "@storybook/react"
-import { Icon } from "@kaizen/component-library"
-import translationIcon from "@kaizen/component-library/icons/translation.icon.svg"
+import { IconButton } from "@kaizen/button"
+import addIcon from "@kaizen/component-library/icons/add.icon.svg"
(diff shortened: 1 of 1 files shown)


draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
```diff
@@ -66,12 +67,27 @@ const SingleCollapsibleCustomHeader = () => (
     variant="default"
     renderHeader={title => (
       <>
-        <Icon icon={translationIcon} />
-        <div style={{ flex: "1 0 auto", marginLeft: "1rem" }}>
-          <Paragraph tag="span" variant="body">
+        <div style={{ flex: "1 0 auto" }}>
```
JakePitman, line 70: Should these just be in the stylesheet if we already have one? Or is it cool because it's a storybook file

dougmacknz: I tend to prefer to keep styles inline in stories, otherwise when someone is looking at this story to understand how to use something, they can't easily see what the class is actually doing without digging into the source code.

JakePitman (approved):

PR #3155, opened 2022-12-08
refactor(draft-loading-placeholder): remove component

NatalieLoyzaga:
…placeholder

## Why

- Deprecated component and it not being used within the codebase

## What

- Delete all code


dougmacknz:
I think the emoji in your branch name (wow it actually lets you do that) is causing build issues.

PR #3157, opened 2022-12-09
feat(date-picker): add inputs to FilterDateRangePicker

HeartSquared:
## What

Add inputs to FilterDateRangePicker

|OLD|NEW|
|---|---|
|image|image|

### Changes
- Fix focus rings for Calendar
- Move input description utils to shared dir
- Move the padding from Calendar to FloatingWrapper
- Add date range input field
	- Overlapping input styles
		- Note: Could not add pseudo state stories to Sticker Sheet as it requires refactoring the underlying Input component to have the story global styles (added new Jira card)
- Add hook `useDateInputHandlers` to control input behaviour (should be reusable for DatePicker - out of scope for this PR to update, will put in the backlog for if this approach is approved)

### Out of scope
- Validation and edge cases (eg. selecting start date that is after end date)
- Changing the calendar to move months according to updated valid input

## Why

[KDS-998](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-998)
FilterDateRangePicker iteration


packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.module.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  border-start-start-radius: 0;
+  border-end-start-radius: 0;
+
+  &:focus + div {
+    border-start-start-radius: 0;
+    border-end-start-radius: 0;
+
+    // Override position of focus ring
+    /* stylelint-disable-next-line declaration-no-important */
```
HeartSquared, line 44: I know this is a smell, but unless I refactor the Input component, I can't do anything about it :'(

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const shouldShowValidationMessage = !disabled && validationMessage
+
+    const inputDescribedBy = shouldShowValidationMessage
+      ? `${errorMessageId} ${descriptionId}`
+      : descriptionId
+
+    return (
+      <FieldGroup {...fieldGroupProps}>
+        <fieldset className={styles.dateRangeInputContainer}>
```
HeartSquared, line 78: @dougmacknz should I be wrapping the `fieldset` around the error message/description too? If so, I'm tempted to remove the `FieldGroup` wrapper, but is there something special about it that I should know making it worth keeping? I noticed some of our form components have it (although not all).

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <legend className={styles.legend}>{legend}</legend>
```
HeartSquared, line 79: I tried to use our `VisuallyHidden` component and it wrapped the `legend` in a `span` which removed the semantics :'D Thus I just copied the styles.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
dougmacknz: I think this is fine the way it is. Looking at `FieldGroup`, it's purely a styling thing, so feel free to remove it anyway if you want.

packages/date-picker/src/_subcomponents/Calendar/Calendar.module.scss
```diff
@@ -20,12 +46,16 @@ $rdp-cell-size: 40px;
 
 .button {
   cursor: pointer;
-  border-radius: 3px;
+  border-radius: $button-border-radius;
 
   &:focus {
     outline: none;
     border: 2px solid transparent;
```
fycdev, line 53: Is this a magic number? 🤔

🏃‍♂️

packages/date-picker/src/_subcomponents/FloatingCalendarWrapper/FloatingCalendarWrapper.module.scss
```diff
@@ -1,9 +1,11 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/design-tokens/sass/shadow";
+@import "~@kaizen/design-tokens/sass/spacing";
 
 .calendarWrapper {
   background-color: $color-white;
   z-index: 1010;
```
fycdev, line 7: This `z-index` appears arbitrary. Would this cause concern when laid out with other components?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
HeartSquared: It works! Updated :)

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.module.scss
```diff
@@ -0,0 +1,3 @@
+.filterDateRangePickerContainer {
+  display: contents;
```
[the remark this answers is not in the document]
HeartSquared: Hmm...In my case, the div is really just an invisible container because I wanted to put a data-testid on it for a story 😂 If I'm reading it right, the concern is that the semantics of the children may disappear? Or is it only for the container, in which case, it really is meant to not really be there. If I was to change it to not display contents, then I'd have to put extra styles on it and the children to ensure that the children still span the full width of the container. Not a lot of work, but since the intent is for it to simulate not existing, maybe it's acceptable to keep?

packages/date-picker/src/_subcomponents/Calendar/CalendarSingle.module.scss
```diff
@@ -1,8 +1,7 @@
 @import "~@kaizen/design-tokens/sass/color";
-@import "~@kaizen/design-tokens/sass/spacing";
 
 .month {
-  padding: $spacing-md;
+  padding: 0;
```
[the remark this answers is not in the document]
HeartSquared: Nope doesn't appear to have it :) I can remove it

packages/date-picker/src/_subcomponents/Calendar/Calendar.module.scss
```diff
@@ -20,11 +46,16 @@ $rdp-cell-size: 40px;
...
 
   &:focus {
     outline: none;
-    border: 2px solid transparent;
+    border: $border-borderless-border-width $border-borderless-border-style
+      $border-borderless-border-color;
+  }
+
+  &:focus-visible {
+    outline: none;
```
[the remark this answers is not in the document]
HeartSquared: Nope when I was tabbing through with my keyboard I found my focus rings to be extra thick and found it was caused by the outline on `:focus-visible`😅

packages/date-picker/src/_subcomponents/Calendar/Calendar.module.scss
```diff
@@ -4,8 +4,34 @@
 @import "~@kaizen/design-tokens/sass/border";
 @import "../../mixins";
 
+$button-border-radius: 3px;
 $rdp-cell-size: 40px;
 
+%focus-ring {
```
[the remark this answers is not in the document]
HeartSquared: Hmm...I could, but some of the places the mixins are used are within combined classes where things like `position: relative` and `&:focus { outline: none }` doesn't need to be repeated (eg. `.dayDisabled` is always used in combination with `.day`).

Although I guess it doesn't hurt to duplicate the styles 🤔

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+  describe("Disabled", () => {
+    it("disables both inputs", () => {
+      render(<DateRangeInputFieldWrapper disabled />)
+      const inputStart = screen.getByRole("textbox", { name: "Date from" })
+      const inputEnd = screen.getByRole("textbox", { name: "Date to" })
+      expect(inputStart).toBeDisabled()
+      expect(inputEnd).toBeDisabled()
+    })
+
+    it("does not show validation message when disabled", () => {
```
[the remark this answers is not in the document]
gyfchong: Lol same debate over on Select, I chose to allow error messages when the field is disabled, betting that this is a scenario that consumers (or form libs) would logically avoid anyway.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -0,0 +1,129 @@
...
+            status={status}
+            reversed={isReversed}
+          />
+        )}
+        <FieldMessage
+          id={descriptionId}
+          message={formatInputDescription(description, locale)}
+          reversed={isReversed}
+          classNameOverride={disabled ? styles.disabled : undefined}
+        />
```
gyfchong, line 124: Shouldn't descriptions come before validation messages?

HeartSquared: I don't believe that's how our components are? 🤔

gyfchong: Oops no you're right, this PR was wrong haha https://github.com/cultureamp/kaizen-design-system/pull/3169

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.spec.tsx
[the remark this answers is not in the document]
HeartSquared: Okay okay you both win. I'll remove it :P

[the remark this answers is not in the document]
HeartSquared: It is done.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.module.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.inputRangeEnd {
+  position: relative;
+  inset-inline-start: calc(-1 * #{$border-solid-border-width});
+
+  border-start-start-radius: 0;
+  border-end-start-radius: 0;
+
+  &:focus + div {
```
dougmacknz, line 29: Nit pick: these could be applied without the `:focus` part (I think).

Also, obviously the element reference here is prone to regressing, if someone were to change that div to a span with no knowledge of this separate component referencing this. You could do a wildcard selector for the class name "focus ring", which is still not ideal but a bit less prone to changing.

```
+ [class*="focusRing"] {
```

HeartSquared: I think ideally we'd refactor the Input component so I don't even need to do something horrible like this 😂 But yeah I suppose that's something that could be done 😬

dougmacknz (dismissed):

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  margin: 0;
+  padding: 0;
+  border: none;
+}
+
+.inputRangeStart {
+  border-start-end-radius: 0;
+  border-end-end-radius: 0;
+
+  &:focus + div {
```
dougmacknz, line 16: and this one? 😁

dougmacknz (approved):

PR #3163, opened 2022-12-12
WIP - flaky tooltips

HeartSquared:
## What

Testing to try fix Tooltip flakiness in Chromatic tests


HeartSquared:
Delays don't seem to fix the problem. The calculation just chooses to vary. Threshold diffs can't be applied to only a section of a story, so it's not really worth having these stories.

Closing in favour of https://github.com/cultureamp/kaizen-design-system/pull/3167

PR #3164, opened 2022-12-12
fix: RichTextEditor empty state value is now type safe

Danny-Duck:
For some reason ProseMirror would return an undefined when passed an empty array, which is technically allowed from the types. This change is just a guard rail.


packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -85,12 +85,12 @@ export const RichTextEditor: React.VFC<RichTextEditorProps> = props => {
...
-            ? ProseMirrorModel.Node.fromJSON(schema, {
-                type: "doc",
-                content: value,
-              })
-            : undefined,
+          doc: ProseMirrorModel.Node.fromJSON(schema, {
+            type: "doc",
+            // we're converting empty arrays to the ProseMirror default "empty" state because when
+            // given an empty array ProseMirror returns undefined, breaking the type
+            content: value.length > 0 ? value : [{ type: "paragraph" }],
```
mcwinter07, line 92: Can we add a null check against the value as well to make sure that isn't undefined
```
content: value?.length > 0 ? value : [{ type: "paragraph" }],
```

mcwinter07 (commented):
Hey @Danny-Duck   sorry for the delayed review of this, but I dug into why the tests are breaking this morning and we may just need to tweak them a little to get our checks to pass.

The main culprit is the `userEvent.click` in the `RichTextEditor.spec.tsx`:
```
...
// UL test
userEvent.click(screen.getByRole("button", { name: "Bullet List" }))
...
// OL test
userEvent.click(screen.getByRole("button", { name: "Numbered List" }))
```

If we change these to fireEvent.click that should resolve the issue.

The underlying cause seems to be how ProseMirror interacts with `testing-library`'s event API, which we've had some issues with before when trying to simulate user interaction. In this case, I think it's fine to change this since we do more extensive tests in Playwrite (which pass with this.

Give me a holler if you want to pair on this sometime this week, otherwise feel free to push up those changes and I'll review async :)

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
Danny-Duck: :galaxy-brain:

Danny-Duck:
Thanks for your guidance and patience @mcwinter07 💯

mcwinter07 (approved):

PR #3165, opened 2022-12-13
feat: update tokens presets

mcwinter07:
## Why
Update tokens for spacing, height and width following token catch up

## What
- Updates tailwind package to use the proposed new spacing tokens
- minor clean up and update to other spacing tokens


christian314159 (dismissed):
Happy to 🚢 !

packages/tailwind/src/kz-spacing.ts
```diff
@@ -0,0 +1,34 @@
+import { defaultTheme } from "@kaizen/design-tokens"
+
+export const kzSpacing = {
+  "spacing-2": ".125rem",
```
christian314159, line 4: I think we should add `spacing-0` (0px) and `spacing-1` (1px)

packages/tailwind/src/kz-spacing.ts
```diff
@@ -0,0 +1,34 @@
...
+  "spacing-320": "20rem",
+  "spacing-xs": defaultTheme.spacing.xs,
+  "spacing-sm": defaultTheme.spacing.sm,
+  "spacing-md": defaultTheme.spacing.md,
+  "spacing-lg": defaultTheme.spacing.lg,
+  "spacing-xl": defaultTheme.spacing.xl,
+  "spacing-xxl": defaultTheme.spacing.xxl,
+  "spacing-xxxl": defaultTheme.spacing.xxxl,
+  "spacing-xxxxl": defaultTheme.spacing.xxxxl,
+  "spacing-xxxxxl": defaultTheme.spacing.xxxxxl,
```
HeartSquared, line 33: Ohhh are we still keeping these? I thought we were replacing our tshirt tokens with the new ones

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -103,30 +95,58 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
...
+    "letter-spacing-normal": "normal",
   },
   // A mix of layout styles
   maxWidth: {
-    ...twDefaultTheme.maxWidth,
-    content: defaultTheme.layout.contentMaxWidth,
-    "content-with-side": defaultTheme.layout.contentMaxWidthWithSidebar,
+    "w-0": "0",
+    "w-25": "25%",
+    "w-50": "25%",
```
HeartSquared, line 104:
```suggestion
    "w-50": "50%",
```

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -103,30 +95,58 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
...
+    "w-min": "min-content",
+    "w-max": "max-content",
+    "w-fit": "fit-content",
   },
   width: {
-    ...twDefaultTheme.width,
-    "content-side-margin": defaultTheme.layout.contentSideMargin,
+    "w-0": "0",
+    "w-25": "25%",
+    "w-50": "25%",
```
HeartSquared, line 115:
```suggestion
    "w-50": "50%",
```
Since width and maxWidth seem to share the same values, it might be worth making an object?

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -103,30 +95,58 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
...
+    "w-100": "100%",
+    "w-auto": "auto",
+    "w-min": "min-content",
+    "w-max": "max-content",
+    "w-fit": "fit-content",
+  },
+  maxHeight: {
+    "h-0": "0",
+    "h-25": "25%",
+    "h-50": "25%",
```
HeartSquared, line 126:
```suggestion
    "h-50": "50%",
```

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -103,30 +95,58 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
...
+    "h-fit": "fit-content",
   },
   height: {
-    ...twDefaultTheme.height,
-    "mobile-actions-drawer-height":
-      defaultTheme.layout.mobileActionsDrawerHeight,
-    "navigation-bar-height": defaultTheme.layout.navigationBarHeight,
+    "h-0": "0",
+    "h-25": "25%",
+    "h-50": "25%",
```
HeartSquared, line 137:
```suggestion
    "h-50": "50%",
```

packages/tailwind/src/kz-spacing.ts
mcwinter07: While I'd like to remove it, I'm at two minds for it 🧠 🧠  If someone uses this but has a design that says spacing-xl will that cause confusion if it's not there?

My assumption is that spacing tokens will eventually be rolled out to the entire platform in figma but maybe leave this here as a stop gap?

packages/tailwind/src/tailwind-presets.ts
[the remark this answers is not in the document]
mcwinter07: Considered it but wasn't sure if it'd be necessary (though certainly couldn't hurt)

packages/tailwind/src/kz-spacing.ts
dougmacknz: > has a design that says spacing-xl will that cause confusion if it's not there

I don't there's any references to spacing tokens in Figma, if that's what you mean. So we'll be fine there.

I'd be strongly in favour of removing these. We can create a map from old token to new token if we want, but even then I kinda doubt that would end up being useful to people.

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -18,29 +20,29 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
     ...defaultTheme.dataViz,
   },
   spacing: {
-    ...defaultTheme.spacing,
-    ...twDefaultTheme.spacing,
+    ...kzSpacing,
   },
```
HeartSquared, line 24: You could actually do this to save a few lines :) Not against it if you find it more readable though. Just a suggestion 😄 

```suggestion
spacing: ksSpacing,
```

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -18,29 +20,29 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
...
-    "width-focus-ring": "2px",
+    "w-none": "0px",
+    "w-default": "2px",
+    "w-focus-ring": "2px",
   },
   // how do we want to handle our focus rings?
   // tailwind used a `ring` property
   borderColor: {
-    "color-solid": `${defaultTheme.border.solid.borderColor}`,
+    DEFAULT: `${defaultTheme.border.solid.borderColor}`,
```
HeartSquared, line 45: Curious about this. Does this mean the token is just `border-color` (or whatever it is) without a suffix?

Almost seems strange to even have a default solid border colour but that's a whole different conversation I think.

packages/tailwind/src/tailwind-presets.ts
```diff
@@ -52,81 +54,71 @@ export const kaizenTailwindTheme: KaizenTailwindTheme = {
     "family-display": [`${defaultTheme.typography.display0.fontFamily}`],
   },
   fontSize: {
-    "size-display": `${defaultTheme.typography.display0.fontSize}`,
```
HeartSquared, line 55: What's the rationale for removing `size-`? I thought the verbosity was clearer rather than guessing whether `font/text-display` came with anything more than just the size.

packages/tailwind/src/tailwind-presets.ts
mcwinter07: actually `border-color:` translate to `border${your-color-prefix}` in tw so it would just be `border`. We could leave it as 'color-default` and provide DEFAULT as the base one too 🤷 or just not use DEFAULT.

packages/tailwind/src/tailwind-presets.ts
mcwinter07: Based on feedback we've been getting in the interviews it doesn't seem like this is needed, outside of tokens that may collide with each other (ie: font-weight and font-family)

packages/tailwind/src/tailwind-presets.ts
mcwinter07: noice catch will update the pr

packages/tailwind/src/kz-spacing.ts
dougmacknz: We're gonna have to bite the bullet on this sooner or later, and now is gonna be the best time to do it.

packages/tailwind/src/tailwind-presets.ts
HeartSquared: So you're telling me that `border` is only `border-color`, so if we don't provide `DEFAULT` then `border` gives no styles? 😂 
I just wonder if it even makes sense to have a default border colour at all. Is that a pattern/guideline we have? That if we design anything new with a border, it will have this colour unless you choose to override it?

packages/tailwind/src/tailwind-presets.ts
HeartSquared: Interesting. I guess it just depends on if we choose to provide combined styles (eg. if `text-display` ends up being size, family, weight etc rolled in one). If we never plan to do so, then there won't be a conflict :)

packages/tailwind/src/tailwind-presets.ts
mcwinter07: The border utility can take a color, size and style as a prefix

packages/tailwind/src/kz-spacing.ts
JakePitman: We've removed these now 🥳

packages/tailwind/src/tailwind-presets.ts
mcwinter07: lul, just found out that as DEFAULT border does not do what we wanted, also we've both decided we should just flat out avoid defaults

mcwinter07 (commented):
✅  🚢  👯

JakePitman (approved):

PR #3169, opened 2022-12-14
feat(Select): add validation status and message

NatalieLoyzaga:
## Why

- Add Validation props to Select

## What

- Add `status` and `validationMessage` props
- Create validation story
image

### DIFF
--- a/packages/select/src/Select/Select.tsx
+++ b/packages/select/src/Select/Select.tsx
@@ -39,6 +39,14 @@ export interface SelectProps
     ref: React.RefObject<HTMLButtonElement>
   ) => React.ReactNode
   children?: (optionsProps: SelectOptionsProps) => React.ReactNode
+  /**
+   * Updates the styling of the validation FieldMessage.
+   */
+  status?: "error" | "caution"
+  /**
+   * A descriptive message for the 'status' states.
+   */
+  validationMessage?: React.ReactNode | undefined
 }
 
 export const Select: React.FC<SelectProps> & SubComponentProps = ({
@@ -47,6 +55,9 @@ export const Select: React.FC<SelectProps> & SubComponentProps = ({
   description,
   isFullWidth,
   placeholder,
+  isDisabled,
+  status,
+  validationMessage,
   classNameOverride,
   trigger = triggerProps => <TriggerButton {...triggerProps} ref={buttonRef} />,
   children,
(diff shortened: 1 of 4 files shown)


packages/select/src/Select/Select.tsx
```diff
@@ -39,6 +39,14 @@ export interface SelectProps
...
   ) => React.ReactNode
   children?: (optionsProps: SelectOptionsProps) => React.ReactNode
+  /**
+   * Updates the styling of the validation FieldMessage.
+   */
+  status?: "error" | "caution"
+  /**
+   * A descriptive message for the 'status' states.
+   */
+  validationMessage?: React.ReactNode | undefined
```
gyfchong, line 49: Should we make the validation message required if there's a status? There doesn't seem to be a default 🤔

NatalieLoyzaga: `type ValidationProps =
  | {
      status: "error" | "caution"
      validationMessage: React.ReactNode | undefined
    }
  | {
      status?: never
      validationMessage?: never
    }`

NatalieLoyzaga:
```
type ValidationProps =
  | {
      status: "error" | "caution"
      validationMessage: React.ReactNode | undefined
    }
  | {
      status?: never
      validationMessage?: never
    }
    
    ```

Comment (NatalieLoyzaga) [reply packages/select/src/Select/Select.tsx:49]: I could do this and then `export const Select: React.FC<SelectProps & ValidationProps>`

Comment (NatalieLoyzaga) [reply packages/select/src/Select/Select.tsx:49]: thoughts @gyfchong ? Means I can only have status + validation message or niether

Comment (gyfchong) [reply packages/select/src/Select/Select.tsx:49]: Hmm I just had a look at the others, doesn't seem like we add any restriction on this, so nevermind in this case. I'll pose the question for a more universal change 👍

**Review by gyfchong — DISMISSED** [review]

File: packages/select/src/Select/Select.tsx [hunk 2]
```diff
@@ -90,6 +111,8 @@ export const Select: React.FC<SelectProps> & SubComponentProps = ({
     }
   }, [state.isOpen])
 
+  const shouldShowValidationMessage = !isDisabled && validationMessage
```
gyfchong, line 114: Oh, according to design the message should show regardless of disabled state

NatalieLoyzaga: I think this maybe a mistake but I will ask Elliot

NatalieLoyzaga: wait what message are you seeing in the design?

The description message will show regardless the validation won't show if disabled

gyfchong: image

I assume this condition will prevent the 4th column from happening?

gyfchong (approved):

packages/select/src/Select/Select.tsx
dougmacknz: This is something @HeartSquared and I were just talking about. The date picker has this opinion built in, which I'm not a fan of. I reckon we should leave this up to consumers.

HeartSquared: This is a strange thing tbh. If it's disabled, does it even make sense to have any validation message on it? I feel like I remember having a chat with Jason about it earlier when we were working on the DatePicker and we figured it made no sense so we put the conditions in.
But yeah, I guess maybe we should just let them do it; weird as it is. Perhaps worth revisiting the whole validation structure pattern in general.

PR #3170, opened 2022-12-15
feat: Kds 930/disabled button options

NatalieLoyzaga:
## Why

- Add ability to add disabled options and styling

## What

- Allow disabledKeys to be passed as an array to Select
- Add styling
- Add stories for the Button and option 
image
image


packages/select/src/Select/Select.tsx
```diff
@@ -31,7 +31,10 @@ export const getSelectChildren: CollectionChildren<SingleItemType> = item => (
   <Item key={item.value}>{item.label}</Item>
 )
 export interface SelectProps
-  extends OverrideClassName<Omit<AriaSelectProps<SingleItemType>, "children">> {
+  extends OverrideClassName<
+    Omit<AriaSelectProps<SingleItemType>, "children" | "disabledKeys">
+  > {
+  disabledKeys?: React.Key[]
```
NatalieLoyzaga, line 37: I re-instantiated the disabledKeys prop as the type from React Aria was Iterable<T> - I felt it would be easier for the consumer to pass in an array of Keys

gyfchong: How are consumers currently disabling their select options?

NatalieLoyzaga: In filterMultiSelect...
image

NatalieLoyzaga: (we dont have any consumers actually using the Select lol)

gyfchong: They are using the old one aren't they?

NatalieLoyzaga: image

NatalieLoyzaga: Gets passed through the items 🤮

gyfchong: Hmm this is a tough one, but I guess no matter the format keys/items you can still apply more or less the same methods to determine what are disabled and use that in your keys.

gyfchong: One thing though, the term "keys" is confusing, cause when defining the items, none of them are keys, if it maps to value, I'd go with `disabledValues`

NatalieLoyzaga: I updated to `disabledValues`

gyfchong (approved):

PR #3172, opened 2022-12-15
docs(tailwind): extract psuedo stories & add TW focus rings

JakePitman:
## Why
Adding focus rings has always been a tricky task. This story shows consumers of our TW package that they can simply use a few of our TW classes to add focus rings to components.

## What
- Extracts the Psuedo Selector TW stories into their own file.
- Adds the `@tailwind base` directive to our TW stories' css file.
- Un-comments our `ring` presets. There is actually nothing wrong with the presets; they were failing because we hadn't added the `@tailwind base` directive to our story.


packages/tailwind/docs/PsuedoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      "https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=14473%3A63845"
+    ),
+  },
+  decorators: [withDesign],
+}
+
+export const TailwindPsuedoStates = () => (
+  <div className="p-12">
+    <h1 className="flex flex-col items-center text-heading-1 font-weight-heading font-family-heading text-purple-700">
+      Tailwind Psuedo states
```
christian314159, line 28:
```suggestion
      Tailwind pseudo states
```

packages/tailwind/docs/PsuedoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      },
+    },
+    ...figmaEmbed(
+      "https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%E2%9D%A4%EF%B8%8F-UI-Kit%3A-Heart?node-id=14473%3A63845"
+    ),
+  },
+  decorators: [withDesign],
+}
+
+export const TailwindPsuedoStates = () => (
```
christian314159, line 25:
```suggestion
export const TailwindPseudoStates = () => (
```

packages/tailwind/docs/PsuedoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+          )}
+        >
+          Focus me
+        </button>
+      </div>
+    </div>
+  </div>
+)
+
+TailwindPsuedoStates.storyName = "Tailwind psuedo selectors"
```
christian314159, line 57:
```suggestion
TailwindPseudoStates.storyName = "Tailwind pseudo selectors"
```

packages/tailwind/docs/PsuedoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { figmaEmbed } from "../../../storybook/helpers"
+import styles from "./styles.module.scss"
+
+export default {
+  title: "Tailwind",
+  parameters: {
+    docs: {
+      description: {
+        component:
+          'Add `presets: [require("@kaizen/design-tokens").TailwindPreset]` into your tailwind config',
```
christian314159, line 15:
```suggestion
          'Add `presets: [require("@kaizen/design-tokens").TailwindPreset]` into your Tailwind config',
```

packages/tailwind/docs/styles.module.scss
```diff
@@ -1,3 +1,7 @@
+// Would normally be imported in project's main css file.
+// Needed for shadow styles etc.
+@tailwind base;
```
christian314159, line 3: Just wondering, this is not the TW CSS reset, is it? I think we made a decision that we're not endorsing it for projects in favor of the already baked in reset.
So if we were using it in our own docs, that might mean we would want to revisit that decision?

packages/tailwind/docs/PsuedoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
christian314159, line 1: File name also needs to change to `Pseudo...`

christian314159 (commented):
Sorry, mainly just nitpicking on some spelling ;)

packages/tailwind/docs/styles.module.scss
JakePitman: No the CSS reset is called [Preflight](https://tailwindcss.com/docs/preflight). If you look at the css `base` compiles to, it's basically just a bunch of `--tw` prefixed css variables.
The docs [here](https://tailwindcss.com/docs/upgrade-guide#base-layer-must-be-present) also mention that it's a requirement from `3.0` onwards

packages/tailwind/docs/PseudoSelectors.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <div className="inline-flex flex-col ml-12">
+        <h2 className="text-heading-2 font-family-heading-2 font-weight-heading-2 text-purple-800 mb-12">
+          Focus
+        </h2>
+        <button
+          className={classnames(
+            "focus:ring-default-width ring-default-color ring-offset-default-width focus:outline-none",
+            "rounded-default border-none border-width-default",
+            " p-12 bg-purple-600 text-white font-family-paragraph font-weight-paragraph-bold text-paragraph "
+          )}
```
HeartSquared, line 48: Is this what the consumer needs to put in for a focus ring?! 😱 
Could we not make a special class that deals with the ring as a whole (position, direction offsets, width, colour, radius)?

gyfchong:
What's the use case for consumers applying focus rings manually?

packages/tailwind/docs/PseudoSelectors.stories.tsx
JakePitman: Easily.

We've been pretty against using the `component` layer so far, but this could be a good use case for it.
Even if we don't, 3 TW classes feels more straightforward than 3 imported sass variables on an `after` pseudo element.

https://cs.github.com/cultureamp/murmur/blob/2439fc43b3b2fa1c9688c86b04ede8a021beb047/app/assets/stylesheets/global/ca-button.scss?q=owner%3Acultureamp+focus-ring#L112

JakePitman:
> What's the use case for consumers applying focus rings manually?

Not sure I understand the question. Are you asking when someone would use these?
Just when being asked by a designer to put a focus ring on something they built themselves I suppose. I might be behind - is there a non-manual way to apply focus ring styles now?

packages/tailwind/docs/PseudoSelectors.stories.tsx
mcwinter07: Yeah, most teams have been pretty vocal in not wanting to use tailwind components too. We could release it as is and then do both could be an option if people find it too verbose.

packages/tailwind/docs/styles.module.scss
mcwinter07: Yeah, this caught us out for a while when we were exploring focus rings. It kind of silently does nothing with no real errors to let the user. We should add a line into the docs just to call that out :)

HeartSquared:
> Just when being asked by a designer to put a focus ring on something they built themselves I suppose.

I think the main thing with this is that if a consumer needs a component with a focus ring, then it's probably too complex and shouldn't be built with TW. That being said, it's a middle ground I suppose. Although encouraging it probably isn't good for us.

There's probably a hidden question here about whether we should be giving this now or whether we should only add these classes when there's a request for it (which hopefully would never be the case).

dougmacknz:
I quite like the idea of offering a `FocusRing` component helper, potentially using https://github.com/discord/focus-rings. I did a brief investigation on it when that library first came out and ran into some issues, but would like to have another go at it at some point.

gyfchong:
Hm I think focus rings are a little too low level for consumer's to be concerned about? If they're making something interactive, Kaizen already provides the primitive components needed which already have focus rings.

I'd go with leaving the focus rings out for now until someone asks for them 😄

JakePitman:
Closing this - we've decided not to expose these through TW.

PR #3174, opened 2022-12-16
refactor: remove VFC and FC types

gyfchong:
## What

- Remove usage of VFC/FC types
- Add eslint rule `@typescript-eslint/explicit-function-return-type`
	- Currently disabled for stories and tests; to be done in a separate PR

## Why

Removes the reliance on helper types which may change in the future (potentially with breaking changes).


gyfchong:
Lol even CRA have decided to remove it https://github.com/facebook/create-react-app/pull/8177

PR #3175, opened 2022-12-17
feat: (tailwind) Add prettier package

JakePitman:
## Why
Consistency in TW class order means increased readability. (Positioning classes first, color classes last etc.)

## What
Adds Tailwind linting via the `prettier-plugin-tailwindcss` package.
Works out of the box with `className=""` and `classnames(...)`

tw-formatting


gyfchong:
I actually don't like it's order of classes, can we change this? 😅

JakePitman:
> I actually don't like it's order of classes, can we change this? 😅

[Sadly not](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#customization). There's some good justifications in there though.

gyfchong (dismissed):

gyfchong (approved):

PR #3176, opened 2022-12-18
Kds 1061 remove tooltip icon

kaydbwithers:
## Why
This PR adds a `showTooltipIcon` prop which is set to true by default.

The `showTooltipIcon` prop is handy for situations where there is not a lot of room in the TableHeaderRowCell as shown in the example below and in ticket [KDS-1061](https://cultureamp.atlassian.net/browse/KDS-1061).

## What
Example of how the `showTooltipProp` will look when set to false.
image

Example of a Table with not enough room for the tooltip icon in the header.
image

### DIFF
--- a/draft-packages/table/KaizenDraft/Table/Table.tsx
+++ b/draft-packages/table/KaizenDraft/Table/Table.tsx
@@ -115,6 +115,7 @@ type TableHeaderRowCellProps = OverrideClassName<
   wrapping?: "nowrap" | "wrap"
   align?: "start" | "center" | "end"
   tooltipInfo?: string
+  showTooltipIcon?: boolean
   sortingArrowsOnHover?: "ascending" | "descending" | undefined
 }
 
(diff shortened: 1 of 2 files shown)


draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -115,6 +115,7 @@ type TableHeaderRowCellProps = OverrideClassName<
   wrapping?: "nowrap" | "wrap"
   align?: "start" | "center" | "end"
   tooltipInfo?: string
+  showTooltipIcon?: boolean
```
dougmacknz, line 118: We prefer our props to be false by default. See https://github.com/cultureamp/kaizen-design-system/blob/main/docs/basic-component.md#props

So I'd go for `isTooltipIconHidden` for this.

dougmacknz:
You'll also need to adjust the PR title with a conventional commit prefix

draft-packages/table/KaizenDraft/Table/Table.tsx
kaydbwithers: Thanks @dougmacknz, that's done now.

dougmacknz (approved):
Thanks Kayd. I might just adjust the PR title to have a bit more info, seeing as this ends up in the changelog as the main summary.

PR #3177, opened 2022-12-21
docs: update button to include badge examples

NatalieLoyzaga:
## Why

- missing Badge prop examples in the sticker sheet

## What

- Add examples of Badges


packages/button/docs/Button.stories.tsx
```diff
@@ -110,6 +111,19 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
     iconPosition: "end",
   }
 
+  const BADGE_LEFT_PROPS: ButtonProps = {
+    label: "Label",
+    icon: filterIcon,
+    badge: { text: "4" },
+  }
+
+  const BADGE_RIGHT_PROPS: ButtonProps = {
```
HeartSquared, line 120: Looking at Figma, there's probably no use case for a badge with an icon on the right. I'd err away from an example of something people shouldn't be doing (especially since the gap looks wrong).

I do wonder if there's a case when we'd want a badge and no icon though (a design question).

NatalieLoyzaga: There is a use case for this! It was actually what sparked this PR. Elliot brought it to mine attention. It used in survey results

NatalieLoyzaga: image

HeartSquared: Hmm interesting... Do you know if there's any case where we would have a badge but not an icon? Would we need an example for that too?

But fair enough to show this case since there's a real scenario for it. The spacing looks off to me though, but happy for that to be fixed in a separate PR :)

packages/button/docs/Button.stories.tsx
```diff
@@ -200,9 +214,19 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
...
+      >
         <StickerSheet.Header
-          headings={["Working", "Working (Focus)", "Icon Left", "Icon Right"]}
+          headings={[
+            "Working",
+            "Working (Focus)",
+            "Icon Left",
+            "Icon Right",
+            "Badge Left",
+            "Badge Right",
```
HeartSquared, line 228:
```suggestion
            "Icon Left with Badge",
            "Icon Right with Badge",
```

Something like that. `Badge Left` made me think the badge should show on the left of the text.

gyfchong (approved):

PR #3178, opened 2022-12-22
feat(GuidanceBlock): Add content prop to allow for custom markup

hqta:
## Why

Team Home has a requirement to add an Early Access tag above the GuidanceBlock heading. Currently GuidanceBlock only allows for title and description text field. This limits our ability to add necessary markup to achieve the product vision.

## What

By adding `content` prop for custom markup, consumers are able to better match design requirements to custom markup.

image

### DIFF
--- a/draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
+++ b/draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { Story } from "@storybook/react"
 import { withDesign } from "storybook-addon-designs"
+import { Box } from "@kaizen/component-library"
 import { GuidanceBlock } from "@kaizen/draft-guidance-block"
 import {
   Informative,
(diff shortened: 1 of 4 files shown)


StaphSynth (approved):
Nice

draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
```diff
@@ -154,6 +177,26 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
         persistent
       />
     </StoryWrapper.Row>
+    <StoryWrapper.Row rowTitle="Custom Content">
+      <GuidanceBlock
+        illustration={<Informative alt="" />}
+        content={<CustomContent />}
```
christian314159, line 183: image

Can we add a description text here for the new prop as well please, so when people look at it understand better when/how to use it. I think it has to be added to the `DefaultStory.argTypes` on line 78.
My idea would be something along these lines:
```
If you need to render custom content insight the `GuidanceBlock` use this prop instead of the default `text` option.
```

christian314159 (changes requested):
Looking good - great work, just some minor tweaking of the Story that I think we should do.

draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
```diff
@@ -98,6 +101,10 @@ DefaultStory.argTypes = {
     description:
       "This takes a scene scene or spot element, ie: `<Informative />`. This radio button implementation is a storybook only representation to toggle between the two illustration styles.",
   },
+  content: {
+    description:
+      "If you need to render custom content insight to the `GuidanceBlock` that is more than just a title and description use this prop instead of the default `text` option.",
```
christian314159, line 106: I think you meant to write this, didn't you?
```suggestion
      "If you need to render custom content inside of the `GuidanceBlock` that is more than just a title and description use this prop instead of the default `text` option.",
```

draft-packages/guidance-block/docs/GuidanceBlock.stories.tsx
```diff
@@ -114,6 +121,24 @@ const PROPS: GuidanceBlockProps = {
...
 }
 
+const CustomContent = () => (
+  <>
+    <Box mb={0.75}>
+      <Tag variant="statusLive" size="small">
+        Early Access
+      </Tag>
+    </Box>
+    <div className={styles.headingWrapper}>
```
christian314159, line 131: This should be replaced with `Box` as well, because we don't want to encourage people reaching into the CSS file.
The class is using  the `spacing-md` token, so it would probably be `<Box mb={1}>...`

christian314159 (commented):
Sorry, found 2 other minor things :(

christian314159 (approved):

PR #3181, opened 2023-01-04
WIP: new modal

dougmacknz:
# Blessed Modal Proposal

## Rundown of the existing (draft) modal and its issues
Consumers choose between 3 components

* Confirmation Modal
 
  * The most used pattern across our platform.
  * Matches W3's [Alert Dialog](https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/alertdialog) pattern.
  * The `prominent` variant appears to be used inconsistently.
  * Occasional usage that goes against the intended usage, with just a single action button.

* Input/Edit Modal

  * Second most used across our platform.
  * Works ok, but there have been cases where consumers would like more control over the footer actions, and the width of the modal.
  * Occasionally this is used with no form controls.

* Context Modal

  * Very few usages across the platform.
  * Doesn't seem to have much point to the opinions that are baked in, and will be obsolete once there's a better way to do generic modals.

In general, we found that there's a bit of leaking between the 3 patterns. Sometimes the `ContextModal` was used where an `InputEditModal` should have been used, and so on.

If none of these components work for whatever reason, there is an unadvertised `GenericModal` component available from the draft-modal package, which is the base component used to compose the 3 main patterns. This comes with no styling, so consumers import the `GenericModal.module.scss` file and extend the styling from that. This is something we want to completely get rid of across our repos, so this is why we've been hesitant to advertise `GenericModal`.

### Additional issues 
* Breaks in React strict mode
* We use `headlessui/react` for its `Transition` component.
* Built as a class component
* Some design tweaks required
* Still has the `draft` namespace

### Composibility
```
<Modal>
  <Modal.Header>
    <Modal.Title>Some title</Modal.Title>
  </Modal.Header>
  <Modal.Body>
    <Paragraph variant="body">Some content</Paragraph>
  </Modal.Body>
  <Modal.Footer>
    <Modal.ActionButton secondary>Cancel</Modal.ActionButton>
    <Modal.ActionButton primary>Continue</Modal.ActionButton>
  </Modal.Footer>
</Modal>
```

### Opinionated patterns

Out of the 3 existing patterns we have for modals, the `ConfirmationModal` is the only one that will remain as a separate component because of the `alertdialog` pattern that it matches, and to make life easy with the large amount of usage this has.

Otherwise, both the Input/Edit and Context patterns become very easy to compose with the new `Modal` component.

Context has nothing particularly special about it. Tailwind can be used to create a flexbox inside of the modal body to add an image if needed.

The existing Input/Edit component has the following opinions baked in:
* Gray background in the modal body to contrast.
* Closing the modal via backdrop click has been disabled.
* It has 2 action buttons: a submit and cancel (plus a hidden feature to replace the cancel button with a secondary action).

The gray background can be applied with a `hasFormControls` prop on `Modal.Body`, and the backdrop click can be removed with a `dismissOnBackdropClick={false}` prop on `Modal`.

`ConfirmationModal` becomes a little bit more locked down than before. You must have both a 'yes' and 'no' button, to avoid it being used as a context modal.

### Controlling the different ways to dismiss the modal
Consumers provide a single `onDismiss` prop which is used across the 3 different ways a modal can be dismissed. They can then adjust the defaults for these 3 
* Backdrop click: enabled by default. Can be adjusted with a `dismissOnBackdropClick` boolean prop.
* Escape key press: enabled by default. Can be adjusted with a `dismissOnEscPress` boolean prop.
* Close 'X' button: enabled by default. Can be hidden with a `hideCloseButton` boolean prop.

### Sizing
Consumers can choose between 3 main sizes (small, medium, large), or override the width with Tailwind if needed.

### Libraries used
[react-focus-on](https://github.com/theKashey/react-focus-on) has been used for focus locking, scroll locking, and putting aria-hidden on all siblings when the modal is open.
[framer-motion](https://github.com/framer/motion) has been used for the animation and transition logic. This is a bit of a hefty dependency wi

## Considerations
* There will no longer be anything stopping consumers from using the disabled submit validation pattern (accessibility anti-pattern).

  * This limitation on the existing InputEditModal has proved useful to prevent people from implementing this pattern, but this should be tackled at a higher level anyway rather than relying on locked down components.

## Still to do 
- [ ] Scrolling behaviour
- [ ] Close button styling
- [ ] Implement the correct animation
- [ ] Responsiveness
- [ ] Extend native props on everything

- https://github.com/cultureamp/kaizen-discourse/issues/46


packages/modal/src/Modal/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { OverrideClassName } from "@kaizen/component-base"
+import { useMediaQueries } from "@kaizen/responsive"
+import styles from "./ActionButton.scss"
+
+export interface ActionButtonProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
+  isAlignedStart?: boolean
+}
+
+export const ActionButton: React.FC<ButtonProps & ActionButtonProps> = ({
```
gyfchong, line 13: We aren't using FC anymore 😄

NatalieLoyzaga: I had a mental note to check what we are using again!  I knew it wasn't VFC lol but now I see it's neither!

gyfchong: Simples! Haha

packages/modal/src/Modal/components/Root/Root.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export interface RootProps extends HTMLAttributes<HTMLDivElement> {
+  isOpen: boolean
+  accessibleLabelId: string
+  children: React.ReactNode
+  onDismiss: () => void
+  dismissOnBackdropClick?: boolean
+  dismissOnEscPress?: boolean
+  hideCloseButton?: boolean
```
gyfchong, line 38: Maybe instead of 3 separate props, we could do a single one that takes an array? Especially if they all turn false if provided

eg.
```
disableDismiss={[“dismissOnBackDropClick”, "dismissOnEscPress"]}
```

dougmacknz: I don't mind that. Maybe we can slim that down a little bit to:
`disableDismiss={['backdrop', 'closeButton']}`

Following the discussion we had in the show and tell, we can leave out the esc press one. I don't think we've seen any cases of people wanting to remove it in the years we've had our current modal, and it would go against accessibility guidelines.

It's worth calling out that the close button one is a bit different in that it determines whether a UI element will show/hide, compared to the backdrop one which doesn't change anything visually, only the behaviour. Do we care that we're tying together two things that aren't consistent in what they do?

packages/modal/src/Modal/components/Body/Body.tsx
```diff
@@ -0,0 +1,21 @@
...
+export interface BodyProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
+  children: React.ReactNode
+}
+
+export const Body: React.FC<BodyProps> = ({
+  children,
+  classNameOverride,
+}): JSX.Element => {
+  const something = "asdasd"
```
dougmacknz, line 15: Oops, another one of these.

packages/modal/src/Modal/components/Root/Root.tsx
gyfchong: Hmm yeah tricky one, what was the scenario for removing backdrop click? Is it better to enforce a particular pattern rather than give the option?

NatalieLoyzaga: What would the type of `disableDismiss` be? .....

``` 
type => disableDismiss: [backdrop: boolean, closeButton: boolean]

use => onClickOutside={disableDismiss[0] ? onDismiss : undefined}

story => disableDismiss={[true, true]}

```

NatalieLoyzaga: @gyfchong

gyfchong: Oh that's true, a bit complex in real usage huh. Seems less necessary if the only options are to hide the close button and disable the backdrop; so feel free to ignore my suggestion 😄

dougmacknz:
Squashed the commits on this, rebased and started a new PR: https://github.com/cultureamp/kaizen-design-system/pull/3933

PR #3191, opened 2023-01-05
docs(split-button): use custom controls for actionButtonProps

HeartSquared:
## Why

POC attempt at better Storybook controls.

Inspired by https://github.com/cultureamp/reports-components/blob/307aa00e7ac87b9809cd84e2bd3936aa51b1d396/src/components/QuestionScore/QuestionScore.stories.tsx#L12-L48

### Existing

|Screenshot|
|---|
|image|

|Pros|Cons|
|---|---|
||TypeScript error with mapping confusion|
|Shows structure of how the consumer passes in the prop|Consumer cannot adjust attributes independently|

### New

|Screenshot|
|---|
|image|

|Pros|Cons|
|---|---|
|TypeScript is happy as transforms are done by us|More code required for types and transforms|
|Consumer can adjust attributes independently|Custom controls cannot be marked as required|
||May not be obvious the controls are split out from a single prop|
||Cannot customise sort order (`"alpha"` is likely the best option to group them)|

### DIFF
--- a/packages/split-button/docs/SplitButton.stories.tsx
+++ b/packages/split-button/docs/SplitButton.stories.tsx
@@ -42,20 +42,40 @@ export default {
     dropdownButtonProps: DropdownButton,
   },
   argTypes: {
-    actionButtonProps: {
-      options: ["Button", "Anchor"],
-      control: {
-        type: "select",
-        labels: {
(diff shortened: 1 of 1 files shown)


gyfchong:
Closed until becomes relevant

PR #3186, opened 2023-01-04
fix(deps): upgrade stylelint to 14.x

gyfchong:
## Why
Keep on top of latest package versions

## What
Upgraded stylelint to latest

PR #3192, opened 2023-01-05
fix(deps): downgrade postcss version to 8.4.19 for causing tsc and jest errors

gyfchong:
## Why
Upgrading postcss to the latest patch causes yarn lock issues which for unknown reasons breaks typescript's ability to parse stylelint-plugin and fails some tests for the rules.

## What
Downgrade the version of postcss to 8.4.19

### DIFF
--- a/packages/stylelint-plugin/package.json
+++ b/packages/stylelint-plugin/package.json
@@ -19,7 +19,7 @@
     "lodash.flatmap": "^4.5.0",
     "lodash.kebabcase": "^4.1.1",
     "nano-memoize": "^2.0.0",
-    "postcss": "^8.4.20",
+    "postcss": "8.4.19",
     "postcss-less": "^6.0.0",
     "postcss-scss": "^4.0.6",
     "postcss-value-parser": "^4.2.0",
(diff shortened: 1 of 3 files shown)


packages/stylelint-plugin/package.json
```diff
@@ -19,7 +19,7 @@
     "lodash.flatmap": "^4.5.0",
     "lodash.kebabcase": "^4.1.1",
     "nano-memoize": "^2.0.0",
-    "postcss": "^8.4.20",
+    "postcss": "8.4.19",
```
HeartSquared, line 22: Rather than locking the version, perhaps you could just downgrade and not lock it.

Renovate would create a new upgrade PR and it would fail CI once your new stylelint config has updated.
However, there would also be the chance that the other deps get updated and the versions become realigned, which then it wouldn't be a problem to upgrade.

gyfchong: Yeah, not locking it works but I figured it was safer to lock, can unlock.

gyfchong:
Consolidated https://github.com/cultureamp/kaizen-design-system/pull/3186

PR #3194, opened 2023-01-06
WIP - Button refactor POC

HeartSquared:
## What

Solution preview POC for new Kaizen Button/Link components.

---

New abstraction method where the base is a Styled[Button/Link] component that adds styles to the child element.

More specific components supplied by Kaizen in this concept:
- Button (looks like a button, semantically a `button`)
- IconButton (looks like a button, semantically a `button`)
- Link (looks like a link, semantically an `a`)

All other variations (eg. looks like a button, semantically an `a`) would require the consumer to use `StyledButton` and supply the underlying element.

---

- StyledButton pattern would be the same for creating a StyledLink component
- Not all variants are added in this POC
- May be missing some a11y?
- Unsure if we need a Working state with the label showing or not (UI Kit does not show it, but there is code in the old one)
- Badge has not been added in this POC, but it should be rather straightforward to add

---

There are two proposals for abstraction in this POC:
1. Underlying element is supplied in `children`
2. Underlying element is supplied in an `element` prop

### DIFF
--- a/packages/button/docs/StyledButton.module.scss
+++ b/packages/button/docs/StyledButton.module.scss
@@ -1,0 +1,3 @@
+.red {
+  border-color: red;
+}
--- a/packages/button/docs/StyledButton.module.scss
+++ b/packages/button/docs/StyledButton.module.scss
@@ -1,3 +1,7 @@
+// Does not override combined styles (eg. .default + .isReversed)
+// and not StyledIconButton styles
 .red {
   border-color: red;
+  color: magenta;
+  padding: 2rem;
 }
--- a/packages/button/src/StyledButton/StyledButton.tsx
+++ b/packages/button/src/StyledButton/StyledButton.tsx
@@ -5,16 +5,20 @@ import styles from "./StyledButton.module.scss"
 
 interface BaseStyledButtonProps extends OverrideClassName<unknown> {
   variant: "default" | "primary" | "secondary" | "secondaryDestructive"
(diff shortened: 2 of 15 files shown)


packages/button/docs/StyledButton.module.scss
```diff
@@ -0,0 +1,7 @@
+// Does not override combined styles (eg. .default + .isReversed)
+// and not StyledIconButton styles
```
HeartSquared, line 2: Unlikely to actually be a problem though - how often does one need to override the colours and padding? Worst case scenario, they could use `!important`

packages/button/src/StyledButton/StyledButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+// V1: Using children
+export interface StyledButtonProps extends BaseStyledButtonProps {
+  children: React.ReactElement
+}
+
+export const StyledButton: React.VFC<StyledButtonProps> = ({
+  children,
+  isWorking,
+  contentsPropName = "children",
```
HeartSquared, line 90: Because we construct the inner content (for the working state), we need to know where the contents go within the child. In most cases, it should be the children (eg. `<button>Contents</button>`), but if there's a custom component structure (eg. `<CustomComponent contents="Contents" />`, this can still work.

HeartSquared: Yes it will 😄

PR #3196, opened 2023-01-10
feat: handle long text multiline or truncated

NatalieLoyzaga:
## Why

to fit text within the button and not compromise on a11y

## What

- Truncate the button value always 
- Allow long paragraph options to go on multiline
- If long single word option - do not word-break and allow horizontal scroll

**Button Value**

image

**Dropdown**

image

image

### DIFF
--- a/draft-packages/select/docs/Select.stories.tsx
+++ b/draft-packages/select/docs/Select.stories.tsx
@@ -18,6 +18,7 @@ const OPTIONS = [
   { value: "Prince", label: "Prince" },
   { value: "Charith", label: "Charith" },
   { value: "Nick", label: "Nick" },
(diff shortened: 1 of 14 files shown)


gyfchong:
Should the chevron stay aligned to the top? :D

gyfchong:
I'm a bit on the fence about showing the dropdown variations separated from the select, technically they should belong to the not-yet-built popover component.

Is it possible to have the select default open to show them?

gyfchong (approved):

gyfchong:
I don't see the focus ring anymore...?
image

NatalieLoyzaga:
@gyfchong Buildkite is still building so the dev link is old eeek sorry I was to prompt to ping you

gyfchong (approved):
Good stuff!

PR #3199, opened 2023-01-12
feat(date-picker): add RTL support to FilterDateRangePicker

HeartSquared:
## What

Add RTL support to FilterDRP.

- Remove x-axis padding from Calendar weekdays (layout was shifted when testing in Arabic)
- Rotate Calendar nav arrows
- Restructured trigger button label
- Restructured input description
- Removed usage of FieldGroup and replaced with div
- Updated sticker sheets
	- Added new RTL sticker sheet for Calendar and DateRangeInputField

|OLD|NEW|
|---|---|
|image|image|

|OLD|NEW|
|---|---|
|image|image|

## Why

[KDS-1005](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1005)

Part of the requirements for Pathfinder (i18n to be introduced separately).

[KDS-1005]: https://cultureamp.atlassian.net/browse/KDS-1005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/date-picker/src/FilterDateRangePicker/utils/getDateRangeDisplayLabel.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+    ).toBeUndefined()
+    expect(
+      getDateRangeDisplayLabel(
+        { from: undefined, to: new Date("2022-05-01") },
+        locale
+      )
+    ).toBeUndefined()
+  })
+
+  it("formats valid date range in en-AU text format", () => {
```
gyfchong, line 25: Wouldn't this be the responsibility of the translation library?

packages/date-picker/src/_subcomponents/Calendar/Calendar.module.scss
```diff
@@ -103,6 +103,10 @@ $selectors--button-focus: "&:focus-visible, #{$polyfill--focus-visible}";
     background-color: $color-blue-100;
     color: $color-blue-500;
   }
+
+  [dir="rtl"] & {
+    transform: rotate(180deg);
```
gyfchong, line 108: Cool trick haha

packages/date-picker/src/FilterDateRangePicker/utils/getDateRangeDisplayLabel.spec.tsx
HeartSquared: Yes :P It's more just testing two to ensure we actually have some i18n going since we were somewhat home-rolling it. We only require the consumer to pass in a string, and then we choose a locale object on their behalf :)

packages/date-picker/src/FilterDateRangePicker/utils/getDateRangeDisplayLabel.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+import { DateRange } from "../../types"
+import { DateRangeDisplayLabel } from "../components/DateRangeDisplayLabel"
+
+export const getDateRangeDisplayLabel = (
```
gyfchong, line 5: Is this function necessary? Hesitant to have a `getX` function that returns a JSX element

HeartSquared: I did try to put it in the other one, but TS didn't want to acknowledge the `undefined` return value in the consuming line (`<DateRangeDisplayLabel />` is considered defined), and since the value is passed into `selectedValue` of the button, `null` wasn't a suitable value either as it doesn't make sense api-wise.

If you can think of a better name for this function, I'm happy for that too :)

gyfchong: Perhaps the return being `undefined` is the red flag here, could you push this logic upwards and rely on the consuming component to determine whether or not to render the `DateRangeDisplayLabel` component?

HeartSquared: Hmm... I suppose I could change this into a conditional instead 🤔 But then it would mean I would need more tests on the consuming component too, whereas now I am able to keep it a little more self-contained, which was my preference (although I'm not too strongly tied to this). Thoughts?

gyfchong: I think the use case is specific enough that the self-containment doesn't provide much value to the added complexity, and leaving the decision of showing/hiding a label to the parent is a IMO a logical pattern to test for.

But if you had to refactor this using the abstracted pattern, would the conditional get in your way?

HeartSquared: The abstracted pattern doesn't care about what gets passed into the `selectedValue` - all it cares is that either it exists or not, and if it does, it renders it. 

I did actually try to return `null` in the component and have the button not render if it was `null`, but unfortunately returning `null` in a `JSX.Element` actually still returns a `JSX.Element` just with some attributes set to `null`.

Although now that I say that, I think there is an `isValidElement` function or something which _maybe_ it might be able to help solve it? 🤞🏻

HeartSquared: Sadly the above did not work :( returning `null` within a JSX.Element is still a valid JSX.Element 😞

HeartSquared: Moved the conditional to the component :)

packages/date-picker/src/FilterDateRangePicker/utils/isCompleteDateRange.ts
```diff
@@ -0,0 +1,6 @@
+import { DateRange } from "../../types"
+
+export const isCompleteDateRange = (
+  dateRange: DateRange | undefined
+): dateRange is { from: Date; to: Date } =>
```
gyfchong, line 5: Could barely recognise this was typescript lol

gyfchong:
Looks good, but can you fix the Storybook example to display all props?
image

gyfchong (dismissed):

gyfchong (approved):

PR #3201, opened 2023-01-16
feat: remove useFocusRing for the Select button

NatalieLoyzaga:
## Why

The focus on the Select button **styles** were not showing we the manually calling butttonRef.current.focus()

## What

Remove the use of useFocusRing for the trigger button and the variable focusVisible that is produced which determined the styling.


gyfchong (approved):
O'Naturale

PR #3207, opened 2023-01-17
feat: Fix custom component prop render on mobile menu items

mcwinter07:
## Why
The `component` prop in secondary, primary and default actions was not rendering as expected on mobile because of the way the TitleBlock component outputs to the mobile drawer content. See full issue reported by Oli [here](https://github.com/cultureamp/kaizen-discourse/issues/2)

## What
- Recreates Menu Item in Titleblock to allow for custom render prop to be passed in
  - This was decided as a team to be the best route rather than updating the menu item component
- Update types for Titleblock to allow for custom component props to be used in Button
- Refactoring of render method for default link and action to make code work well with mobile action render
- Adds new stories to showcase component render prop and how the menu ordering working

## Notes mobile menu hierarchy
Titleblock has a specific hierarchy for the primary actions on the mobile menu. It should render as follows:

1. primary actions that are links
2. default actions
3. primary actions that are not links (ie: anything without an href)

Screenshot 2023-02-20 at 3 43 15 pm

I've maintained this sorting logic but it does constrain the way the user has to use the component render prop.

Screenshot 2023-02-20 at 3 44 46 pm


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockMenuItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+): JSX.Element | null => {
+  const className = classnames(styles.menuItem, {
+    [styles["menuItem--destructive"]]: styles.destructive,
+    [styles["menuItem--disabled"]]: styles.disabled,
+    [styles["menuItem--active"]]: styles.isActive,
+  })
+  if ("component" in props) {
+    return (
+      <li className={styles.menuListItem}>
+        <props.component
```
OliBridgman: 🤨 Does this work? I thought you'd need to capitalise it?

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
```diff
@@ -18,44 +20,31 @@ import styles from "./MobileActions.module.scss"
...
   primaryAction: PrimaryActionProps
 ): JSX.Element[] | null | undefined => {
   if (!primaryAction) return null
   if (isMenuGroupNotButton(primaryAction)) {
-    return primaryAction.menuItems
-      .filter(item => typeof item.action === "string")
-      .map((item, idx) => (
-        <MenuItem
+    return primaryAction.menuItems.map((item, idx) => {
+      const itemType =
```
OliBridgman, line 35: See, I _like_ this stuff, but people keep telling me to make this sort of stuff a function. I think in a design system context you probably wanna be explicit about what this is trying to find out?

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
```diff
@@ -18,44 +20,31 @@ import styles from "./MobileActions.module.scss"
...
+          ? "link"
+          : "action"
+
+      return (
+        <TitleBlockMenuItem
           {...item}
-          key={`title-block-mobile-actions-primary-link-${idx}`}
-          automationId={`title-block-mobile-actions-primary-link-${idx}`}
+          key={`title-block-mobile-actions-primary-custom-${idx}`}
+          automationId={`title-block-mobile-actions-primary-${itemType}-${idx}`}
```
OliBridgman, line 46: How much do you consider this stuff an internal API? I have a feeling you could break peoples tests with these sorts of changes. Might wanna consider a it a Breaking Change?

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
```diff
@@ -79,7 +68,7 @@ const renderDefaultLinkOrAction = (
   if (kind === "link" && buttonIsLink(defaultAction) && defaultAction.href) {
     return (
       <MenuItem
-        action={defaultAction.href}
+        href={defaultAction.href}
```
OliBridgman: Can MenuItems have `onClicks` _and_ `hrefs`? Like, it's a link style menu item but also I wanna do some analytics stuff onClick or maybe load a modal under certain conditions.

Or actually, even in the case of performance, if there's a basepath-ified href we pass, but the onClick we pass has a different URL cos we're using the basepath aware history. 😖

OliBridgman: Ah yeah nvm ignore me, it's covered below.

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.spec.tsx
```diff
@@ -245,6 +245,49 @@ describe("<TitleBlockZen />", () => {
     })
   })
 
+  // TODO: fix up these tests and add coverage across the other actions
```
OliBridgman, line 248: I'd also suggest mocking the window dimensions? Given how buttons become different components under mobile conditions, I'd add a test for each button-like prop of the TitleBlock (primary, secondary, secondaryOverflow, etc), for each render-like prop of the button (`href`, `onClick`, `component`), for each dimension size that changes how it render  (large and small, maybe medium?).

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
mcwinter07: Fair call, I support it runs the risk of breaking tests. I could leave all components listed as actions, but it is kind of its own thing. 🤔 While I've tried to make it as compatible as possible, it might be safer to make this a breaking change

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
mcwinter07: Fair, I can probably break it into smaller pieces to try to increase legibility

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/MobileActions.tsx
```diff
@@ -13,77 +15,88 @@ import {
   SecondaryActionsProps,
   TitleBlockMenuItemProps,
   convertSecondaryActionsToMenuItems,
+  DefaultActionProps,
 } from "./TitleBlockZen"
 import styles from "./MobileActions.module.scss"
 
-const buttonIsLink: (action: TitleBlockButtonProps) => boolean = action =>
+const menuItemIsLink: (item: TitleBlockMenuItemProps) => boolean = item =>
+  "href" in item || (!("component" in item) && typeof item.action === "string")
```
gyfchong, line 23: Lol this is quite verbose

mcwinter07: Yeah it is a bit ott. I mainly whipped that up to not break the old deprecated "action" prop, which still has a bit of legacy code in there. We actually push peeps to use href and onclick in its place, which would remove the need to have such a gnarly check

mcwinter07: I'm leaning towards this being a breaking change for safety so potentially room to remove it entirely

gyfchong: Breaking change might guarantee no one upgrades haha, keep it backwards compat. It's Almost Out The Door™

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react"
+import { Button } from "@kaizen/button"
+import {
+  TitleBlockCustomButtonProps,
+  TitleBlockButtonProps,
+} from "./TitleBlockZen"
+
+export const TitleBlockButton = (
+  props: TitleBlockCustomButtonProps | TitleBlockButtonProps
+): JSX.Element | null => {
```
JakePitman, line 10: Can we use React.ComponentType<> or React.FC<> to specify return type

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockMenuItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export type TitleBlockMenuItemProps =
+  | ({
+      component: (props: CustomButtonProps) => JSX.Element
+    } & MenuItemProps)
+  | MenuItemProps
+
+export const TitleBlockMenuItem = (
+  props: TitleBlockMenuItemProps
+): JSX.Element | null => {
```
JakePitman, line 16: Maybe React.ComponentType<TitleBlockMenuItemProps> ?

JakePitman (approved):

JakePitman (approved):

PR #3211, opened 2023-01-18
feat: add reversed styles to Select

NatalieLoyzaga:
## Why

Add reversed styles to Select

## What

- add `isReversedProp`
- Add reversed styling
- add reversed stickersheet

### DIFF
--- a/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
+++ b/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
@@ -14,7 +14,7 @@
   flex: 1;
   min-width: 0;
 
-  &:not(:disabled) {
+  &:not(.disabled) {
     &:hover .icon,
     &:global(.story__button-hover) .icon,
     &:active .icon,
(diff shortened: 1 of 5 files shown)


packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
```diff
@@ -14,15 +14,17 @@
   flex: 1;
   min-width: 0;
 
-  &:hover .icon,
-  &:global(.story__button-hover) .icon,
-  &:active .icon,
-  &:global(.story__button-active) .icon,
-  &:focus-visible .icon {
-    color: $color-gray-600;
+  &:not(:disabled) {
```
HeartSquared, line 17: Is it possible to not have this? The specificity can sometimes prevent overriding/extending styles sometimes. 

Global selectors are also a little dangerous as they might sneak in and add styles where you don't want them, so a class would be safer :)

NatalieLoyzaga: Hey i replaced the` :disabled` with `.disabled`

HeartSquared (approved):
If we can reduce the number of `:not`s (easier to maintain and refactor) that would be great. If not, that's fine in the meantime too until someone finds it doesn't work for them.

PR #3214, opened 2023-01-19
feat: update focus selector to :focus

NatalieLoyzaga:
## Why

To allow consumers to call `.focus()` and the focus ring to show

## What

Update to psudeo selector for the focus to be `:focus` rather than `:focus-visible`

### DIFF
--- a/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
+++ b/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
@@ -18,7 +18,7 @@
   &:global(.story__button-hover) .icon,
   &:active .icon,
   &:global(.story__button-active) .icon,
-  &:focus-visible .icon {
+  &:focus .icon {
     color: $color-gray-600;
   }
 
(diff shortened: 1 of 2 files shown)


gyfchong (approved):

sentience:
The purpose of `:focus-visible` is that if a user clicks a button with the mouse, it shouldn't show a focus ring.

@NatalieLoyzaga @gyfchong Do we definitely want buttons to display focus rings when the user is using a mouse?

I would suggest that if we need buttons to display a focus style when using a mouse, it should probably be more subtle than a focus ring.

NatalieLoyzaga:
@sentience Yep! Not only the focus-ring indicator needed to visually show when a consumer calls `.focus()` but also the guidance for form fields actually says the focus-ring indicator on form fields show on both keyboard and mouse -[Interaction states](https://cultureamp.design/guidelines/interaction-states/#focus-indicator). I will bring up your suggestion about a more subtle focus ring and whether this is something we should do :)

gyfchong:
@sentience also want to note that this ticket only makes this change on the select's trigger button, not all buttons 😄

PR #3217, opened 2023-01-19
WIP feat: allow grouped sections for Select

NatalieLoyzaga:

### DIFF
--- a/packages/select/docs/Select/Select.stories.tsx
+++ b/packages/select/docs/Select/Select.stories.tsx
@@ -1,4 +1,5 @@
 import React from "react"
+import { Section } from "@react-stately/collections"
 import { Node } from "@react-types/shared"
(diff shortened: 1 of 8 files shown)


gyfchong:
Is it possible to show the `groupedMockItems` in this code snippet? I'd prefer we have an example so consumers know how to use this feature 😄 
image

packages/select/docs/Select/Select.stories.tsx
```diff
@@ -61,6 +62,15 @@ DefaultStory.parameters = {
   docs: { source: { type: "code" } },
 }
 
+export const WithSections: ComponentStory<typeof Select> = () => (
+  <Select
+    id="select-grouped"
+    label="label"
+    items={groupedMockItems}
```
gyfchong, line 69: Is it possible to show the `groupedMockItems` in this code snippet? I'd prefer we have an example so consumers know how to use this feature 😄 

image

PR #3220, opened 2023-01-20
feat: Kaizen all in one components package

gyfchong:
This PR focuses on the creation of the package itself and providing the basic structure for ongoing development.


packages/components/future/README.md
```diff
@@ -0,0 +1,27 @@
+# Entrypoint: future
+
+This is directory is here to redirect any imports from `@kaizen/components/future` to the correct distribution of compiled code to allow for treeshaking of future components.
```
HeartSquared, line 3:
```suggestion
This directory is here to redirect any imports from `@kaizen/components/future` to the correct distribution of compiled code to allow for treeshaking of future components.
```

packages/components/future/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+More details: [https://github.com/theKashey/multiple-entry-points-example](https://github.com/theKashey/multiple-entry-points-example)
+
+## Contribution
+
+All "future" components are to be added to the `src/__future__` directory.
+
+### What is a "future" component?
+
+These are components which are scheduled to replace an already existing component. **It is not** to house unfinished components.
```
HeartSquared, line 13: So where do unfinished components go?

packages/components/src/ExampleComponent/__tests__/ExampleComponent.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react"
```
HeartSquared, line 1: What is this structure you've built for this example component? 👀 You might want to check the plop templates and docs :)

Also, the reason why we had a `docs` folder as opposed to having the stories next to their component file is because the distribution structure wasn't great and it was the easiest way to exclude them from being compiled and included in the dist.

If your AIO can give us easier control over this, then we do not necessarily need so many levels of nesting (personally on the train of keeping tests/docs next to their associated files rather than detached; much more self-contained).

packages/components/tsconfig.dist.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.dist.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
+  "include": ["src/**/*"],
```
HeartSquared, line 6: So `future/` will never have any TS right?

packages/components/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "url": "git+https://github.com/cultureamp/kaizen-design-system.git"
+  },
+  "author": "Geoffrey Chong <geoff.chong@cultureamp.com>",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/cultureamp/kaizen-discourse/issues"
+  },
+  "homepage": "https://github.com/cultureamp/kaizen-design-system#readme",
+  "devDependencies": {
+    "rimraf": "^3.0.2"
```
HeartSquared, line 29: Side thought: I wonder if it's possible to just have this in the root of the repo? 🤔

HeartSquared (commented):
I wonder why there were yarn.lock changes in this PR 😬

packages/components/package.json
gyfchong: Oh, maybe 🤔

packages/components/tsconfig.dist.json
gyfchong: Yeah

packages/components/future/README.md
gyfchong: We shouldn't have "unfinished" components in the library ie. components that don't match definition of done or components we aren't happy for people to start using yet.

packages/components/package.json
gyfchong: Yeah that works, removing the dep here.

packages/components/src/ExampleComponent/__tests__/ExampleComponent.spec.tsx
gyfchong: > easiest way to exclude them from being compiled and included in the dist.

Exactly, I think this structure makes it easier to configure the tooling, rather than maintaining the regex split between the tsconfig, jest and playwright, we just ignore the folders that don't need to be compiled.

Generally I prefer to have everything that isn't compiled nested away so that I can immediately see the files that I can work on. We're also moving to having both `stories.tsx` and `stories.mdx` which adds more clutter.

Happy to move this to a discussion though.

packages/components/future/README.md
HeartSquared: Then how do we do iterative work on it? :'( You want 6 month old branches with 5000 lines of code? 😭

gyfchong: Well scoped work with clear definition of done, so we can release something as soon as possible 😄 -- edge cases can go ahead and be added, but imo anything taking 6mo and 5k lines is a flag.

gyfchong:
> I wonder why there were yarn.lock changes in this PR 😬

@HeartSquared this is definitely perplexing... Seems to get worse every time I try to fix it lol

packages/components/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "url": "git+https://github.com/cultureamp/kaizen-design-system.git"
+  },
+  "author": "Geoffrey Chong <geoff.chong@cultureamp.com>",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/cultureamp/kaizen-discourse/issues"
+  },
+  "homepage": "https://github.com/cultureamp/kaizen-design-system#readme",
+  "peerDependencies": {
+    "@kaizen/design-tokens": "^2.10.3 || ^3.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0 || ^9.0.0 || ^10.0.0",
```
HeartSquared, line 29: Would the design-tokens be moved into this package at some point? So that they don't need to be a peerDep.

packages/components/src/ExampleComponent/ExampleComponent.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react"
```
HeartSquared, line 1: Even if we disregard the structure, the contents of your templates need a review (check or use the plop templates)😉

packages/components/package.json
gyfchong: I did this cause the yarn error log seemed to complain about this...

gyfchong: But yeah it will move to this package, although not sure how the import experience will be for those just yet.

gyfchong:
https://github.com/cultureamp/kaizen-design-system/pull/3229

PR #3225, opened 2023-01-23
fix: update 1.25 padding prop to correct rem value

mcwinter07:
## Why
The value output for the 1.25 prop for the box component has incorrectly been rendering this as the `"0-point-25"` rem value.

## What
- corrects the utili on ln18 to `"1-point-25"`
- remove dependence on the deprecated component lib helper for ca-grid
- cleans up and adds calc value on the sass calculations

### DIFF
--- a/packages/component-library/components/Spacing/Base.module.scss
+++ b/packages/component-library/components/Spacing/Base.module.scss
@@ -1,19 +1,19 @@
 @import "~@kaizen/component-library/styles/fonts";
-@import "~@kaizen/deprecated-component-library-helpers/styles/type"; // just for the grid unit @TODO - update when we move to dart-sass
+@import "~@kaizen/design-tokens/sass/spacing";
 
 // this should always be parity with util.ts
 $spacing-list: (
   ("0", 0),
-  ("0-point-25", 0.25 * $ca-grid),
-  ("0-point-5", 0.5 * $ca-grid),
-  ("0-point-75", 0.75 * $ca-grid),
-  ("1", $ca-grid),
-  ("1-point-25", calc(1.25 * calc(#{$ca-grid}/ 4))),
-  ("1-point-5", 1.5 * $ca-grid),
-  ("1-point-75", 1.75 * $ca-grid),
-  ("2", $ca-grid * 2),
-  ("2-point-5", 2.5 * $ca-grid),
-  ("3", 3 * $ca-grid),
-  ("3-point-5", 3.5 * $ca-grid),
-  ("4", $ca-grid * 4)
+  ("0-point-25", calc(0.25 * #{$spacing-md})),
+  ("0-point-5", calc(0.5 * #{$spacing-md})),
+  ("0-point-75", calc(0.75 * #{$spacing-md})),
+  ("1", $spacing-md),
+  ("1-point-25", calc(1.25 * #{$spacing-md})),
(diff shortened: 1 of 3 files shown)


packages/component-library/components/Spacing/util.ts
```diff
@@ -14,7 +14,7 @@ export const convertFractionToString = (fraction: GridFractions): string => {
     case 1:
       return "1"
     case 1.25:
-      return "0-point-25"
+      return "1-point-25"
```
mcwinter07, line 17: This appears to be the root of the issue and might have been bust since the original component was created

packages/component-library/components/Spacing/Base.module.scss
```diff
@@ -1,19 +1,19 @@
...
-  ("1", $ca-grid),
-  ("1-point-25", calc(1.25 * calc(#{$ca-grid}/ 4))),
-  ("1-point-5", 1.5 * $ca-grid),
-  ("1-point-75", 1.75 * $ca-grid),
-  ("2", $ca-grid * 2),
-  ("2-point-5", 2.5 * $ca-grid),
-  ("3", 3 * $ca-grid),
-  ("3-point-5", 3.5 * $ca-grid),
-  ("4", $ca-grid * 4)
+  ("0-point-25", calc(0.25 * #{$spacing-md})),
```
mcwinter07, line 7: Update this to use calc for safety since we've had issues in the past with sass math (re: division) and interpolated variables

packages/component-library/components/Spacing/Base.module.scss
```diff
@@ -1,19 +1,19 @@
 @import "~@kaizen/component-library/styles/fonts";
-@import "~@kaizen/deprecated-component-library-helpers/styles/type"; // just for the grid unit @TODO - update when we move to dart-sass
+@import "~@kaizen/design-tokens/sass/spacing";
```
mcwinter07, line 2: I've just updated this to remove dependence on deprecated comp lib helpers

packages/component-library/components/Spacing/util.ts
gyfchong: Is there a risk people have used this thinking it was 1.25 and are happy with the result?

HeartSquared: I'd say it's unlikely. If they were after 0.25, they would've gone with 0.25. If anything, I expect if they saw the 1.25 didn't look right, they'd just not use the token.

packages/component-library/components/Spacing/Base.module.scss
HeartSquared: It looks so weird to have them all reference `$spacing-md`, but I guess it's the same for now 🤷🏻 

I wonder if we'd still have this list when we have the new tokens 🤔

HeartSquared (approved):

packages/component-library/components/Spacing/Base.module.scss
mcwinter07: hahaha yeah, we'll probably remove the need for this in its entirety 😆

packages/component-library/components/Spacing/util.ts
mcwinter07: Yeah, feel like peeps would have avoided it or used classnameOverride to fix it if they really needed it 🤷 technically still a fix so here's hoping people haven't done something strange 😬

PR #3227, opened 2023-01-24
feat: add renderSectionTitle render prop

jwmccowan:
to title block zen

## Why

Our team has a design to have a status Pill (Tag) in the section title area of the title block component.
See [discussion here](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1669611247289339).

## What
This PR adds a render prop, that, if provided, renders the provided component in the space where the section title is normally rendered.  All relevant section title props are provided through the callback (`sectionTitle`, `sectionDescription`, and the `data-*` attributes).

## Screenshots
Screen Shot 2023-01-24 at 2 35 06 pm

## Thoughts
This implementation was a tougher decision than I thought.  Initially, I was just going to change the `sectionTitle` prop type from a `string` to `ReactNode`.  The issue I came across was, if the `{sectionTitle}` section is rendered in a heading, which has style and accessibility implications.  

If we split the render conditionally, for example just rendering`{sectionTitle}` if the prop is a function, and the normal heading if it's a string, then we ignore the other props related to `sectionTitle`, such as `sectionDescription`.

I'm happy to switch back to the `React.Node`, but the render prop seemed the simplest solution that:
- Did not just fix my particular problem, but would allow future designs without another PR
- Continued to use all the available props
- Is opt in, and backwards compatible
- Fixes my issue ;)

We can discuss more in the PR.

### DIFF
--- a/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
+++ b/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
@@ -557,7 +557,7 @@ const TitleBlockZen = ({
           [styles.adminVariant]: variant === "admin",
           [styles.collapseNavigationArea]:
             collapseNavigationArea &&
-            !(sectionTitle || sectionTitleDescription),
+            !(sectionTitle || sectionTitleDescription || renderSectionTitle),
           [styles.hasLongTitle]: title && title.length >= 30,
           [styles.hasLongSubtitle]:
             subtitle && typeof subtitle === "string" && subtitle.length >= 18,
--- a/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
+++ b/draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
@@ -266,14 +266,14 @@ const renderSubtitle = (
 )
 
 const defaultRenderSectionTitle = (
-  sectionTitle?: React.ReactNode,
+  sectionTitle?: string,
   sectionTitleDescription?: string,
   variant?: Variant,
   sectionTitleAutomationId?: string,
   sectionTitleDescriptionAutomationId?: string
 ): JSX.Element => (
   <>
-    {sectionTitle && typeof sectionTitle === "string" && (
+    {sectionTitle && (
       <div className={styles.sectionTitle}>
         <Heading
           variant="heading-2"
@@ -284,9 +284,6 @@ const defaultRenderSectionTitle = (
           {sectionTitle}
         </Heading>
       </div>
-    )}
-    {sectionTitle && typeof sectionTitle !== "string" && (
-      <div className={styles.sectionTitle}>{sectionTitle}</div>
     )}
     {sectionTitleDescription && (
       <div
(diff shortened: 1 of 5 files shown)


draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -256,39 +265,40 @@ const renderSubtitle = (
...
-          {sectionTitleDescription}
-        </div>
-      )}
-    </div>
-  </div>
+          {sectionTitle}
+        </Heading>
+      </div>
+    )}
+    {sectionTitle && typeof sectionTitle !== "string" && (
```
mcwinter07, line 288: Just a wondering if this is here to showcase the use of the ReactNode approach over the custom render prop? My confusion just sprouts from React Node not being a type that can be passed to section title (at least for a TS perspective)
Screenshot 2023-01-26 at 11 14 18 am

mcwinter07: Could we potentially extend what the `sectionTitle`prop can take and not introduce a new render prop? ie:
```
sectionTitle?:
    | string
    | React.ReactNode
    | ComponentType<SectionTitleRenderProps>
```
Then it could look like this:
sectionTitle={<CustomComponent />}
or this:
sectionTitle={<div >This may be a bit of chaos by won't clamp down on types we assume?</div>}

Curious to hear peeps thoughts 🤔

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
```diff
@@ -643,14 +654,28 @@ const TitleBlockZen = ({
...
             <div className={styles.rowBelowSeparatorInnerContent}>
-              {(sectionTitle || sectionTitleDescription) &&
-                renderSectionTitle(
-                  sectionTitle,
-                  sectionTitleDescription,
-                  variant,
-                  sectionTitleAutomationId,
-                  sectionTitleDescriptionAutomationId
-                )}
+              {(sectionTitle ||
```
mcwinter07, line 657: One callout I wanted to make on ln 564 that kind of relates to this:
```
          [styles.collapseNavigationArea]:
            collapseNavigationArea &&
            !(sectionTitle || sectionTitleDescription),
```
If we went the render prop route we'd probably want a check there too ^ since you could theoretically have a custom render with out the sectionTitle and sectionTitleDescription prop

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
jwmccowan: This is a great callout, that `defaultRenderSectionTitle` shouldn't accept a `ReactNode`.. That's an artifact from the previous approach.

I've pushed a fix that leaves the `defaultRenderSectionTitle` exactly the same as before, barring the name, and it leaves out the two styling `div`s, since those are handled in the main title block render.

I like the idea of simply injecting the `sectionTitle` as a ReactNode directly in the previous render function, but conditionally rendering it in a `Heading` in the local render function.  Typing can be a little silly here, since `ReactNode` composes `string`, but I'm sure we can work around that.

Glad to hear any feedback

draft-packages/title-block-zen/KaizenDraft/TitleBlockZen/TitleBlockZen.tsx
jwmccowan: Great catch 🎉

mcwinter07 (approved):
Raised the render prop with the team and we're happy to add this in 👍 Thanks for clearing up those other small fixes

PR #3229, opened 2023-01-24
feat: add components AIO

gyfchong:


packages/components/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "build": "yarn clean && yarn prepublish",
+    "build:watch": "yarn clean && yarn prepublish --watch",
+    "clean": "rimraf 'dist'"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cultureamp/kaizen-design-system.git"
+  },
+  "author": "Geoffrey Chong <geoff.chong@cultureamp.com>",
+  "license": "ISC",
```
HeartSquared, line 23: Just curious - what's with the license change? The others say MIT

packages/components/src/ExampleComponent/ExampleComponent.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
+import { render } from "@testing-library/react"
+import { ExampleComponent } from "."
+import "@testing-library/jest-dom"
```
HeartSquared, line 4:
```suggestion
```

Shouldn't be needed now since we should have global config to apply to all our test files

packages/components/src/__future__/ExampleComponent/docs/ExampleComponent.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { ExampleComponent } from "../ExampleComponent"
+
+export default {
+  title: "AIO/ExampleComponent (Do Not Use)",
```
HeartSquared, line 6: I think this will conflict with the other story since the title dictates the path and it's exactly the same

packages/components/src/ExampleComponent/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./ExampleComponent"
```
HeartSquared, line 1: This file should be `index.ts`

packages/components/src/__future__/ExampleComponent/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./ExampleComponent"
```
HeartSquared, line 1: This file should be `index.ts`

packages/components/src/__future__/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./ExampleComponent/ExampleComponent"
```
HeartSquared, line 1:
```suggestion
export * from "./ExampleComponent"
```

This should work?

packages/components/package.json
gyfchong: I just didn't use the plop template and ran npm init 😅

packages/components/src/__future__/ExampleComponent/ExampleComponent.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
+import { render } from "@testing-library/react"
+import { ExampleComponent } from "./ExampleComponent"
+import "@testing-library/jest-dom"
```
HeartSquared, line 4: You missed one 😗

packages/components/src/__future__/ExampleComponent/docs/ExampleComponent.stories.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export default {
+  title: "AIO/future/ExampleComponent (Do Not Use)",
+  component: ExampleComponent,
+  parameters: {},
+} as ComponentMeta<typeof ExampleComponent>
+
+export const DefaultStory: ComponentStory<typeof ExampleComponent> = args => (
+  <ExampleComponent />
+)
```
HeartSquared, line 13: Probably for the other one too. Although this is nitpicking I guess.

```suggestion
export const DefaultStory: ComponentStory<typeof ExampleComponent> = () => (
  <ExampleComponent />
)
```
or
```suggestion
export const DefaultStory: ComponentStory<typeof ExampleComponent> = args => (
  <ExampleComponent {...args} />
)
```

packages/components/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@kaizen/components",
+  "version": "1.0.0",
```
HeartSquared, line 3: Hmm...since this isn't really housing anything, should we make it `0.0.0`? 🤔 Your choice though.

packages/components/src/__future__/ExampleComponent/docs/ExampleComponent.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
```
HeartSquared, line 1: I just realised, this structure actually doesn't match the other existing packages either. Those are something like:

```
package-name-here
├── docs
|   ├── ExampleComponent.stories.tsx
├── src
│   ├── ExampleComponent
|   │   ├── ExampleComponent.tsx
|   |   ├── ExampleComponent.spec.tsx
|   │   ├── ExampleComponent.module.scss
│   └──index.ts
```

Does this mess everything up? I do admit I don't like it and I much prefer the docs being within the component at the very least (I think I read your initial proposal wrong and I am less against it now).

gyfchong: Ah my bad, I should've framed the proposal from the perspective of the AIO. But yeah, the docs dir sits with the component.

packages/components/package.json
gyfchong: Hm don't think anyone can install a `0.0.0` so it'd have to be `0.0.1`, but if that's the case the follow up question would be what would make it `1.0.0`? The first component? Despite it not breaking anything.

packages/components/src/__future__/ExampleComponent/docs/ExampleComponent.stories.tsx
HeartSquared: Shippit. I do like the `__docs__` syntax too to surface them. Not used to `__tests__` but I could live with it. Although the subcomponent structure still makes me worry about having a million nested folders, but this is for the other discussion.

packages/components/package.json
HeartSquared: I think `0` to `1` is a special case, is it not? It's essentially saying it's in Beta.
Although one caveat is that any breaking changes made while in `v0` is that it does not get bumped to `v1`. That must be done manually.

Up to you :)

HeartSquared (approved):
I could nitpick forever I reckon, but I cbb anymore :P Just as long as no one copies your component as a base and it gets deleted at some point, then it's fine :)

packages/components/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "Kaizen component library",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "future",
+    "src",
+    "dist"
```
HeartSquared, line 10: We'll probably need to double check this. Does this include the source code for stories and tests? I wonder if consumer config would yell at it 🤔

gyfchong: Yep, should be able to: `import { FilterDRP } from "@kaizen/components/src/FilterDRP"`

HeartSquared: Well they shouldn't really exist in a distributed package usually right? I'm more worried about it making their CIs explode with unnecessary errors is all.

gyfchong: Fair enough!

gyfchong:
> Just as long as no one copies your component as a base and it gets deleted at some point, then it's fine :)

The moment a component goes in, it should go out 😄

HeartSquared (approved):

PR #3235, opened 2023-01-25
feat: add isContentFullWidth boolean

kaydbwithers:
## Why
Given a modal has the isProminent boolean, 
then the Modal Body does not appear as full width. 

This new boolean allows content to be full width which is handy for Modal Body's with a lot of content. https://cultureamp.atlassian.net/browse/KDS-1156

## What
**An example without isContentFullWidth**
image

**An example with isContentFullWidth**
- Icon is slightly smaller (130px), 
- Content is full width, 
- Modal title is moved to the left
- Modal body top padding has been increased (xl)
image


gyfchong:
We have chosen to steer away from this design pattern for now. Thanks @kaydbwithers!

PR #3238, opened 2023-01-25
feat: Swap action and draft colors in tag element

jwmccowan:
## Why
After a [discussion with designers](https://cultureamp.slack.com/archives/C04JJQN9066) about which color tags should be paired with which statuses in `develop`, the decision was made to make `statusDraft` blue, and the ready for review status orange.

Since there was an undocumented `statusAction`, I've swapped `statusAction` with `statusDraft`, so the colors match the design decisions.

Some quotes from the conversation:

> Draft becomes Blue-100 (Informative)
> Ready for Review is Orange-100 (Assertive)
> Closed is Grey-200 (Neutral)

> i think we want to look at changing this globally and having it effect existing tags as the current color/status pairings dont make a lot of sense.

## What
I noticed a few of the statuses did not have documentation in storybook (`statusAction`, `statusClosed`), a problem which started this conversation in the first place, since we did not think we had enough status tags for each of our use cases.

This PR
1. Swaps the `statusAction` and `statusDraft` colors
2. Documents all of the available statuses in storybook

## Screenshot
Screen Shot 2023-01-25 at 3 50 04 pm


mcwinter07 (approved):
Looks hunky dory to me based on the conversations in that thread ☑️ I'll approve this now but going to wait till I get a designer sign-off before I merge it in :)

mcwinter07:
Conversation thread (slack) on updating the prop name from `action` to `actionRequired` can be found [here](https://cultureamp.slack.com/archives/C02NUQ27G56/p1674695423396599)

draft-packages/tag/docs/Tag.stories.tsx
```diff
@@ -36,10 +36,10 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
   <>
     <StoryWrapper isReversed={isReversed}>
       <StoryWrapper.RowHeader
-        gridColumns={10}
-        headings={["Default", "Live", "Draft"]}
+        gridColumns={8}
+        headings={["Default", "Live", "Draft", "Action", "Closed"]}
```
mcwinter07, line 40: @jwmccowan Cleared it with the team and since it's already a breaking change would be great if we could update the vairant from `statusAction` to `statusActionRequired` to make it more explicit with its status. Also updating this heading to "Action Required" above :)

mcwinter07 (dismissed):
Awesome stuff and thanks again for the contribution :shipit:

mcwinter07 (dismissed):

draft-packages/tag/docs/Tag.stories.tsx
mcwinter07: Thanks for reverting this

mcwinter07 (approved):

mcwinter07:
Note from the Kaizen team - we are not considering this to be a breaking change since it will only this is an intentional design decision and will not change the API. Please be aware this may cause chromatic tests to fail - if you have any queries on what status your tag should be please reach out in #prod_design_system

PR #3247, opened 2023-01-31
docs: update filterMultiSelect story example

NatalieLoyzaga:
## Why

- The consumer should be able to display selected items separated from the unselected option aka float selected to the top. So this within example

## What

- Updated the Async story to show the `selectedItems` within a ListBoxSection and the `unselectedItems` in another


packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -465,15 +469,25 @@ export const Async: ComponentStory<typeof FilterMultiSelect> = args => {
     [data]
   )
 
+  // Combine current and cached and remove the duplicates
```
gyfchong, line 472: Include a description of the feature that this code is supporting

NatalieLoyzaga: updated :)

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -465,15 +469,31 @@ export const Async: ComponentStory<typeof FilterMultiSelect> = args => {
...
+   * and cached people, to be passed as the items.
+   * Make sure we remove the duplicates.
+   */
+  const mergedPeople = [...currentPeople, ...cachedPeople].filter(
+    (item, index, a) =>
+      a.findIndex(currItem => currItem.value === item.value) === index
+  )
+
+  /**
+   * Show the only the filtered current people when there is a search query
```
gyfchong, line 483:
```suggestion
   * Only show the current filtered people when there is a search query
```

gyfchong (approved):

PR #3249, opened 2023-01-31
feat: Add connect icon

dscultureamp:
## Why
There’s a need by designer Daniel Spagnolo from team Mercury for a connect icon which is available in Figma but missing in the codebase. This will be used in the Analytics or Retention application, refer here for more details: https://www.figma.com/file/ssgJiCIMdGdvTpJblDXQKG/Mercury%3A-Continuous-Reporting---Turnover-Insights?node-id=7726%3A326241&t=p4LMDLnhb8vkGp9d-0

Usage:
Screen Shot 2023-02-01 at 9 11 45 am

## What
Add a new connect icon which is available in Figma but missing in the codebase


packages/component-library/icons/connect.icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M14.01 5L14 1H12V5H8V1H6V5H5.99C5 4.99 4 5.99 4 6.99V12.48L7.5 16V19H12.5V16L16 12.49V6.99C16 5.99 15 4.99 14.01 5Z" fill="black"/>
```
mcwinter07, line 2: Could we change the `fill="black"` to `fill="#000000"` - I believe it will currently work with our process of inheriting the `color` CSS property from its parent (see below), but for the sake of continuity of our icons going forward we're trying to only use the hex value for true black.

```
// using an inline style as an example, IRL you'd probably have a className on the div and pass color through as with CSS modules

<div style="color:red">
    <Icon
      icon={connect}
      title="A descriptive title"
      role="img"
      inheritSize={true}
    />
</div>
```

mcwinter07 (commented):
Hey @dscultureamp 👋  Thanks for the contribution - I left a small comment on the fill valve in the SVG.

I also realized we should probably import the connect icon into `stories/Icons.ts`(ln6) and add it to the Actions object (ln188). This is used to build out the story that showcases all of our icons

dscultureamp:
> Hey @dscultureamp 👋 Thanks for the contribution - I left a small comment on the fill valve in the SVG.
> 
> I also realized we should probably import the connect icon into `stories/Icons.ts`(ln6) and add it to the Actions object (ln188). This is used to build out the story that showcases all of our icons

@mcwinter07 Thanks for reviewing my PR. I have fixed the fill hex value and also added the connect icon in Icons.ts.

packages/component-library/stories/Icons.ts
```diff
@@ -69,6 +69,7 @@ import maximize from "../icons/maximize.icon.svg"
 import repeats from "../icons/repeats.icon.svg"
 import zoomIn from "../icons/zoom-in.icon.svg"
 import zoomOut from "../icons/zoom-out.icon.svg"
+import connect from "../icons/connect.icon.svg"
```
mcwinter07, line 72: Brill, thanks for that. The final step is to add the `connect` icon into the `Actions` object located on line 188, which will add it to the [All Icons story](https://cultureamp.design/storybook/?path=/story/components-icon--sticker-sheet-default). Probably best we put it at the bottom of the object too since its a new one 👍

dscultureamp: Fixed. Sorry missed that line, thanks!

mcwinter07 (approved):
Thanks for adding that to the story @dscultureamp ✅ will merge it in a holler when it's passed CI

PR #3252, opened 2023-02-02
docs(tailwind): fix missing link and add classnameOverride example

JakePitman:
## Why
- Because I left out a link
- Because the current classnameOverride example isn't representative of the most common use case.

## What
- Adds a missing link
- Adds a more common example of classnameOverride

New example:
image


christian314159 (dismissed):

docs/Systems/Tailwind/WorkingWithTailwind.stories.mdx
```diff
@@ -30,13 +36,34 @@ When building a page, ideally [Kaizen](https://github.com/cultureamp/kaizen-desi
...
 
 ## Altering Kaizen components with classnameOverride
 
-If a Kaizen component supports classNameOverride, Tailwind classes can also be used.
+If a Kaizen component supports classNameOverride, Tailwind classes can also be used.  
+This is particularly useful for adding margins to Kaizen components. In this example, our preset is being used to add `margin-right: 1.5rem` to Kaizen's `<Card/>` component, without the need to wrap it in a div.
+
+<Canvas>
+<div className="flex justify-center">
+  <Card classNameOverride="mr-24"> {/* Adds margin-right: 24px */}
```
gyfchong, line 50: IMO this isn't an ideal use case, to add a gap between items in flex they should use `.gap`. A bit more of a simplified example I think would be more effective, something like your CardContent where someone would want to add spacing between a header and paragraph:
```
<Heading variant="heading-3" classNameOverride="mb-24">{title}</Heading>
<Paragraph variant="body"> Lorem ipsum oogum boogum. Tutant meenage neetle teetles.</Paragraph>
```

JakePitman: Yeah I see.

It looks like passing a `margin` rule to `classNameOverride` doesn't work for `<Heading/>`, it must be set already.

image

We'll have to come up with a different example

gyfchong (approved):

PR #3254, opened 2023-02-03
fix: add box-sizing:content-box to modal to prevent incorrect sizing in environments that use a border-box default

diffidentDude:
fix: add box-sizing:content-box to modal to prevent incorrect sizing in environments that use a border-box default

## Why

Murmur has a style rule of
`*, :before, :after { box-sizing: border-box; }` which breaks the modal's width calculations as it assumes a default of content-box.

## What

I've added a default of content-box to the element with the width calculations.

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Presets/ContextModal.module.scss
+++ b/draft-packages/modal/KaizenDraft/Modal/Presets/ContextModal.module.scss
@@ -42,6 +42,7 @@
 .portraitContentlayout {
   padding: $spacing-md;
   display: block;
+  box-sizing: content-box;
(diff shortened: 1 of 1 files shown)


HeartSquared (approved):
Fingers crossed this works! 🤞🏻

kaydbwithers:
I feel like everything should be `box-sizing: border-box;`. Fixing the width might be better?

diffidentDude:
> 

From the component lib perspective, we have no way of knowing the inherited box sizing. Given the width rules have been written for content-box I'm inclined to leave it alone

HeartSquared (approved):

PR #3255, opened 2023-02-03
feat: remove max-width in form groups

jwmccowan:
BREAKING CHANGE: Layouts that rely on a `max-width: 400px` checkbox or radio groups may be affected

## Why
The radio group has a css rule that all labels have a max-width of `400`.  Something like:

```
.radio-group {
  label {
    max-width: 400px;
  }
}
```

This number seems arbitrary and we have needed to add special css overrides in our app in order to have full width radio buttons.

## What
This change removes the max-width css rule.

### Before
Screen Shot 2023-02-03 at 3 08 58 pm

### After
Screen Shot 2023-02-03 at 3 09 16 pm

## Notes
We don't have any issues with the checkbox group, but I noticed that has a `width: 400px` css rule for the label.  I'm happy to remove that as part of this PR.

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/CheckboxGroup/CheckboxGroup.module.scss
+++ b/draft-packages/form/KaizenDraft/Form/CheckboxGroup/CheckboxGroup.module.scss
@@ -10,8 +10,4 @@
   &.noBottomMargin {
     margin-bottom: 0;
   }
-
-  label {
-    width: 400px;
-  }
 }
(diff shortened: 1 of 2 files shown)


gyfchong:
> We don't have any issues with the checkbox group, but I noticed that has a `width: 400px` css rule for the label. I'm happy to remove that as part of this PR.

@jwmccowan sounds good 😄

HeartSquared:
You'll want to add the `BREAKING CHANGE: {description}` to the commit itself under the main message, rather than just the PR :)
It needs to be specifically in that format too, so if you're unsure, you can use `yarn commit` to help you :)

gyfchong (approved):

PR #3257, opened 2023-02-03
feat: export context for Select

NatalieLoyzaga:
## Why

Export the context for Single Select so consumers can use `useSelect()`

## What

Export the context for Single Select


gyfchong:
I think this is a little too much work for the consumer to create custom options. Is it possible to expose the item's object through the children?

Given: `{ title: "hello", subtitle: "sub" }`
or even: `{ label: { title: "hello", subtitle: "sub" } }`

```
<Select>
  {({item,  ...props}) => (
     <Option {...props}>
        item.title
        item.subtitle
    </Option>
  )}
</Select>
```

NatalieLoyzaga:
@gyfchong I think we run into another TS conflict when exposing the Item. Happy to pair on this if you want. Ideally this PR was a fix to expose the context which was missed in the initial release.

gyfchong (approved):

PR #3259, opened 2023-02-08
docs(tailwind): fix invalid important field in config docs

JakePitman:
## Why
Arrays of strings aren't actually valid.

## What
Changes the `important` field example in the Tailwind config docs to a string. Fleshes out the explanation slightly.


mcwinter07 (approved):
🦖 🐜  Looks good!

PR #3264, opened 2023-02-08
chore: remove stale bot

gyfchong:
To keep the warning issue alive, I am removing the stale bot.


.github/workflows/stale.yml
```diff
@@ -1,22 +0,0 @@
-name: "Handle stale PRs and issues"
```
HeartSquared, line 1: Are you sure you don't want to keep it for PRs?

gyfchong: Hm, has it proven useful in your opinion?

HeartSquared: I suppose not recently since we've been diligent in keeping our open PRs relevant 🤔

HeartSquared (approved):

PR #3265, opened 2023-02-09
fix(filtermultiselect): trigger button tooltip consistency

Andsbf:
## Why
Update component tooltip to match others filters tooltips as per:

https://github.com/cultureamp/kaizen-design-system/blob/main/packages/date-picker/src/FilterDateRangePicker/components/Trigger/RemovableFilterTriggerButton/RemovableFilterTriggerButton.tsx#L38

Ticket: https://cultureamp.atlassian.net/browse/KZN-962

## What
Update tooltip interpolation


HeartSquared (approved):

HeartSquared:
Ah, you'll need to update your commit to be a conventional commit (use `yarn commit` if you don't remember the syntax) :)

Andsbf:
> Ah, you'll need to update your commit to be a conventional commit (use `yarn commit` if you don't remember the syntax) :)

🙈

HeartSquared (approved):

PR #3267, opened 2023-02-10
feat: add visible focus state to close button & centre close icon

caitlinmwillington:
## Why

The focus outline for the inline & toast notification close icon is not very visible in Firefox, causing an accessibility issue. This change brings the close icon focus state inline with the design system. 
The close icon was also not perfectly centred in its position, so this PR also minutely adjusts it so that it is better positioned.

## What

- Add better focus outline for close icon
- Centre the close icon (was a tiny bit out of alignment)


packages/notification/src/_styles.scss
```diff
@@ -328,22 +328,6 @@ $notification-slide-right: transform 300ms ease-out;
   }
 }
 
-%ca-notification__cancel__inner {
```
caitlinmwillington, line 331: This was only used in the Generic Notification component and is what was causing the misalignment

NatalieLoyzaga (approved):
Looks great! Awesome work thanks again :)

PR #3269, opened 2023-02-13
feat(tailwind): Add spacing token for 6px/.375 rem

JakePitman:
## Why
6px is a spacing value that comes up a lot in designs.

## What
Adds a 6px spacing token to the Tailwind package.

### DIFF
--- a/packages/tailwind/src/kz-spacing.ts
+++ b/packages/tailwind/src/kz-spacing.ts
@@ -5,6 +5,7 @@ export const kzSpacing = {
   1: ".0625rem",
   2: ".125rem",
   4: ".25rem",
+  6: ".375rem",
   8: ".5rem",
   12: ".75rem",
   16: "1rem",


NatalieLoyzaga:
Did we want to also include all the other missing spacing tokens ? Or are we trying to move away from them fully
e.g 
- `2.25rem` = `$spacing-lg`
- `3.75rem` = `$spacing-xxl`
- `5.25rem` = `$spacing-xxxxl`

NatalieLoyzaga (approved):

JakePitman:
> Did we want to also include all the other missing spacing tokens ? Or are we trying to move away from them fully e.g
> 
>     * `2.25rem` = `$spacing-lg`
> 
>     * `3.75rem` = `$spacing-xxl`
> 
>     * `5.25rem` = `$spacing-xxxxl`

Looks like we just want the `xs` one, and the higher values are being replaced. Reasoning is that a small difference is hardly noticeable at the higher end of the spectrum.

PR #3271, opened 2023-02-14
chore(docs): add migration to KaizenProvider

gyfchong:


packages/components/src/KaizenProvider/docs/KaizenProvider.stories.mdx
```diff
@@ -1,11 +1,15 @@
+
```
HeartSquared, line 1: 👀

packages/components/src/KaizenProvider/docs/code-diffs/migration-bonus.diff
```diff
@@ -0,0 +1,19 @@
...
+-     services={[...defaultPreset, ThemeProvider]}
++     services={[...defaultPreset, KaizenProvider]}
+      options={{
+        ...
+-       themeManager,
+      }}
+  >
+   ...
+  </FrontendServices>
+)
```
HeartSquared, line 19: New line EOF (other diff file too)? :)

packages/components/src/KaizenProvider/docs/code-diffs/migration-bonus.diff
```diff
@@ -0,0 +1,19 @@
...
++ import { KaizenProvider } from "@kaizen/components"
+
+- const themeManager = new ThemeManager(defaultTheme)
+
+...
+
+return(
+  <FrontendServices
+-     services={[...defaultPreset, ThemeProvider]}
++     services={[...defaultPreset, KaizenProvider]}
```
HeartSquared, line 11: Is `defaultPreset` meant to be `defaultTheme`?

packages/components/src/KaizenProvider/docs/code-diffs/migration-to-v1.diff
```diff
@@ -0,0 +1,20 @@
...
++ import { KaizenProvider, ThemeManager } from "@kaizen/components"
+
+const themeManager = new ThemeManager(defaultTheme)
+
+...
+
+return(
+  <FrontendServices
+-     services={[...defaultPreset, ThemeProvider]}
++     services={[...defaultPreset, KaizenProvider]}
```
HeartSquared, line 12: Is `defaultPreset` meant to be `defaultTheme`?

packages/components/src/KaizenProvider/docs/code-diffs/migration-bonus.diff
gyfchong: Nope, that's the FEF preset

HeartSquared: Oh. That's very confusing in this example. Maybe add a comment or put an example value for the variable?

gyfchong: Hm, that value comes from a separate package so I don't think I want to elude people to thinking they should be managing that value. I'll add a description to mention this example is from the FE-Template, hopefully less confusion.

HeartSquared (approved):

PR #3272, opened 2023-02-14
feat(select): Add tech docs KZN-991-tech-docs

NatalieLoyzaga:
## Why

- Provide technical documentation 

## What

- Add prop descriptions
- Expose arg types to the storybook docs


packages/select/docs/Select/Select.stories.tsx
```diff
@@ -3,12 +3,15 @@ import { Section } from "@react-stately/collections"
 import { Node } from "@react-types/shared"
 import { ComponentMeta, ComponentStory, Story } from "@storybook/react"
 import { withDesign } from "storybook-addon-designs"
+import { CodeBlock } from "@kaizen/design-tokens/docs/DocsComponents"
+import { Paragraph } from "@kaizen/typography"
```
gyfchong, line 7: I think reference the path instead of the package

gyfchong (approved):

PR #3277, opened 2023-02-15
feat: add validation to DateRangePicker

NatalieLoyzaga:
## Why

Add validation to FilterDateRangePicker

Figma
image (1)

Storybook - Date Range Picker Field

image

## What

- Allow `dateStart` and `dateEnd` to have a `status`
- add `validationMessage` to allow `dateStart` or `dateEnd` messages
- Add special typing so that if status: {dateStart: ""} exists validationMessage: {dateStart:""} is also required

https://user-images.githubusercontent.com/48232362/220523149-02b5316b-b278-4130-96f0-08b70ab8e483.mov


draft-packages/form/KaizenDraft/Form/Primitives/Input/Input.module.scss
```diff
@@ -19,6 +19,10 @@ $input-with-icon-padding: calc(
 .wrapper {
   position: relative;
 
+  &.hasStatus {
```
NatalieLoyzaga, line 22: To allow the validation border of dateStart to display at all times

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -28,11 +28,12 @@ export interface DateRangeInputFieldProps
   /**
    * Updates the styling of the validation FieldMessage
    */
-  status?: FieldMessageStatus
+  status?: { dateStart?: FieldMessageStatus; dateEnd?: FieldMessageStatus }
```
NatalieLoyzaga, line 31: Separate the `dateStart` and the `dateEnd` status due to ability to have separate validation

HeartSquared: Nice, but I do wonder what we expect the consumer to do when:
- There is an error for the field as a whole (you'd want both inputs highlighted)
- Both inputs have different errors and require different validation messages

Thoughts?

gyfchong: > Both inputs have different errors and require different validation messages

I'd argue that this would be an edge case since if both start/end are invalid, what are the chances that they'd need separate messages if the collective data is invalid they'd use a collective message?

An example scenario could be that the start/end dates are too far apart, and the message would be "Please enter dates which are closer together".

NatalieLoyzaga: I will be raising this with @jmitsios tomorrow for a designers perspective

HeartSquared: I can see how we could use this api as it is to solve it (get the consumer to figure it out), but it's worth thinking about anyway 😉 

However, when we put in the inbuilt validation, we might have someone type gibberish in one input, then type a disabled date with the other input, so we'll end up with different errors for each input. How should we go about piecing them together? Do we only show one message at a time?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.module.scss
```diff
@@ -43,3 +43,7 @@ $input-disabled-opacity: 0.3;
 .disabled {
   opacity: $input-disabled-opacity;
 }
+
+.fieldMessageIcon > span {
+  align-self: flex-start;
```
HeartSquared, line 48: Is this unique to the DatePicker error messages, or should it be a global change to the FieldMessage?

packages/date-picker/docs/DateRangeInputField.stories.tsx
```diff
@@ -127,15 +127,85 @@ const StickerSheetTemplate: Story<{
...
+                dateEnd: "error",
+                dateStart: "error",
+              }}
+              validationMessage={
+                <ul style={{ margin: "0px", paddingInlineStart: "24px" }}>
+                  <li>
+                    "Date to" cannot be earlier than the "Date from" selection.
+                  </li>
+                  <li>"Date from" is not a valid date selection.</li>
+                </ul>
```
HeartSquared, line 190: Since this will be a consistent way of displaying multiple messages, I suggest this be a subcomponent where the consumer can just pass in: 
```
status={{
  dateEnd: "error",
  dateStart: "error",
}}
validationMessage={{
  dateStart: '"Date from" is close to the submission date.',
  dateEnd:
    '"Date to" cannot be earlier than the "Date from" selection.',
}}
```
and it renders the list for them.

Of course, we should still allow them to render whatever they want if they choose not to pass in the object but instead just a JSX.Element (which it seems you've already done).

packages/date-picker/src/types.ts
```diff
@@ -66,3 +66,25 @@ export interface DisabledDayMatchers {
...
+  dateStart?: FieldMessageStatus
+  dateEnd?: FieldMessageStatus
+}
+
+export type DateRangeMessageType = {
+  dateStart: React.ReactNode
+  dateEnd: React.ReactNode
+}
+
+export type DateRangeValidatioMessage = React.ReactNode | DateRangeMessageType
```
HeartSquared, line 80:
```suggestion
export type DateRangeValidationMessage = React.ReactNode | DateRangeMessageType
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -107,13 +112,31 @@ export const DateRangeInputField = React.forwardRef<
...
+            />
+            <FieldMessage
+              id={errorMessageId}
+              message={validationMessage.dateEnd}
+              status={status?.dateEnd}
+              reversed={isReversed}
+              classNameOverride={styles.fieldMessageIcon}
+            />
+          </>
         )}
```
HeartSquared, line 140: This is getting too complex. Best to make a subcomponent (and add tests) :)

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.module.scss
NatalieLoyzaga: I plan on moving this and the icon update to FieldMessage in a separate PR

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+          status={{
+            dateEnd: "error",
+          }}
+          validationMessage={<p>Validation message</p>}
+        />
+      )
+      expect(screen.getByText("Validation message")).toBeVisible()
+    })
+
+    it("when consumer passes validationMessage object but has the same status", () => {
```
HeartSquared, line 52:
```suggestion
    it("combines multiple messages with the same status", () => {
```

Something on the lines of that.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+            dateStart: "error",
+          }}
+          validationMessage={{
+            dateStart: '"Date from" is not a valid date selection.',
+            dateEnd:
+              'Date to" cannot be earlier than the "Date from" selection.',
+          }}
+        />
+      )
+      expect(container.getElementsByClassName("error").length).toBe(1)
```
HeartSquared, line 67: Is it possible to check the contents rendered is a multi-item list? The current expectation just tells me there's a single error box.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <FieldMessage
+        id={errorMessageId}
+        message={validationMessage}
+        status={status?.dateStart || status?.dateEnd}
+        reversed={isReversed}
+        classNameOverride={styles.fieldMessageIcon}
+      />
+    )
+  } else if (status?.dateEnd === status?.dateStart) {
+    return (
```
HeartSquared, line 40:
```suggestion
  } 
  
  if (status?.dateEnd === status?.dateStart) {
    return (
```

When you have early returns, you don't need to use `else` :)

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <li>{validationMessage.dateStart}</li>
+          </ul>
+        }
+        status={status?.dateStart || status?.dateEnd}
+        reversed={isReversed}
+        classNameOverride={styles.fieldMessageIcon}
+      />
+    )
+  } else {
+    return (
```
HeartSquared, line 55:
```suggestion
  }
    
  return (
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,76 @@
...
+  errorMessageId?: string
+  isReversed?: boolean
+}
+
+export const DateRangeValidationMessage = ({
+  status,
+  validationMessage,
+  errorMessageId,
+  isReversed,
+}: DateRangeValidationMessageProps): JSX.Element | null => {
```
HeartSquared, line 28:
```suggestion
}: DateRangeValidationMessageProps): JSX.Element => {
```

It looks like you'll always be returning a JSX.Element :)

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.spec.tsx
```diff
@@ -45,15 +45,33 @@ describe("<DateRangeInputField />", () => {
...
     const inputStart = screen.getByRole("textbox", { name: "Date from" })
     expect(inputStart).toHaveAccessibleDescription(
-      "Error message This is an error Input format : dd/mm/yyyy"
+      "Error message Date Start has an error Input format : dd/mm/yyyy"
     )
-    expect(screen.getByText("This is an error")).toBeVisible()
+    expect(screen.getByText("Date Start has an error")).toBeVisible()
+  })
+
+  it("adds validation message to description if it exists", () => {
```
HeartSquared, line 61: You probably don't need this test since the logic is done within `DateRangeValidationMessage`.

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
HeartSquared: Alternatively, you could also add something like this to the top:

```
if (!validationMessage) return null
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,76 @@
...
+export type DateRangeValidationMessageProps = {
+  /**
+   * Updates the styling of the validation FieldMessage
+   */
+  status?: DateRangeValidationStatus
+  /**
+   * A descriptive message for `status` states
+   */
+  validationMessage?: DateRangeValidationMessageType
+  errorMessageId?: string
```
HeartSquared, line 19:
```suggestion
  id?: string
```

This is pretty agnostic, and could potentially not be an error :)

packages/date-picker/src/types.ts
```diff
@@ -66,3 +66,27 @@ export interface DisabledDayMatchers {
...
+export type DateRangeValidationMessageType =
+  | React.ReactNode
+  | DateRangeMessageType
+
+export function isDateRangeMessageType(
+  message: DateRangeValidationMessageType
+): message is DateRangeMessageType {
+  return (
+    message != undefined &&
+    (message as DateRangeMessageType).dateStart !== undefined &&
```
HeartSquared, line 89: Shouldn't it be `or`? You can probably write a unit test for this to confirm it's doing what you're expecting 😉 

```suggestion
    (message as DateRangeMessageType).dateStart !== undefined ||
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.module.scss
```diff
@@ -0,0 +1,10 @@
+@import "~@kaizen/design-tokens/sass/border";
+.fieldMessageList {
+  margin: 0px;
+  padding-inline-start: 24px;
```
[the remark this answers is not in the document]
NatalieLoyzaga: w@t R th0se

packages/date-picker/src/types.ts
NatalieLoyzaga: How I thought the API would be that the consumer either passes a React.ReactNode or the 
```
{
  dateStart: React.ReactNode
  dateEnd: React.ReactNode
}
```

where both `dateStart` and `dateEnd` are required. Otherwise if they only wanted one then they can pass it via the `React.ReactNode`?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const DateRangeValidationMessage = ({
+  status,
+  validationMessage,
+  id,
+  isReversed,
+}: DateRangeValidationMessageProps): JSX.Element | null => {
+  if (!validationMessage) return null
+
+  if (status?.dateEnd !== undefined && status?.dateStart !== undefined) {
```
HeartSquared, line 30: So there's no way to have a single message but have both fields have an error border?

NatalieLoyzaga: I think that is not a pattern we actually want to encourage - There should be a message for each to provide enough feedback to the user

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
```diff
@@ -15,6 +15,8 @@ const openFilter = async (): Promise<void> => {
 
 const FilterDateRangePickerWrapper = ({
   selectedRange,
+  status,
+  validationMessage,
```
HeartSquared, line 19: I think we should leave these as part of the spread props :)

packages/date-picker/docs/controls/validationControls.ts
```diff
@@ -0,0 +1,76 @@
+export const validationControls = {
```
HeartSquared, line 1: Any possibility of making a custom control which is the combination of the valid combinations instead? And disable the controls for the individual props.

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
NatalieLoyzaga: It causes a TS error otherwise

HeartSquared: Hmm...that's a problem though. We should be able to safely pass in these values without an error.

NatalieLoyzaga: It says we can't pass undefined to it

NatalieLoyzaga: image

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  describe("will render a single validation field message", () => {
+    test("when the consumer has given it a node", () => {
+      render(
+        <DateRangeValidationMessage
+          id="error-message-id"
+          status={{
+            dateEnd: "error",
+          }}
+          validationMessage={{
+            dateEnd: <p>Validation message</p>,
```
gyfchong, line 15: Do we need separate tests for string vs react component rendering?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+    validationMessage={{
+      dateStart: '"Date from" is close to the submission date.',
+      dateEnd: '"Date to" cannot be earlier than the "Date from" selection.',
+    }}
+  />
+)
+
+describe("<DateRangeValidationMessage />", () => {
+  describe("will render a single validation field message", () => {
+    it("when the consumer has given it a string", () => {
```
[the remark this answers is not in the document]
NatalieLoyzaga: @gyfchong I had originally had separate tests here but removed it for just one

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
NatalieLoyzaga: https://github.com/cultureamp/kaizen-design-system/pull/3277#discussion_r1118180643

gyfchong (dismissed):

packages/date-picker/docs/controls/validationControls.ts
```diff
@@ -0,0 +1,113 @@
...
```
HeartSquared, line 1: I'm still not a fan of being able to choose incompatible combinations using the controls, especially since there isn't any description or anything telling me I shouldn't (and it renders unexpectedly too, showing the start date having the error border, but rendering only the end date message).

image

But if it's going to be merged in as is and fixed later, I think you also need to add these controls to DateRangeInputField.

NatalieLoyzaga: @HeartSquared I have updated both FilterDRP and the Input stories :)

packages/date-picker/docs/controls/validationControls.ts
```diff
@@ -0,0 +1,61 @@
...
```
HeartSquared, line 1: Just need to disable the controls for the OG props (you can do this with `{ control: "disabled" }`) :)

NatalieLoyzaga: Updateddd :) @HeartSquared

HeartSquared (approved):

PR #3278, opened 2023-02-15
feat: Convert Select to be a generic component

NatalieLoyzaga:
## Why

- The Select component is currently tied to the item type `SingleItemType`.
```
export interface SingleItemType {
  label: string
  value: ValueType | SectionChildType[]
}

export type SectionChildType = {
  label: string
  value: ValueType
}
```

This restricts the consumer's data set and therefore does not allow the consumer to pass any other data structures.

## What

- Transform the `Select` component into a generic component. 
- Allow the consumer to specify their item type (which extends the `SingleItemType`)

### DIFF
--- a/packages/select/docs/Select/CustomItem/CustomItem.stories.tsx
+++ b/packages/select/docs/Select/CustomItem/CustomItem.stories.tsx
@@ -1,0 +1,124 @@
+import React from "react"
+import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { withDesign } from "storybook-addon-designs"
+import { CATEGORIES, SUB_CATEGORIES } from "../../../../../storybook/constants"
+import { figmaEmbed } from "../../../../../storybook/helpers"
+import { Select } from "../../../src/Select"
+import { ListBoxSection } from "../../../src/Select/components/ListBoxSection"
+import {
+  CustomOptionWithSubtitle,
+  WithSubtitleType,
+} from "./OptionWithSubtitle"
+
+export default {
+  title: `${CATEGORIES.components}/${SUB_CATEGORIES.select}/Select/Custom Item`,
+  component: Select,
+  parameters: {
+    actions: {
+      argTypesRegex: "^on.*",
+    },
+    docs: {
+      source: { type: "code" },
+      description: {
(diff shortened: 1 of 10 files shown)


packages/select/docs/Select/OptionWithSubLabel.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { HTMLAttributes } from "react"
```
NatalieLoyzaga, line 1: This would be done on a consumer level - here for example and testing purposes only

packages/select/docs/Select/CustomItem/ListBoxSectionWithSubtitle.tsx
```diff
@@ -0,0 +1,53 @@
+import React from "react"
```
NatalieLoyzaga, line 1: This would be done on a consumer level - here for example and testing purposes only

packages/select/docs/Select/CustomItem/CustomItem.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+        optionsProps.items.map(item =>
+          item.type === "section" ? (
+            <ListBoxSectionWithSubtitle key={item.key} section={item} />
+          ) : (
+            <CustomOptionWithSubtitle
+              {...optionsProps}
+              key={item.key}
+              item={item}
+            />
+          )
```
gyfchong, line 108: I think this level of detail for the consumer is too tedious, we're relying on the consumers to remember this is needed otherwise a feature is no longer available to them. We also lose some control over this feature.

Could we use the [same pattern from `react-aria`](https://react-spectrum.adobe.com/react-aria/useSelect.html#usage) where they collect a single child from the consumer? That way we can internally replace the default `Option` with their custom one?
```
<Select label="Pick an engineering major" items={options}>
        {(item) => <Item>{item.name}</Item>}
</Select>
```

NatalieLoyzaga: Yeah its not very intuitive but is also not necessarily a pattern that I have seen but doing it for testing purposes.

In terms of your suggestion - you will have to run me through your idea

PR #3279, opened 2023-02-15
feat: Backport new pixel-based spacing tokens to design tokens

JakePitman:
## Why
So that consumers who do not use Tailwind can still use the new spacing token system.
Tailwind is also generating spacing utilities based off of these tokens now, for consistency between TW and design-tokens.

## What
- Adds the new spacing tokens to design-tokens package
- Generates Tailwind utilities from design-tokens programatically instead of from a hard-coded ts file


packages/design-tokens/sass/spacing.scss
```diff
@@ -1,5 +1,53 @@
...
+$spacing-80: var(--spacing-80, 5rem);
+$spacing-96: var(--spacing-96, 6rem);
+$spacing-112: var(--spacing-112, 7rem);
+$spacing-128: var(--spacing-128, 8rem);
+$spacing-160: var(--spacing-160, 10rem);
+$spacing-200: var(--spacing-200, 12.5rem);
+$spacing-240: var(--spacing-240, 15rem);
+$spacing-280: var(--spacing-280, 17.5rem);
+$spacing-320: var(--spacing-320, 20rem);
+$spacing-0-id: --spacing-0;
```
HeartSquared, line 27: These look strange 🤔

packages/design-tokens/src/themes/heart.ts
```diff
@@ -146,6 +146,31 @@ export const heartTheme: Theme = {
     xxxl: "4.5rem",
     xxxxl: "5.25rem",
     xxxxxl: "6rem",
+
```
HeartSquared, line 149: Is this blank line safe?

packages/tailwind/src/kz-spacing.ts
```diff
@@ -1,28 +1,10 @@
...
-  240: "15rem",
-  280: "17.5rem",
-  320: "20rem",
-}
+const allSpacingTokenPairs: Array<[string | number, string]> = Object.entries(
+  heartTheme.spacing
+)
+
+// Filters out non-pixel-based spacing tokens
+export const kzSpacing = Object.fromEntries(
```
HeartSquared, line 8: Having a little trouble understanding the intent here. Could you write a unit test? 😂

packages/design-tokens/tokens/spacing.json
```diff
@@ -1,5 +1,53 @@
 {
   "spacing": {
+    "0": "var(--spacing-0, 0)",
+    "1": "var(--spacing-1, .0625rem)",
+    "2": "var(--spacing-2, .125rem)",
+    "4": "var(--spacing-4, .25rem)",
+    "6": "var(--spacing-6, .375rem)",
```
gyfchong, line 7: Are these put here for our own internal use?

JakePitman: Not sure, this file is autogenerated

packages/design-tokens/src/themes/heart.ts
JakePitman: Not sure, I've removed it now though

packages/tailwind/src/kz-spacing.ts
JakePitman: Good call, added one in.

packages/design-tokens/sass/spacing.scss
JakePitman: Yeah agreed. But this file is autogenerated and the previous shirt sizes get the `-id` treatment too for whatever reason.

packages/tailwind/src/kz-spacing.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const dummyData = {
+  ...numericKeyPairs,
+  ...nonNumericKeyPairs,
+}
+
+describe("filterOutNonPixelTokens", () => {
+  const result = filterOutNonPixelTokens(dummyData)
+
+  test("It removes pairs with non-numeric keys", () => {
```
christian314159, line 21: For future reference, the way you're writing the test, you could use the `it()` syntax here:
```suggestion
  it("removes pairs with non-numeric keys", () => {
```

packages/design-tokens/sass/spacing.scss
HeartSquared: Ohhhh they do. I completely missed it since they were slotted in between. I wonder why these ones don't slot between each other like the others do 🤔

packages/design-tokens/tokens/spacing.json
HeartSquared: Apparently this file is created, then used to create the `less` and `scss` variables, but being `json` we can't write a comment to say this is auto-generated.

packages/tailwind/src/kz-spacing.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+describe("filterOutNonPixelTokens", () => {
+  const result = filterOutNonPixelTokens(dummyData)
+
+  it("It removes pairs with non-numeric keys", () => {
+    Object.keys(nonNumericKeyPairs).forEach(key => {
+      expect(result).not.toHaveProperty(key)
+    })
+  })
```
HeartSquared, line 25: This isn't needed since the test below has an exact match

packages/tailwind/src/kz-spacing.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  it("Keeps all numeric keys", () => {
+    expect(result).toEqual(numericKeyPairs)
```
HeartSquared, line 27:
```suggestion
const result = filterOutNonPixelTokens({
	0: "0",
	xs: "0.375rem",
	1: ".0625rem",
	sm: "0.75rem",
	2: ".125rem",
})

expect(result).toEqual({
	0: "0",
	1: ".0625rem",
	2: ".125rem",
})
```

I think it's actually a little easier to read without too many variables in a test, especially since there aren't many in this test. Something like this is very explicit and clear what we're expecting to get.

packages/tailwind/src/kz-spacing.ts
```diff
@@ -1,28 +1,14 @@
-import { defaultTheme } from "@kaizen/design-tokens"
+const { heartTheme } = require("@kaizen/design-tokens")
```
HeartSquared, line 1:
```suggestion
import { heartTheme } from "@kaizen/design-tokens"
```

Using `require` is an old pattern. Did it not work with `import`?

packages/tailwind/src/kz-spacing.ts
```diff
@@ -1,28 +1,14 @@
...
-  128: "8rem",
-  160: "10rem",
-  200: "12.5rem",
-  240: "15rem",
-  280: "17.5rem",
-  320: "20rem",
+// Filters out non-pixel-based spacing tokens, such as shirt sizes
+export const filterOutNonPixelTokens = (spacingTokensObject: {
+  [key: string]: string
+}): { [key: string]: string } => {
```
HeartSquared, line 6:
```suggestion
export const filterOutNonPixelTokens = (spacingTokensObject: Record<string, string>): Record<string, string> => {
```

This should work :)

packages/design-tokens/tokens/spacing.json
gyfchong: Sorry, I meant to question their existence as a value we can use.

HeartSquared: Yeah I question their existence too. A card for the backlog? 😉

gyfchong: I'm assuming this came from chatting with Elliot and what to do about values we use internally but aren't part of the original scale?

gyfchong: I'd rather not expose what could potentially spawn a bunch of "off-scale" usage

packages/tailwind/src/kz-spacing.spec.ts
JakePitman: Changed to "it" now

packages/tailwind/src/kz-spacing.ts
JakePitman: It was throwing an error... But not any more. Weird.

Changed to import now

packages/tailwind/src/kz-spacing.ts
JakePitman: Done

packages/design-tokens/sass/spacing.scss
mcwinter07: hmmm maybe numbers have a strange sequential sorting logic? In any case, looks legit 🤷 RE: why they output `-id` I would assume it's some kind of convention around using global and locally scoped CSS variables, though that's a stab in the dark

packages/design-tokens/tokens/spacing.json
mcwinter07: By off-scale usage @gyfchong do you mean adding the sizes that align to the older t-shirt equivalents?

gyfchong: Not necessarily, if we're promoting a specific scale which consumers should use to space things apart (ie. the multiples of 4) then anything that doesn't conform to the scale is considered "off-scale". In this example, `1,2,6` are considered off scale. The scale is inherently there to make spacing math easier, so having these odd numbers might mess with that.

Although I just realised these are part of the Tailwind (new) spacing, which I thought hadn't included these so it didn't make sense to me as to why they'd be included in the design-tokens, so yeah we can take this discussion elsewhere.

packages/tailwind/src/kz-spacing.spec.ts
JakePitman: Removed now

packages/tailwind/src/kz-spacing.spec.ts
JakePitman: Alrighty 👍

packages/design-tokens/src/types.ts
```diff
@@ -174,6 +174,31 @@ export type Theme = {
     xxxl: string
     xxxxl: string
     xxxxxl: string
+
```
HeartSquared, line 177: Eh? How did this blank line appear again? O_O

packages/design-tokens/tokens/spacing.json
mcwinter07: Ah gotcha, thanks for the clarification 👍

packages/design-tokens/src/types.ts
JakePitman: No idea. But he's been dealt with now 🔫 😎

gyfchong: What's dead may never die

HeartSquared (approved):
LGTM :)

PR #3256, opened 2023-02-03
feat: add css pre processing to AIO

gyfchong:

PR #3282, opened 2023-02-16
fix: components bundling

gyfchong:
## Why
Our `@kaizen/components` AIO package is made up of both Typescript and SCSS modules files, bundling to a `dist` folder requires the bundler to be able to handle both the Typescript files as well as the SCSS files as they both need to be moved to the `dist` file for components to properly reference this.

In the rest of the repo, component's bundles are outputted into the same directory which lets the bundles access the SCSS files. This makes various config difficult to manage/has unintended consequences (eg. [gitignore](https://github.com/cultureamp/kaizen-design-system/blob/main/.gitignore) needs specific lines to allow for config js files)

## What

Introduced `esbuild` which will handle Typescript, SCSS and Tailwind bundling all in one config. This outputs a `dist` directory and encapsulates cjs, esm and any other future bundle format necessary into folders, and supports the next step which is to [pre-process our CSS](https://github.com/cultureamp/kaizen-discourse/issues/30).

## Version
Placing this as a patch because technically this should've been working to begin with 😅 

## Additional context
- https://github.com/cultureamp/kaizen-design-system/pull/3256
- https://github.com/cultureamp/kaizen-discourse/issues/30


packages/components/esbuild.config.js
```diff
@@ -1,7 +1,74 @@
...
+
+// const entryPoints = readdirSync(join(process.cwd(), "src"))
+//   .filter(
+//     file =>
+//       file.endsWith(".ts") &&
+//       statSync(join(process.cwd(), "src", file)).isFile()
+//   )
+//   .map(file => `src/${file}`)
+
+// esm output bundles with code splitting
```
HeartSquared, line 25: What's happening here?

packages/components/esbuild.config.js
```diff
@@ -1,7 +1,74 @@
...
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+;(async () => {
```
HeartSquared, line 27: Starting with a `;`? 😛

gyfchong:
Although, wondering what value is ESM when we only bundled CJS before and we got no complaints 😛 

@jakubriedl any thoughts?

packages/components/esbuild.config.js
gyfchong: Ah leftover code from an example, I resolved this with the glob

packages/components/tsconfig.dist.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "../../tsconfig.dist.json",
   "compilerOptions": {
-    "outDir": "dist"
+    "declarationDir": "dist/types",
+    "emitDeclarationOnly": true,
+    "declaration": true
```
gyfchong, line 6: `tsc` now relegated to simply produce types

packages/components/package.json
```diff
@@ -33,7 +36,9 @@
     "@kaizen/tailwind": "^0.4.0",
     "autoprefixer": "^10.4.13",
     "esbuild": "^0.17.8",
-    "tailwindcss": "^3.2.6"
+    "fs": "^0.0.1-security",
```
gyfchong, line 39: Actually I don't think this is right lol

packages/components/esbuild.config.js
HeartSquared: For real? 🥲 What if we remove both `;(` (wow, maybe the linter is telling us it's not happy)

gyfchong: Seems it's some kind of JS optimisation.

> The leading `;`in front of immediately-invoked function expressions is there to prevent errors when appending the file during concatenation to a file containing an expression not properly terminated with a `;`.

https://stackoverflow.com/questions/1873983/what-does-the-leading-semicolon-in-javascript-libraries-do

HeartSquared (dismissed):
#trustfall

HeartSquared (approved):

PR #3285, opened 2023-02-16
feat: re-export-all-existing-components-from-the-new-aio

gyfchong:
## Why

To allow consumers to gradually adopt the all-in-one package `@kaizen/components` we are exporting all existing component packages from this package. Consumers can now safely switch their imports.

## What

All packages/draft-packages are added as a dependency and exported from the appropriate indexes.

## Additional context

This is still experimental, but as we don't have any consumers yet, it should be safe to test with versions for now. Worst case if someone decides to try use this, their import just won't work 😅 -- it all seems to work locally at the very least.


packages/components/src/index.ts
```diff
@@ -1,2 +1,57 @@
...
+export * from "@kaizen/brand"
+export * from "@kaizen/brand-moment"
+export * from "@kaizen/button"
+export * from "@kaizen/component-base"
+export * from "@kaizen/date-picker"
+
+// Part of design tokens is already exported from KaizenProvider
+export {
+  makeCssVariableDefinitionsMap,
+  makeCSSVariableTheme,
```
christian314159, line 15: The inconsistency here is killing me!

gyfchong: I can put it at the bottom, out of sight, out of mind lol

packages/components/src/index.ts
```diff
@@ -1,2 +1,59 @@
...
+export * from "@kaizen/responsive"
+export * from "@kaizen/rich-text-editor"
+export * from "@kaizen/split-button"
+
+// Special cases
+
+// Part of design tokens is already exported from KaizenProvider
+export {
+  makeCssVariableDefinitionsMap,
+  makeCSSVariableTheme,
```
christian314159, line 49: @gyfchong I don't think you got what it was that is killing me here ;)
`makeCssVar...` vs `makeCSSVar...` - I wish those were consistent.

gyfchong: OH haha

mcwinter07 (commented):
Looks hunky dory to me, though i did try to test it out with yalc to see if it would work with the FE template and got a bunch of esbuild errors on all the SVGs 🤔 Feels like it might be more a yalc issue since all out build seems to be green 🤷 If that's not an issue happy to ✅ 
Screenshot 2023-02-20 at 12 19 59 pm

gyfchong:
Thanks @mcwinter07 I can test further once this is published. All good for the ✅

packages/components/package.json
```diff
@@ -29,6 +29,48 @@
   },
   "homepage": "https://github.com/cultureamp/kaizen-design-system#readme",
   "dependencies": {
+    "@kaizen/a11y": "^1.6.22",
+    "@kaizen/brand": "^1.5.4",
+    "@kaizen/brand-moment": "^1.12.55",
+    "@kaizen/button": "^2.0.17",
+    "@kaizen/component-base": "^1.1.3",
+    "@kaizen/date-picker": "^5.17.9",
+    "@kaizen/design-tokens": "^10.2.0",
```
HeartSquared, line 38: Hmm...so if we're exporting `@kaizen/design-tokens` from here, do we therefore remove the peerDep? Or we still need to keep it?

yarn.lock
```diff
@@ -2101,6 +2101,19 @@
     "@jridgewell/resolve-uri" "3.1.0"
     "@jridgewell/sourcemap-codec" "1.4.14"
 
+"@kaizen/draft-form@^9.0.1":
```
HeartSquared, line 2104: What is this doing here by itself? 😬 
I think it should be v10 now

packages/components/package.json
gyfchong: Oh good point

yarn.lock
gyfchong: Ah, it's not up-to-date. I'm wondering if I should use `*` instead.

gyfchong: Shouldn't lerna bump my package though...

HeartSquared: `*` sounds risky somehow. But I think as long as when you merge this in it's up to date, Lerna should help us keep it up to date moving forward.

HeartSquared: Lerna will bump dependencies on release. You're adding this for the first time, and it's still in a PR, so it doesn't know to do it here :)

gyfchong: Stick with version numbers for now, thanks!

packages/components/src/__future__/index.ts
```diff
@@ -1 +1,5 @@
 export * from "./ExampleComponent"
```
HeartSquared, line 1: Should we remove this? Or keep it?

gyfchong: Need to keep it for testing purposes or until there's a future component

HeartSquared: Oh right. Because the exports don't count. Gotcha 👍🏻

HeartSquared (approved):

PR #3287, opened 2023-02-16
fix: allow Button to take HTML Button Element attributes, change existing "form" variant to compact prop

jshli:
…


gyfchong:
Fixed with https://github.com/cultureamp/kaizen-design-system/pull/3576

PR #3289, opened 2023-02-16
fix(brandmoment): change stylesheet to module.scss

hqta:
## Why

Team Home are doing initial work to upgrade NextJS and noticed BrandMoment component has stylesheet in `.scss` no `.module.scss` which caused some styling issues. 

## What

Change `BrandMoment.scss` to `BrandMoment.module.scss`


NatalieLoyzaga (approved):
Thanks!

PR #3299, opened 2023-02-20
feat: Update Toast Notification to handle long strings KZN-1005

NatalieLoyzaga:
## Why

- Wrap long single words within the title - most likely caused through translations

## What

- Update the text container to include a max-width and allow word breaks


packages/notification/docs/ToastNotification.stories.tsx
```diff
@@ -224,6 +224,20 @@ const StickerSheetTemplate: Story = () => {
       ),
       autohide: !IS_CHROMATIC,
     })
+    addToastNotification({
+      type: "positive",
+      title: "VerylongTitleExampleVerylongtitleExample ",
```
gyfchong, line 229: Needs to be longer!
image

gyfchong (dismissed):

NatalieLoyzaga:
@gyfchong I had to update this so that the styles only target the `Toast Notification` - Chromatic picked up changes on the `Inline Notification`

packages/notification/src/_styles.scss
```diff
@@ -201,6 +201,13 @@ $notification-slide-right: transform 300ms ease-out;
     margin-left: 6px;
   }
 
+  %ca-notification---toast & {
+    margin-left: 6px;
+    max-width: calc(100% - #{$spacing-xl});
+    overflow-wrap: break-word;
+    hyphens: auto;
```
gyfchong, line 208: Maybe we can apply these styles based on a `@container` query? I'm not entirely sold on having a `isToast` prop for this 😅

NatalieLoyzaga: Hm Im getting a lint issue when trying to use it - unknown at rule container css(unknownAtRules)

gyfchong: Probably worth digging a little into why, we'll need this for future work too. If it ends up being too deep, make a ticket and we can move on

NatalieLoyzaga: I added a container query - thoughts?

gyfchong: That's like magic lol

gyfchong (approved):

PR #3300, opened 2023-02-20
chore(deps): Bump react-day-picker to 8.6.0

dougmacknz:


dougmacknz:
Needs https://github.com/gpbl/react-day-picker/issues/1688 to be resolved before updating.

PR #3301, opened 2023-02-20
feat(filter-drp): update the calendar when start input is changed

HeartSquared:
## What

Update the calendar to reflect the same start month as the input start date.
- If there is no start date selected, fall back onto the `defaultMonth` provided by the consumer
- If there is no `defaultMonth`, fall back to today

Also adjusted the existing `useEffect` for resetting input values to only trigger when the popover is closed, as it is used to simulate a re-render (based on external triggers modifying the values).
**Note:** This issue does not exist in the [new Filter abstraction pattern](https://cultureamp.atlassian.net/wiki/spaces/TV/pages/2997683785/Solution+preview+Kaizen+Filter+components).

## Why

[KZN-387](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?modal=detail&selectedIssue=KZN-387)

Better UX for the consumer.

[KZN-387]: https://cultureamp.atlassian.net/browse/KZN-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -112,16 +124,18 @@ export const FilterDateRangePicker = ({
...
-      setInputRangeStartValue(transformDateToInputValue(selectedRange?.from))
-    }
-    if (rangeEnd !== inputRangeEndValue) {
-      setInputRangeEndValue(transformDateToInputValue(selectedRange?.to))
+    if (!isOpen) {
+      const rangeStart = transformDateToInputValue(selectedRange?.from)
+      const rangeEnd = transformDateToInputValue(selectedRange?.to)
+
+      if (rangeStart !== inputRangeStartValue) {
+        setInputRangeStartValue(transformDateToInputValue(selectedRange?.from))
```
gyfchong, line 132: Can you use the `rangeStart` variable here instead of `transformDateToInputValue` again?

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -90,6 +90,18 @@ export const FilterDateRangePicker = ({
       : null
   )
   const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [startMonth, setStartMonth] = useState<Date>(
+    selectedRange?.from || defaultMonth || new Date()
```
gyfchong, line 94: `selectedRange` get's referenced a lot, could it be placed in a variable at the very top and referenced all the way down?

gyfchong:
Should this work on blur as well? I found it only works on enter keypress

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
HeartSquared: I most certainly can 😂 Derp

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
HeartSquared: You mean `selectedRange?.from`? Probably 🤔

HeartSquared:
> Should this work on blur as well? I found it only works on enter keypress

It works for me :'( What are you doing specifically?

gyfchong (approved):

PR #3306, opened 2023-02-21
refactor: add base directive to tailwind.scss

mcwinter07:
## Why
We had declared the utilities directive twice instead of base, which should be present

## What
- adds `@tailwind base` to tailwind.scss


gyfchong (approved):
Lols

PR #3307, opened 2023-02-21
chore: add preflight to tailwind config

mcwinter07:
## Why
validate UI changes against chromatic and test Tailwind Preflight

## What
- remove flag in tailwind config to disable pre-flight


gyfchong:
Yeahhh there's a lot of chromatic regression lol...

mcwinter07:
> Yeahhh there's a lot of chromatic regression lol...

@gyfchong think its worth creating a tickets to insulate the styles from these specific components from the preflight or just tackle them in this PR?

gyfchong:
@mcwinter07 I think we're going to have to park this ticket and assess the solution options for insulation. But this was good as an exploration, I'll reply to Jakub's GH discussion outlining our next steps.

mcwinter07:
> @mcwinter07 I think we're going to have to park this ticket and assess the solution options for insulation. But this was good as an exploration, I'll reply to Jakub's GH discussion outlining our next steps.

Cool beans - just as it relates to some of the work I've been doing on updating the FE template tailwind config - I should avoid adding the preflight into that for now then @gyfchong?

gyfchong:
> Cool beans - just as it relates to some of the work I've been doing on updating the FE template tailwind config - I should avoid adding the preflight into that for now then @gyfchong?

Yeah that's correct, let's put a pin on anything related to Preflight support

PR #3311, opened 2023-02-21
fix: replace esbuild with tsup

gyfchong:
Replacing ESBuild with tsup has fixed direct SCSS imports + CJS/ESM is correctly chosen from the consumers (idk how this happened, just did lol)

Added icons to the lib cause can't quite re-export them from the original, still contemplating moving icons to it's own package 🤔


mcwinter07 (commented):
Doesn't break yalc when I publish it this time though I did try to test out the Example component from future on the FE template - not sure if I'm doing something wrong but it doesn't look like the CSS modules came through 🤔 

Example:
Screenshot 2023-02-22 at 4 03 29 pm
Example DOM:
Screenshot 2023-02-22 at 4 03 44 pm

I may not be using / importing it incorrectly? 🤷 

```
import { ExampleComponent } from "@kaizen/components/future"
```

packages/components/src/__future__/ExampleComponent/ExampleComponent.tsx
```diff
@@ -3,7 +3,7 @@ import styles from "./ExampleComponent.module.scss"
 
 // Placeholder for illustration DO NOT COPY
 export const ExampleComponent = (): JSX.Element => (
-  <div className={`${styles.exampleComponent} bg-purple-600`}>
+  <div className={`${styles.exampleComponent} bg-green-300`}>
```
HeartSquared, line 6: Haha why?

packages/components/src/KaizenProvider/index.ts
```diff
@@ -1,2 +1 @@
 export * from "./KaizenProvider"
-export * from "./ThemeProvider"
```
HeartSquared, line 2: We don't want this anymore?

packages/components/src/index.ts
```diff
@@ -7,6 +7,14 @@ export * from "@kaizen/brand"
 export * from "@kaizen/brand-moment"
 export * from "@kaizen/button"
 export * from "@kaizen/component-base"
+export {
+  Icon,
+  Box,
+  Dropdown,
+  Layout,
+  Spacing,
+  Text,
```
HeartSquared, line 16: Do we want all of these even though most are deprecated patterns? Especially `Text`, which is replaced by `@kaizen/typography`

packages/components/src/__future__/ExampleComponent/ExampleComponent.tsx
gyfchong: Page Templates was already using `bg-purple-600` so it was in their CSS bundle and I needed to test whether or not our tailwind styles were making it through, and the component was just working cause the consumer had the class already 😄

packages/components/src/KaizenProvider/index.ts
gyfchong: Nope, `KaizenProvider` replaces it and it's deprecated from `design-tokens` with suggesting `KaizenProvider` as the replacement.

packages/components/src/index.ts
gyfchong: HMMMM this is a tough one. On one hand I want everyone to be able to adopt, but on the other, it would take quite a number of majors to delete them...

Okay lets leave out deprecated ones haha

HeartSquared (approved):
Seems to work (aside from the TW styles, but that's [a different issue](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?modal=detail&selectedIssue=KZN-1017))

PR #3313, opened 2023-02-22
feat: update field message caution icon

NatalieLoyzaga:
## Why

- Update to use new Caution Icon for Caution status

## What

- Add Caution to Field Message
- Show the Caution icon when status = caution
- Update the icon alignment to start

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/FieldMessage.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/FieldMessage.spec.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { cleanup, render } from "@testing-library/react"
+import { cleanup, render, screen } from "@testing-library/react"
 import { FieldMessage, FieldMessageProps } from "./index"
 
 afterEach(cleanup)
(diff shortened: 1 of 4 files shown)


mcwinter07 (approved):
Nice, I like that FieldMessage has its own Story now

draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/FieldMessage.spec.tsx
```diff
@@ -51,5 +51,13 @@ describe("<FieldMessage />", () => {
     const { container } = renderFieldMessage({ status: "error" })
 
     expect(container.querySelector(".warningIcon")).toBeTruthy()
+    expect(screen.getByTitle("Error message"))
+  })
+
+  it("should render a warning icon with an error status", () => {
```
christian314159, line 57: Just nitpicking, but I against `shoulds` ;) It "does it" or "does not do it".
```suggestion
  it("renders a warning icon with an error status", () => {
```

I think as a team we should (ha!) make a decision here.

NatalieLoyzaga:
@mcwinter07 Hey I made a quick update to remove "should" from the test messages - Could you please approve again Thank you!!

mcwinter07 (approved):

PR #3320, opened 2023-02-24
fix: Add aria-label for close button in popover

lavanya-everest:
## Why

- Popover cancel/dismiss button - Currently screen readers are reading it as just button instead of cancel button.
  - It will be great if we can add a name/aria-label to that button so that screen readers will identify correctly
More details [here](https://github.com/cultureamp/kaizen-discourse/issues/56)

## What

- Add `aria-label` to the cancel button so that screen readers will read that as `cancel button`


dougmacknz (approved):
Thanks Lavanya :)

PR #3323, opened 2023-02-27
fix: design-tokens-aio-import

gyfchong:
Testing out resolving an issue with design-tokens importing with esm by removing the need to import.


packages/components/src/KaizenProvider/ThemeProvider/themes/types.ts
```diff
@@ -0,0 +1,250 @@
...
+ * type Font = {
+ *   fontSize: `${number}rem`,
+ *
+ *  }
+ * type HexDigit = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | A | B | C | D | E | F ;
+ * type Color = `rgba(${number}, ${number}, ${number}, ${number}) | #${HexDigit}${HexDigit}${HexDigit}` // You get the point
+ *
+ * ```
+ * */
+type KaizenCSSProperties = ReactCSSProperties
```
HeartSquared, line 17: Just realised how weird this is. We just create a bunch of aliases lol
```
type KaizenCSSProperties = ReactCSSProperties = CSSProperties
```

packages/components/src/KaizenProvider/ThemeProvider/themes/types.ts
```diff
@@ -0,0 +1,250 @@
...
+
+type Hex = string
+// Once TypeScript is upgraded in the repo, you can use Lowercase<>
```
HeartSquared, line 20: Has TS been upgraded yet? 😂

packages/components/src/index.ts
```diff
@@ -51,8 +51,6 @@ export {
   makeCSSVariableTheme,
   mapLeafsOfObject,
 } from "@kaizen/design-tokens"
-export * from "@kaizen/design-tokens/src/themes"
-export * from "@kaizen/design-tokens/src/types"
```
HeartSquared, line 55: Was this referencing `themes/types.ts`? Or are we missing something?

HeartSquared (approved):

packages/components/src/index.ts
gyfchong: Oh you're right, that doesn't exist haha, it should've been `types.ts` yes. Another way to tackle the issue might've been to changes these to export from `dist/themes` and not `src`. But I chose the nuclear option lol

HeartSquared: Perhaps exporting these from `dist` may have prevented the problem? 😂

packages/components/src/KaizenProvider/ThemeProvider/themes/types.ts
gyfchong: Lol touching this broke everything, maybe later :P

packages/components/src/index.ts
gyfchong: We'll never know now 😈

PR #3325, opened 2023-02-28
feat: css-processing

gyfchong:
Process CSS Modules and Tailwind into a single consumable styles.css.


packages/components/package.json
```diff
@@ -71,15 +75,18 @@
     "@kaizen/select": "^6.13.6",
     "@kaizen/split-button": "^1.3.33",
     "@kaizen/tabs": "^1.5.30",
-    "@kaizen/tailwind": "^0.5.1",
     "@kaizen/typography": "^2.3.6",
     "classnames": "^2.3.2"
   },
   "devDependencies": {
-    "@deanc/esbuild-plugin-postcss": "^1.0.2",
+    "@kaizen/tailwind": "^0.5.1",
```
gyfchong, line 82: Moved this back to devDep cause the library doesn't require this to function, it only uses it at build time.

package.json
```diff
@@ -26,7 +26,8 @@
     "chromatic": "chromatic",
     "commit": "cz",
     "plop": "plop",
-    "pkg:aio": "yarn workspace @kaizen/components"
+    "pkg:aio": "yarn workspace @kaizen/components",
+    "build:deps": "yarn workspace @kaizen/design-tokens build && yarn workspace @kaizen/tailwind build"
```
HeartSquared, line 30: Make sure to fix the storybook/release scripts :)

Actually, since building the deps is only required for releasing the AIO, could you move this into the AIO's `prepublish`?

packages/components/package.json
```diff
@@ -71,15 +75,20 @@
...
     "autoprefixer": "^10.4.13",
+    "concat-cli": "^4.0.0",
+    "cssnano": "^5.1.15",
```
HeartSquared, line 85: What's this for? I only see it as a key in the postcss config (or is having it as a key enough to reference the package?)

packages/components/plugins/esbuild-css-module.ts
```diff
@@ -0,0 +1,63 @@
...
+      pluginData: {
+        pathDir: path.join(args.resolveDir, args.path),
+      },
+    }))
+    build.onLoad(
+      { filter: /#css-module$/, namespace: "css-module" },
+      async args => {
+        const { pluginData } = args as {
+          pluginData: { pathDir: string }
+        }
```
HeartSquared, line 22: Does this work? To avoid typecasting. Although of course, I don't know if `args` should be something else (ideally we'd have an actual type from somewhere).

```suggestion
      async ({ pluginData }: { pluginData: { pathDir: string } }) => {
```

packages/components/plugins/esbuild-css-module.ts
```diff
@@ -0,0 +1,63 @@
...
+          }"; export default ${JSON.stringify(cssModule)}`,
+        }
+      }
+    )
+    build.onResolve(
+      { filter: /\.module\.css$/, namespace: "css-module" },
+      args => ({
+        path: path.join(args.resolveDir, args.path, "#css-module-data"),
+        namespace: "css-module",
+        pluginData: args.pluginData as { css: string },
```
HeartSquared, line 48: Any way to avoid type casting?

packages/components/plugins/esbuild-css-module.ts
```diff
@@ -0,0 +1,63 @@
...
+      })
+    )
+    build.onLoad(
+      { filter: /#css-module-data$/, namespace: "css-module" },
+      args => {
+        const contents = (args.pluginData as { css: string }).css
```
HeartSquared, line 54: Any way to avoid type casting?

packages/components/tsup.config.ts
```diff
@@ -1,14 +1,19 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { defineConfig } from "tsup"
+import cssModules from "./plugins/esbuild-css-module"
 
 export default defineConfig({
   entry: ["src/**/*.ts"],
-  format: ["cjs", "esm", "iife"],
+  outDir: "./dist",
+  format: ["cjs", "esm"],
```
HeartSquared, line 8: Oh yay you got rid of the weird iife that no one understood :P

package.json
gyfchong: Ah good idea, can do! I only put this here to help me develop lol

packages/components/package.json
gyfchong: PostCSS is a strange and wonderful config file where merely referencing the key enables the plugin 😅

packages/components/plugins/esbuild-css-module.ts
gyfchong: Can try, it's a copy pasta 😛

packages/components/tsup.config.ts
gyfchong: Yeahhh it actually took the longest to compile and it didn't actually work with the SCSS files referencing `../variables` files haha

packages/components/plugins/esbuild-css-module.ts
gyfchong: Prepare to weep: https://github.com/evanw/esbuild/blob/main/lib/shared/types.ts#L325

PR #3326, opened 2023-02-28
docs: Kzn 1025 use utilities in utility class references

JakePitman:
## Why
Tailwind doesn't allow for utility classes to be generated from dynamically constructed classnames. This means that when we loop over the preset values to create utility classes for our Utility Class References docs, the css doesn't get generated.
As a workaround, we've been setting the equivalent css through inline styles.

This works visually, but leads to confusion when developers inspect our Utility Class Reference docs, and see inline styles instead of utility classes.

## What
[x] Leverages Tailwind's `safelist` config option to force compilation of documented utility styles in our storybook build
[x] Swaps out inline styles for the correct utility classes for each entry in Utility Class References

## Incidental bug fixes
[x] Border spacing had the wrong utility classes listed. It should be `border-spacing-SUFFIX` instead of `border-SUFFIX`
[x] Background color stories were using the color values from the preset's `borderColor` field, instead of `colors`, resulting in incorrect utility classes being documented


tailwind.config.js
```diff
@@ -15,4 +15,10 @@ module.exports = {
     preflight: false,
   },
   plugins: [],
+  safelist: [
+    {
+      pattern:
+        /(bg|border|rounded|shadow|h|w|m|p|font-family|font-weight|text|leading)-.*/,
```
JakePitman, line 21: Note that most of these are safe from creating unused utilities. For example, `p-SUFFIX` utils will be generated, but not `pl-SUFFIX`

image

One exception is the border utils though. With this regex, `border-y-SUFFIX`, `border-t-SUFFIX` etc will be generated even though they aren't used in our docs.
It would be good to exclude these somehow, but I haven't managed to find a regex expression that can:
- Exclude `border-[y|x|t|l|r|b]-SUFFIX`
- Match `border-SUFFIX`

It's not the end of the world adding in colors for 6 more prefixes, but if anyone is a regex wizard let me know

mcwinter07: So you should be able to do with with the [block list](https://tailwindcss.com/docs/content-configuration#discarding-classes), though that may mean we can never generate them so no sure how worth it that is

mcwinter07 (approved):
Looks hunky dory to me 👍 goodbye inline styles 🕺

PR #3327, opened 2023-02-28
feat: update components to use Caution Icon

NatalieLoyzaga:
## Why

UI kit has been updated to use a new Caution Icon

## What

- Update the components which have a cautionary variant to use the new Icon
- Update the use of `exclamationIcon` import to `errorIcon` to match UI kit terminology


christian314159 (approved):
👍 LGTM!

PR #3328, opened 2023-03-01
docs: Jp/tw docs update

JakePitman:
Makes some fixes to the Tailwind docs:

- Removes the no-longer-needed caveat about how we were emulating utilities with inline styles in Utility Class References. This is no longer the case
- Points users to Utility Class References when explaining utility classes  in the Overview docs
- Updates the TW Play url. I've created a new instance with some utilities that were missing, and some commented-out extensions in the config for consumer reference


docs/Systems/Tailwind/GettingStarted.stories.mdx
```diff
@@ -71,7 +71,7 @@ These directives inject classes that are needed for certain Tailwind features. F
 
 ### Tailwind Play
 
-Tailwind Play is an online code sandbox with Tailwind set up out of the box. The Kaizen Team has created an instance with the Kaizen Preset values [here](https://play.tailwindcss.com/dy7PzIcLed)
+Tailwind Play is an online code sandbox with Tailwind set up out of the box. The Kaizen Team has created an instance with the Kaizen Preset values [here](https://play.tailwindcss.com/AAjA6HqO5Y)
```
christian314159, line 74: Sorry to be nitpicking here, but the example in the Playground is using text color with opacity 
image
which is something we wanted to discourage (and remove from our current `Typography` components).
Thus I think we shouldn't have it in our official examples.

JakePitman: Thanks that's interesting... Let me see where that's coming from.

mcwinter07 (approved):
✅ Merge on my friend

PR #3329, opened 2023-03-01
feat: update aria labels for validation

NatalieLoyzaga:
## Why

- Update the DateRangeFields to have separate errorMessage ids which will correspond to dateStart or dateEnd for a11y

## What

- 2 separate errorMessageIds and aria-describeBy for dateStart and dateEnd


packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -71,10 +71,23 @@ export const DateRangeInputField = React.forwardRef<
...
 
-    const inputDescribedBy = validationMessage
-      ? `${errorMessageId} ${descriptionId}`
+    // Date Start aria labels
+    const dateStartErrorMessageId = validationMessage?.dateStart
+      ? `${id}--date-start-error-message`
+      : undefined
+
+    const dateStartInputDescribedBy = dateStartErrorMessageId
+      ? `${dateStartErrorMessageId} ${descriptionId}`
```
NatalieLoyzaga, line 81: We still need to include the error message within the aria-describeby as aria-errormessage is not yet compatible for "pertinent" messages  see [here](https://github.com/cultureamp/kaizen-discourse/issues/29#issuecomment-1449290697)

I still wanted to add them as the basic functionality is supported

draft-packages/form/KaizenDraft/Form/Primitives/FieldMessage/FieldMessage.tsx
```diff
@@ -63,9 +63,13 @@ export const FieldMessage = ({
         </span>
       )}
       <div className={styles.message}>
-        <Paragraph variant="small" color={textColor}>
-          {message}
-        </Paragraph>
+        {typeof message === "string" ? (
```
NatalieLoyzaga, line 66: Is this the best way to do this?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeInputField/DateRangeInputField.tsx
```diff
@@ -86,7 +99,9 @@ export const DateRangeInputField = React.forwardRef<
           <DateInput
             ref={inputRangeStartRef}
             id={`${id}--from`}
-            aria-describedby={inputDescribedBy}
+            aria-describedby={dateStartInputDescribedBy}
+            aria-errormessage={dateStartErrorMessageId}
+            aria-invalid={dateStartErrorMessageId ? true : false}
```
christian314159, line 104: Small tip:
This is a bit "magical" here, without much context I must assume reading this that the presence of a error message ID means that the field is invalid.
To make this more readable you could direct this into a variable and use that instead like so:
```
const isInvalid = dateStartErrorMessageId !== undefined;
...
aria-invalid={isInvalid}
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
```diff
@@ -22,7 +22,8 @@ describe("<DateRangeValidationMessage />", () => {
     it("combines multiple validation messages with the same status", () => {
       const { container } = render(
         <DateRangeValidationMessage
-          id="error-message-id"
+          dateStartId="date-start-error-message-id"
```
christian314159, line 25: I see you adding these props to the render function, but they don't seem to be used in the spec.
Are those just mandatory props?

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -36,11 +38,10 @@ export const DateRangeValidationMessage = ({
     if (status.dateEnd === status.dateStart) {
       return (
         <FieldMessage
-          id={id}
           message={
             <ul className={styles.fieldMessageList}>
-              <li>{validationMessage.dateEnd}</li>
-              <li>{validationMessage.dateStart}</li>
+              <li id={dateStartId}>{validationMessage.dateEnd}</li>
+              <li id={dateEndId}>{validationMessage.dateStart}</li>
```
HeartSquared, line 44: Could you switch these around please? My brain expects the Start one to come first 😂 

image

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.spec.tsx
NatalieLoyzaga: They are not mandatory but I have added a test to confirm the ids are being passed down

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -36,11 +38,10 @@ export const DateRangeValidationMessage = ({
     if (status.dateEnd === status.dateStart) {
       return (
         <FieldMessage
-          id={id}
           message={
             <ul className={styles.fieldMessageList}>
-              <li>{validationMessage.dateEnd}</li>
-              <li>{validationMessage.dateStart}</li>
+              <li id={dateEndId}>{validationMessage.dateStart}</li>
+              <li id={dateStartId}>{validationMessage.dateEnd}</li>
```
HeartSquared, line 44:
```suggestion
              <li id={dateStartId}>{validationMessage.dateStart}</li>
              <li id={dateEndId}>{validationMessage.dateEnd}</li>
```
😛

HeartSquared: Side note (cos I can't comment on it), it looks like you don't use `styles.fieldMessageIcon` anymore, so you can remove it from all the places in this file :)

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -69,7 +70,7 @@ export const DateRangeValidationMessage = ({
 
   return (
     <FieldMessage
-      id={id}
+      id={validationMessage.dateStart ? dateStartId : dateEndId}
```
HeartSquared, line 73: Instead of these ternaries, it might actually be clearer to have another early return instead. Otherwise you should apply the same ternary `message` and `status` (although I'd say it's less favourable than an if statement since there's 3).

```
if (validationMessage.dateStart) {
  return <FieldMessage id={dateStartId} ...etc />
}

return <FieldMessage id={dateEndId} ...etc />
```

packages/date-picker/src/FilterDateRangePicker/components/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -67,15 +67,20 @@ export const DateRangeValidationMessage = ({
...
   }
 
+  if (validationMessage.dateStart)
+    <FieldMessage
+      id={dateStartId}
+      message={validationMessage.dateStart}
+      status={status?.dateStart}
+      reversed={isReversed}
+    />
+
```
HeartSquared, line 77:
```suggestion
  if (validationMessage.dateStart) {
    return <FieldMessage
      id={dateStartId}
      message={validationMessage.dateStart}
      status={status?.dateStart}
      reversed={isReversed}
    />
}
```

HeartSquared (approved):

PR #3330, opened 2023-03-01
docs: update guidance on the important selector

mcwinter07:
## Why
Updating documentation to provide guidance following exploration into this [discussion](https://github.com/cultureamp/kaizen-discourse/discussions/58) 

## What
- add new section of guidance to the getting started TW doc


JakePitman:
Nice 👍 

I think it makes sense as `Customising the important selector` section, but what do you think about moving it over to the `Preset configuration` section?
- It's technically a configuration option,
- That page is less bloated than getting started

mcwinter07:
> I think it makes sense as a `Customising the important selector` subsection instead of its own page, but what do you think about moving it over to the `Preset configuration` section?

That's a good point, It'd props be a better place for it since it's more tailored to configuration

JakePitman:
I guess it depends on how opinionated we are about what their selector should be...

I wonder if it'd be best to put whatever our recommended selector is in the `2. Implement the Preset` section, but leave a stronger note about configuration options.

I love this notification you added in:
image
Maybe we could use it to replace this?
image

But that's just my opinion - your call ~

gyfchong:
I agree, move this to the configuration page, as I think we'd want to de-emphasise the customising of the important selector.

I think you should replace the first part with some kind of scenario/reason why you would want to customise the selector and leave out "how it works", cause I don't think the consumers necessarily need to know this, just the affects of the change.

mcwinter07:
> I think you should replace the first part with some kind of scenario/reason why you would want to customise the selector and leave out "how it works", cause I don't think the consumers necessarily need to know this, just the affects of the change.

Fair point 👍 will make a couple o' tweaks and drive the why

JakePitman (dismissed):
Awesome work

JakePitman (approved):

PR #3338, opened 2023-03-02
fix: a11y and component-library package-locks + build

gyfchong:
I was so cowboy in my last release that I forgot a few edits to make it even work lol, need some kinda package.json linter


NatalieLoyzaga (approved):
yeehaw

PR #3341, opened 2023-03-02
feat(filter): Add Filter component into the AIO

NatalieLoyzaga:
## Why

- Add new Filter component

## What

- Extract Solution 2 from previous branch see [here](https://github.com/cultureamp/kaizen-design-system/pull/3340/files)
- Move Cass' Filter abstraction into the AIO package
- Add additional stories and tests


packages/components/package.json
```diff
@@ -73,14 +73,18 @@
...
+    "react-focus-on": "^3.7.0"
   },
   "devDependencies": {
     "@deanc/esbuild-plugin-postcss": "^1.0.2",
     "autoprefixer": "^10.4.13",
     "tailwindcss": "^3.2.7",
     "tiny-glob": "^0.2.9",
-    "tsup": "^6.6.3"
+    "tsup": "^6.6.3",
+    "esbuild": "^0.17.8"
```
HeartSquared, line 87: Pretty sure this is just part of Geoff's stuff and we don't need it ;)

packages/components/src/Filter/components/FilterPopover/FilterPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+import styles from "./FilterPopover.module.scss"
+
+export interface FilterPopoverProps
+  extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
+  children: React.ReactNode
+  referenceElement: HTMLElement | null
+  popperOptions?: Partial<Options>
+}
+
+export const FilterPopover = ({
```
gyfchong, line 15: Can this be a generic `Popover` to be (later) used in other places like `Menu`?

HeartSquared: We actually have a Popover component which is a little more like a tooltip.
This is certainly rather generic (if you consider the positioning has the offset and is only used in a context where it's `role="dialog"`), but we can't use the Popover name currently.
Since this is the AIO though, and this component is used by the Filter but not exported, we can rename and move it around later when we know what we want.

packages/components/src/Filter/docs/Filter.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  component: Filter,
+  parameters: {
+    docs: {
+      description: {
+        component:
+          '`import { Filter, FilterContents, FilterButton, FilterButtonRemovable } from "@kaizen/components"`<br />Use a Filter Button for the renderTrigger.<br />Use FilterContents to wrap the contents within a Filter.',
+      },
+    },
+  },
+} as ComponentMeta<typeof Filter>
```
HeartSquared, line 22: Note: Tried to use `satisfies` instead of `as` and the code snippet died 😂 Something to check for v7

HeartSquared (approved):

PR #3344, opened 2023-03-03
feat(filter-drp): add inbuilt validation

HeartSquared:
## What

Add inbuilt validation for FilterDateRangePicker.
Uses existing `validateDate` util.
Adds additional validation for checking whether End Date is before Start Date.

## Why

[KZN-397](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?modal=detail&selectedIssue=KZN-397)

[KZN-397]: https://cultureamp.atlassian.net/browse/KZN-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
+++ b/packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
@@ -30,6 +30,7 @@ import {
 } from "./components/Trigger"
 import { DateRangeValidationStatus } from "./types"
 import { isCompleteDateRange } from "./utils/isCompleteDateRange"
+import { validateEndDateBeforeStartDate } from "./utils/validateEndDateBeforeStartDate"
 import styles from "./FilterDateRangePicker.module.scss"
 
 type InputRangeStartProps = DateRangeInputFieldProps["inputRangeStartProps"]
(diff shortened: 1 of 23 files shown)


packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -142,17 +158,101 @@ export const FilterDateRangePicker = ({
...
+  >()
+  const [
+    inbuiltStartDateValidationMessage,
+    setInbuiltStartDateValidationMessage,
+  ] = useState<string | undefined>()
+  const [inbuiltEndDateStatus, setInbuiltEndDateStatus] = useState<
+    ValidationResponse["status"] | undefined
+  >()
+  const [inbuiltEndDateValidationMessage, setInbuiltEndDateValidationMessage] =
+    useState<string | undefined>()
```
gyfchong, line 172: ...I think we need to kill this message/status pattern. If the crazy types Nat had to do to make hers work wasn't a sign already, this is just so verbose lol

gyfchong: And I think consumers would agree, most of their app errors should be in the shape of
```
{ status, message }
```

HeartSquared: Yes I completely agree. Definitely keen to get rid of it if we could.
I could probably change these state objects though anyway, at least for inbuilt. It did come to me over the weekend, so I was planning to refactor to do it :)

NatalieLoyzaga:
@HeartSquared - The tests are failing did you still want me to review :)

packages/date-picker/src/FilterDateRangePicker/utils/validateEndDateBeforeStartDate.ts
```diff
@@ -0,0 +1,51 @@
...
+    inputValue: endDateInputValue,
+    status: undefined,
+    validationMessage: undefined,
+    isInvalid: false,
+    isDisabled: false,
+    isEmpty: false,
+    isValidDate: true,
+  }
+
+  if (endDate < startDate) {
```
NatalieLoyzaga, line 34: No idea this was possible - very cool

packages/date-picker/src/FilterDateRangePicker/utils/validateEndDateBeforeStartDate.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      isInvalid: false,
+      isDisabled: false,
+      isEmpty: false,
+      isValidDate: true,
+    })
+
+    expect(newDate).toBe(endDate)
+  })
+
+  test("end date equals start date", () => {
```
NatalieLoyzaga, line 30: Confirming we didn't want to have an opinion on this and leave it up to the consumer if they wanted to make this an error or warning

HeartSquared: Yeah I think so. It's a valid selection for react-day-picker, so I think it makes sense to replicate it. Better than over-restricting :)

NatalieLoyzaga: Yep makes sense!

NatalieLoyzaga (approved):
LGTM! Awesome work

NatalieLoyzaga (approved):
Nice!

PR #3345, opened 2023-03-05
feat: add caption to table

kaydbwithers:
## Why
This PR adds an optional caption for screen readers. 

This caption allows screen readers to read a description of what the table contains.

I have also changed `<div role="table">` to `<table>` as the caption element does not work correctly without a table element.

## What
It does not effect the current behavior and there is no change to the UI, it is screen reader only.

### DIFF
--- a/draft-packages/table/KaizenDraft/Table/Table.tsx
+++ b/draft-packages/table/KaizenDraft/Table/Table.tsx
@@ -11,6 +11,7 @@ import { Heading } from "@kaizen/typography"
 import styles from "./Table.module.scss"
 
 type TableContainerProps = {
+  caption?: string
   children?: React.ReactNode
   variant?: "compact" | "default" | "data"
 }
@@ -20,19 +21,20 @@ type TableContainerProps = {
  */
 export const TableContainer = ({
   variant = "compact",
+  caption,
   children,
   ...otherProps
 }: TableContainerProps): JSX.Element => (
-  <div
-    role="table"
+  <table
     className={classNames(styles.container, {
       [styles.defaultSpacing]: variant === "default",
       [styles.dataVariant]: variant === "data",
     })}
     {...otherProps}
   >
+    {caption && <caption className="sr-only">{caption}</caption>}
     {children}
-  </div>
+  </table>
 )
 
 /**
@@ -350,6 +352,7 @@ export const TableCard = ({
       href={href}
       className={className}
       onClick={onClick as AnchorClickEvent}
+      role="button"
       {...otherProps}
     >
       {children}
--- a/draft-packages/table/docs/Table.stories.tsx
+++ b/draft-packages/table/docs/Table.stories.tsx
@@ -184,7 +184,7 @@ const Multiline = ({ isReversed }: { isReversed: boolean }): JSX.Element => (
 
 export const DataVariant: Story = () => (
   <Container>
-    <TableContainer variant="data">
+    <TableContainer caption="A data variant table example." variant="data">
       <TableHeader>
         <ExampleTableHeaderRow reversed={false} />
       </TableHeader>
@@ -204,7 +204,7 @@ DataVariant.parameters = { chromatic: { disable: false } }
 
 export const IconVariant: Story = () => (
   <Container>
-    <TableContainer>
+    <TableContainer caption="An icon variant table example.">
       <TableHeader>
         <TableHeaderRow>
           <TableHeaderRowCell
@@ -266,7 +266,7 @@ IconVariant.parameters = { chromatic: { disable: false } }
 
 export const LinkVariant: Story = () => (
   <Container>
-    <TableContainer>
+    <TableContainer caption="A link variant table example.">
       <TableHeader>
         <TableHeaderRow>
           <TableHeaderRowCell labelText="Header A" width={1 / 3} />
@@ -291,6 +291,19 @@ export const LinkVariant: Story = () => (
         <TableRow>
           <TableRowCell width={1 / 3}>
             <Paragraph variant="body">This row has a href</Paragraph>
+          </TableRowCell>
+          <TableRowCell width={1 / 3}>
+            <Paragraph variant="body">24</Paragraph>
+          </TableRowCell>
+          <TableRowCell width={1 / 3}>
+            <Paragraph variant="body">48</Paragraph>
+          </TableRowCell>
+        </TableRow>
+      </TableCard>
+      <TableCard>
+        <TableRow>
+          <TableRowCell width={1 / 3}>
+            <Paragraph variant="body">This is a default row</Paragraph>
           </TableRowCell>
           <TableRowCell width={1 / 3}>
             <Paragraph variant="body">24</Paragraph>


draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -350,6 +352,7 @@ export const TableCard = ({
       href={href}
       className={className}
       onClick={onClick as AnchorClickEvent}
+      role="button"
```
dougmacknz, line 355: What's the thinking on this change @kaydbwithers?

gyfchong:
Just curious, wouldn't this caption also be useful for sighted people?

draft-packages/table/KaizenDraft/Table/Table.tsx
kaydbwithers: Thanks @dougmacknz, that slipped in as a mistake, it's been removed now.

kaydbwithers:
> Just curious, wouldn't this caption also be useful for sighted people?

I haven't encountered any designs that show a description of the table. They might have a heading, but these are usually different from a description of the table.

kaydbwithers:
@gyfchong if we were to make the caption visible, then would this "Heading" look okay? (Ignore the export in the top right) 

image

gyfchong:
> @kaydbwithers if we were to make the caption visible, then would this "Heading" look okay? (Ignore the export in the top right)

Isn't this caption supposed to serve as a description? Using it as a heading will certainly clash with tables that already have headings.

kaydbwithers:
Looking at some of the government tables, it looks like you are correct. The description should just be the heading.

UK.GOV says..
> Use the <caption> element to describe a table in the same way you would use a heading. A caption helps users find, navigate and understand tables.

Here is that example.
https://design-system.service.gov.uk/components/table/

gyfchong:
Right, well then it seems okay to include it as a caption heading then. Let's just get some design eyes on the screenshot.

@dougmacknz is this something we should promote wide usage of? ie. would it be a bit odd to have some tables with captions and without.

gyfchong:
@kaydbwithers push up the new heading style, we can use Chromatic to get a design review

kaydbwithers:
Thanks @gyfchong, I've added a visible caption in the latest commit.

dougmacknz:
> @dougmacknz is this something we should promote wide usage of? ie. would it be a bit odd to have some tables with captions and without.

My understanding is that every table should have one.

gyfchong:
> My understanding is that every table should have one.

We should make this mandatory then! Will probably have to be tabled at the UI Working Group then

draft-packages/table/KaizenDraft/Table/Table.tsx
```diff
@@ -20,23 +21,32 @@ type TableContainerProps = {
...
+  <table
     className={classNames(styles.container, {
       [styles.defaultSpacing]: variant === "default",
       [styles.dataVariant]: variant === "data",
     })}
     {...otherProps}
   >
+    {caption && (
+      <Heading
+        classNameOverride="pb-12 text-left"
```
gyfchong, line 37: Unfortunately you cannot use Tailwind here, as there are code bases that don't have Tailwind configured, and this lives outside the `components` package that enables Tailwind distribution.

Regardless, this should be a margin.

gyfchong (changes requested):

draft-packages/table/KaizenDraft/Table/Table.tsx
kaydbwithers: Yep sure, I'll do that. I did try `mb-12` but due to a specificity issue it gets overridden by the Heading's `margin: 0;`. But not using Tailwind should resolve that.

kaydbwithers: @gyfchong I've added the changes you requested in the latest commit.

gyfchong:
Thanks @kaydbwithers, let's see what happens with the UI Working Group. May have to offer more heading sizes.

kaydbwithers:
Hey @gyfchong, are there any updates from the UI Working Group?

gyfchong:
@kaydbwithers This is likely going to be rolled into a bigger tables initiative run by the design team as multiple teams are looking into this.

kaydbwithers:
@gyfchong since caption is an optional prop, can we please merge this? I really need `<Heading tag="caption" />` for my tables.

gyfchong:
> @gyfchong since caption is an optional prop, can we please merge this? I really need `<Heading tag="caption" />` for my tables.

@kaydbwithers can you help us understand why? From our perspective design hasn't needed titles for tables, why is it a requirement for your use case?

kaydbwithers:
@gyfchong this is my use case. "City" should be a heading with `tag="caption"` I could create a seperate PR just for the heading. But since caption is optional on the table, I feel like it could be merged.

image

gyfchong:
@kaydbwithers I don't think you're going to get the affect you want with this PR: the caption will sit above the table and won't line up with your export button.

It's probably better to create your own header and use `aria-describedby` instead: https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/

Because I think the most optimal solution here would be to provide a render prop for a custom header to a) support multiple heading sizes and b) allow for any layout a consumer needs for their header. But then a consumer may as well just place their header above the table since that's all we'll be doing internally anyway 😅 

```
<div>
    <CustomHeader />
    <Table />
</div>
```

I agree with the concept of a caption though, and it will be considered an option for the future tables, but for your use case I don't think it provides that much value.

dougmacknz:
> It's probably better to create your own header and use aria-describedby instead

`aria-labelledby` is what you want for the caption alternative.

`aria-describedby` can be used to provide a table summary

e.g. for the screenshot below:
`Availability of holiday accommodation` is the caption/aria-labelledby
`Column one has...` is the summary/aria-describedby

image

https://www.w3.org/WAI/tutorials/tables/caption-summary/

I agree that we should leave it up to the consuming side to create and link up the header, at least for now.

I'd recommend we make `aria-labelledby` a required prop in a breaking change, so that we can make sure people are adding an accessible title for every table.

sentience:
> I'd recommend we make aria-labelledby a required prop in a breaking change, so that we can make sure people are adding an accessible title for every table.

Does every table need an accessible title? I thought this was purely optional in HTML, and I also suspect we have plenty of tables without a dedicated title in our UI designs.

For that matter, does _this_ use case need an accessible title? Can't the heading just be a standard `h2` (or whatever level) above the table, providing a navigation landmark and implicitly describing the content that follows it? In this use case, the title does not just describe the table, it also describes the "Export" button. So marking up the title as a table caption seems semantically incorrect to me.

kaydbwithers:
@gyfchong yes we can do `aria-labelledby` that should suffice.

All of these arias and roles are a bit smelly though. Native elements provide the functionality I'm looking for with less code. Is there a reason we always use `role=""` and `aria-` in our code? Maybe for the next table we could simplify it by using native elements?

dougmacknz:
> Does every table need an accessible title?

@sentience this is what we heard about our tables in the audit. I will double check all of this with the auditors when I do the debrief session, but I _think_ we might be failing `4.1.2` here because of no name for the table. It might be because the table is interactive, with sortable headings. The table on it's own never gets focus, but maybe 4.1.2 still applies if a child element can receive focus.

image
image
image

So it may be that in terms of hard requirements, just the tables that have sorting turned on need the headings. But it certainly doesn't hurt to add this every time, even if it's a visually hidden aria label.

dougmacknz:
I found a table from the audit results that isn't interactive. No failure of 4.1.2, but still a big emphasis on making sure the table has a caption. Maybe it's 1.3.1.

image
image
image

gyfchong:
> @gyfchong I was able to line up the Export button with some negative margin.

@kaydbwithers I'd advise against using negative margins to achieve alignment, as generally it's hard to manage when dealing with a scaling UI (zoom/responsiveness) combined with extensibility eg. adding more elements to the same area.

> All of these arias and roles are a bit smelly IMO. Native elements provide the functionality I'm looking for with less code. Is there a reason we always use `role=""` and `aria-` in our Kaizen code? Maybe for the next table we could simplify it by using native elements?

I imagine that the table was developed using aria and roles to facilitate the many non-standard table features it provides, to simplify the feature set so we can potentially utilise native elements, would be the goal of the Table initiative.

> A good example is that we would now have to add a uniqueId for the `aria-labelledby`. If we were to use `<caption>` inside a `<table>` this would not be needed.

For your case, I don't think there's a way around creating something custom, as the optimal way to build your header would be using a heading + button in a flex container. But I agree that we should provide some API to cater to the simpler headings and the more complex and maintain the accessibility internally so consumers don't have to worry about it for the most part, perhaps a `heading` prop that takes either a string or a react component:

```
// Adds the <caption>
<Table heading="City"></Table>

// Replaces <caption> internally and attaches aria-labelledby
const CustomHeading = () => ( ... )
<Table heading={CustomHeading}></Table>
```

kaydbwithers:
Thanks @gyfchong, in the latest commit I've added heading instead of caption and I've given it a uniqueId.

gyfchong:
Hey @kaydbwithers the structure you've provided doesn't quite solve the overall problem as it doesn't provide the flexibility for your use case (heading + button), and the default heading needs to be a caption if it sits inside the `table` itself.

So the two internal structures would be:

1. If the consumer provides a string as a heading
```html
<table>
    <caption>{heading}</caption>
</table>
```

2. if the consumer provides a component as a heading (your use case):
```html
<UserCreatedHeading id="heading_1234"  />
<table aria-labelledby="heading_1234">
    ...
</table>
```

However the issue with this approach is that there's no natural way to control what the `caption` appearance would be, as we can't assume that every design will use a `heading-2` style 🤔

One approach would be to put it back on the consumer and say "here's the default heading style for our tables, if you want to change it then pass a `<Heading>` component in with the style you want", but I imagine at the end of the day we wouldn't have saved anyone's time given the variety of designs that could be made, 9/10 times the consumer would have to pass a component anyway -- which we try avoid in our API.

Having said all that, I don't think it's worth progressing with this PR anymore as there's no clear path that greatly improves productivity or any consistent design standard to support. At most we could provide an example in the docs of how to do headings using `aria-labelledBy` so that it's a documented pattern people can reference. 

@dougmacknz I'm not so sure requiring it works how we think it will, given history, it seems leaving a blank string when unsure why something is required would be the most likely outcome. I think we should tackle this from a design angle, if designers know to always add headings to their tables (via the Table discovery Jason M is doing) and we advertise the documentation along with the design push, we might get better results.

kaydbwithers:
Thanks @gyfchong, in reports-components we have multiple tables with headings not always with a button on the right. Here is an example.

image

I think it would be good to have a consistent heading style. As you can see here the "Has mentor" and "Manager" headings are slightly different but I think this is a consistency mistake.
image

We can leave this if you like and I can add the headings manually. Adding aria-labelledby to the Table docs is a good idea.

gyfchong:
Thanks @kaydbwithers, we'll take all this into account for the next iteration of Tables 👍

PR #3346, opened 2023-03-06
Rollup test

ckychris:
Adding rollup and compiling the components package.

**Reference**
https://github.com/cultureamp/rollup-sass-ui-library/tree/extract
https://cultureamp.atlassian.net/wiki/spaces/TV/pages/2911633606/UI+package+publishing


packages/components/src/Badge/KaizenDraft/Badge/Badge.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { HTMLAttributes, useLayoutEffect, useState } from "react"
+import classNames from "classnames"
+import { OverrideClassName } from "@t/overrideClassName"
```
gyfchong, line 3: Hmm seems the linter doesn't like this...

gyfchong: Solved this, because the global `tsc` command uses a different config (without the paths in config) this doesn't work.

gyfchong:
Main concepts have been added to `main`, thanks @ckychris

PR #3347, opened 2023-03-07
docs: Kzn 1058 how can we style storybook create a poc

JakePitman:
## Why
For the aesthetic.

## What
- Themes our storybook app with CA design tokens

image
image

## Out of scope
- Storybook icons
- Hover states on left-navbar

These are not modifiable with the `create` or `addons` apis, and will require hacks.


gyfchong:
Bit of a logo problem
image

storybook/main.ts
```diff
@@ -47,4 +47,5 @@ module.exports = {
     "storybook-addon-designs",
   ],
   presets: [path.resolve("./storybook/header-preset/preset")],
+  staticDirs: ["./assets"],
```
mcwinter07, line 50: If we change this to:
 ```
  staticDirs: [{ from: "./assets", to: "/static/media" }],
```
then it will move those files to the static media folder at build 🕺

Comment (mcwinter07) [reply storybook/main.ts:50]: still need to figure out the janky reference of [object Module] at build but figure we could do something like this [example](https://github.com/storybookjs/storybook/issues/9443#issuecomment-574713070) but inversed 🤷 
```
  brandImage: (process.env.NODE_ENV === 'production') ? logoUrl : '/logo.svg',`
// to
    brandImage: (process.env.NODE_ENV === 'production') ?  'static/media/logo.svg' : importedInSVG`

storybook/manager.tsx
```diff
@@ -0,0 +1,37 @@
...
+  theme: KaizenTheme,
+  analyticsGTM: {
+    gtmId: "GTM-KS4VWLT",
+  },
+  sidebar: {
+    renderLabel: item =>
+      item.type === "root" ? (
+        <span
+          style={{
+            color: colors.purple["800"],
```
mcwinter07, line 21: Would be awesome to loop back an maybe tail-windify this as some point but rather than sink any more time into it lets forge ahead with this :shipit:

storybook/manager.css
```diff
@@ -0,0 +1,4 @@
+.sidebar-item {
```
mcwinter07, line 1: We'll probably want to remove this file now we're no longer using it :)

package.json
```diff
@@ -8,7 +8,7 @@
     "node": ">=v16.13.1"
   },
   "scripts": {
-    "storybook": "start-storybook -p 6006 -c storybook",
+    "storybook": "IS_DEV=true start-storybook -p 6006 -c storybook",
```
mcwinter07, line 11: For context for anyone else looking at this, this is to allow us to use this environment variable for local dev to see the brand image asset

mcwinter07 (commented):
We'll probably want to remove the png version of the logo and the manage.css since we don't use it :) After that, it's a ✅ from me.

PR #3348, opened 2023-03-08
chore(docs): storybook addons

gyfchong:
- Table of contents
- Hiding stickersheets


packages/button/docs/Button.stickersheet.stories.tsx
```diff
@@ -0,0 +1,247 @@
+import React from "react"
```
NatalieLoyzaga, line 1: Do these Button story updates need to be here?

gyfchong: I think it's better that they go into their own category rather than being hidden by config, we still want to be able to access them.

PR #3349, opened 2023-03-08
chore(docs): storybook ia

gyfchong:
## Why
Consumers/Kaizen Team can't find things

## What
Currently docs are arranged such that they are (more or less) grouped by their **package name**. eg. you have a `date-picker` package that exports 3 different components, `Date Picker` would be the top level link.

image

The new arrangement is (mostly) about the **components** being at top level links and any components that are related to the main component are nested, irrespective of their package.

For example, `FilterDateRangePicker` used to be nested under `date-picker`, but is now accessible from the top level:

image

or in some cases components that share the same name are grouped together, this way if someone were thinking "I need a Loading component" and simply scanned to `Loading` they'd easily discover there's multiple types.

image

Whilst it looks very verbose, I think it improves on the previous arrangement as there's less requirements for a visitor to understand our package structure to be able to find the component they're looking for.


HeartSquared:
One main thing is when components share subcomponents. FilterDRP uses the same Calendar as DatePicker, so where does that go?

gyfchong:
> One main thing is when components share subcomponents. FilterDRP uses the same Calendar as DatePicker, so where does that go?

@HeartSquared in this case I'd opt for the component that uses it in the "most pure" way to be the one that has the subcompnents, where DatePicker is Field + Calendar versus the FilterDRP being a combination of multiple Calendars and Fields. This also helps us simplify our documentation because we have a page that documents the shared component in the simplest way, FilterDRP could have a section saying something like "X Prop accepts all options from Calendar" and cross link it so we don't repeat ourselves.

HeartSquared:
(Why is there no thread for this?)

> in this case I'd opt for the component that uses it in the "most pure" way to be the one that has the subcompnents, where DatePicker is Field + Calendar versus the FilterDRP being a combination of multiple Calendars and Fields. This also helps us simplify our documentation because we have a page that documents the shared component in the simplest way, FilterDRP could have a section saying something like "X Prop accepts all options from Calendar" and cross link it so we don't repeat ourselves.

Not opposed to it, but I guess this would be where categories would benefit :P Pick your poison I suppose.

gyfchong:
> Not opposed to it, but I guess this would be where categories would benefit :P Pick your poison I suppose.

Actually, Calendar is in the wrong place, it should be top level. I'd want to reserve sub components for those that inherently depend on the parent to work, eg. `MenuItem`

HeartSquared (approved):

PR #3352, opened 2023-03-09
fix: Add aria-current=page to title block nav active item

dougmacknz:
https://cultureamp.atlassian.net/browse/KZN-1069

image

image


gyfchong:
Curious, how does this work in conjunction with the top level nav? (assuming it has a `current-page=active` as well)

dougmacknz:
What this does is just announce 'current page' when you navigate to the item. So just like there's no issues with having multiple `nav`s on your page, there's no issues with having multiple of these on a single page.

PR #3353, opened 2023-03-09
feat: KDS-1055 | Dropzone component

jared-luke:
## Why

**This is a work in progress.**
Introducing the new `Dropzone` component.
It is leveraging a library called `react-dropzone` (link [here](https://react-dropzone.js.org/)).
The reason behind this component is to create a consistent experience for our customers and engineers when using a file uploader. 

**Note: the functionality of uploading a file is yet to be worked on.** Below is a checklist of things that need to be done before this PR is ready for review.

Currently, there have been four storybook stories to demonstrate the different scenarios of the `Dropzone` component. These include and can be seen in the video below:

- Default component
- Adjusting the `maxFileSize` prop. Defaults to `40kb`
- Passing state back up to the parent component. This is useful for forms, buttons, and [workflows](https://cultureamp.design/guidelines/workflow/) 
- Custom tailwind styles being applied to the parent container

Design file: [Found here](https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%A7%AA-Kaizen-Labs?node-id=27341-158623&t=5eVPmMUZkg1UByNt-0)

## What

https://user-images.githubusercontent.com/43884313/223904242-97e4d6c4-c104-4397-aa14-0430ddf7b54a.mp4

## Checklist
A Basic list of to-do's that will be worked on over time:
- [ ] Write a complete suite of tests
- [ ] Write a complete set of storybook stories
- [ ] Internationalisation 
- [ ] Accessibility review
- [ ] Finalise the documentation for release
- [ ] Responsiveness review with designers

### DIFF
--- a/packages/components/src/Dropzone/Dropzone.tsx
+++ b/packages/components/src/Dropzone/Dropzone.tsx
@@ -2,7 +2,7 @@
 import React, { Dispatch, SetStateAction, useEffect, useState } from "react"
 import classNames from "classnames"
 import { useDropzone, FileRejection, DropzoneOptions } from "react-dropzone"
-import { ACCEPTED_FILE_TYPES, DEFAULT_MAX_FILE_SIZE_40KB } from "./constants"
+import { ACCEPTED_FILE_TYPES, DEFAULT_MAX_FILE_SIZE_40KB, DROPZONE_AUTOMATION_ID } from "./constants"
 import { DefaultMessage } from "./sub-components/DefaultMessage/DefaultMessage"
 import { DropzoneErrors } from "./sub-components/DropzoneError/DropzoneError"
 import { DropzoneIcon } from "./sub-components/DropzoneIcon/DropzoneIcon"
@@ -86,6 +86,7 @@ export const Dropzone = ({ setEnableFileUpload, maxFileSize, className }: Dropzo
   return (
     <div className={classNames(className)}>
       <div
+        data-automation-id={DROPZONE_AUTOMATION_ID}
         className={classNames(
           {
             "rounded-default": true,
--- a/packages/components/src/Dropzone/Dropzone.tsx
+++ b/packages/components/src/Dropzone/Dropzone.tsx
@@ -1,0 +1,114 @@
+/* eslint-disable no-console */
+import React, { Dispatch, SetStateAction, useEffect, useState } from "react"
+import classNames from "classnames"
+import { useDropzone, FileRejection, DropzoneOptions } from "react-dropzone"
+import { ACCEPTED_FILE_TYPES, DEFAULT_MAX_FILE_SIZE_40KB } from "./constants"
+import { DefaultMessage } from "./sub-components/DefaultMessage/DefaultMessage"
+import { DropzoneErrors } from "./sub-components/DropzoneError/DropzoneError"
+import { DropzoneIcon } from "./sub-components/DropzoneIcon/DropzoneIcon"
+import { OnDragMessage } from "./sub-components/OnDragMessage/OnDragMessage"
+import { SelectedFileMessage } from "./sub-components/SelectedFileMessage/SelectedFileMessage"
+
+type DropzoneProps = {
+  /**
+   * The setter function that allows a piece of state to be sent back up to the parent component.
+   * This will be useful for disabling buttons within a [workflow](https://cultureamp.design/guidelines/workflow/) or a submit button within a form.
+   */
+  setEnableFileUpload: Dispatch<SetStateAction<boolean>>;
+
+  /** this unit is in bytes: defaults to `40000` = `40kb` */
+  maxFileSize?: number;
+
+  /** Apply tailwind classnames which will be applied to the container */
+  className?: string;
+}
+
+export const Dropzone = ({ setEnableFileUpload, maxFileSize, className }: DropzoneProps): JSX.Element => {
+  const [selectedFileName, setSelectedFileName] = useState<string>("")
+  const [dropzoneErrors, setDropzoneErrors] = useState<FileRejection[]>()
+
+  const defaultConfigurationOptions: DropzoneOptions = {
+    noClick: true,
+    noKeyboard: true,
+    multiple: false,
+    maxFiles: 1,
+    accept: {
+      "application/type": ACCEPTED_FILE_TYPES,
+    },
+    onDragEnter: () => {
+      setSelectedFileName("")
+    },
+    onDropAccepted(files, event) {
+      setSelectedFileName(files[0].name || "")
+      setDropzoneErrors(undefined)
+    },
+    onError(err) {
+      console.error("onError", err);
+    },
+    onDrop(acceptedFiles, fileRejections, event) {
+      console.log("onDrop", acceptedFiles, fileRejections, event);
+    },
+    onDropRejected(fileRejections, event) {
+      console.warn("onDropRejected", fileRejections, event);
+      setDropzoneErrors(fileRejections)
+    },
+  }
+
+  /**
+   * Merge default config options with user config options
+   * in the future we can add more config options
+   */
+  const maxSize = maxFileSize ? maxFileSize : DEFAULT_MAX_FILE_SIZE_40KB
+  /** prevent an empty file from being uploaded */
+  const minSize = 1000
+  const options: DropzoneOptions = {
+    ...defaultConfigurationOptions,
+    maxSize,
+    minSize
+  }
+
+  /** initialise the dropzone hook */
+  const { open, isDragActive, acceptedFiles, getRootProps, getInputProps, } = useDropzone(options)
+
+  /**
+   * Set the parent `enableFileUpload` state to allow the parent component to update its UI
+   * Note: this logic will change when we allow multiple files in the future
+   */
+  useEffect(() => {
+    const hasAcceptedFile = acceptedFiles.length >= 1
+    if (hasAcceptedFile) {
+      setEnableFileUpload(true)
+    } else {
+      setEnableFileUpload(false)
+    }
+  }, [acceptedFiles])
+
+  return (
+    <div className={classNames(className)}>
+      <div
+        className={classNames(
+          {
+            "rounded-default": true,
+            "border-w-default border-dashed border-gray-400": true,
+            "transition-colors": true,
+            "text-body text-center text-purple-800 transition-colors": true,
+            "w-100 mb-4": true,
+            "!border-solid !border-blue-500 !bg-blue-100": isDragActive,
+          }
+        )}
+        {...getRootProps()}
+      >
+        <div className="flex items-center justify-center h-[120px] px-24 py-32">
+          <input type="file" {...getInputProps()} />
+          <div>
+            <DropzoneIcon fileName={selectedFileName} isDragActive={isDragActive} />
+            {isDragActive && <OnDragMessage />}
+            {selectedFileName && <SelectedFileMessage fileName={selectedFileName} handleOpenDialog={open} />}
+            {!isDragActive && !selectedFileName && <DefaultMessage handleOpenDialog={open} />}
+          </div>
+        </div>
+      </div>
+      {dropzoneErrors && <DropzoneErrors dropzoneErrors={dropzoneErrors} acceptedFileTypes={ACCEPTED_FILE_TYPES} maxFileSize={maxSize} minFileSize={minSize} />}
+    </div>
+  )
+}
--- a/packages/components/src/Dropzone/sub-components/DefaultMessage/DefaultMessage.tsx
+++ b/packages/components/src/Dropzone/sub-components/DefaultMessage/DefaultMessage.tsx
@@ -1,0 +1,21 @@
+import React from "react"
+import { Paragraph } from "@kaizen/typography"
+import { Hyperlink } from "../Hyperlink/Hyperlink"
+
+interface DefaultMessageProps {
+  handleOpenDialog: () => void
+}
+
+const DefaultMessage = ({ handleOpenDialog }: DefaultMessageProps): JSX.Element => (
+  <div className="flex flex-col items-center justify-center">
+    <span className="inline-block mb-8">
+      <Paragraph variant="body">
+        <span className="text-purple-800">Drag and drop file or{" "}</span>
+        <Hyperlink text="Select a file" handleOnClick={handleOpenDialog} />
+      </Paragraph>
+    </span>
+    <Paragraph variant="extra-small"><span className="text-purple-800/70">.xlsx format only</span></Paragraph>
+  </div>
+)
+
+export { DefaultMessage }
--- a/packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
+++ b/packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
@@ -1,0 +1,61 @@
+import React from "react"
+import { ErrorCode, FileError, FileRejection } from "react-dropzone";
+import { InlineNotification } from "@kaizen/notification";
+import { getKaizenDropzoneError, prettifyErrorCode } from "../../utils";
+
+export const DropzoneError = ({ code, message }: FileError): JSX.Element => {
+  /**  updating this specific error code because the language is awkward. Default code from react-dropzone is 'file-invalid-type'. The other error codes are fine. */
+  const errorCode = code === ErrorCode.FileInvalidType ? "incorrect-file-type" : code
+  const notificationTitle = prettifyErrorCode(errorCode)
+
+  return (
+    <InlineNotification
+      type="negative"
+      title={notificationTitle}
+      persistent
+      noBottomMargin={true}
+    >
+      <span
+        tabIndex={0}
+        role="alert"
+        aria-live="polite"
+        data-automation-id={""}
+      >
+        {message}
+      </span>
+    </InlineNotification>
+  )
+}
+
+export interface DropzoneErrorsProps {
+  acceptedFileTypes: string | string[],
+  maxFileSize: number,
+  /** May implement this in the future. Adding now to cover all four error code provided from react-dropzone library */
+  minFileSize: number,
+  dropzoneErrors: FileRejection[]
+}
+
+export const DropzoneErrors = ({
+  acceptedFileTypes,
+  maxFileSize,
+  minFileSize,
+  dropzoneErrors
+}: DropzoneErrorsProps): JSX.Element => (
+  <span>
+    {dropzoneErrors.map(dropzoneError => {
+      const { errors } = dropzoneError
+      return errors.map((error: FileError) => {
+        const errorMessage = getKaizenDropzoneError({
+          errorCode: error.code as ErrorCode,
+          acceptedFileTypes,
+          maxFileSize,
+          minFileSize
+        })
+        return (
+          <span key={`dropzone-error-${error.code}`} className={"block mb-4"}>
+            <DropzoneError code={errorMessage.code} message={errorMessage.message} />
+          </span>)
+      })
+    })}
+  </span>
+)
--- a/packages/components/src/Dropzone/sub-components/DropzoneIcon/DropzoneIcon.tsx
+++ b/packages/components/src/Dropzone/sub-components/DropzoneIcon/DropzoneIcon.tsx
@@ -1,0 +1,26 @@
+import React from "react"
+import classNames from "classnames"
+import { Icon } from "@kaizen/component-library"
+import uploadFileSvgIcon from "@kaizen/component-library/icons/import.icon.svg"
+import { AcceptedFileIcon } from "../SelectedFileMessage/SelectedFileMessage"
+
+const UploadFileIcon = ({ isHover = false }: { isHover?: boolean }): JSX.Element => (
+  <span className={classNames({
+    "inline-block text-purple-800/70 mb-8 transition-transform -translate-y-0": true,
+    "!text-blue-500 !-translate-y-6": isHover
+  })}>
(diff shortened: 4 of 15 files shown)


packages/components/src/Dropzone/sub-components/DefaultMessage/DefaultMessage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const DefaultMessage = ({ handleOpenDialog }: DefaultMessageProps): JSX.Element => (
+  <div className="flex flex-col items-center justify-center">
+    <span className="inline-block mb-8">
+      <Paragraph variant="body">
+        <span className="text-purple-800">Drag and drop file or{" "}</span>
+        <Hyperlink text="Select a file" handleOnClick={handleOpenDialog} />
+      </Paragraph>
+    </span>
+    <Paragraph variant="extra-small"><span className="text-purple-800/70">.xlsx format only</span></Paragraph>
```
gyfchong, line 17: We recommend that this should be plain text from the consumer as it would be difficult to generalise this statement when designers may want to abbreviate this message to say "All images" or not say anything at all when it comes to allowing all file types.

gyfchong: As a side note, you can use the `color` and `tag` props on `Paragraph` to achieve these styles.
```suggestion
    <Paragraph variant="extra-small" tag="span" color="dark-reduced-opacity">.xlsx format only</Paragraph>
```

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { InlineNotification } from "@kaizen/notification";
+import { getKaizenDropzoneError, prettifyErrorCode } from "../../utils";
+
+export const DropzoneError = ({ code, message }: FileError): JSX.Element => {
+  /**  updating this specific error code because the language is awkward. Default code from react-dropzone is 'file-invalid-type'. The other error codes are fine. */
+  const errorCode = code === ErrorCode.FileInvalidType ? "incorrect-file-type" : code
+  const notificationTitle = prettifyErrorCode(errorCode)
+
+  return (
+    <InlineNotification
```
gyfchong, line 12: Please replace all `InlineNotification` with `FieldMessage` as this is what we use in our form fields.
https://cultureamp.design/storybook/?path=/docs/components-form-subcomponents-field-message--default-kaizen-site-demo

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
...
+      type="negative"
+      title={notificationTitle}
+      persistent
+      noBottomMargin={true}
+    >
+      <span
+        tabIndex={0}
```
gyfchong, line 19: To ensure that all error messages are read by the screen reader, this `tabIndex` needs to be applied to your container. eg.
```
<div tabIndex={-1}>
    <FieldMessage />
    <FieldMessage />
</div>
```

gyfchong: Also generally this `span` (and it's remaining props) won't be needed when converted to use `FieldMessage`

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
...
+  dropzoneErrors: FileRejection[]
+}
+
+export const DropzoneErrors = ({
+  acceptedFileTypes,
+  maxFileSize,
+  minFileSize,
+  dropzoneErrors
+}: DropzoneErrorsProps): JSX.Element => (
+  <span>
```
gyfchong, line 44: This will need to be a `div` to accommodate my other comment regarding the `tabIndex`

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react"
+import { ErrorCode, FileError, FileRejection } from "react-dropzone";
+import { InlineNotification } from "@kaizen/notification";
+import { getKaizenDropzoneError, prettifyErrorCode } from "../../utils";
+
+export const DropzoneError = ({ code, message }: FileError): JSX.Element => {
+  /**  updating this specific error code because the language is awkward. Default code from react-dropzone is 'file-invalid-type'. The other error codes are fine. */
+  const errorCode = code === ErrorCode.FileInvalidType ? "incorrect-file-type" : code
```
gyfchong, line 8: I'd prefer we stick with what comes out of the library as something like this transformation is easily overlooked when debugging

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
...
+      return errors.map((error: FileError) => {
+        const errorMessage = getKaizenDropzoneError({
+          errorCode: error.code as ErrorCode,
+          acceptedFileTypes,
+          maxFileSize,
+          minFileSize
+        })
+        return (
+          <span key={`dropzone-error-${error.code}`} className={"block mb-4"}>
+            <DropzoneError code={errorMessage.code} message={errorMessage.message} />
```
gyfchong, line 56: As per my previous comments in this file, it might be easier to use the `FieldMessage` directly in here rather than creating a new component.

packages/components/src/Dropzone/sub-components/DropzoneIcon/DropzoneIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+import classNames from "classnames"
+import { Icon } from "@kaizen/component-library"
+import uploadFileSvgIcon from "@kaizen/component-library/icons/import.icon.svg"
+import { AcceptedFileIcon } from "../SelectedFileMessage/SelectedFileMessage"
+
+const UploadFileIcon = ({ isHover = false }: { isHover?: boolean }): JSX.Element => (
+  <span className={classNames({
+    "inline-block text-purple-800/70 mb-8 transition-transform -translate-y-0": true,
+    "!text-blue-500 !-translate-y-6": isHover
+  })}>
```
gyfchong, line 11: You can pass `classNames()` into `Icon` via the `classNameOverride`, so you don't need this span

packages/components/src/Dropzone/sub-components/Hyperlink/Hyperlink.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <a
+        href={href}
+        aria-label={text}
+        target={openNewTab ? "_blank" : "_self"}
+        className="text-blue-500 underline hover:underline"
+      >
+        {text}
+      </a>
+    )
+  }
```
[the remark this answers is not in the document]
gyfchong: If it's in case your error message needs a link, this should be handled by the consumer passing in a custom error message in the form of a ReactNode.

packages/components/src/Dropzone/sub-components/Hyperlink/Hyperlink.tsx
```diff
@@ -0,0 +1,45 @@
...
+import classNames from "classnames"
+
+interface HyperlinkProps {
+  text: string,
+  href?: `https://${string}`
+  openNewTab?: boolean,
+  handleOnClick?: () => void
+}
+
+export const Hyperlink = ({
```
gyfchong, line 11: Because this component's styles are strictly for Dropzone, we recommend calling it something very specific like `SelectFileButton`

packages/components/src/Dropzone/sub-components/Hyperlink/Hyperlink.tsx
```diff
@@ -0,0 +1,45 @@
...
+        tabIndex={0}
+        className={classNames({
+          "p-0 rounded-[2px] focus:border-w-default focus:border-blue-500 focus:outline-none focus:ring focus:ring-blue-500 bg-transparent border-none outline-none cursor-pointer text-blue-500 underline hover:underline": true,
+        })}
+      >
+        {text}
+      </button>
+    )
+  }
+  return <span className="text-blue-500 underline">{text}</span>
```
gyfchong, line 43: We shouldn't have text that looks interactive but isn't 😅

packages/components/src/Dropzone/sub-components/OnDragMessage/OnDragMessage.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { Paragraph } from "@kaizen/typography"
+
+export const OnDragMessage = (): JSX.Element => (
+  <div className="flex flex-col items-center justify-center">
+    <span className="mb-8">
```
gyfchong, line 6: You can add this class to `Paragraph` via `classNameOverride`

packages/components/src/Dropzone/sub-components/OnDragMessage/OnDragMessage.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { Paragraph } from "@kaizen/typography"
+
+export const OnDragMessage = (): JSX.Element => (
+  <div className="flex flex-col items-center justify-center">
```
gyfchong, line 5: You shouldn't need to use `flex` for this. Even if you remove all the spans, a `text-center` should align nicely.

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Paragraph } from "@kaizen/typography"
+import { trunicateFilename } from "../../utils"
+import { Hyperlink } from "../Hyperlink/Hyperlink"
+
+interface SelectedFileMessageProps {
+  fileName: string
+  handleOpenDialog: () => void
+}
+
+export const AcceptedFileIcon = (): JSX.Element => (
```
gyfchong, line 15: We suggest making all the Icons separate files, then imported where needed.

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  <span className="inline-block mb-8 text-green-500">
+    <Icon
+      height={20}
+      width={20}
+      icon={acceptedFileSvgIcon}
+      title="Selected File"
+      role="presentation"
+      inheritSize={true}
```
gyfchong, line 23: Wrapping this icon in `inline-block` isn't necessary if you remove the `inheritSize`

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { trunicateFilename } from "../../utils"
+import { Hyperlink } from "../Hyperlink/Hyperlink"
+
+interface SelectedFileMessageProps {
+  fileName: string
+  handleOpenDialog: () => void
+}
+
+export const AcceptedFileIcon = (): JSX.Element => (
+  <span className="inline-block mb-8 text-green-500">
```
gyfchong, line 16: I noticed all your icons are wrapped in an `mb-8` class. Instead of doing this individually, you should add a `div` container in the `Dropzone.tsx` to house all the icons:

```
<div className="mb-8">
    <DropzoneIcon />
</div>
```

This way the overall content structure is easily manipulated (eg. changing the spacing between icons and content).

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const maxCharacterLimit = 50
+  let selectedFileName = fileName
+
+  if (fileName?.length > maxCharacterLimit) {
+    selectedFileName = trunicateFilename(fileName, maxCharacterLimit)
+  }
+
+  return (
+    <div className="flex flex-col items-center justify-center">
+      <span className="mb-8">
```
gyfchong, line 38: Similar to the icons, it might be simpler to lift this wrapper to `Dropzone.tsx` and have all messages live inside it:

```
<div className="mb-8">
    <SelectedFileMessage />
    <DefaultMessage />
</div>
```

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Paragraph variant="body">
+          <span className="block mb-12 break-all"><strong>{selectedFileName}</strong></span>
+        </Paragraph>
+        <Paragraph variant="body">
+          <span className="text-purple-800">Drag and drop to replace file or {" "}</span>
+          <Hyperlink text="Select another File" handleOnClick={handleOpenDialog} />
```
gyfchong, line 44: Both the `Default` and this component have the same message, maybe simplify the overall number of components and abstract the logic for `selectedFileName`, then lift all this markup to `Dropzone.tsx`:

```
<div>
    {selectedFileName && parseFileName(selectedFileName)}
    {condition && defaultMessage}
    <div>
            Drag and drop to replace file or
            <SelectFileButton />
    </div>
</div>
```
This will save us from having to update multiple places if this button or text needs to change.

packages/components/src/Dropzone/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+export const prettifyErrorCode = (errorCode: ErrorCode | string): string => {
+  let title: string = ""
+  /** Capitalise first letter */
+  title = errorCode.charAt(0).toUpperCase() + errorCode.slice(1)
+  /** Replace all hyphens with spaces */
+  title = title.replace(/-/g, " ")
+  return title
+}
+
+export const printAcceptedFileTypes = (
```
gyfchong, line 32: This function would need to work on the `accept` object rather than an array of strings as we'd want to make sure this scales to more than just one file type.

packages/components/src/Dropzone/sub-components/DropzoneError/DropzoneError.tsx
```diff
@@ -0,0 +1,61 @@
...
+        data-automation-id={""}
+      >
+        {message}
+      </span>
+    </InlineNotification>
+  )
+}
+
+export interface DropzoneErrorsProps {
+  acceptedFileTypes: string | string[],
```
gyfchong, line 31: Should make this only accept a `string[]` to simplify logic

packages/components/src/Dropzone/utils.ts
```diff
@@ -0,0 +1,101 @@
+import { ErrorCode, FileError } from "react-dropzone"
+import { DropzoneErrorsProps } from "./sub-components/DropzoneError/DropzoneError"
+
+export const trunicateFilename = (
```
gyfchong, line 4: Each of these should be it's own file in a `utils` folder

packages/components/src/Dropzone/sub-components/DropzoneIcon/DropzoneIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+      icon={uploadFileSvgIcon}
+      title="Upload File"
+      role="presentation"
+      inheritSize={true}
+    />
+  </span>
+)
+
+export const DropzoneIcon = ({ isDragActive, fileName }: { isDragActive: boolean, fileName: string | null }): JSX.Element => {
+  if (fileName) return <AcceptedFileIcon />
```
gyfchong, line 24: Best to avoid passing in a string to be treated as a boolean, a better way would be to have a `hasFile` prop and leave it up to the consumer to define how to achieve that boolean.

packages/components/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,115 @@
...
+      setEnableFileUpload(true)
+    } else {
+      setEnableFileUpload(false)
+    }
+  }, [acceptedFiles])
+
+  return (
+    <div className={classNames(className)}>
+      <div
+        data-automation-id={DROPZONE_AUTOMATION_ID}
```
gyfchong, line 89: This is an outdated method of applying test id's as the react-testing-library opts to use `data-test-id`

Having said that we prefer to allow the consumers to pass any extra prop they want through to the component, so we would spread the rest of the props onto your top level wrapping div (once extracting all the props required for the dropzone config).

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  handleOpenDialog: () => void
+}
+
+export const AcceptedFileIcon = (): JSX.Element => (
+  <span className="inline-block mb-8 text-green-500">
+    <Icon
+      height={20}
+      width={20}
+      icon={acceptedFileSvgIcon}
+      title="Selected File"
```
gyfchong, line 21: There's an interesting issue where after the file has been attached, focus is lost completely and the screen reader jumps back to the browser window. Ideally what we expect to happen is the focus to shift to this icon (or it's container) and read out this title to let the SR user know that what they have done is successful.

packages/components/src/Dropzone/sub-components/OnDragMessage/OnDragMessage.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { Paragraph } from "@kaizen/typography"
+
+export const OnDragMessage = (): JSX.Element => (
```
gyfchong, line 4: This should be renamed to sound more like a component, as we use the "On" prefix for functions.

packages/components/src/Dropzone/sub-components/SelectedFileMessage/SelectedFileMessage.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react"
+import { Button } from "@kaizen/button"
+import { Icon } from "@kaizen/component-library"
+import selectedFileSvgIcon from "@kaizen/component-library/icons/file.icon.svg"
+import acceptedFileSvgIcon from "@kaizen/component-library/icons/success.icon.svg"
```
gyfchong, line 5: Use local paths for these

packages/components/src/Dropzone/sub-components/Hyperlink/Hyperlink.tsx
jared-luke: Great call out, and I have refactored the component to only do one thing instead of account for the 'link'. Which was mentioned in this comment here: https://github.com/cultureamp/kaizen-design-system/pull/3353#discussion_r1136501150

packages/components/src/Dropzone/sub-components/Hyperlink/Hyperlink.tsx
jared-luke: This has been resolved

packages/components/src/Dropzone/sub-components/OnDragMessage/OnDragMessage.tsx
jared-luke: If I do this the tailwind styles are the paragraph is inheriting the `Paragraph.module.scss` over the tailwind class.
Screenshot 2023-03-23 at 4 42 49 pm

To get around this, I need to use the `!important` flat. Is this something that you're okay with using in this component @gyfchong?

packages/components/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,115 @@
...
+import classNames from "classnames"
+import { useDropzone, FileRejection, DropzoneOptions } from "react-dropzone"
+import { ACCEPTED_FILE_TYPES, DEFAULT_MAX_FILE_SIZE_40KB, DROPZONE_AUTOMATION_ID } from "./constants"
+import { DefaultMessage } from "./sub-components/DefaultMessage/DefaultMessage"
+import { DropzoneErrors } from "./sub-components/DropzoneError/DropzoneError"
+import { DropzoneIcon } from "./sub-components/DropzoneIcon/DropzoneIcon"
+import { OnDragMessage } from "./sub-components/OnDragMessage/OnDragMessage"
+import { SelectedFileMessage } from "./sub-components/SelectedFileMessage/SelectedFileMessage"
+
+type DropzoneProps = {
```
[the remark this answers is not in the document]
jared-luke: What are you thoughts on extending or `Pick`ing some of the `InputHTMLAttributes` as well. This can allow attributes like `name` when used inside of a `form` element being passed down to the dropzone component as well

packages/components/src/Dropzone/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+  const k = 1000
+  const dm = decimals < 0 ? 0 : decimals
+  const sizes = ["Bytes", "kb", "mb", "gb"]
+
+  const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+  return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))}${sizes[i]}`
+}
+
+interface GetKaizenDropzoneError
```
gyfchong, line 57: We're inside Kaizen already, no need to namespace these 😅

gyfchong:
Let's re-open this PR if needed because it would be better placed in Tier 3

PR #3351, opened 2023-03-08
chore: upgrade storybook to v7

HeartSquared:
## What

Storybook changes:
- Update Storybook to v`7.0.0`
- Update `preview.tsx` to use default export
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#default-export-in-previewjs
- Remove `--no-manager-cache` flag from storybook build as it no longer exists (broke the build)
- Add new local script `storybook:preview` to locally develop with docs view (same nav structure as release build)
- Remove `preview-head.html` as the font source has been deactivated and the TS hack doesn't appear needed
- Update all titles to not use string literals
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#string-literal-titles
	- Delete constants used for titles as they are no longer usable
	- Fix sort order to not use constants
- Remove usage of `withDesign` and `figmaEmbed`
	- Broken with update. Discussed with Kaizen team and we would prefer it gone instead.
	- Remove `storybook-addon-designs` (plugin for this feature)
- Fix stories using imported markdown
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#importing-plain-markdown-files-with-transcludemarkdown-has-changed
- Update type `ComponentMeta` to `Meta`
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#componentstory-componentstoryobj-componentstoryfn-and-componentmeta-types-are-deprecated
- Update types `ComponentStory` and `Story` to `StoryFn`
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#componentstory-componentstoryobj-componentstoryfn-and-componentmeta-types-are-deprecated
- Replace type `DecoratorFunction` with `Decorator`
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#renamed-decoratorfn-to-decorator
- Move query decorator from global to `FilterMultiSelect` `Async` story
- Exclude decorators from code snippets
	- https://github.com/storybookjs/storybook/issues/21456#issuecomment-1458268788
- Replace usages of `ArgsTable` with `Controls`
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#argstable-block
- Wrap examples that do not live within a `Canvas` with `Unstyled`
	- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#default-docs-styles-will-leak-into-non-story-user-components
- Update `.stories.mdx` files to `.mdx`
[...]

PR #3355, opened 2023-03-09
docs(tailwind): add links to spacing system docs in spacing utils docs

JakePitman:
## Why
To help consumers get more familiar with our new spacing system.

## What
Adds links to the spacing system docs from our spacing utils docs (padding and margin), like so:
image


gyfchong:
The light blue contrast on the white and the light grey shadow is hard to look at, can we use one of the notifications instead?

HeartSquared:
Note that this file has been renamed in https://github.com/cultureamp/kaizen-design-system/pull/3351.

To whoever merges later, this needs to be fixed (may not create a conflict).

JakePitman:
> Note that this file has been renamed in #3351.
> 
> To whoever merges later, this needs to be fixed (may not create a conflict).

Which file?

gyfchong:
> > Note that this file has been renamed in #3351.
> > To whoever merges later, this needs to be fixed (may not create a conflict).
> 
> Which file?

Both! https://github.com/cultureamp/kaizen-design-system/pull/3351/files#diff-5d0897a23051da4e30d53656b17ec5dbe6882fb3995b0964bc7d867b0866d546L20

PR #3359, opened 2023-03-10
feat: add tw styles for toc component

mcwinter07:
## Why
Make TOC align with designs - [ticket](https://cultureamp.atlassian.net/browse/KZN-1075)

## What
- adds tw utilities to link and list class
- updates toc title to h2 for a11y rotor findability
- adds ts ignore and Unstyled tag to wrap the TableOfContents
  - this is here as `@storybook/blocks` is built on a lower version of react (not 18) and FC's do not require children to be explicitly added to a type

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "@kaizen/design-tokens": "^10.0.11",
     "@kaizen/tailwind": "*",
+    "@storybook/blocks": "^7.0.0-beta.64",
     "@storybook/addons": "^7.0.0-beta.64",
     "@storybook/api": "^7.0.0-beta.64",
     "@storybook/components": "^7.0.0-beta.64",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
   "dependencies": {
     "@kaizen/design-tokens": "^10.0.11",
     "@kaizen/tailwind": "*",
-    "@storybook/blocks": "^7.0.0-beta.64",
     "@storybook/addons": "^7.0.0-beta.64",
     "@storybook/api": "^7.0.0-beta.64",
     "@storybook/components": "^7.0.0-beta.64",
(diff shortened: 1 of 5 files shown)


gyfchong:
You can merge in the `v7-addons` to fix the TS error now 😄

storybook/preview.tsx
```diff
@@ -28,7 +27,11 @@ export const PageContainer = ({
   <DocsContainer {...props}>
     <Main>
       <Sidebar>
-        <TableOfContents />
+        {/* This is due to Unstyled typed as a being a FC pre-react-18 */}
+        {/* @ts-ignore */}
+        <Unstyled>
```
gyfchong, line 32: Can this go inside the `TableOfContents.tsx`?

mcwinter07: Should be able to, will move it and make sure it's all hunky dory

mcwinter07: Interesting, moving this to the TOC now has a eslist error where it doesn't in preview 🤔 
Screenshot 2023-03-10 at 2 01 28 pm
I mean we could add it as a dependency, which should clear this up.

mcwinter07: Just tried and in breaks on the latest alpha, though interestingly if I lock it into `7.0.0-beta.64` like the rest of the beta releases its hunky dory. Maybe they intend to remove Unstyled in alpha .8? Don't know if that will spell trouble for us 🤷

package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@kaizen/design-tokens": "^10.0.11",
     "@kaizen/tailwind": "*",
+    "@storybook/blocks": "^7.0.0-beta.64",
```
gyfchong, line 35: This comes with `@storybook` so you don't need to import it.

mcwinter07: If I don't have it as a dependency I get [this eslint](https://github.com/cultureamp/kaizen-design-system/pull/3359#discussion_r1131905717) issue I mentioned

mcwinter07: @gyfchong we can always move it back for now and then address this later if it is an issue - though I do wonder about the next release if the next alpha was anything to go by

gyfchong: Yeah put the Unstyled back to Preview and move on lol

PR #3364, opened 2023-03-15
feat: KZN-1068 - Expose rte internal state

maiconmborges:
## Why
The object current returned on the RTE's `onChange` handler doesn't have all properties needed by some ProseMirror modules to content to/from other formats. Currently, the Performance team is using a different text editor, and saving the format as markdown. As a short term goal, we want to replace this editor with Kaizen's RTE and use a parser to convert to/from markdown, so that no BE and database changes are needed.

**Tasks**:
https://cultureamp.atlassian.net/browse/TX-48

**Solution preview**
https://cultureamp.atlassian.net/wiki/spaces/TV/pages/3066855433/Solution+preview+Use+Kaizen+Rich+Text+Editor+in+Performance

**Discussion:**
https://cultureamp.atlassian.net/browse/KZN-1068

## What
**THIS IS A BREAKING CHANGE.** Instead of returning a content array, we are now returning the whole editorState, giving developers access to important information about the current state of the editor, like its schema. The style of the `ul` and `ol` are also being explicitly set, to avoid conflict with possible CSS resets from other codebases.

As an extra, I also added a secondary commit which fixes the style of the editor's height based on the `rows` prop (now it takes in consideration the margin of the paragraphs and lists).

### DIFF
--- a/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
+++ b/packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
@@ -23,7 +23,7 @@ import styles from "./RichTextEditor.module.scss"
 
 export interface BaseRichTextEditorProps
   extends OverrideClassName<Omit<HTMLAttributes<HTMLDivElement>, "onChange">> {
-  onChange: (content: EditorContentArray) => void
+  onChange: (content: ProseMirrorState.EditorState) => void
   value: EditorContentArray
   controls?: ToolbarItems[]
   /**
@@ -117,7 +117,7 @@ export const RichTextEditor = (props: RichTextEditorProps): JSX.Element => {
   const controlMap = buildControlMap(schema, editorState, controls)
 
   useEffect(() => {
-    onChange(editorState.toJSON().doc.content)
+    onChange(editorState)
     // Including `onContentChange` in the dependencies here will cause a 'Maximum update depth exceeded' issue
   }, [editorState])
 
--- a/packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
+++ b/packages/rich-text-editor/docs/EditableRichTextContent.stories.tsx
@@ -40,6 +40,8 @@ function InlineEditor(props: {
   const [rteData, setRTEData] = useState<EditorContentArray>(
     props.content || dummyContent
   )
+  const handleOnChange = (editorState): void =>
+    setRTEData(editorState.toJSON().doc.content)
   const handleContentClick = (): void => setEditMode(true)
   const handleCancel = (): void => setEditMode(false)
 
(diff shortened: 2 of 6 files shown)


gyfchong:
You'll need the proper conventional commit to apply the breaking change, run `yarn commit` and it will guide you. I think your Readme also needs to start with **BREAKING CHANGE**

packages/rich-text-editor/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -23,7 +23,7 @@ import styles from "./RichTextEditor.module.scss"
 
 export interface BaseRichTextEditorProps
   extends OverrideClassName<Omit<HTMLAttributes<HTMLDivElement>, "onChange">> {
-  onChange: (content: EditorContentArray) => void
+  onChange: (content: ProseMirrorState.EditorState) => void
```
hyunjae-lee-cultureamp, line 26: I am hesitant to introduce this breaking change. I'd just provide an additional data leaving 'content' alone as we have a number of other systems integrating RTE, which means they will have to update them too when they upgrade.

hyunjae-lee-cultureamp: Cool thanks. Just saw that thread now.

@dougmacknz @gyfchong 
Some further comments on MD vs Prosemirror, not particularly relating to this PR:

One thing that popped up during the solution preview design meeting was that Performance products rely on MD quite a lot due to how it's supported out of the box in Slack etc. When we send slack message with the RTE content, it's rendered correctly in most part according to the MD format.

We also had opinions saying that MD is the more accepted schema out in the wild than prose-mirror.

gyfchong: I think this needs further thought as to the future of MD, something we can't resolve here. But my overall thought is to answer the question "what does CultureAmp need?" rather than follow a more accepted schema because it's more accepted. For example, If our use case for MD stops at Slack and the rest of the platform needs ProseMirror features, then we can concentrate our adapters in that specific communication area, converting ProseMirror in a very isolated circumstance.

Regarding the breaking change, we should add migration notes for those upgrading on how to achieve the same outcome as they had before this change.

hyunjae-lee-cultureamp: @gyfchong 
Sure we can have more discussion about MD. Not in the scope for this PR.

mcwinter07 (approved):
Nice work @maiconmborges appreciate the PR! The breaking change doesn't introduce too much complexity so the PR and migration notes should suffice for those needing to migrate to the latest version. I'll write up a post in #updates_design_system channel once this gets merged further extrapolating on this

PR #3367, opened 2023-03-20
fix(tailwind): remove w- from borderWidth utilities

JakePitman:
We've decided that the clarity of adding w- to borderWidth isn't worth the busy utility classes. For example, border-b-w-default isn't as nice as border-b-default.

BREAKING CHANGE: Any border width utilities will need to be renamed without the `w-`. These include the following prefixes, where `?` represents a pixel-based spacing token.
`border-w-?`  -> `border-?`
`border-x-w-?` -> `border-x-?`
`border-y-w-?` -> `border-y-?`
`border-t-w-?` -> `border-t-?`
`border-b-w-?` -> `border-b-?`
`border-r-w-?` -> `border-r-?`
`border-l-w-?` -> `border-l-?`

## Why
Adding `w-` (to represent 'width') is redundant, considering that these only get applied to borderWidth anyway. The unnecessary clarity isn't worth the extra utility-class-bulk.

## What
- Removes `w-` from borderWidth utilities.
- Adds an info card explaining that there are other borderWidth prefixes that can be applied to our Kaizen suffixes

### Before
image

### After
image

### DIFF
--- a/docs/Systems/Tailwind/UtilityClassReferences/Borders/BorderWidth.stories.tsx
+++ b/docs/Systems/Tailwind/UtilityClassReferences/Borders/BorderWidth.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { Story } from "@storybook/react"
 import classnames from "classnames"
+import { Card } from "@kaizen/draft-card"
 import { kaizenTailwindTheme } from "@kaizen/tailwind"
 import { CATEGORIES } from "../../../../../storybook/constants"
 import { UtilityClassTemplate } from "../../components/UtilityClassTemplate"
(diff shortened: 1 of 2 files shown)


kaydbwithers:
@JakePitman if possible could we add a border bottom example to the docs? It's something I struggled with last time and I think will help others out.

JakePitman:
> @JakePitman if possible could we add a border bottom example to the docs? It's something I struggled with last time and I think will help others out.

Thanks Kayd that's good feedback. These docs are all generated by mapping over our preset to create the utility classes, but I've looked into adding some radio buttons at the top of the docs to toggle prefixes (`border`, `border-b` etc.). Couldn't make it work before, but that was back when we were emulating utilities with inline styles due to a restriction. We've since fixed that issue, so it may be possible. Planning on looking into it in a P&L day or health week.

The main goal of the Utility Class References docs is to provide all suffixes that are different from the Tailwind defaults, and prefixes are lower priority because they're unchanged from the default. But in the mean-time, I've added an info card to the top of the document explaining that the listed border-width suffixes can be attached to other prefixes, and provided a link to the list of these prefixes in the Tailwind docs.

NatalieLoyzaga (approved):

PR #3368, opened 2023-03-20
ci(chromatic): run chromatic on changes to the /docs folder

JakePitman:
## Why
To detect changes in the docs folder.

## What
Adds the /docs folder to the chromatic process.


.github/workflows/chromatic.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
               - 'packages/**'
               - 'draft-packages/**'
               - 'storybook/**'
+              - 'docs/**'
```
HeartSquared, line 28: You'll only want docs that are stories since there are `.md` files in this directory which will never affect Chromatic, so you'll want to check for `.stories.tsx` or `stories.mdx` only.
I'm guessing it will be something like this `'docs/**/*.stories.(mdx|tsx)'` but I'm not 100% sure (maybe it would need to be separate lines per extension), so you'll want to check using `run chromatic` :)

HeartSquared:
I actually just realised that there are helper `tsx` components within `docs` so we can adjust the check to allow all `mdx|tsx` files (not just `.stories`).

You'll want to test it like this (unfortunately it's manual, because it's testing a GH Action in combination with Chromatic).

For each of the below, you'll want to have them all in a separate commit, and add the `run chromatic` label between each of them (remember to remove the label first, it doesn't auto run just because it exists):

1. Check MDX story runs - Update a `.mdx` in the `docs` dir
2. Check TSX story runs - Update a `.tsx` in the `docs` dir
3. Check non-story doesn't run - Update a `.md` or `.ts` in the `docs` dir
	- Note that the GH Action runs, but you want to check the details and make sure it doesn't end up doing anything in Chromatic
	- You only need to run one of them as long as 1 and 2 are successful

JakePitman:
> I actually just realised that there are helper `tsx` components within `docs` so we can adjust the check to allow all `mdx|tsx` files (not just `.stories`).
> 
> You'll want to test it like this (unfortunately it's manual, because it's testing a GH Action in combination with Chromatic).
> 
> For each of the below, you'll want to have them all in a separate commit, and add the `run chromatic` label between each of them (remember to remove the label first, it doesn't auto run just because it exists):
> 
>     1. Check MDX story runs - Update a `.mdx` in the `docs` dir
> 
>     2. Check TSX story runs - Update a `.tsx` in the `docs` dir
> 
>     3. Check non-story doesn't run - Update a `.md` or `.ts` in the `docs` dir
>        
>        * Note that the GH Action runs, but you want to check the details and make sure it doesn't end up doing anything in Chromatic
>        * You only need to run one of them as long as 1 and 2 are successful

Yeah good pickup.

Thanks for the step-by-step - that's really useful.
I also just noticed we've renamed the files from `mdx` to `md` - I'll update that in the check too.

PR #3369, opened 2023-03-20
feat: Remove built-in margins from our components

christian314159:
## Why
- Laying out and spacing components on a page is the responsibility of the application that uses them
- Any built-in margins take away from this responsibility and might make spacing inconsistent
- We want to focus on tools like Tailwind CSS to add consistent margins to our components

## What
- Removing any built-in margins that are usually set by props
- Making sure the same can still be achieved by using `classNameOverride`


gyfchong:
Should we roll this change in with the one about form input refs? #2628

PR #3370, opened 2023-03-20
fix(form): Link RadioGroup/CheckboxGroup label to input group

dougmacknz:
## Why
Groups of radios or checkboxes should have two things for screenreader functionality:
* An element that groups them. The native element would be `fieldset`, but can also be `role=group/radiogroup`.
* The element that groups the radios/checkboxes needs a label. Natively this is the `legend` in the fieldset. We're using `aria-labelledby` instead for now.

RadioGroup already had `role=radiogroup`, but the label would only be linked up if the consumer provided a `labelId`.

## What
RadioGroup
- creates a unique ID for the radiogroup label, if one hasn't been provided by the consumer with the `labelId` prop.

CheckboxGroup:
- Adds `role=group` to the element grouping the checkboxes
- Adds a `labelId` prop like RadioGroup has.
- Creates a unique ID for the group label, if one hasn't been provided by the consumer with the `labelId` prop. 

Updates tests for both to remove snapshot testing, and add specific tests for checking they have an accessible name and role.


dougmacknz:
If we get a chance to rewrite this, I think we could consolidate down to one `Fieldset` component. For now I'm going with the path of least resistance with the amount of stuff to get through in the audit.

HeartSquared:
What would go wrong if we changed the internal to use `fieldset` and `legend` but kept the props the same? Putting `label` into `legend` for example.

draft-packages/form/KaizenDraft/Form/RadioGroup/RadioGroup.spec.tsx
```diff
@@ -66,21 +14,13 @@ describe("<RadioGroup /> ", () => {
...
-          automationId="test"
-        />
-      )
-      expect(getByRole("radiogroup").getAttribute("aria-labelledby")).toEqual(
-        labelId
-      )
-      expect(
-        getByText("Label title").parentElement?.getAttribute("id")
-      ).toEqual(labelId)
+    it("should have an accessible name when not provided a labelId", () => {
```
christian314159, line 17: Two minor things here:
1. I've been removing all `shoulds` from tests. We are very inconsistent with it, and when I learned testing I was taught not to use "shoulds" because "you assert that the code does do a thing, not that is should be doing it".
2. The way you're using `expect` without a matcher will work - even without the `expect` it will work, because if it can't find the element `getBy...` will throw and hence the test will fail. But for the sake of readability and clarity I'd use the `.tobeInTheDocument()` matcher here.

christian314159 (dismissed):
Looks good, I'd clean up the spec as suggested though before merging.

christian314159 (approved):
Thanks for updating the specs!

PR #3373, opened 2023-03-22
fix: Tooltip named export

gyfchong:


christian314159 (approved):
Hm, I trust you here that it's necessary, but it still looks weird that we're exporting the same component once as a named export and once as the default?

HeartSquared (commented):
Delete the default export please :)

gyfchong:
Generally I think it's fine, handy in some cases, but not necessary 😄

christian314159 (approved):

draft-packages/tooltip/KaizenDraft/Tooltip/index.ts
```diff
@@ -1 +1 @@
-export { default as Tooltip, TooltipProps } from "./Tooltip"
+export { Tooltip, TooltipProps } from "./Tooltip"
```
HeartSquared, line 1: We could do this, but not a big deal currently.

```suggestion
export * from "./Tooltip"
```

HeartSquared (approved):

christian314159 (approved):

PR #3378, opened 2023-03-22
fix: tooltip aio error

gyfchong:


packages/components/package.json
```diff
@@ -40,7 +40,9 @@
     "@popperjs/core": "^2.11.6",
     "classnames": "^2.3.2",
     "react-focus-on": "^3.7.0",
-    "react-popper": "^2.3.0"
+    "react-popper": "^2.3.0",
+    "use-debounce": "^9.0.3",
+    "uuid": "^9.0.0"
```
HeartSquared, line 45: Wait what are these doing here? 😂

gyfchong: Oh right, from Tooltip deps

HeartSquared (approved):

PR #3383, opened 2023-03-24
feat(components): add FilterSelect component

HeartSquared:
## Why

[KZN-1044](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?selectedIssue=KZN-1044)

Requirement for Team Pathfinder feature work.

## What

Add `FilterSelect` component, leveraging `Select` and `Filter`.

Changes:
- add deps to AIO:
	- `@kaizen/select`
	- `@react-aria/button`
	- `@react-aria/select`
	- `@react-stately/select`
	- `@react-types/shared`
- Refactor Select to move `ListBox` and `Overlay` styles to `Select` as they clashed with Filter
- Add new callback prop to `Filter` to return `ref` to consuming component
	- Ref is required for using aria-select hooks
- Abstract Select item transformer
- Abstract Select option render function into `ListItem` and `ListItems`

## Notes

- Work in progress (separate branch) to make FilterSelect a generic for better user experience (will likely just copy/paste things from Select into AIO so as not to break the existing Select). Api may change slightly.
- The reason for the StickerSheet not having anything open is because the focus states all fight each other, then somehow all your `onClick`s will fail to work for any FilterSelect, so I made them only open for Chromatic.

[KZN-1044]: https://cultureamp.atlassian.net/browse/KZN-1044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/package.json
```diff
@@ -36,8 +36,14 @@
   },
   "dependencies": {
     "@kaizen/component-library": "^16.5.1",
+    "@kaizen/design-tokens": "^10.3.1",
```
gyfchong, line 39: It's already a peer dep, is this necessary?

packages/components/src/Filter/Filter.tsx
```diff
@@ -8,11 +8,12 @@ export interface FilterProps
   extends OverrideClassName<HTMLAttributes<HTMLDivElement>> {
   children: React.ReactNode
   isOpen: boolean
-  setIsOpen: React.Dispatch<React.SetStateAction<boolean>>
+  setIsOpen: (isOpen: boolean) => void
```
gyfchong, line 11: Why do we need this prop? Wouldn't it be possible to achieve the same affect toggling `isOpen`?

packages/components/src/FilterSelect/FilterSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+        setIsOpen={state.setOpen}
+        renderTrigger={(): JSX.Element =>
+          renderTrigger({
+            selectedValue: state.selectedItem?.textValue || undefined,
+            label,
+            isOpen,
+            ...buttonProps,
+          })
+        }
+        onTriggerLoaded={setTriggerRef}
```
gyfchong, line 90: Introducing API for `react-aria` 😢

packages/components/src/FilterSelect/FilterSelect.spec.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  it("shows the options initially when isOpen is true", async () => {
+    render(<FilterSelectWrapper isOpen />)
+    await waitFor(() => {
+      expect(screen.queryByRole("listbox")).toBeVisible()
+    })
+  })
+
+  describe("Trigger", () => {
+    it("shows the selected option in the button", () => {
```
gyfchong, line 53: Is it necessary to test this since it's tested in `Filter`? ie. if it fails in `Filter` then we can assume it failed here too.

packages/components/src/FilterSelect/FilterSelect.tsx
HeartSquared: Yeah I really hated that but unfortunately all the hooks rely on each other too much and we couldn't get away with not doing it 😭

packages/components/src/Filter/Filter.tsx
HeartSquared: The prop exists because we expect the consumer to hold the state for `isOpen`. The toggle functionality is used because we pass them to the trigger so the consumer doesn't need to do it o3o

packages/components/package.json
HeartSquared: Nope! I don't know where this came from :D

packages/components/src/FilterSelect/FilterSelect.spec.tsx
HeartSquared: In a normal abstraction, we absolutely do not need to test it. But of course, react-aria is so opinionated and does funny stuff in its state that I can't ensure won't change, I figured I'd add a test.

That being said, I realise I have an example of a selected option in the Sticker Sheet, so I suppose we could probably get away without this unit test?

packages/components/src/Filter/Filter.tsx
gyfchong: Wouldn't this then be more accurately named `handleOnClick`?

packages/components/src/FilterSelect/FilterSelect.spec.tsx
gyfchong: Hm I think keep it, probably a tighter feedback loop compared to waiting on Chromatic

packages/components/src/FilterSelect/FilterSelect.tsx
gyfchong: What do you think about adding something to signify this is a specific prop for react aria?
eg. `setReactAriaTriggerRef`

Or even introduce a `reactAria` prop that accepts this?

HeartSquared: Nah I don't want to be too tied into saying a prop only exists for a library. If some other library decides it also needs a ref, it can use it too.
That being said, if we want to call it something more specifically related to having a `ref` that's fine too. I also considered `onMount` lol, cos that's what it's doing.

packages/components/src/Filter/Filter.tsx
HeartSquared: `onTriggerClick`? :P

The problem with that is it isn't obvious that I need the open state handler in there. Then you'll just end up with something that appears borked. Not to mention, because it's passed back to the trigger as props, you can actually extend it as the consumer through there, and not put it into the `Filter` prop.

packages/components/src/FilterSelect/FilterSelect.tsx
gyfchong: Ah yeah, bring back ol' class terms, `onTriggerMount`

gyfchong: Actually, just being wary, but we're stacking a lot of props for the trigger. Should we consider an object prop for `trigger` to encapsulate it a bit?

```
trigger: {
   render: () => {},
   onMount: () => {}
}
```

HeartSquared: The mounting is an extremely unlikely case tbh. Adding another layer to the trigger prop is probably not very nice. Also we're back at the confusing `trigger` prop naming 😂 
I think if we keep it as `onMount` it could be more extensible moving forward if we want more things later since it's attached to an agnostic event as opposed to tied to an event for an element.

gyfchong: Hmm I suppose at this stage, all the props are technically for the trigger lol

HeartSquared:
Converting the component to a generic had much more changes than I anticipated, so closing this PR to open another for the new branch.

PR #3389, opened 2023-03-27
fix: github public key

gyfchong:
## Why
Github screwed up

## What
https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/


HeartSquared (approved):
Lul

PR #3390, opened 2023-03-28
add FilterSelect component

HeartSquared:
## Why

[KZN-1044](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?selectedIssue=KZN-1044)

Requirement for Team Pathfinder feature work.

## What

Add `FilterSelect` component, leveraging `Filter`.

Changes:
- add deps to AIO:
	- `@react-aria/button`
	- `@react-aria/select`
	- `@react-stately/select`
	- `@react-types/shared`
	- `@react-stately/collections`
	- `@react-aria/focus`
	- `@react-aria/overlays`
	- `@react-aria/listbox`
	- `@react-aria/utils`
	- `@kaizen/typography`
	- `@kaizen/draft-divider`
- Add new callback prop to `Filter` to return `ref` to consuming component
	- Ref is required for using aria-select hooks

## Notes

- The reason for the StickerSheet not having anything open is because the focus states all fight each other, then somehow all your `onClick`s will fail to work for any FilterSelect, so I made them only open for Chromatic.

[KZN-1044]: https://cultureamp.atlassian.net/browse/KZN-1044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/FilterSelect/subcomponents/ListBoxSection/ListBoxSection.tsx
```diff
@@ -0,0 +1,42 @@
...
+    heading: section.rendered,
+    "aria-label": section["aria-label"],
+  })
+
+  return (
+    <>
+      <li {...itemProps} className={styles.sectionWrapper}>
+        {section.rendered && (
+          <Heading
+            variant="heading-6"
```
gyfchong, line 25: @dougmacknz should this be a `h6`? This was copied from `Select` but just wanted to double check.

dougmacknz: Hm, this actually ends up with `aria-hidden=true`.
image

The react-aria docs have these as spans https://react-spectrum.adobe.com/react-aria/useListBox.html#sections

I think we should switch this to a span just to make it clear we don't want a header here.

gyfchong:
I want to test Changeset with this, could you do `yarn changeset` to add one and amend your title to break conventional commits? (I assume Lerna will ignore/fail on non-conventional titles)

packages/components/src/FilterSelect/subcomponents/ListBoxSection/ListBoxSection.tsx
dougmacknz: I reckon this was just another case of forgetting to use the `tag` prop. It happens all the time, even in our team.

gyfchong (approved):

PR #3391, opened 2023-03-28
chore: Adding ESLint rules for our tests

christian314159:
## Why
Found some tests that
- had matchers that weren't invoked
- had no matchers at all

## What
- Added new ESLint rules specifically for our tests that now lint for a variety of things that will improve our tests:
  - Matchers are now required and checked to be invoked
  - Focused tests aren't allowed anymore (ensuring we run all tests in CI)
  - Expects are required
  - Top level describes are required
  - Enforces more appropriate matchers (e.g. `expect(foo).toBe(5)` over `expect(foo === 5).toBe(true)`)
  - Many other small things

### What not
- I replaced "shoulds" where I came across them, but there's still many tests using them in the descriptions; I'm not enforcing a removal, it'll hopefully happen over time


.eslintrc.js
```diff
@@ -198,6 +199,19 @@ module.exports = {
         pathGroupsExcludedImportTypes: [],
       },
     ],
+    "jest/consistent-test-it": ["error", { fn: "it" }],
```
christian314159, line 202: Explanations for all these rules can be found here: https://github.com/jest-community/eslint-plugin-jest

packages/date-picker/src/FilterDateRangePicker/hooks/useRangeDateValidation.spec.tsx
```diff
@@ -1,99 +1,74 @@
...
+          inputLabel: "Start date",
+        })
+      )
+      const { validateDate, updateValidation } = result.current
+      const { validationResponse, newDate } = validateDate({
+        date: new Date("potato"),
+        inputValue: "potato",
+      })
+
+      act(() => {
```
HeartSquared, line 37: Should it be this instead? Or are hooks different? You'd need to `await` regardless though I think (at least, I know that `waitFor` without it would often lead to false positives).

```suggestion
      await waitFor(() => {
```

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
```diff
@@ -14,6 +16,7 @@ describe("no-invalid-use-of-var-tokens-in-equations rule", () => {
         },
       }
     )
+    /* eslint-disable-next-line jest/valid-expect */
     return expect(reported === 0)
```
HeartSquared, line 20: Wait, we're returning an expect? 😵

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -144,7 +144,6 @@ describe("<DatePicker /> - Focus element", () => {
 
       const input = screen.getByLabelText("Input label")
       await userEvent.tab()
-      expect(input).toHaveFocus()
```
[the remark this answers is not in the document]
christian314159: One of my rules says no `expect`s outside of `it`s... this is in a `beforeEach`.
I think I just made a quick call that it might not be super important with it being in that `before`.
If you think it is, we can either disable the rule for that line or move things out of the `before` into the actual test.

packages/stylelint-plugin/src/rules/no-invalid-use-of-var-tokens-in-equations.spec.ts
christian314159: If you expand this, you'll see it's essentially a custom `expect` that's used further down.
The spec is way too wild, I'm not refactoring it (in this PR).

packages/button/src/Button/IconButton.spec.tsx
```diff
@@ -2,31 +2,41 @@ import React from "react"
...
-  render(
-    <IconButton
-      onClick={(): void => undefined}
-      label="Accessible label on the button version"
-    />
-  )
-  screen.getByLabelText("Accessible label on the button version")
-})
+    expect(
+      screen.getByLabelText("Accessible label on the link version")
```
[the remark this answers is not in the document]
christian314159: Good suggestion, but scoping this out for this PR as it's a general improvement to the existing test.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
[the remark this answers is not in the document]
HeartSquared: Ohhhh. I think the idea of this check is to ensure that the input is the one we're adding the keyboard action to, not another element. Given the dialog can be opened by the button within the input as well, I suppose it's to prevent a false positive for the first test?

packages/components/src/__future__/ExampleComponent/ExampleComponent.spec.tsx
```diff
@@ -4,6 +4,7 @@ import { ExampleComponent } from "./ExampleComponent"
 
 describe("<ExampleComponent />", () => {
   it("renders", () => {
-    render(<ExampleComponent />)
+    const { getByTestId } = render(<ExampleComponent />)
```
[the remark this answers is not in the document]
christian314159: Thanks!

packages/date-picker/src/FilterDateRangePicker/hooks/useRangeDateValidation.spec.tsx
christian314159: iirc The test was working but complaining that something was doing something to the DOM and should be wrapped in `act` - so I did that to remove the warning.
I just tested it with `waitFor` - it actually makes the test fail, so we might just leave it with `act`.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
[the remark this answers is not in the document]
christian314159: Cool. I might change it so it's part of the test.

packages/date-picker/src/FilterDateRangePicker/hooks/useRangeDateValidation.spec.tsx
HeartSquared: Maybe hooks need to be tested differently. All good!

HeartSquared (approved):

PR #3403, opened 2023-03-29
feat: Pass HTMLAttributes to content in EditableRichTextContent

harish-everest:
## Why
- This PR is created to allow passing html attributes to `RichTextContent` component inside the `EditableRichTextContent`.
- This is required for adding id attribute to content div in `EditableRichTextContent`, with this we can use that id for accessible purposes. 
- Without this change if I just pass id to `EditableRichTextContent` and use that id for `aria-labelledby` it will read the button label as well when using voice over.

https://cultureamp.atlassian.net/browse/COMP-664

## What
- Added a separate prop for passing the html attributes to `RichTextContent` inside `EditableRichTextContent`.
- This change does not effect any existing functionality.

### DIFF
--- a/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
+++ b/packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
@@ -3,7 +3,7 @@ import classnames from "classnames"
 import { VisuallyHidden } from "@kaizen/a11y"
 import { OverrideClassName } from "@kaizen/component-base"
 import { Label } from "@kaizen/draft-form"
-import { EditorContentArray, RichTextContent } from "../../"
+import { EditorContentArray, RichTextContent, RichTextContentProps } from "../../"
 import styles from "./EditableRichTextContent.module.scss"
 
 export interface EditableRichTextContentProps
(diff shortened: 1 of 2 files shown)


packages/rich-text-editor/src/EditableRichTextContent/EditableRichTextContent.tsx
```diff
@@ -12,6 +12,7 @@ export interface EditableRichTextContentProps
   content: EditorContentArray
   labelText: string
   isLabelHidden?: boolean
+  contentHTMLAttributes?: HTMLAttributes<HTMLDivElement>
```
gyfchong, line 15:
```suggestion
  contentProps?: Omit<RichTextContentProps, "contents">```

We want to make this as generic as possible, utilising the `RichTextContentProps` from the original `RichTextContent` component so that any of the props can be passed (including your aria attributes!)

harish-everest: Thanks @gyfchong. I have updated the PR.

gyfchong:
Also, please write a test to confirm that the attributes are being passed to the correct element

gyfchong (approved):

PR #3405, opened 2023-03-30
fix: changeset-script

gyfchong:
## Why

Scripts relied on permissions to access private npm packages, which is unnecessary. 

## What

Replaced yarn installs with npx to avoid needing to install private cultureamp packages just to do changeset releases/versions


HeartSquared (approved):
Let's Gamble, Try Merging.

PR #3406, opened 2023-03-30
feat: 71-working-button-state-needs-to-be-announced-with-a-live-region

gyfchong:
https://github.com/cultureamp/kaizen-discourse/issues/71

## What

Add a live region polite to the contents of `GenericButton`

### DIFF
--- a/.changeset/fresh-cats-train.md
+++ b/.changeset/fresh-cats-train.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/button": minor
+---
+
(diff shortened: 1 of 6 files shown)


HeartSquared (approved):
...is that it? Somehow seems too easy 😂

gyfchong:
> ...is that it? Somehow seems too easy 😂

@HeartSquared a11y is easy™

gyfchong:
Dumbass snapshot tests

PR #3407, opened 2023-03-30
fix: npx script

gyfchong:
Because I don't know how NPX works

### DIFF
--- a/packages/button/CHANGELOG.md
+++ b/packages/button/CHANGELOG.md
@@ -1,11 +1,19 @@
 # Change Log
 
+## 2.2.0
+
+### Minor Changes
+
+- ae55e7acd: Add aria-live to GenericButton to support screen readers reading the working state label.
+
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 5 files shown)


packages/button/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@kaizen/button",
-  "version": "2.1.0",
+  "version": "2.2.0",
```
HeartSquared, line 3: Is this meant to be bumping?

packages/button/CHANGELOG.md
```diff
@@ -1,834 +1,435 @@
 # Change Log
 
+## 2.2.0
```
gyfchong, line 3: @HeartSquared how do I revert this file to `main` lol

HeartSquared: https://stackoverflow.com/questions/37972753/git-reset-single-file-in-feature-branch-to-be-the-same-as-in-master

This? I wonder if you could reset hard on the single file 🤔 

Alt, you barely have any changes, so you could always just soft reset to main, then only commit the changes you want :)

gyfchong: Yeah this worked:
```git checkout origin/master packages/button/CHANGELOG.md```

HeartSquared (approved):

PR #3420, opened 2023-03-31
feat(components): add FilterDateRangePicker to AIO

HeartSquared:
## Why

[KZN-1108](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/491?modal=detail&selectedIssue=KZN-1108)

Rebuild FilterDRP in AIO to use Filter abstraction.

### BREAKING CHANGES 

Not applicable to AIO, but in terms of comparison with the old FilterDRP

- Validation structure changed from `status: FieldMessageStatus` & `validationMessage: string` to `validationMessage: { status: FieldMessageStatus; message: string }`
- Use `Filter` structure where open state and Filter Button are passed in by consumer
- Rename `inputRangeStartProps` to `inputStartDateProps`
- Rename `inputRangeEndProps` to `inputEndDateProps`
- `disabledDays` replaces the multiple disabled props to closer match the `react-day-picker` api

### Non-breaking changes

- Add deps to AIO:
	- `@kaizen/a11y`
	- `@kaizen/date-picker`
	- `@kaizen/draft-form`
	- `date-fns`
- Add `FilterDateRangePicker`
- Add comment to deprecate FilterDRP in `@kaizen/date-picker` (package not deprecated)
	- Move main stories under `Deprecated` category
	- Delete subcomponent stories
- Add `LabelledMessage` to correctly render dir="rtl" text with `:` in between

[KZN-1108]: https://cultureamp.atlassian.net/browse/KZN-1108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gyfchong:
You'll need to add a changeset 😄

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/utils/getDateValidationHandler.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export type GetDateValidationHandlerArgs = {
+  onValidate: ((validationResponse: DateValidationResponse) => void) | undefined
+  setInbuiltValidationMessage: (
+    validationMessage: ValidationMessage | undefined
+  ) => void
+  inputLabel: React.ReactNode
+}
+
+export const getDateValidationHandler =
```
gyfchong, line 14: Could this be `dateValidationHandler`? Doesn't seem to return a function that would be considered a handler of any kind.

packages/components/src/LabelledMessage/LabelledMessage.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+import styles from "./LabelledMessage.module.scss"
+
+export interface LabelledMessageProps {
+  label: string | React.ReactElement
+  message: string | React.ReactElement
+}
+
+export const LabelledMessage = ({
```
gyfchong, line 9: Why is this considered a standalone component? Maybe belongs in `Filter`?

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -39,6 +39,9 @@ type InputRangeEndProps = DateRangeInputFieldProps["inputRangeEndProps"]
 type FilterInputProps<InputProps> = Omit<Partial<InputProps>, "value"> &
   DataAttributes
 
+/**
+ * @deprecated Use FilterDateRangePicker from @kaizen/components
```
gyfchong, line 43: Can we include a link to the storybook page here?

packages/components/src/FilterDateRangePicker/subcomponents/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+import styles from "./DateRangeValidationMessage.module.scss"
+
+export interface DateRangeValidationMessageProps {
+  validationMessage?: DateRangeFieldValidationMessage
+  dateStartId?: string
+  dateEndId?: string
+  isReversed?: boolean
+}
+
+export const DateRangeValidationMessage = ({
```
gyfchong, line 13: Looking clean ✨

packages/components/src/FilterDateRangePicker/subcomponents/DateRangeValidationMessage/DateRangeValidationMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+  if (dateStart && dateEnd) {
+    if (dateStart.status === dateEnd.status) {
+      return (
+        <FieldMessage
+          message={
+            <ul className={styles.fieldMessageList}>
+              <li id={dateStartId}>{dateStart.message}</li>
+              <li id={dateEndId}>{dateEnd.message}</li>
+            </ul>
+          }
```
gyfchong, line 33: Out of scope, but maybe `message` could accept an array and pump this out for us.

gyfchong: Although the `id` would be harder to do

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -0,0 +1,248 @@
...
+          value: inputRangeEndValue,
+          ...inputRangeEndProps,
+          // The below props extend the values from inputRangeEndProps, therefore must be below the spread
+          ...inputRangeEndHandlers,
+        }}
+        locale={locale}
+        description={description}
+        validationMessage={{
+          dateStart: startDateValidation.validationMessage,
+          dateEnd: endDateValidation.validationMessage,
```
gyfchong, line 232: Could you match the terminology up between `dateStart` and `startDateValidation` and have it `dateStart.validationMessage`?

packages/components/src/LabelledMessage/LabelledMessage.tsx
HeartSquared: I guess the use case is too generic that I didn't feel it made sense to belong in there 😬 We could use it for any text with the format of `{label}:{message}` (maybe even allow changing the separator later). It's moreso a i18n helper 😬 

We could actually use it for FilterButton too that I look at it lol. But any place that has that pattern it can be used. Might be useful for consumers too?

packages/components/src/FilterDateRangePicker/subcomponents/DateRangeValidationMessage/DateRangeValidationMessage.tsx
HeartSquared: Definitely more for a generic component to handle, but not right now 😉

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -0,0 +1,248 @@
...
+  locale: propsLocale,
+  defaultMonth,
+  selectedRange,
+  onRangeChange,
+  disabledDates,
+  disabledDaysOfWeek,
+  disabledRange,
+  disabledBeforeAfter,
+  disabledBefore,
+  disabledAfter,
```
gyfchong, line 83: Might be also a good opportunity to try reduce these repetitive props, maybe convert it into an object syntax instead.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -0,0 +1,248 @@
...
+  inputRangeStartProps,
+  inputRangeEndProps,
```
gyfchong, line 85: Could these be `InputStartDateProps` and `inputEndDateProps`? Using `Range` seems unnecessary.

packages/components/src/LabelledMessage/LabelledMessage.tsx
gyfchong: Perhaps it needs to be named `Format*` component or maybe `use*` to make it a hook... Just feels like a lost component without something to make it sound more like a utility or helper, cause you can't really just use it by itself outside the context of another component.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/utils/getDateValidationHandler.tsx
HeartSquared: This is a curry function. We are returning a function 😄 Do you want me to make a type for the return value instead of typing the inner function's args and return value?

packages/components/src/LabelledMessage/LabelledMessage.tsx
HeartSquared: Sure you can :D Who says you can't just stick it on your page as contents? Lol

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
HeartSquared: Haha true! Now that I'm breaking everything, might as well remove the over-abstraction! Just gonna make it use what the library natively wants. Hopefully this doesn't mean I need to change the inner component lol.

packages/components/src/LabelledMessage/LabelledMessage.tsx
gyfchong: Lol okay. But what if I wanted to use this to label a list of `Tag` components? `Message` would be incorrect 😄

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
HeartSquared: You really want to break this api don't you 😛 But yeah should be okay. I just need to remember to tell people about it :)

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/utils/getDateValidationHandler.tsx
gyfchong: Ohh I totally missed the double `=>` nah it's fine.

packages/components/src/LabelledMessage/LabelledMessage.tsx
HeartSquared: Hmm... What do you propose that part be called then? It's a labelled something. I don't want to name it `Item` because I have that associated with Select 😂 LabelledNode? LabelledText?

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
gyfchong: Haha yeah, I'd feel worse if it weren't just one team using a "beta" component 😛

packages/components/src/LabelledMessage/LabelledMessage.tsx
gyfchong: `LabelledList`? A list can be of a singular item 😈

packages/date-picker/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -39,6 +39,10 @@ type InputRangeEndProps = DateRangeInputFieldProps["inputRangeEndProps"]
...
+ * https://cultureamp.design/storybook/?path=/docs/components-filter-date-range-picker--default-story
```
gyfchong, line 44:
```suggestion
 * https://cultureamp.design/storybook/?path=/docs/components-filter-date-range-picker
```

😬

packages/components/src/LabelledMessage/LabelledMessage.tsx
HeartSquared: That is 100% worse than `Message` 😂

gyfchong: Damn, I was aiming for 150% worse! Let's go with `Message` 😄

gyfchong:
Can you include one hard coded example that someone can copy paste to get the most simple version working? Then we should be good.

gyfchong (approved):

.changeset/afraid-pumas-enjoy.md
```diff
@@ -0,0 +1,10 @@
+---
+"@kaizen/date-picker": patch
+---
+
+* Deprecate `FilterDateRangePicker`
```
gyfchong, line 5: Include a note for where to find the replacement. I think this is also a good place to note the replacement's API changes given we haven't got the flexibility on the `components/FilterDRP` docs to add migration notes yet.

gyfchong (approved):

gyfchong (approved):

PR #3428, opened 2023-04-03
Kzn 540/add react intl to components

JakePitman:
## Why
So that we can translate KZ components when necessary, such as the Filter DRP.

## What
- Adds a function to safely translate strings, even when the consumer doesn't have an intl config object set up in the context of their app


JakePitman:
Closing this as we've opted for a different strategy - adding the OptionalIntlProvider to KaizenProvider means we don't need the graceful error handling approach.

PR #3429, opened 2023-04-03
refactor(a11y): KZN-1111 Move package to components

JakePitman:
Moves a11y package to `packages/components/src`, as per [this issue](https://github.com/cultureamp/kaizen-discourse/issues/46)


gyfchong:
Don't remove the files from `packages/a11y` just yet, we need them to apply deprecation notices.
https://github.com/cultureamp/kaizen-design-system/pull/3420/files#diff-ab5e36a50e3591b03747fca4bed7291bee9e1de6fb0d14247a5f0be043276705R43

You can remove the docs though.

packages/components/package.json
```diff
@@ -35,6 +35,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@kaizen/component-base": "^1.1.5",
```
gyfchong, line 38: The `OverrideClassName` type is available in `~types/OverrideClassName`

gyfchong:
It's probably not worth nesting the `SkipLink` and `VisuallyHidden` components inside the `/a11y` dir anymore, so just shift them up a level and we should be good.

Also don't forget to export them in `components/src/index.ts` 😄

gyfchong:
Also, not sure you're allowed unicode characters in the branch name, might be what's blowing storybook deploy up haha

PR #3435, opened 2023-04-03
feat(components): Kzn 1111 the great migration

JakePitman:
Moves a11y package to packages/components/src, as per [this issue](https://github.com/cultureamp/kaizen-discourse/issues/46)


packages/components/package.json
```diff
@@ -35,6 +35,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@kaizen/component-base": "^1.1.5",
```
HeartSquared, line 38: We don't use `component-base` (used for `OverrideClassName`) in AIO. This has been copied in and is aliased, so you'll update the import instead to be from `~types/OverrideClassName` :)

HeartSquared (commented):
Pretty sure we want to copy the files over, not actually move them :)

yarn.lock
```diff
@@ -2430,6 +2430,14 @@
     "@jridgewell/resolve-uri" "3.1.0"
     "@jridgewell/sourcemap-codec" "1.4.14"
 
+"@kaizen/a11y@^1.7.7":
```
HeartSquared, line 2433: We definitely don't want this change either. This probably showed up because the package was considered deleted from the repo, but is used as a dep for some of the other packages.

JakePitman:
Not sure what's causing [this error](https://github.com/cultureamp/kaizen-design-system/actions/runs/4603631240/jobs/8133807794?pr=3435#step:6:12), but I noticed all other packages so far are exporting something with the same name from their index.ts ([here](https://github.com/cultureamp/kaizen-design-system/blob/main/packages/components/src/Filter/index.ts#L1) for example).
Do we need to do that here too?

Also interesting that this passes locally...

gyfchong:
Ah what I meant was that we shouldn’t have an A11y directory at all, these two components should stand on their own at src/ level. The nesting doesn’t make sense anymore as we aren’t organising things by package but as standalone components.

.changeset/orange-monkeys-cheat.md
```diff
@@ -0,0 +1,7 @@
+---
+"@kaizen/components": minor
+"@kaizen/a11y": minor
```
gyfchong, line 3: a11y gets covered by Lerna for now and shouldn’t be added in changesets

gyfchong:
Unfortunately, I think we had to redo this cause there could be changes made to these components before we do the full migration

PR #3436, opened 2023-04-04
refactor(brand): Kzn 1111 Move package to components

JakePitman:
Moves Brand package to packages/components/src, as per https://github.com/cultureamp/kaizen-discourse/issues/46


packages/components/src/Brand/README.mdx
```diff
@@ -0,0 +1,15 @@
+---
```
HeartSquared, line 1: Oh yeah, you can remove these `README.mdx` files :)

packages/components/src/Brand/docs/Brand.stories.tsx
```diff
@@ -0,0 +1,72 @@
+import React from "react"
+import { ComponentStory, Story } from "@storybook/react"
+import { StoryWrapper } from "../../../storybook/components/StoryWrapper"
+import { CATEGORIES } from "../../../storybook/constants"
+import { figmaEmbed } from "../../../storybook/helpers/figmaEmbed"
```
HeartSquared: Drat...just realised these conflict with Storybook 7 :') It's a race!!
Maybe just make sure you don't fix up anything in any stories you copy over so if I copy/paste the fixes later I don't undo anything.

packages/components/src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from "./KaizenProvider"
 export * from "./Filter"
 export * from "./FilterButton"
 export * from "./FilterSelect"
+export * from "./Brand"
```
HeartSquared, line 5: Any chance we could order these alphabetically please 🙏🏻

packages/components/src/Brand/docs/Brand.stories.tsx
```diff
@@ -0,0 +1,72 @@
+import React from "react"
```
HeartSquared, line 1: Almost missed this. For the AIO, we want the docs to go into `_docs/` instead of `docs/` :)

HeartSquared (commented):
One last change, then I think we're good!

packages/components/src/Brand/docs/Brand.stories.tsx
JakePitman: Yep, missed that step

HeartSquared (approved):

HeartSquared:
Oops we forgot to delete the old stories :'D Something to remember for the next one (they fight each other; breaks V7).

PR #3437, opened 2023-04-04
fix: A11y issue when using keyboard with Likert Scale

christian314159:
## Why
- Legend would not update when using keyboard navigation
- See [Kaizen Discourse issue #69](https://github.com/cultureamp/kaizen-discourse/issues/69)

## What
- Added events for `onFocus` and `onBlur`
- Some general cleanup

Fixes cultureamp/kaizen-discourse#69


draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -158,34 +157,34 @@ export const LikertScaleLegacy = ({
...
                 className={classnames(
                   styles.likertItemFill,
                   styles[`field${item.value}`],
                   {
-                    [styles.pop]:
-                      selectedItem && item.value === selectedItem.value,
+                    [styles.pop]: isSelectedItem,
                   }
                 )}
               />
```
gyfchong, line 186: This is a self-closing `div`...? If it's not too much hassle, I think you should open this `div` and place the `Icon` inside it to avoid having to mess with `absolute` CSS

christian314159: I did that at some point and tried to use `flex` to put the icon in the right position, had some weird side effects on the sizing of the likert scale :( This is all very complex and fragile.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.module.scss
```diff
@@ -249,6 +265,47 @@ $fifth: $color-red-500;
...
+  color: black;
+}
+
+:is(.likertItem4.suggested, .likertItem5.suggested) .check {
+  color: white;
+}
+
+.hovered.reversed .itemContainer > :not(.suggested.likertItem) .check {
+  color: white;
+}
```
gyfchong, line 307: Hmm might be cleaner to use JS to conditionally add a `checkmark-white` classname when given X condition.

```
likertId < 6 && .checkmark-white
```

Then you can contain your styles:

```
.checkmark-white .check {}
.checkmark-white:hover .check {}
```

gyfchong: Or go for a `likertItemDark` and `likertItemLight`, applied on the appropriate items. This could also help you reduce the classes you need to reference + scales better if more items were to be added or the colour needs to change, you'll have a better indicator of where to apply the right changes, rather than just numbers

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -41,32 +47,20 @@ export const LikertScaleLegacy = ({
     value: s.value,
     ref: createRef<HTMLDivElement>(),
   }))
+  const notRatedStateItem = scale.find(s => s.value === -1) || null
```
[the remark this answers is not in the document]
christian314159: Yes, the name's not perfect. I was trying to make something clearer that took me a bit to understand:

A scale is an array of objects with a `name` and a `value`. `Strongly disagree` is *1* and `Strongly agree` is *5*... but `Not rated` is also included in that scale as an item with `value` of *-1*.

I thought directing this into a variable would make things clearer (it did so while I was working on the code). But the name's not perfect.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -41,32 +47,20 @@ export const LikertScaleLegacy = ({
...
-    // Remove hover state
+    onSelect(newItem)
     setHoveredItem(null)
   }
 
   const handleRadioMouseEnter = (item: ScaleItem): void => {
-    if (selectedItem && selectedItem.value > 0) {
-      return // Disable hover when there is a selection
-    }
     setHoveredItem(item)
```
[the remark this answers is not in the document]
christian314159: Yep, will do. That was a late change and I missed the opportunity to move it.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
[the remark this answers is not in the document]
christian314159: I think this might be one of those occasions where a code comment is warranted?

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -41,32 +47,20 @@ export const LikertScaleLegacy = ({
...
   }))
+  const notRatedStateItem = scale.find(s => s.value === -1) || null
 
   const handleRadioClick = (item: ScaleItem): void => {
-    const newValue =
-      selectedItem && item.value === selectedItem.value
-        ? scale.find(s => s.value === -1) || null
-        : item
-    onSelect(newValue)
+    const isSameAsCurrentlySelectedItem = selectedItem?.value === item.value
```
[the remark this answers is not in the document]
christian314159: I did this because the previous ternary with fallbacks was too confusing and this makes it easier to follow (IMO).

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.module.scss
christian314159: image
image

Some background, I don't think it's obvious what's going on here:

The check on the last two item (4 + 5) is white, because of their darker background colors. But when you hover on the scale and they potentially have a transparent background, they need to be dark.

But when `reversed` those last two stay white (because the bg is dark) and also the other ones that are usually dark need to be white 🤯

The JS logic for that is also not going to be trivial, but at least testable - which is my biggest gripe with a pure CSS solution.

I'll give it a quick go, but not sure I can finish it today.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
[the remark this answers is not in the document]
gyfchong: Hmm, well if it's supposed to actually return a group of items, then perhaps `unratedItems` is simple enough.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.module.scss
gyfchong: Maybe not worth the stab then, at least we have chromatic.

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.tsx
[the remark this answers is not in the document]
christian314159: Did you see the latest update? I tried to make it as clear as possible.
It's not a group of items, it's just the one that is:
```
const scale = [
{
  name: "Not rated",
  value: -1,
},
{
  name: "Strongly disagree",
  value: 1,
},
...
]
```

draft-packages/likert-scale-legacy/KaizenDraft/LikertScaleLegacy/LikertScaleLegacy.module.scss
christian314159: I was just giving it a go, but the function would need:
- current item's value
- value of hovered item (or null)
- value of selected item (or null)
- and reversed state
From there it would determine the check mark's color.

Just the test cases will take a while to write, unless I can come up with something cleverer than in the past 10 min ;)

PR #3438, opened 2023-04-04
feat(notification): add isSubtle prop to inline notification

jshli:
## Why
Adding the 'subtle' variant to the inline notification, as specified in the [UI kit](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%F0%9F%92%9C-UI-Kit%3A-Heart?node-id=1929-21594&t=FqKRwn2ERtHBUSQr-0)
[Kaizen Discourse](https://github.com/cultureamp/kaizen-discourse/issues/79)

## What
- Add an `isSubtle` prop to the `InlineNotification` and `GenericNotification` components, along with matching styles that remove internal padding, background colour and borders from the component
image

### DIFF
--- a/packages/notification/docs/InlineNotification.stories.tsx
+++ b/packages/notification/docs/InlineNotification.stories.tsx
@@ -97,6 +97,18 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
             <a href="/">View all</a>
           </InlineNotification>
         </StoryWrapper.Row>
+        <StoryWrapper.Row rowTitle="Subtle">
+          <InlineNotification
+            type="informative"
+            isSubtle
+            title="Informative title"
+            hideCloseIcon
+          >
+            "All Employees - North America" status has been changed to
+            'Archived'.
+            <a href="/">View all</a>
+          </InlineNotification>
+        </StoryWrapper.Row>
         <StoryWrapper.Row rowTitle="Positive">
           <InlineNotification type="positive" title="Positive title">
             Emails will be sent notifying coaches and inviting reviewers to give
--- a/packages/notification/docs/InlineNotification.stories.tsx
+++ b/packages/notification/docs/InlineNotification.stories.tsx
@@ -105,8 +105,7 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
             hideCloseIcon
           >
             "All Employees - North America" status has been changed to
-            'Archived'.
-            <a href="/">View all</a>
+            'Archived'. <a href="/">View all</a>
           </InlineNotification>
         </StoryWrapper.Row>
         <StoryWrapper.Row rowTitle="Positive">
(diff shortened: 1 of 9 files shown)


packages/notification/docs/InlineNotification.stories.tsx
```diff
@@ -97,6 +97,18 @@ const StickerSheetTemplate: Story<{ isReversed: boolean }> = ({
...
+        <StoryWrapper.Row rowTitle="Subtle">
+          <InlineNotification
+            type="informative"
+            isSubtle
+            title="Informative title"
+            hideCloseIcon
+          >
+            "All Employees - North America" status has been changed to
+            'Archived'.
+            <a href="/">View all</a>
```
HeartSquared, line 109: It's `children`, but the story looks like a space is missing so probably better to just add it 🤷🏻 
```suggestion
            "All Employees - North America" status has been changed to
            'Archived'. <a href="/">View all</a>
```

packages/notification/src/components/GenericNotification.tsx
```diff
@@ -30,6 +30,7 @@ type Props = {
   noBottomMargin?: boolean
   forceMultiline?: boolean
   headingProps?: HeadingProps
+  isSubtle?: boolean
```
HeartSquared, line 33: Since we want it confined, I'd actually propose we add `classNameOverride` here and add the class to the InlineNotification instead.

If we later find that we want it for the other notifications, we will move it then.

packages/notification/src/types.ts
```diff
@@ -29,3 +29,6 @@ export type AddToastNotification = (
 export type RemoveToastNotification = (notificationId: string) => void
 
 export type ClearToastNotifications = () => void
+
+export type OverrideClassName<T extends Record<string, any>> =
```
HeartSquared, line 33: You'll want to import this from `@kaizen/component-base` instead :)

packages/notification/src/components/GenericNotification.tsx
```diff
@@ -17,7 +18,8 @@ export type NotificationType =
   | "cautionary"
   | "negative"
 
-type Props = {
+interface GenericNotificationProps
```
HeartSquared, line 21: Nice! Given we're adding the attributes though, you'll want to spread them into the div :)

packages/notification/src/InlineNotification.tsx
```diff
@@ -3,6 +3,7 @@ import { HeadingProps } from "@kaizen/typography"
 import GenericNotification, {
   NotificationType,
 } from "./components/GenericNotification"
+import styles from "./components/GenericNotification.module.scss"
```
HeartSquared, line 6: You'll actually want to create the class in a `InlineNotification.module.scss` file since it's self-contained :)

HeartSquared (approved):

PR #3442, opened 2023-04-05
feat(components): Kzn 1111/loading skeleton migrate

JakePitman:
Moves LoadingSkeleton package's components to packages/components/src, as per [this issue](https://github.com/cultureamp/kaizen-discourse/issues/46)


packages/components/src/LoadingGraphic/loading-skeleton.module.scss
JakePitman, line 1: Would be good to establish a pattern for shared resources between components. Currently have three of these loading-skeleton scss files

PR #3443, opened 2023-04-05
feat: KZN-1056 remove lerna stack

gyfchong:
https://cultureamp.atlassian.net/browse/KZN-1056

Complete transition from "lerna stack" (lerna + commitizen + conventional commits/changelog) to changesets


.buildkite/scripts/release.sh
```diff
@@ -43,29 +43,18 @@ release() {
   git checkout main && git pull
 
   yarn install --frozen-lockfile
-
-  # Certain packages rely on these two packages to be built so they can access dist files
-  yarn workspace @kaizen/design-tokens prepublish
-  yarn workspace @kaizen/tailwind prepublish
```
gyfchong, line 49: Not entirely sure why I don't need this for changesets...

.buildkite/scripts/release.sh
```diff
@@ -43,29 +43,18 @@ release() {
...
+  yarn ci:publish
 }
 
 release_canary() {
   git checkout canary && git pull
 
   yarn install --frozen-lockfile
-
-  yarn lerna publish --canary --preid canary --yes
+  yarn ci:publish --tag "canary"
```
gyfchong, line 53: Untested for now, but should yield the same results.

package.json
```diff
@@ -21,10 +21,10 @@
     "lint": "yarn eslint --max-warnings=0 && yarn prettier && yarn stylelint",
     "lint:fix": "yarn stylelint --fix && yarn eslint --max-warnings=0 --fix && yarn prettier --write",
     "clean:storybook": "rimraf 'storybook/public'",
-    "clean": "lerna run clean && yarn clean:storybook",
+    "clean": "yarn clean:storybook",
```
HeartSquared, line 24: We should actually be looking through all the repos and cleaning out their dist files with this command 😬

package.json
```diff
@@ -134,16 +131,9 @@
     "webpack": "^5.77.0"
   },
   "resolutions": {
-    "@types/react": "^18.0.32",
-    "commitizen": "^4.3.0"
```
HeartSquared, line 138: We don't want this for anyone who wants to use it? :') I guess it's only me who would miss it (although I don't need it haha).

HeartSquared (commented):
We need to ensure we test that updates within any package that is a dep of another also get updated (eg. if we update `@kaizen/button` all the consuming packages get their `package.json` updated).

package.json
gyfchong: Haha, time to rely on our team's natural discipline

package.json
gyfchong: Ah right, Lerna did that didn't it... But I can only think of introducing a task runner nx/turbo to execute the same thing

package.json
```diff
@@ -21,10 +21,10 @@
     "lint": "yarn eslint --max-warnings=0 && yarn prettier && yarn stylelint",
     "lint:fix": "yarn stylelint --fix && yarn eslint --max-warnings=0 --fix && yarn prettier --write",
     "clean:storybook": "rimraf 'storybook/public'",
-    "clean": "lerna run clean && yarn clean:storybook",
+    "clean": "yarn turbo clean && yarn clean:storybook",
```
gyfchong, line 24: Using turbo to go through all directories and execute their internal clean functions, this is to support the different clean configs in different packages.

packages/select/src/FilterMultiSelect/components/ListBox/ListBox.tsx
```diff
@@ -65,7 +65,7 @@ export const ListBox = ({ children }: ListBoxProps): JSX.Element => {
   if (hasNoItems) {
     return (
       <>
-        <div>{children(itemsState)}</div>
+        <div>{!itemsState && children(itemsState)}</div>
```
gyfchong, line 68: Not entirely sure why, but typescript started throwing errors for various parts of `FilterMultieSelect`

gyfchong:
This has broken types for some reason, moved to: https://github.com/cultureamp/kaizen-design-system/pull/3480

PR #3448, opened 2023-04-06
docs: inter font for storybook

gyfchong:
## Why
Designers wanted it

## What
Styles added to preview body to override all other styles.


storybook/preview-body.html
```diff
@@ -0,0 +1,5 @@
+<style>
```
HeartSquared, line 1: Isn't this something we should be able to put into the theme file? 🤔

gyfchong: So apparently to affect the styles within the preview, you have to add them here. The theme file is only for everything outside the preview window -- cause that makes _total_ sense!

mcwinter07: Oh Storybook, you make perfect sense until you don't 😆

mcwinter07: That said I think the main reason they don't make it intuitive to do this is to make most Storybook out in the wild look the same #stickyBranding

mcwinter07 (approved):

gyfchong:
Unfortunately we can't move forward with this since it impacts the component's demo styles as well. I have attempted to make a wrapping `Content` component to try isolate the styles, but needless to say it would be tedious to have to apply a wrapper around every content block we want to create.

Going to close this for now while we decide where some of this content is going to end up living. Maybe we design our way out of needing to update the fonts on storybook.

PR #3481, opened 2023-04-11
docs: Update tech docs with Storybook v7

JakePitman:
## Why
For a better docs experience, inline with the new Storybook 7.0 release

## What
Adds autodocs for the following components:

- Brand
- LoadingGraphic
- LoadingHeading
- LoadingInput
- LoadingParagraph
- LoadingSpinner
- Pagination
- GlobalNotification
- InlineNotification

### DIFF
--- a/packages/components/src/Brand/_docs/Brand.stories.tsx
+++ b/packages/components/src/Brand/_docs/Brand.stories.tsx
@@ -1,8 +1,8 @@
 import React from "react"
-import { Brand } from "../index"
-import { Meta, StoryFn, StoryObj } from "@storybook/react"
+import { StoryFn, StoryObj } from "@storybook/react"
 import { ComponentDocsTemplate } from "../../../../../storybook/components/DocsContainer"
 import { StickerSheet } from "../../../../../storybook/components/StickerSheet/StickerSheet"
+import { Brand } from "../index"
 
 const meta = {
   tags: ["autodocs"],
--- a/packages/components/src/Brand/_docs/Brand.stories.tsx
+++ b/packages/components/src/Brand/_docs/Brand.stories.tsx
@@ -1,0 +1,89 @@
+import React from "react"
+import { Brand } from "../index"
+import { Meta, StoryFn, StoryObj } from "@storybook/react"
+import { ComponentDocsTemplate } from "../../../../../storybook/components/DocsContainer"
+import { StickerSheet } from "../../../../../storybook/components/StickerSheet/StickerSheet"
+
+const meta = {
+  tags: ["autodocs"],
+  title: "Components/Brand",
+  component: Brand,
+  parameters: {
+    chromatic: { disable: false },
+    docs: {
+      container: ComponentDocsTemplate,
+    },
+    installation: [
+      "npm install @kaizen/Brand",
+      "import { Brand } from `@kaizen/Brand`",
+    ],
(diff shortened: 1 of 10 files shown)


packages/components/src/Brand/_docs/Brand.stories.tsx
```diff
@@ -10,12 +11,18 @@ export default {
...
+    ],
+    packageName: "@kaizen/Brand",
+    sourceCodeLink:
+      "https://github.com/cultureamp/kaizen-design-system/tree/main/packages/brand",
+    figmaLink: "Add Figma link here",
   },
-}
+} as Meta<typeof Brand>
 
 export const DefaultStory: StoryFn<typeof Brand> = args => <Brand {...args} />
```
gyfchong, line 27: We don't need this one anymore, just make a note for what is default in the Variants description.

gyfchong:
Just needs the playground code too: https://github.com/cultureamp/kaizen-design-system/blob/main/packages/button/docs/Button.stories.tsx#L45

packages/components/src/Brand/_docs/Brand.stories.tsx
JakePitman: Sweet. This one doesn't have a default as far as I can tell (you always have to pass a `variant` prop). Do I leave it blank in cases like this?

gyfchong: Yeah, don't mention a default if there is none 😄

packages/components/src/Brand/_docs/Brand.stories.tsx
```diff
@@ -25,42 +48,42 @@ DefaultStory.args = {
...
-        />
-      </div>
-    </StoryWrapper.Row>
-  </StoryWrapper>
+/**
+ * `logo-horizontal`, `logo-vertical`, `enso`, `collective-intelligence`
+ */
+export const Variants: StoryFn = () => (
+  <StickerSheet>
+    <StickerSheet.Row rowTitle="Logo Horizontal">
```
gyfchong, line 56: I think make it a single row if possible

JakePitman: Can't seem to get the `'collective-intelligence'` variant to render this way... Even if I give the first three their own row collective intelligence disappears 🤔 
image

gyfchong: That's super weird... Maybe separate collective intelligence to it's own heading, cause it probably needs a little sentence for why it's used.

gyfchong:
Just having a look through the stories, sorry if I maybe wasn't clear enough at the start (probably should've paired) but you need to split up/make stories for essentially each of the props, similar to [Tooltip](https://dev.cultureamp.design/geoff/KZN-1125-update-docs/?path=/docs/components-tooltip--docs), you also need to create a [separate file for stickersheets](https://github.com/cultureamp/kaizen-design-system/blob/3eca87a72f28f67b114103f50c5f6d67c2569cce/draft-packages/tooltip/docs/Tooltip.stickersheet.stories.tsx) which get placed in a separate `/stickersheets` directory.

Let me know if you want to pair on this a bit

gyfchong:
Sorry, we'll have to do this again later once the packages are re-added to KAIO, the conflicts will be huge.

PR #3482, opened 2023-04-11
Restrict publishing of storybook

gyfchong:
## Why
Currently every merge to main will trigger a deploy to storybook, but since Changeset will prevent packages from publishing until the ChangesetPR is merged, Storybook will be updated but the packages won't.

We want to only deploy storybook once packages are released to ensure our docs match the features at all times, but also have the option of publishing storybook right away (for visual or information changes to docs)

## What

Added an if condition to deploying storybook which checks that the build author is the github action bot (which manages the changesets) and then triggers the publish, or if the "force publish storybook" label is present.


gyfchong:
Hmm this might be an issue if we want to release only docs...

.buildkite/pipeline.yaml
```diff
@@ -45,12 +45,12 @@ steps:
...
   - wait
 
-  - name: ":seedling: Publish: ${KAIZEN_DOMAIN_NAME}"
+  - name: ":books: Publish Storybook: ${KAIZEN_DOMAIN_NAME}"
     key: "publish-step"
     command: ".buildkite/scripts/publish.sh"
     timeout_in_minutes: 15
     # The canary branch is not able to update dev.cultureamp.design
-    branches: "!canary"
+    if: build.branch == "main" && (build.author.name != "github-actions[bot]" || build.pull_request.labels includes "force publish storybook")
```
HeartSquared, line 53: This isn't a GH action though? I don't think it has access to these?

gyfchong: Yeah Buildkite actually gets some of the GH PR details
https://buildkite.com/docs/pipelines/conditionals

HeartSquared: Ohhh snazzy.

I don't think your labels condition works though. The `main` branch isn't a PR 😅

gyfchong: Oh good point...

gyfchong:
Solution will change once we move our release and publishing pipeline into Github Actions rather than having half in GH and the other in BK.

PR #3483, opened 2023-04-12
Remove-lerna-config

gyfchong:
## Why
Missed this in the Lerna purge

## What
Removed the config file and gitignore references to lerna.

Looked for an equal option to replicate Lerna config in changesets, but you can only ignore packages. I think Changesets looks at all workspaces by default, so it's not necessary to include, but not entirely sure about the tests and docs files.


mcwinter07 (approved):
love the description in the why 😆

PR #3488, opened 2023-04-12
feat(filterMultiselect): Allow defaultSelectedKeys prop from react aria

NatalieLoyzaga:
## Why

Currently working within performance-insights repo where they are trying to force the initial selectedKeys value to be all via a useEffect. Ideally this fix will mean we do not need this anymore.

Allow `defaultSelectedKeys` prop to be passed in as it is already an exisiting prop from React Aria.

## What

- Add `defaultSelectedKeys` props
- Update the Demographic Filter story to use the defaultSelectedKeys as an example which is reflective of the use case.


packages/select/src/FilterMultiSelect/provider/SelectionProvider/SelectionProvider.tsx
```diff
@@ -12,7 +12,10 @@ export interface SelectionProviderProps {
   children: React.ReactNode // control how menu should look like
   items?: ItemType[]
   onSelectionChange?: (keys: Selection) => void
+  /** The currently selected keys in the collection (controlled). */
   selectedKeys?: Selection
+  /** The initial selected keys in the collection (uncontrolled). */
```
gyfchong, line 17: Ooo I dunno, none of our other components support being uncontrolled 🤔

NatalieLoyzaga: Yeah I was wondering this - Is there a reason that its not recommended??

NatalieLoyzaga: Technically our Select and filterSelect allow `defaultSelectedKey` because we are extending the props :P

gyfchong: I think @HeartSquared mentioned that no one wanted to use our components uncontrolled, so it never got built? 🤷‍♂️

In any case, seems we inadvertently support it with React Aria components, so why not. Filters = uncontrolled lol

gyfchong:
You'll need a changeset nowadays 😄 
https://github.com/cultureamp/kaizen-design-system/blob/main/CONTRIBUTING.md#creating-a-changeset

NatalieLoyzaga:
@gyfchong Can you check I did the change set thing correctly?

.changeset/hungry-bees-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/select": minor
+---
+
+Include defaultSelectedkeys prop from React-aria on FilterMultiSelect
```
gyfchong, line 5: Maybe add a little extra description on what the prop does

PR #3507, opened 2023-04-17
update setup-node to v3

gyfchong:
## Why
Github has been throwing warnings saying that Node 12 is deprecated and to update the `setup-node` action as it still used Node 12 as of `v2`

## What
Update the `setup-node` action to `v3`


HeartSquared (commented):
Chromatic and Danger need to be updated too :)

HeartSquared (approved):

PR #3511, opened 2023-04-19
Fix-changset-build-release

gyfchong:
Will manually re-run the gh action

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -24,14 +24,13 @@
     "build": "yarn build:clean && yarn prepublish",
     "build:clean": "rimraf 'dist'",
     "build:components": "rollup -c",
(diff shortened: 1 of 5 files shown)


packages/components/package.json
```diff
@@ -24,6 +24,7 @@
     "build": "yarn build:clean && yarn prepublish",
     "build:clean": "rimraf 'dist'",
     "build:components": "rollup -c",
+    "clean": "yarn build:clean && rimraf '.turbo'",
```
HeartSquared, line 27: Are these changes here intentional?

mcwinter07 (approved):

packages/components/package.json
gyfchong: Yeah, I wanted to sneak in some QoL 😄

PR #3514, opened 2023-04-19
Run chromatic when a PR is updated with changes from main

gyfchong:
## Why
We want chromatic to automatically run again when a branch is updated with code from `main`.

## What

Removal of branch restriction on push event so that the job will detect pushes on all branches.

### DIFF
--- a/.github/workflows/chromatic.yaml
+++ b/.github/workflows/chromatic.yaml
@@ -5,7 +5,7 @@ name: "chromatic"
 
 on:
   pull_request:
-    types: [labeled, synchronize]
+    types: [labeled]
     branches-ignore:
       - changeset-release/main
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yaml
```diff
@@ -5,12 +5,10 @@ name: "chromatic"
 
 on:
   pull_request:
-    types: [labeled, synchronize]
+    types: [labeled]
     branches-ignore:
       - changeset-release/main
-
   push:
```
HeartSquared, line 11: I believe we wanted to have an exemption for `canary` and the changeset branches didn't we?

gyfchong: The conditions should prevent that, `canary` doesn't sync `main` and neither does `changeset`

HeartSquared: I suppose if someone adds `run chromatic` on those branches they messed up anyway...

HeartSquared (approved):

PR #3523, opened 2023-04-23
GenericModal - Add support for server-side hydration

lisatassone:
## Why
Tried to use ContextModal in the performance-review-cycles-ui project that uses Next.js. It encountered errors when accessing `document`
image

As `ContextModal` is built off `GenericModal,` I have added a client-only wrapper for `GenericModal` that renders after mount, which ensures its on the client at that point. This should allow the modals to be used in Next.js apps. I couldn't directly test this in the project as linking the package caused issues with `Illustration` and `lottie` dependencies.

If there's some other global way Kaizen approaches this, happy to be guided.

### DIFF
--- a/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
+++ b/draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
@@ -289,4 +289,28 @@ const createAriaHider = (dialogNode: HTMLElement): (() => void) => {
   }
 }
 
-export default GenericModalContainer
+const ClientOnly = ({
+  children,
+}: {
+  readonly children: React.ReactNode
+}): JSX.Element => {
+  const [hasMounted, setHasMounted] = React.useState(false)
+
+  React.useEffect(() => {
+    setHasMounted(true)
+  }, [])
+
+  if (!hasMounted) {
+    return <></>
+  }
+
+  return <>{children}</>
+}
+
+const ClientModal = (args: GenericModalContainerProps): JSX.Element => (
+  <ClientOnly>
+    <GenericModalContainer {...args} />
+  </ClientOnly>
+)
+
+export default ClientModal
--- a/.changeset/sixty-rivers-care.md
+++ b/.changeset/sixty-rivers-care.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/draft-modal": patch
+---
+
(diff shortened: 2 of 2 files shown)


draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
```diff
@@ -289,4 +289,28 @@ const createAriaHider = (dialogNode: HTMLElement): (() => void) => {
   }
 }
 
-export default GenericModalContainer
+const ClientOnly = ({
```
christian314159, line 292: Would you see this wrapper being used with more components in the future? If so, it might be worth to at least extract it out to its own file (even if it stays in the same package for now).

Would you be able to add a test for this new functionality as well, please.

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
```diff
@@ -289,4 +289,28 @@ const createAriaHider = (dialogNode: HTMLElement): (() => void) => {
...
+  children,
+}: {
+  readonly children: React.ReactNode
+}): JSX.Element => {
+  const [hasMounted, setHasMounted] = React.useState(false)
+
+  React.useEffect(() => {
```
ckychris, line 299: Can we please avoid using side effects to deal with server-side rendering and client-side rendering hydration errors?

Next.js recommends using https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr. This simple solution can be used on any component. 

If there are reusability concerns, can we please create a util function like this

```
export const isSSR = () => typeof window === "undefined"
```

and use it directly without using `useEffect`

```
return <>{!isSSR && children}</>
```

ckychris (changes requested):

ckychris:
This is a common error when developing with Next.js. I can further explain the why, what and how in [this doc](https://cultureamp.atlassian.net/wiki/spaces/S2/pages/3116401178/Migration+map+from+perf-ui+to+review-cycles-ui) that I keep track of everything for now.

draft-packages/modal/KaizenDraft/Modal/Primitives/GenericModal.tsx
lisatassone: Oh I forgot about dynamic imports. That's all that's needed for now. I'll close this PR

lisatassone:
Will use dynamic imports instead.

PR #3524, opened 2023-04-23
Icon title now rendered as aria-label instead of title tag for improved A11y

christian314159:
## Why
- Accessibility audit identified issue with the `Icon` component where `title` is not always read out, suggestion is to use `aria-label` instead

Fixes cultureamp/kaizen-discourse#74

## What
- Rendering `title` prop in `arial-label` attribute
- Cleaned up specs & component a little


packages/component-library/components/Icon/Icon.spec.tsx
```diff
@@ -9,13 +9,13 @@ const svgIcon = {
 
 describe("<Icon />", () => {
   describe("presentational", () => {
-    it("does not render a title", () => {
+    it("does not render an arial label", () => {
```
dougmacknz, line 12:
```suggestion
    it("does not render an aria label", () => {
```

dougmacknz (approved):
Nice! Tiny typo on the tests if you can be bothered

packages/component-library/components/Icon/Icon.spec.tsx
christian314159: Haha, that keeps on happening to me :)

dougmacknz (approved):

PR #3525, opened 2023-04-24
fix: RTL Table styling fixes

kaydbwithers:
## Why
This PR adds styling fixes to the Table when RTL dir is active.

## What
These changes should not affect anything in LTR, they are just fixes for RTL.

**Before**
LTR
image

RTL
In RTL the tooltip icon appeared too close to the text.
image

In RTL when a `TableHeaderRowCell` had an alignment of `center` or `end` then the sorting chevron did not appear correctly.
image

**After**
LTR
image

RTL
image
image


draft-packages/table/KaizenDraft/Table/Table.module.scss
```diff
@@ -101,7 +101,13 @@ $row-height-data-variant: 48px;
 
 .headerRowCellTooltipIcon {
   color: $color-blue-500;
-  margin-right: $spacing-xs;
+
+  [dir="ltr"] & {
+    margin-right: $spacing-xs;
```
christian314159, line 106: Btw, as we're not supporting IE11 anymore, you should be able to use `margin-inline/block` here instead 😃

kaydbwithers: Thanks @christian314159, I've updated it.

.changeset/smooth-dogs-cheat.md
```diff
@@ -2,4 +2,4 @@
 "@kaizen/draft-table": patch
 ---
 
-Align the tooltip to the center of the Table header text
+* Align the tooltip to the center of the Table header text
```
gyfchong, line 5: Different PR?

kaydbwithers: Okay, it's gone now.

christian314159:
Hey @kaydbwithers I just had a look at the branch Storybook and found this small issue where the tickbox in the header sits too close to the description:
image
I checked and this is an existing issue, not caused by your PR, but I wonder if we could incorporate a fix as well while we're touching the table headings? Feel free to push back.

kaydbwithers:
@christian314159 sure, I've also updated the checkbox margin.

christian314159 (approved):
🚢 Thanks for contributing!

PR #3526, opened 2023-04-24
Add pause/play button to animations [KZN-1157]

christian314159:
## Why
- See [Accessibility Audit issue](https://github.com/cultureamp/kaizen-discourse/issues/70) here
- Animations need a pause button to stop them from being distracting to some people
Fixes cultureamp/kaizen-discourse#70

## What
- WIP: Added rudimentary pause button


draft-packages/illustration/KaizenDraft/Illustration/Players/VideoPlayer.spec.tsx
```diff
@@ -155,4 +156,24 @@ describe("<VideoPlayer />", () => {
...
+      const { container, getByRole } = render(
+        <VideoPlayer
+          fallback="illustrations/heart/spot/moods-cautionary.svg"
+          source="illustrations/heart/spot/moods-cautionary.webm"
+        />
+      )
+
+      const videoElement = container.querySelector("video")!
+
+      expect(videoElement.paused).toEqual(true)
```
christian314159, line 171: All HTMLVideoElement prototype functions have been mocked out by the previous test, so I can't really play a video in here - not sure it would even work in a Jest test.
There's a proper test for the hook though and this at least tests the overall wiring.

draft-packages/illustration/KaizenDraft/Illustration/Base.module.scss
```diff
@@ -1,8 +1,40 @@
...
+  position: relative;
+
+  // nested to get more specificity, beat out the generic button styles
+  .pausePlayButton {
+    position: absolute;
+    right: 1rem;
+    bottom: 1rem;
+    /* stylelint-disable declaration-no-important */
+    background-color: $color-white !important;
+    border: 1px solid $color-gray-400;
```
dougmacknz, line 20: This border is a 1.5:1 ratio against white. We can try to get away with it if we want, I'm a bit hazy about whether the icon by itself is enough here.

dougmacknz (approved):
Nice. One little concern about the IconButton's border contrast ratio, but we could follow up later if we need.

dougmacknz:
We could also try to get away with only showing the button on hover of the video container if we wanted.

jmitsios:
> We could also try to get away with only showing the button on hover of the video container if we wanted.

I think this was ultimately what I was hoping to do, but just had a strong assumption that displaying content only on hover was a danger zone?   I wouldn't mind learning the Do's and Dont's for that kind of interaction.

If we can, then it would be aesthetically much better.

draft-packages/illustration/KaizenDraft/Illustration/Base.module.scss
```diff
@@ -21,6 +21,10 @@
     border: 1px solid $color-gray-400;
     transition: all $animation-duration-immediate;
 
+    @media (hover: none) and (pointer: coarse) {
```
gyfchong, line 24: `coarse` is a new one to me, nice find!

draft-packages/illustration/KaizenDraft/Illustration/Base.module.scss
```diff
@@ -1,8 +1,51 @@
...
+    &:hover {
+      background-color: $color-gray-200 !important;
+
+      svg {
+        opacity: 100%;
+      }
+    }
+  }
+
+  &:hover {
```
dougmacknz, line 42: Should just be able to do this I think

```suggestion
  &:hover,
  &:focus {
```

dougmacknz (approved):

PR #3527, opened 2023-04-24
fix: Table tooltip appears centered above header text

kaydbwithers:
## Why
This PR fixes Table styling to allow the tooltip to appear correctly above Table header text.

## What
Before
image
image

After
image
image


draft-packages/table/KaizenDraft/Table/Table.module.scss
```diff
@@ -72,8 +68,6 @@ $row-height-data-variant: 48px;
 // The .headerRowCell wrapper is required to increase the specificity, as so
 // we can override the `Tooltip` component's styling.
 .headerRowCell .headerRowCellTooltip {
```
gyfchong, line 70: Have you had a look at changing the [Tooltip display prop](https://cultureamp.design/storybook/?path=/docs/components-tooltip--docs)? Might remove the need to apply any extra styling.

kaydbwithers: Thanks @gyfchong, I tried but it doesn't seem to help. The problem is with the header row expanding the full width.

gyfchong (approved):

.changeset/smooth-dogs-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/draft-table": minor
```
gyfchong, line 2: I'd say this is a `patch` since no new feature was added.

kaydbwithers: Thanks @gyfchong, that's done.

gyfchong (approved):

gyfchong (approved):

PR #3529, opened 2023-04-26
chore(plop): fix stories template imports

HeartSquared:
## Why

Forgot to update the import after renaming the template component.

## What

Fix docs template import in plop template for stories.


gyfchong (approved):
Oops! Will remember to update this in the future.

PR #3530, opened 2023-04-26
update select deps and add type fixes

mcwinter07:
## Why
Bundling all the @kaizen/select deps upgrades along with type fixes Cass and I implements into one PR since the previous [PR](https://github.com/cultureamp/kaizen-design-system/pull/3476) for renovate was getting messy / grew in scope.

## What
- upgrade all of the Select's aria, stately, and shared types packages
- adds narrower types to use in place of the generic Node type that could be nullable (which causes type issues)
  - This is further exacerbated (and required) by the function `collection.getItem()` now having a nullish return value

Old PR's for reference:
Select:
https://github.com/cultureamp/kaizen-design-system/pull/3476 
https://github.com/cultureamp/kaizen-design-system/pull/3477 

Filter Multi-Select:
https://github.com/cultureamp/kaizen-design-system/actions/runs/4646068795/jobs/8222275479?pr=3474


mcwinter07:
Will close off the old PRs for renovate that relate to this and rebase it to main once this merged 👍

packages/select/src/FilterMultiSelect/components/ListBox/ListBox.tsx
```diff
@@ -1,18 +1,34 @@
...
-    disabledItems: Array<Node<ItemType>>
-    allItems: Array<Node<ItemType>>
-    hasNoItems: boolean
-  }) => React.ReactNode
+  children: (items: ListBoxItems) => React.ReactNode
+}
+
+/** get items from keys and filter out nullish or undefinded values
+ ** this is required now `collection.getItem()` has a nullish return value
+ */
```
HeartSquared, line 22: No need for this comment :) It's enough to explain it in the PR desc, but not needed in the codebase since the function is pretty self-explanatory.

.changeset/empty-cheetahs-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/select": minor
+---
+
+Updating the `@kaizen/select` dependancies (specifically react-aria, shared-types and react-stately) caused a number of type errors that needed to be fixed. This related to the Node<T> generic that was being pulled in from shared-types that allowed a return value for null while our component filters out null values.
```
HeartSquared, line 5:
```suggestion
Updating the `@kaizen/select` dependencies (specifically react-aria, shared-types and react-stately) caused a number of type errors that needed to be fixed. This related to the Node<T> generic that was being pulled in from shared-types that allowed a return value for null while our component filters out null values.
```

Although reading this, I'd probably expect the changelog to tell me something more along the lines of what was changed rather than the problem we came across.

Something like:
```
* Upgrade `react-aria`, `shared-types`, `react-stately`
  * Fix type errors where value can now be `null`
* Add type `MultiSelectItem` and replaced usage of `Node<T>`
```

mcwinter07: Fair point, I suppose anyone looking at this could refer to the PR to get the context if they really needed it

HeartSquared (approved):

PR #3531, opened 2023-04-26
feat: set Table Tooltip animationDuration to 0

kaydbwithers:
## Why
This PR sets animationDuration to 0 which makes the tooltip disappear more quickly. 

## What
An example of the table tooltip. Previously it stayed around for a bit of time which wasn't a good user experience. Now it will disappear as soon as the user has left the tooltip hover area. 
image


gyfchong:
Don't forget your changeset, read our [contribution guidelines](https://github.com/cultureamp/kaizen-design-system/blob/main/CONTRIBUTING.md#creating-a-changeset) for more detail

.changeset/red-parrots-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/draft-table": minor
```
gyfchong, line 2: This is a `patch` as it doesn't introduce any new features 😄

kaydbwithers: Thanks @gyfchong, that has been updated.

gyfchong (approved):

PR #3535, opened 2023-04-27
fix changelog config to hopefully update releases

gyfchong:
## Why
Packages versions and changelogs aren't being updated in github.

## What
Update the changeset config to point changelogs to our repo.


HeartSquared (approved):
🤞🏻

PR #3539, opened 2023-04-27
fix: remove negative margin hover styles for Table

kaydbwithers:
## Why
- This PR addresses an issue in reports-components where a sorting chevron did not appear correctly due to the negative margin.
- I've also added a Storybook example of `sortingArrowsOnHover`.

## What
I did some experimenting and it appears that we longer need the hover fix for left alignment due to a recent fix.

I experimented with the sortingArrowsOnHover and the label was no longer affected when left aligned, so I have removed this code.


draft-packages/table/docs/Table.stories.tsx
```diff
@@ -463,10 +467,7 @@ export const Tooltip: StoryFn = () => (
             wrapping="wrap"
             tooltipInfo="This is a tooltip"
             sorting="ascending"
-            onClick={(e: React.MouseEvent): void => {
-              e.preventDefault()
```
gyfchong, line 467: Why are you not preventing default anymore?

kaydbwithers: This was a mistake by me in the previous update, I copied the code from the anchor click example. I've now used the sorting click example code instead.

gyfchong (approved):

PR #3540, opened 2023-04-27
chore(eslint): add react rules

HeartSquared:
## Why

We kept forgetting to put `type` on our `<button>`s, which leads to accidental form submissions for our consumers.

## What

- Add eslint rule `react/button-has-type` and fix issues
- Extra rules added as a side-effect of enabling `react/button-has-type`:
	- `react/no-unescaped-entities` (fixed)
	- `react/jsx-key` (fixed)
	- `react/no-children-prop` (fixed)
	- `react/jsx-no-target-blank` (fixed)
	- `react/display-name` (fixed)
	- `react/prop-types` (disabled)
- Loosened `@typescript-eslint/explicit-function-return-type` to allow expressions to not have to declare the return type
	- eg. `onClick={(): void => doSomething()}` can now be `onClick={() => doSomething()}`

### DIFF
--- a/draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
+++ b/draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
@@ -52,7 +52,7 @@ const SingleCollapsibleNoPadding = (): JSX.Element => (
     </ListItem>
     <ListItem>
       <Paragraph variant="body">
-        In that case you should use the 'noSectionPadding' prop.
+        In that case you should use the &#39;noSectionPadding&#39; prop.
       </Paragraph>
     </ListItem>
   </Collapsible>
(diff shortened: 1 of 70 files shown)


draft-packages/collapsible/docs/SingleCollapsible.stories.tsx
```diff
@@ -52,7 +52,7 @@ const SingleCollapsibleNoPadding = (): JSX.Element => (
     </ListItem>
     <ListItem>
       <Paragraph variant="body">
-        In that case you should use the 'noSectionPadding' prop.
+        In that case you should use the &#39;noSectionPadding&#39; prop.
```
christian314159, line 55: Is this due to the `no-escaped-entities` rule? I am unsure if this is the best developer experience having to write them like this?
How necessary is it as well? I'd expect 🤞 we're all using the same charset so there shouldn't be any problems?

HeartSquared: As discussed with @gyfchong, I have changed them to `&quot;` and `&apos;` :)

christian314159 (approved):

PR #3542, opened 2023-04-28
POC: Filters / FilterBar

HeartSquared:


packages/components/src/FilterBar/_docs/FilterBar.stories.tsx
```diff
@@ -91,105 +95,87 @@ export const Playground: StoryFn<typeof FilterBar> = () => {
...
+    {
+      id: "coffee",
+      label: "Coffee",
+      Component: <FilterPancake id="coffee" />,
+    },
+    {
+      id: "carrots",
+      label: "Carrots",
+      Component: <FilterPancake id="carrots" />,
+      isInitHidden: true,
```
gyfchong, line 149: Perhaps we can operate under the assumption that all stateful props here are the "initial" ones, so we don't end up with `initialValue` `isInitHidden` etc.

Although, the `Component` could be responsible for `defaultValue` as part of their standard "form-like" API.

PR #3557, opened 2023-05-02
docs(filter-drp): update stickersheets to all have dates in the past

HeartSquared:
## What

Update FilterDateRangePicker sticker sheet examples to use dates in the past.

## Why

"Today" has a different style, so the snapshots will change every day if the date is within the shown calendars.


gyfchong (approved):
Swear I made a ticket for this, but can't find it. Oh well!

PR #3559, opened 2023-05-02
Upgraded Storybook to latest

christian314159:
## Why
- New versions are released, old ones have bugs

## What
- Upgrading versions


gyfchong (approved):

storybook/main.ts
```diff
@@ -42,11 +49,12 @@ const config = {
...
         const isHTMLElementProp =
           prop.parent?.fileName.includes("node_modules/@types/react") ?? false
-
         return !isHTMLElementProp
       },
     },
   },
+  docs: {
+    autodocs: true,
+  },
```
HeartSquared, line 58: We don't want this :)

HeartSquared (changes requested):

storybook/main.ts
```diff
@@ -20,16 +20,23 @@ const getStoryPathsFromEnv = (): string[] | false => {
...
 ]
-
 const config = {
   stories: getStoryPathsFromEnv() || defaultStoryPaths,
-  addons: ["@storybook/addon-essentials", "@storybook/addon-a11y"],
-  staticDirs: [{ from: "./assets", to: "/static/media" }],
+  addons: [
+    "@storybook/addon-essentials",
+    "@storybook/addon-a11y",
+    "@storybook/addon-mdx-gfm",
```
HeartSquared, line 32: Pretty sure we don't need this either

christian314159: I assume you specifically mean the `@storybook/addong-mdx-gfm`? I think it was added during the update flow, but I can remove it again (I hope w/o problems ;) )

HeartSquared: Yeah we don't need this plugin added by the migration :)

storybook/main.ts
HeartSquared: I wonder if we should be explicit and put `autodocs: "tag"` (or whatever it should be) in to stop the auto migrate prompt next time 🤔

HeartSquared: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#autodocs-changes

>However, as of 7.0, the new default is 'tag'

christian314159: Ok, I'm then just going to remove the setting, as it will default to `tag`.

HeartSquared (approved):

PR #3560, opened 2023-05-02
Allow custom heading props on MenuList

dougmacknz:
## Why
Closes https://github.com/cultureamp/kaizen-discourse/issues/75

## What
Adds a new prop `headingProps` to the `MenuList` component, allowing people to adjust the props of the heading. Pattern copied from what we did with EmptyState and InlineNotification last year.


draft-packages/menu/KaizenDraft/Menu/components/MenuList.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+      expect(
+        screen.getByRole("list", { name: "A menu section" })
+      ).toBeInTheDocument()
+    })
+
+    it("renders a different element when the tag prop is provided", () => {
+      render(
+        <MenuList heading={{ children: "A menu section", tag: "h2" }}>
+          <MenuItem label="Item one" />
+        </MenuList>
```
gyfchong, line 22: Have you considered exposing the heading component and have consumers compose it instead?

```
<MenuList>
  <MenuListHeading id="list-heading-id" tag="h2">A menu section</MenuListHeading>
  <MenuItem label="Item one" />
</MenuList>
```

Would also simplify your `listHeadingId` implementation.

draft-packages/menu/KaizenDraft/Menu/components/MenuListHeading.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Heading } from "@kaizen/typography"
+import { MenuListProps } from "./MenuList"
+import styles from "./MenuListHeading.module.scss"
+
+export type MenuListHeadingProps = Required<Pick<MenuListProps, "heading">> & {
+  listHeadingId: string
+}
+
+export const MenuListHeading = (props: MenuListHeadingProps): JSX.Element => {
+  const { heading, listHeadingId } = props
```
gyfchong, line 12: Could you apply the `v4()` for the `id` in here to fully encapsulate MenuListHeading's concerns?

draft-packages/menu/KaizenDraft/Menu/components/MenuList.spec.tsx
dougmacknz: Ok just explored that option here: https://github.com/cultureamp/kaizen-design-system/pull/3604

It's a little different from your code sample, the menu needs to sit outside of the list because we can't have it sitting inside the `<ul>`. The MenuList also needs to link to the heading id with `aria-labelledby`

PR #3563, opened 2023-05-03
Force update of @react-stately/datepicker to avoid typescript issue

jasononeil:
## Why

Version 3.3.0 of react-stately/datepicker generated a TS issue, which was handled in Kaizen with a `@ts-expect-error` comment. But newer releases (3.4.0 onwards) don't have the problem, and so no error is found.

Because of a problem with Kaizen packaging accidentally shipping TS source files, consuming projects, which had different versions of this package, got TS errors saying that the "ts-expect-error" was unnecessary.

Fixes FEF-435 - https://cultureamp.atlassian.net/browse/FEF-435

## What

There are two suggested fixes I suggest:

1. Remove the `@ts-expect-error` and force a package upgrade to a version which won't produce an error. (this commit).
2. Fix Kaizen package publishing so that TS source files are either not included, or are included in a separate folder, so that the TS compiler in consuming projects doesn't accidentally parse TS source code.

I'll leave (2) for the Kaizen team as it feels like a change to be more careful with, but (1) will unblock the immediate error being faced in the support thread I linked.


.changeset/cool-rockets-hug.md
```diff
@@ -0,0 +1,11 @@
+---
+"@kaizen/date-picker": patch
+---
+
+Fix a typescript error by bumping the minimum required @react-stately/datepicker package
+
+Version 3.3.0 of react-stately/datepicker generated a TS issue, which was handled in Kaizen with a @ts-expect-error comment. But newer releases (3.4.0 onwards) don't have the problem, and so no error is found.
+
+Because of a problem with Kaizen packaging accidentally shipping TS source files, consuming projects, which had different versions of this package, got TS errors saying that the "ts-expect-error" was unnecessary.
```
gyfchong, line 9: Wouldn't say this is an accident, but more of an existing design 😅

jasononeil: The `files: []` field in the package.json includes `!**/*.ts` which I assume was designed to exclude all TS files (and same for TSX too) but it doesn't work and probably has never worked.

At some point though someone thought "we shouldn't ship TS files" and tried to set it to that. But yeah its been shipping TS for a long time, probably forever

gyfchong: Ohh that makes more sense... I'll try fix this instead of attempting bundling.

packages/components/src/FilterSelect/FilterSelect.spec.tsx
```diff
@@ -185,7 +185,7 @@ describe("FilterSelect generic", () => {
         {({ items }): JSX.Element[] =>
           items.map(item =>
             item.type === "item" ? (
-              <li key={item.key}>{item.value.isRubberDuck}</li>
+              <li key={item.key}>{item.value?.isRubberDuck}</li>
```
HeartSquared, line 188: How did this break by updating the date-picker? 😅

jasononeil: I wrote the reason in the commit message, it's a weird one.

I'm going to ask @gyfchong to take a look and decide if this is the best solution or not

gyfchong (approved):

packages/components/src/FilterSelect/FilterSelect.spec.tsx
gyfchong: If they want to allow `null` then we'll have to check for it. Sounds like good practice to always check for null anyway.

PR #3566, opened 2023-05-03
wip build to dist folder

gyfchong:
## Why

Linters in consumer repos are including Kaizen `tsx` files. 

## What

Move compile files from root directories into dist folders so consumer build tools can target the dist folder where no tsx files are present.

### DIFF
--- a/draft-packages/avatar/tsconfig.dist.json
+++ b/draft-packages/avatar/tsconfig.dist.json
@@ -1,14 +1,7 @@
 {
   "extends": "../../tsconfig.dist.json",
   "compilerOptions": {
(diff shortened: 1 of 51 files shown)


draft-packages/avatar/tsconfig.dist.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../../tsconfig.dist.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
```
gyfchong, line 5: It's weird that I need to do this, even if I put it in the root tsconfig, it won't work.

PR #3567, opened 2023-05-03
updated clean commands

gyfchong:
## Why
Clean commands were not cleaning.

## What
* add -g to rimraf to treat the arguments as globs
* updated some to target dist


HeartSquared (approved):
Tested a few things and seems to work nicely 🌟

PR #3568, opened 2023-05-03
Revert "updated clean commands"

gyfchong:
Reverts cultureamp/kaizen-design-system#3567

Causes type errors when building `FilterMultiSelect`
https://github.com/cultureamp/kaizen-design-system/actions/runs/4868447331/jobs/8681943382#step:5:236


HeartSquared (approved):
Cleaned a little _too_ well 😂

PR #3572, opened 2023-05-04
add slack notify to buildkite on passed release build

gyfchong:
So we know when packages are available to announce


HeartSquared (approved):

gyfchong:
It really didn't like any of the custom config I put in there which was according to [THEIR docs](https://buildkite.com/docs/pipelines/notifications)

PR #3573, opened 2023-05-04
chore(tsconfig): enable noImplicitReturns

HeartSquared:
## What

Enable `noImplicitReturns` in tsconfig.

## Why

Our consumers have stricter linting than us, and our shipping of source files meant they are getting errors.
We should be writing cleaner code anyway, so it's a good motivation to enable it.


gyfchong (approved):
Next PR: Remove build tool from titleblock

PR #3576, opened 2023-05-04
Replace Buttons form prop with size

mcwinter07:
## Why
The `form` property is currently typed as a Boolean within the Button component, which means consumers of the Button package are not able to leverage a native link between the HTML form and button element.

This issue was raised in Kaizen discourse here: https://github.com/cultureamp/kaizen-discourse/issues/50

## What
- Remove the `form` prop typed as a boolean
- add a `size` prop typed as `small | regular`. 
  - Avoided using `default` as this is already tied to a CSS property name and would cause some funky behaviours if we passed it in with classnames `style[props.size]`


packages/button/src/Button/Button.tsx
```diff
@@ -13,7 +13,8 @@ export type ButtonProps = GenericProps &
     primary?: boolean
     destructive?: boolean
     secondary?: boolean
-    form?: boolean
```
gyfchong, line 16: I don't see an extension of HTML Attributes or anything to support adding the real `form` prop

mcwinter07: Nope, you're absolutely right, that's completely missing - I probably should've made this a draft PR -  just wanted to get this up as a save point yesterday 😉 Will add that in this morning and fix the ts issues 👍

mcwinter07: Our button props are pretty gnarly, do you think it's worth just extending the whole of the button HTML attributes? Or just picking the form-related one for now 🤔

gyfchong: Probably form related one for now

mcwinter07: At the very least all these look handy to have if they're planning on using the button to submit forms 🤔 
 Screenshot 2023-05-05 at 9 36 32 am

mcwinter07: > Probably form related one for now
Fair, may be a very large thread to pull at if we were to start refactoring the button type 😆

gyfchong: Let's stick with `form` first

HeartSquared: What are the chances of us just removing the size difference altogether? 😏 It's a breaking change for the consumers anyway, and if it's an anti-pattern maybe we can just say they can't use the smaller one anymore.

gyfchong: Because it's a valid use case, without an intuitive way to communicate yet:
https://github.com/cultureamp/kaizen-discourse/issues/50#issuecomment-1447222226

gyfchong: Stay tuned for a `SubmitButton` solution preview lol

HeartSquared: But... what if they just have a bigger button? Would the design explode?

gyfchong: Probably not, but it could also be used for tables where the information could be dense, so introducing whitespace across the row could be undesirable. There's been no design discovery done on how small small should be yet, just that this is to mitigate the existing issue whilst giving us the option to improve it further down the road.

HeartSquared: Ah... well we'd probably need to note that people should refrain from using it in the meantime until we've figured out what it's meant to be 😬

gyfchong: Designers actually don't know this exists, so no problem on that part! For now it's purely for backwards compatibility with some snowflake designs.

gyfchong (approved):

PR #3580, opened 2023-05-05
Dependency updates

christian314159:


gyfchong:
Upgrading the stylelint package difficult as they have removed a few of the type definitions we use, and we aren't exactly sure how the plugin works, the tests also suck. I believe the types now come from `PostCSS` directly, but they are a different shape to what we need here.|

We should consider looking into if anyone is actually using this and deleting it.

HeartSquared:
I've taken over updates (didn't see this earlier) :)

PR #3582, opened 2023-05-05
feat: set autohide to false by default and depreciate the prop on toasts

caitlinmwillington:
## Why

Toast notifications that disappear automatically violates WCAG ['2.2 Enough Time'](https://www.w3.org/WAI/WCAG21/Understanding/enough-time) criteria

## What

This change sets the autohide to false, sets the prop as optional, and deprecates the prop.


packages/notification/src/types.ts
```diff
@@ -7,11 +7,15 @@ export type ToastNotification = {
...
   title: string
   message: React.ReactNode
-  autohide: boolean
   autohideDelay?: "short" | "long"
   onHide?: () => void
   automationId?: string
   persistent?: boolean
+  /**
+   * **Deprecated:**
+   * @deprecated
```
dougmacknz, line 16: Probably want to put some info in here

```suggestion
   * @deprecated Automatic hiding is an accessibility anti-pattern, violating WCAG '2.2 Enough Time' criteria. This prop will be removed in an upcoming breaking change.
```

packages/notification/src/types.ts
```diff
@@ -7,11 +7,15 @@ export type ToastNotification = {
   type: NotificationType
   title: string
   message: React.ReactNode
-  autohide: boolean
   autohideDelay?: "short" | "long"
```
dougmacknz, line 10: We should list this prop as deprecated too

packages/notification/src/ToastNotification.tsx
```diff
@@ -18,7 +18,7 @@ type ToastNotificationProps = Omit<
 const ToastNotification = ({
   id,
   hideCloseIcon = false,
-  autohide = true,
+  autohide = false,
   autohideDelay = "short",
```
dougmacknz, line 22: Might as well change the default on this while we're here too

```suggestion
  autohideDelay = "long",
```

packages/notification/src/types.ts
caitlinmwillington: Good idea, I went with the same deprecated message as the autohide as it seemed to apply for both but let me know if it should be different!

gyfchong:
You'll need to run `yarn changeset` 😄

caitlinmwillington:
> You'll need to run `yarn changeset` 😄

@gyfchong I tried to run it but was getting error Command "changeset" not found. 🤔 Anything I could be missing?

dougmacknz:
Might just need to `yarn install` @caitlinmwillington ?

caitlinmwillington:
> Might just need to `yarn install` @caitlinmwillington ?
Screenshot 2023-05-05 at 5 26 57 pm

@dougmacknz Yeah I did that :/ 
running `npx changeset` gives me:  `npm ERR! could not determine executable to run`

gyfchong:
Attempt `yarn commit`

dougmacknz (approved):

PR #3591, opened 2023-05-08
Kzn 1194 First Tailwind official release

JakePitman:
## Why
Adoption is high enough, and we've figured out most of the unknowns, including:
- The width and height system we want to go forward with
- A strange bug with missing arbitrary values for media queries

## What
- Settles on the Tailwind default as the width and height system for our TW preset
- Removes max-width breakpoints to fix bug where arbitrary values weren't working for media queries
- Updates the docs in the following ways:
    - Removes Utility Classname Reference docs for `width`, `height`, `maxWidth` and `maxHeight` (these are no longer custom)
    - Adds a paragraph explaining that our reference docs only cover custom utilities
    - Updates the url for TW play


gyfchong (approved):
🌬️

PR #3595, opened 2023-05-08
Optimise-turbo

gyfchong:
## Why

A few unused scripts, optimising turbo's inputs and outputs

## What

* ~fix clean scripts~ #3602 
* remove unused `build:watch`
* remove `clean` from scripts, utilise `yarn clean` from root dir to manage cleaning in CI scripts
* update turbo inputs and outputs on a per package basis


.buildkite/scripts/build-storybook.sh
```diff
@@ -5,7 +5,7 @@ set -e
 . ".buildkite/scripts/helpers/setup-registry.sh"
 
 yarn install --frozen-lockfile
-yarn workspace @kaizen/design-tokens prepublish
-yarn workspace @kaizen/tailwind prepublish
+yarn workspace @kaizen/design-tokens publish
```
HeartSquared, line 8: I think `publish` runs extra stuff?
If this is similar to other package commands, `prepublish` _may_ have been replaced with `version`.

But if you've checked that it doesn't have any side-effects, then it will be fine :)

docs/package-structure.md
```diff
@@ -79,9 +79,9 @@ The base package.json will look like this, where:
   "description": "The {{ COMPONENT_NAME }} component",
   "scripts": {
     "prepublish": "tsc --project tsconfig.dist.json",
-    "build": "yarn clean && yarn prepublish",
-    "build:watch": "yarn clean && yarn prepublish --watch",
-    "clean": "rimraf '**/*.d.ts' '**/*.js' '**/*.map'"
+    "build": "yarn prepublish",
+    
+    "clean": "rimraf -g '**/*.d.ts' '**/*.js' '**/*.map'"
```
HeartSquared, line 84: Didn't think overnuke last time? 😂

.buildkite/scripts/build-storybook.sh
gyfchong: Missed this, I toyed with `publish` but just went back to `prepublish`

packages/a11y/package.json
```diff
@@ -4,9 +4,8 @@
   "description": "A collection of accessibility utils",
   "scripts": {
     "prepublish": "tsc --project tsconfig.dist.json",
-    "build": "yarn clean && yarn prepublish",
-    "build:watch": "yarn clean && yarn prepublish --watch",
-    "clean": "rimraf '**/*.d.ts' '**/*.js' '**/*.map'"
+    "build": "yarn prepublish",
```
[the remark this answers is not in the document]
gyfchong: What do you mean by "it's not using `turbo`"?

docs/package-structure.md
gyfchong: Seems to be happy this time haha

packages/a11y/package.json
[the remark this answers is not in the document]
HeartSquared: Does it still use turbo if you go into the package and run the command by itself?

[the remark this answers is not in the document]
gyfchong: Nope, I think the general pattern (when dealing with many package) is to have them all setup as if they aren't in a monorepo ie. transferrable. `turbo` only runs from root and executes the scripts inside each package.

packages/stylelint-plugin/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
HeartSquared, line 1: Nuke (and the yarn.lock changes) 💣

package.json
```diff
@@ -31,8 +31,8 @@
     "pkg:aio": "yarn workspace @kaizen/components",
     "pkg:aio:test": "yarn pkg:aio test",
     "pkg:aio:compile": "yarn pkg:aio compile",
-    "ci:version": "yarn build && yarn changeset version",
-    "ci:publish": "yarn build && yarn changeset publish"
+    "ci:version": "yarn turbo prepublish && yarn changeset version",
+    "ci:publish": "yarn turbo prepublish && yarn changeset publish"
```
HeartSquared, line 35: Technically these could remain as `yarn build` lol. But not fussed if you don't change it.

HeartSquared (approved):

package.json
gyfchong: Eh, I like this one being explicit. Also won't need to run clean, making CI "faster" 😄

PR #3604, opened 2023-05-09
Create MenuHeading subcomponent

dougmacknz:
## Why
Closes https://github.com/cultureamp/kaizen-discourse/issues/75

## What
Creates a new `MenuHeading` component for use in the `Menu` component.


draft-packages/menu/KaizenDraft/Menu/components/MenuList.tsx
```diff
@@ -1,29 +1,18 @@
...
+import React, { HTMLAttributes } from "react"
+import { OverrideClassName } from "@kaizen/component-base"
 import styles from "./MenuList.module.scss"
 
-export type MenuListProps = {
-  heading?: React.ReactNode
+export interface MenuListProps
+  extends OverrideClassName<HTMLAttributes<HTMLUListElement>> {
   children: React.ReactNode
 }
```
gyfchong, line 8: Would you want to make `aria-labelledBy` mandatory here?

gyfchong:
Overall yeah I think I prefer this one

draft-packages/menu/KaizenDraft/Menu/components/MenuList.tsx
dougmacknz: No because more often than not there's no heading

draft-packages/menu/KaizenDraft/Menu/components/MenuList.tsx
```diff
@@ -1,25 +1,24 @@
...
-  const { heading, children } = props
+  const { heading, children, ...restProps } = props
   return (
     <>
-      {heading && (
-        <span className={styles.header} id={listHeadingID}>
-          {heading}
-        </span>
-      )}
+      {heading && cloneElement(heading, { id: listHeadingID })}
```
gyfchong, line 17: Hm wondering if it's worth checking the element props for a custom id and overriding the internal one?

Just in case someone wants to add their own id and wonders why it's not working.

dougmacknz: This is the `MenuList` component and we're injecting the ID into the cloned `MenuHeading` here. People can add their own ids to either `MenuList` and that works fine.

Now that you say that, I don't want to allow ids for MenuHeading because it will break the link, I'll adjust the TS there.

dougmacknz: Wait never mind I can't do that because it needs to accept the id from here.

gyfchong: Yeah, so I was thinking if consumers do this:
```
<MenuList heading={<Heading id="pie">Hello</Heading>} />
```

Then internally you can go:
```
heading.props.id ? setListId(heading.props.id) : setListId(v4());
```

dougmacknz: Ok that took a bit of rejigging because (currently) from the MenuHeading's perspective, it's always going to get an id and it can't determine if it came from the consumer or if it came from the `cloneElement`.

The ID generation was happening on MenuList rather than MenuHeading. In order to move the ID generation to MenuHeading, I need to pass a set state function through to MenuHeading from MenuList.

Have a look at my latest commit and see what you think. I think it does goes against the React ideology that data should only travel down and not up.

draft-packages/menu/KaizenDraft/Menu/components/MenuHeading.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const MenuHeading = (props: MenuHeadingProps): JSX.Element => {
+  const { id, setMenuListAriaLabelledBy, classNameOverride, ...restProps } =
+    props
+  const [headingId] = useState<string>(id || v4())
+
+  useEffect(() => {
+    setMenuListAriaLabelledBy && setMenuListAriaLabelledBy(headingId)
+  }, [])
```
gyfchong, line 19: Yeah I think your hunch is right, in that it shouldn't accept a setter, apologies for the misguided comment about encapsulating the `id` in here, it makes more sense for the `MenuList` to decide the `id` and pass it down. 😅

dougmacknz: All good, it was worth a shot. I've removed that commit now and fixed the merge conflicts

draft-packages/menu/KaizenDraft/Menu/components/MenuList.tsx
gyfchong: This still needs to happen in `MenuList` I think:
```
heading.props.id ? setListId(heading.props.id) : setListId(v4());
```

dougmacknz: Oops why didn't I think of that, that's much easier

gyfchong (approved):

PR #3607, opened 2023-05-09
Fix form prop in generic button

mcwinter07:
## Why
- The props were not being passed all the way into the Generic Button Component, which means they did not exist on the IconButton or Button Component.
  - It appears that they were filtered out by the getCustomProps function that removes anything not in kebab case that hasn't been restructured from props.

## What
- Destructures the form attribute and passes them into the button element
- Adds test coverage to validate a form submit is possible and that the attributes are there
- Adds story to document update to the Component


packages/button/src/Button/components/GenericButton.tsx
```diff
@@ -169,6 +169,12 @@ const renderButton = (
...
     onBlur,
+    form,
+    formAction,
+    formMethod,
+    formEncType,
+    formTarget,
+    formNoValidate,
     ...rest
   } = props
   const customProps = getCustomProps(rest)
```
mcwinter07, line 180: In an ideal world, we'd probably get rid of this and allow the button to extend all native attributes but I think that's part of a bigger refactor instead of adding a filter for kebab case

packages/button/src/Button/components/GenericButton.spec.tsx
```diff
@@ -86,3 +86,48 @@ describe("<GenericButton />", () => {
...
+  it("will pass form props into the button", () => {
+    render(
+      <GenericButton
+        label="submit button"
+        data-testid="custom-prop"
+        {...buttonFormAttributes}
+      />
+    )
+
+    expect(screen.getByRole("button", { name: "submit button" })).toBeTruthy()
```
christian314159, line 109: I think you can clean this up a little:
- We should use `toBeInTheDocument` here, rather than `toBeTruthy` which can have weird side effects (e.g. imagine a mistake in your expectation that returns `1` - that would be "truthy" but maybe not what you actually wanted to test)
- Having made the above point, you don't actually need the first `expect`, because the 2nd one - where you assert the `form` attribute will fail if the button can't be found

packages/button/src/Button/components/GenericButton.spec.tsx
```diff
@@ -86,3 +86,48 @@ describe("<GenericButton />", () => {
...
+        <GenericButton
+          label="submit button"
+          data-testid="custom-prop"
+          type="submit"
+          {...buttonFormAttributes}
+        />
+      </div>
+    )
+    fireEvent.click(screen.getByRole("button", { name: "submit button" }))
+    expect(onSubmit).toHaveBeenCalledTimes(1)
```
christian314159, line 131: I like the test and would definitely leave it, but you are essentially testing native `button` functionality (ie. it submitting the form in the `form` attribute), which some people would say is superfluous.

christian314159 (commented):
Looks great, just a tiny thing to clean up in the spec IMO.

packages/button/src/Button/components/GenericButton.spec.tsx
gyfchong: Indeed, that person is me 😄

Let's not make it a habit to add superfluous tests. Lest someone returns later and is confused as to why this is tested thinking it's for something important

packages/button/src/Button/components/GenericButton.spec.tsx
mcwinter07: True, will just remove the first one to clean up the test

packages/button/src/Button/components/GenericButton.spec.tsx
mcwinter07: Honestly, I was on the fence about doing this, I wanted to validate that the prop would actually work if it was passed into the component, but I am testing essentially to test that HTML should work as HTML 🤔 So maybe just its existence is enough?

packages/button/docs/Button.stories.tsx
```diff
@@ -158,3 +158,17 @@ export const Working: StoryFn = () => (
  * `import { LoadingInput } from "@kaizen/loading-skeleton"`
  */
 export const Loading: StoryFn = () => <LoadingInput isAnimated width={13} />
+
+/**
+ * <p>Button now extends native HTML "form" attributes for `button`</p>
+ * <p>You can use this to submit a `form` using `Button` with a matching form id</p>
+ */
+export const NativeFormButton: StoryFn = () => (
```
HeartSquared, line 166: This story probably needs a `<form>` in it to showcase the connection :)

mcwinter07: Fair point, probably a nice way to do a quick demo for our consumers 😉

HeartSquared (approved):

PR #3609, opened 2023-05-09
chore(eslint): enable react/jsx-curly-brace-presence

HeartSquared:
## Why

Prop strings don't need to be `prop={"string"}`, but instead can be `prop="string"`.
Add an auto-fixable linting rule to keep our usage consistent :)

## What

Enable eslint rule `react/jsx-curly-brace-presence`


gyfchong (approved):
😎

PR #3611, opened 2023-05-10
fix: revert packages peerDep design-tokens versions

HeartSquared:
## What

- Revert the supported design-tokens versions for packages to the full list instead of just the latest version.
- Add the experimental flag [`onlyUpdatePeerDependentsWhenOutOfRange`](https://github.com/changesets/changesets/blob/main/docs/experimental-options.md) to our changesets config which hopefully will mitigate the issue moving forward

## Why

Our packages support many versions of `design-tokens`, but changesets has been too helpful in bumping all the versions in the peerDeps for us.


.changeset/config.json
```diff
@@ -6,5 +6,8 @@
   ],
   "access": "public",
   "baseBranch": "main",
-  "updateInternalDependencies": "patch"
+  "updateInternalDependencies": "patch",
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
```
mcwinter07, line 11: Reading the changesets docs:
> When set to true, Changesets will only bump peer dependents when peerDependencies are leaving the range.
Just to clarify if I understand this - does this just mean that it will leave our peers deps as is or it will bump the version within each major in that range

HeartSquared: From my understanding, it should leave peerDeps alone until a major outside the range is released, of which in that scenario we should be updating the peerDeps anyway.

I couldn't find an option to exclude only peerDeps from getting updates. It seemed to be an all-types-of-deps-or-nothing 😬

mcwinter07: Fair enough! lets merge it in and see 🤞😆

mcwinter07 (approved):

.changeset/config.json
gyfchong: For what it's worth, [Unified-Home](https://github.com/cultureamp/unified-home/blob/master/.changeset/config.json#L14) do this as well.

gyfchong (approved):

PR #3616, opened 2023-05-10
Ignore source files when publishing

gyfchong:
## Why
For unknown reasons, source files are being published together with compiled files, effectively ignoring the "files" set in package.json. For reasons, consumer apps are picking up source files and checking types/linting.

## What
Add an npmignore file to force npm publish to ignore the files.

To test this, `cd` into any package and run `npm pack --dry-run` which will show you what files are being published.

### DIFF
--- a/draft-packages/avatar/.npmignore
+++ b/draft-packages/avatar/.npmignore
@@ -1,6 +1,0 @@
-**/*.ts
-**/*.tsx
-**/*.spec.*
-.turbo/*
-docs
-tsconfig.dist.json
--- a/draft-packages/avatar/.npmignore
+++ b/draft-packages/avatar/.npmignore
@@ -1,0 +1,6 @@
+**/*.ts
+**/*.tsx
+**/*.spec.*
+.turbo/*
+docs
+tsconfig.dist.json
--- a/draft-packages/badge/.npmignore
+++ b/draft-packages/badge/.npmignore
@@ -1,6 +1,0 @@
-**/*.ts
-**/*.tsx
-**/*.spec.*
-.turbo/*
-docs
-tsconfig.dist.json
--- a/draft-packages/badge/.npmignore
+++ b/draft-packages/badge/.npmignore
@@ -1,0 +1,6 @@
+**/*.ts
+**/*.tsx
+**/*.spec.*
+.turbo/*
+docs
+tsconfig.dist.json
--- a/draft-packages/form/.npmignore
+++ b/draft-packages/form/.npmignore
@@ -1,6 +1,0 @@
-**/*.ts
-**/*.tsx
-**/*.spec.*
-.turbo/*
-docs
-tsconfig.dist.json
--- a/draft-packages/form/.npmignore
+++ b/draft-packages/form/.npmignore
@@ -1,0 +1,6 @@
+**/*.ts
+**/*.tsx
+**/*.spec.*
+.turbo/*
+docs
+tsconfig.dist.json
--- a/draft-packages/illustration/.npmignore
+++ b/draft-packages/illustration/.npmignore
@@ -1,6 +1,0 @@
-**/*.ts
-**/*.tsx
-**/*.spec.*
-.turbo/*
-docs
-tsconfig.dist.json
--- a/draft-packages/illustration/.npmignore
+++ b/draft-packages/illustration/.npmignore
@@ -1,0 +1,6 @@
+**/*.ts
+**/*.tsx
+**/*.spec.*
+.turbo/*
+docs
+tsconfig.dist.json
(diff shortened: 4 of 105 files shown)


draft-packages/avatar/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
+**/*.tsx
+**/*.spec.*
+.turbo/*
+docs
+tsconfig.dist.json
```
HeartSquared, line 6: You should enable your IDE setting that adds the EOF new line 😛

draft-packages/badge/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: This is per package?! Gross :'(

HeartSquared: Do you need to ignore `.npmignore` too? Or is that just done by default?

draft-packages/form/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for form :P

draft-packages/illustration/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for illustration :P

draft-packages/menu/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for menu :P

draft-packages/tabs/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for tabs :P

draft-packages/tile/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for tile :P

packages/brand-moment/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for brand-moment :P

packages/component-library/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for component-library :P

packages/deprecated-component-library-helpers/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for this :P

packages/design-tokens/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for design-tokens :P

packages/hosted-assets/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Is this needed if `files` holds no negatives?

packages/notification/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*.ts
```
HeartSquared, line 1: Where's the `package.json` update for notification :P

draft-packages/badge/.npmignore
gyfchong: I think it needs to be committed, otherwise the pipeline won't be able to read it

HeartSquared: Yeah definitely committed, but does it need to exist in the dist? This file tells it what to include in the distribution right?

gyfchong: Lol actually, this files definition works fine:
```
    "**/*.js",
    "**/*.map.js",
    "**/*.d.ts"
```

HeartSquared: True! As long as we haven't lost anything in between :)

gyfchong:
> One thing to note is that if anyone adds any other type of file to the package (eg. images) then we need to update the `package.json`. Unlikely to happen so often, but we certainly need to keep it in mind (perhaps we need to add some documentation?)

I'm going to optimistically say that we don't need to document cause we're moving to KAIO which puts everything necessary into `dist` 😛

PR #3626, opened 2023-05-11
chore(deps): update stylelint to ^15.6.1

HeartSquared:
## What

Update `stylelint` to `^15.6.1`
- Update the config to suggested format (js => json)
- Install the recommended base config according to docs
- Ignore rules where they seem to want a strange looking format (rules set to `null`)
- Fix all the errors (most have autofix, otherwise I either quickly fixed them or added some ignores)


.stylelintrc.json
```diff
@@ -0,0 +1,99 @@
+{
+  "extends": "stylelint-config-standard-scss",
```
gyfchong, line 2: Do we need one of these for CSS too?

HeartSquared: CSS has a different set of config defaults, but I think this one extends the base one?
When we move away from scss we can change this :)

gyfchong: Can you make a quick ticket for us to make one of these for KAIO? So we don't forget, thanks.

gyfchong (approved):

PR #3630, opened 2023-05-12
Add isAnimated prop to MenuLoadingSkeleton

slacis:
## Why
We are requiring the loading state here in
https://cultureamp.atlassian.net/browse/COSE-1367
and noticed that is can't currently be animated.
This was a simple need of passing the prop in

## What
Add the isAnimated prop to animate the loading skeleton
loadinganimated

### DIFF
--- a/.changeset/nice-taxis-sell.md
+++ b/.changeset/nice-taxis-sell.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/select": minor
+---
+
+Add isAnimated prop to FilterMultiSelect MenuLoadingSkeleton.
--- a/.changeset/nice-taxis-sell.md
+++ b/.changeset/nice-taxis-sell.md
@@ -2,4 +2,4 @@
 "@kaizen/select": minor
 ---
 
-Add isAnimated prop to FilterMultiSelect MenuLoadingSkeleton.
+Add isAnimated prop to FilterMultiSelect MenuLoadingSkeleton. Setting this to true animates all the child loading skeletons.
--- a/packages/select/docs/MenuLoadingSkeleton.stories.tsx
+++ b/packages/select/docs/MenuLoadingSkeleton.stories.tsx
@@ -16,7 +16,6 @@ const meta = {
       "npm install @kaizen/select",
       "import { FilterMultiSelect } from `@kaizen/select`",
     ],
-    chromatic: { disable: false },
     docs: {
       container: ComponentDocsTemplate,
     },
@@ -37,6 +36,7 @@ export default meta
  */
 export const Playground: StoryObj<typeof meta> = {
   parameters: {
+    chromatic: { disable: false },
     docs: {
       canvas: {
         sourceState: "shown",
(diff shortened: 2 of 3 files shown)


.changeset/nice-taxis-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/select": minor
+---
+
+Add isAnimated prop to FilterMultiSelect MenuLoadingSkeleton.
```
gyfchong, line 5: Please add a bit more detail as to what it does 😄

gyfchong (approved):

packages/select/docs/MenuLoadingSkeleton.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { MenuLoadingSkeleton } from "../src/FilterMultiSelect/components/MenuLayout/MenuLoadingSkeleton/index"
+
+const meta = {
+  tags: ["autodocs"],
+  title: "Components/Filter Multi-Select/Menu Loading Skeleton",
+  component: MenuLoadingSkeleton,
+  args: {
+    isAnimated: false,
+  },
+  parameters: {
```
gyfchong, line 12: Ah apologies, you'll need to add a `installation` key here, for example:
```
    installation: [
      "npm install @kaizen/button",
      "import { Button } from `@kaizen/button`",
    ],
```

gyfchong, line 12: Also need to add the chromatic key to push snapshots
```
chromatic: { disable: false },
```

gyfchong (changes requested):

packages/select/docs/MenuLoadingSkeleton.stories.tsx
slacis: I had this one but removed it because it looked a little confusing, i.e. it ends up being 
```
    installation: [
      "npm install @kaizen/select",
      "import { FilterMultiSelect } from `@kaizen/select`",
    ],
```

is it better to have it regardless?

slacis: i've pushed it for now, guess it still can't hurt to know where it exists

packages/select/docs/MenuLoadingSkeleton.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    installation: [
+      "npm install @kaizen/select",
+      "import { FilterMultiSelect } from `@kaizen/select`",
+    ],
+    chromatic: { disable: false },
```
HeartSquared, line 19: This is the wrong place to put this because we don't want to be sending every story from this file as a snapshot (especially not an animated one). If you're not making a sticker sheet, then you need to add this param only to the `Playground` in your case.

packages/select/docs/MenuLoadingSkeleton.stories.tsx
gyfchong: Yeah, we're noting these things to optimise in the next iteration of docs 😄

packages/select/docs/MenuLoadingSkeleton.stories.tsx
slacis: ohhh right, makes sense!

gyfchong (approved):

PR #3632, opened 2023-05-12
KZN-1148/retrofit-filter-multi-select

gyfchong:


packages/components/src/FilterBar/subcomponents/FilterMS.tsx
```diff
@@ -0,0 +1,57 @@
...
+          />
+        ) : (
+          <FilterMultiSelect.TriggerButton
+            selectedOptionLabels={getSelectedOptionLabels(
+              filterState.selectedValue,
+              items
+            )}
+            label={filterState.label}
+          />
+        )
```
HeartSquared, line 50: This component isn't styled correctly. Is it possible to retrofit it to FilterButton?

gyfchong: As discussed, we are descoping retrofitting to use `Filter` Abstraction due to significant rework required to `FilterMultiSelect` and will add this to the roadmap.

To summarise:
- A new `MultiSelect` is on the horizon which can be built with the `Filter` abstraction in mind so we can potentially port the `FilterMultiSelect` to use it.
- We ant to maintain high compatibility with existing implementations of `FilterMultiSelect`, any potential rework of the API will disrupt adoption.

PR #3643, opened 2023-05-17
Add workflow components

mcwinter07:
## Why
We are adopting the Workflow Components into the Design System as part of the Self Reflections Rebuild

## Issue

The workflow pattern is used across several repos and multiple camps but is often built as a snowflake / copied directly into their repo. This could cause inconsistencies in behavior for the end user and increase the difficulty to roll out new features or optimizations in the future.

## What
- Rebuilds the engagement-worflow package into composable chunks that could be used to re-create the component
  - sensible defaults, such as the ExitWorkflow have been created to ensure ease of use
- Addresses a11y issues for heading hierarchy, color contrast ratios, landmarks and accessible names outline in the [accessibility report](https://cultureamp.audit.intopia.digital/accessibility-review-of-culture-amp---engagement-product-and-shared-features.html)

## Jira references
- https://cultureamp.atlassian.net/browse/SR2-63
- https://cultureamp.atlassian.net/browse/SR2-73
- https://cultureamp.atlassian.net/browse/SR2-76


packages/workflow/docs/WorflowHeader.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+  status: {
+    content: "Draft",
+    vairant: "statusDraft",
+  },
+  actions: [],
+} satisfies WorkflowHeaderProps
+
+const meta = {
+  tags: ["autodocs"],
+  title: "Components/Workflow Header",
```
gyfchong, line 20: Might be better to separate it into it's own category,

```suggestion
  title: "Workflow/Header",
```

packages/workflow/docs/WorkflowExample.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+// import VisibleIcon from "@kaizen/component-library/icons/visible.icon.svg"
+import { Button } from "@kaizen/button"
+import { ComponentDocsTemplate } from "../../../storybook/components/DocsContainer"
+import {
+  Workflow,
+  WorkflowExit,
+  WorkflowFooter,
+  WorkflowHeader,
+  WorkflowProps,
+} from "../"
```
gyfchong, line 13: Probably need to have a chat about this naming, cause it's going to be a bit awkward especially when from the consumer side it would look like this:

```
import {
  Workflow,
  WorkflowExit,
  WorkflowFooter,
  WorkflowHeader,
  WorkflowProps,
} from "@kaizen/workflow" // workflow is implied already
```

And in 99% of cases, it looks like this:
```
import { Workflow } from "@kaizen/workflow"
```

mcwinter07: Oh this is just a stand-in while I was building out all the components and testing how they slot together. The final one here would be just using the Workflow component from `Worflow.tsx`. Or did you mean if someone was going to use the individual components that should use dot notation, ie: Workflow.Header?

packages/workflow/docs/WorflowHeader.stories.tsx
mcwinter07: Good catch! That's actually not meant to be there 😆

packages/workflow/docs/WorkflowExample.stories.tsx
gyfchong: Yeah the latter, but not sure the dot notation applies cause it doesn't have a hard dependency on `Workflow` to work eg. context

packages/workflow/docs/WorflowHeader.stories.tsx
mcwinter07: I mean as long as it not confusing to folks when they're looking for it? I know we discussed possibly grouping it under patterns, but I don't know if `patterns/workflow` hierarchy would be more confusing

mcwinter07: Screenshot 2023-05-18 at 2 34 54 pm
What do you think about something like this? @gyfchong - Its similar to the filter button, which is less flat than our standard components, but maybe better for documenting the grouped element. We'd probably also want a progress stepper story that lives inside the footer

mcwinter07: Or do you think we should scrap the "components" and just do "Workflow Exit"

gyfchong: So far my rule has been "If the component can technically be combined with others it should be top level" or rather only nest if it has some kind of hard dependency.

gyfchong: Could Stepper be a non-workflow specific component?

mcwinter07: We could potentially have it as a non-workflow component with some modifications but I don't know if it's something we would want outside of this context. There are limitations from a design perspective where it could be used and I think in time it would be scrapped or revamped - so leaning toward no?

gyfchong: Perhaps since we want to guide people to the primary monolith component, we nest the Header, Footer etc. in `Workflow/components/*`

gyfchong:
I think just to be complete, we should have a `Main` component to encapsulate the styles and `main` landmark 😄

packages/workflow/docs/WorflowHeader.stories.tsx
mcwinter07: I think that's a nice way to describe the relationship / guide folks to the larger wrapper

packages/workflow/docs/WorkflowExample.stories.tsx
mcwinter07: > I think just to be complete, we should have a `Main` component to encapsulate the styles and `main` landmark 😄

Sounds like a way to nudge users to ensure the main landmark is used 👍

packages/workflow/README.md
```diff
@@ -0,0 +1,5 @@
+# Install
+
+```
+yarn add @kaizen/workflow
+```
```
gyfchong, line 5: Don't think you need this, they should be coming from Storybook?

packages/workflow/docs/WorkflowExample.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    },
+  },
+} satisfies Meta<typeof Workflow>
+
+export default meta
+
+/**
+<p>This is a custom version of the `Workflow` assembled using the `Header` and `Footer` exported from `@kaizen/workflow`.</p>
+ */
+export const Playground: StoryFn<WorkflowProps> = ({
```
gyfchong, line 63: Probably want to use `mdx` for this, cause you don't necessarily need an args table/controls for an example of composition. A simple story with the source code open would suffice I think

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+        Lorem ipsum dolor sit amet consectetur adipisicing elit. Soluta error
+        labore dolorum debitis eaque laboriosam qui quidem itaque necessitatibus
+        obcaecati aut earum consectetur excepturi accusamus nulla libero maxime,
+        quibusdam vero?
+      </p>
+    </main>
+  </Workflow>
+)
+
+Playground.args = { ...defaultArgs }
```
gyfchong, line 103: Could do with a few more stories, generally I'd aim for one story per prop which shows what each of the values looks like. eg. what `isComplete` looks like both `true` and `false`.

Admittedly this is a bit big, but give it a try.

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    stepName={stepName}
+    status={status}
+    actions={actions}
+    steps={steps}
+    isComplete={isComplete}
+    nextAction={nextAction}
+    previousAction={previousAction}
+    {...restProps}
+  >
+    <main className="flex flex-col content-center min-h-[100vh]">
```
gyfchong, line 85: Shouldn't this be provided by the `Workflow` component?

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+  workflowName: "Create a self-reflection cycle",
+  stepName: "Settings",
+  steps: ["Settings", "Questions", "Preview", "Employees", "Schedule"],
+  isComplete: false,
+  status: {
+    content: "Draft",
+    vairant: "statusDraft",
+  },
+  previousAction: <Button reversed label="back" />,
+  nextAction: <Button reversed label="next" />,
```
gyfchong, line 17: These don't look very good in the controls, try something like the `validationState` in [FilterSelect](https://dev.cultureamp.design/SR2-73/add-workflow-components/?path=/docs/components-filter-select--docs) where people can toggle between `undefined` and a predetermined component.

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+  title: "Workflow",
+  component: Workflow,
+  parameters: {
+    docs: {
+      source: { type: "code" },
+      container: ComponentDocsTemplate,
+    },
+    installation: [
+      "npm install @kaizen/workflow",
+      "import { Workflow } from `@kaizen/workflow`",
```
gyfchong, line 42:
```suggestion
      "import { Workflow } from '@kaizen/workflow'",
```

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+  tags: ["autodocs"],
+  title: "Workflow",
+  component: Workflow,
+  parameters: {
+    docs: {
+      source: { type: "code" },
+      container: ComponentDocsTemplate,
+    },
+    installation: [
+      "npm install @kaizen/workflow",
```
gyfchong, line 41:
```suggestion
      "yarn add @kaizen/workflow",
```

I think everyone uses `yarn`?

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+  },
+} satisfies Meta<typeof Workflow>
+
+export default meta
+
+/**
+<p>The `Workflow` component offers a page wrapper that props down to it `Header` and `Footer` components.</p>
+<p>Each component is exported from the `@kaizen/workflow` package so it can be assembled individually if a custom solution is needed.</p>
+<p>See the `Custom Workflow` story for an example of how to assemble with a header and footer.</p>
```
gyfchong, line 61: I think it would be better to describe what the Workflow is here, perhaps grab a design description of what it is.

And to encourage more use of the main Workflow component, we shouldn't call out directly that they can create a custom solution, I recommend putting your "custom workflow" example at the bottom of the story ([example](https://ui.compono.dev/?path=/docs/components-menu--interactive#decomposed-usage)), that way it's like a "last resort" once they've read through all the existing API and haven't found their use case covered.

packages/workflow/src/Header/components/Titles/Titles.tsx
```diff
@@ -0,0 +1,57 @@
...
+        color="dark-reduced-opacity"
+      >
+        {workflowName}
+      </Heading>
+      <span>{stepName}</span>
+    </Heading>
+    {status && (
+      // status may need to be update by a fetch
+      <div className={styles.status}>
+        <Tag variant={status?.vairant || "statusDraft"}>{status?.content}</Tag>
```
gyfchong, line 51: `vairant`?

```suggestion
        <Tag variant={status?.variant || "statusDraft"}>{status?.content}</Tag>
```

packages/workflow/docs/Workflow.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const defaultArgs = {
+  workflowName: "Create a self-reflection cycle",
+  stepName: "Settings",
+  steps: ["Settings", "Questions", "Preview", "Employees", "Schedule"],
+  isComplete: false,
+  status: {
+    content: "Draft",
+    vairant: "statusDraft",
+  },
```
gyfchong, line 15: Something good to call out in a story would be that this extends the `Tag` component's variants, so you don't necessarily have to showcase every variant on this page, and they can go check relevant docs.

packages/workflow/src/Header/components/Titles/Titles.tsx
mcwinter07: Ralph Wiggum vibes there 😆 
Screenshot 2023-05-19 at 12 02 16 pm

packages/workflow/docs/Workflow.stories.tsx
mcwinter07: Fair point, might be calling too much attention to it, maybe more fitting as a footnote with caveats *

packages/workflow/docs/Workflow.stories.tsx
mcwinter07: Except for those using pnpm 😉 pretty sure yep. I think this was generated from the plop template automatically so complete glazed over it - should props update that too 👍

packages/workflow/README.md
mcwinter07: I think this came from plop, we could axe that from the template if we don't need those in the future

packages/workflow/docs/Workflow.stories.tsx
mcwinter07: I'll add that <Main> wrapper into the Workflow component as you recommended so that landmark will be present. The one thing I'm hesitant about is applying styles in case to have specific needs for grid or flex layouts / how they want to handle page responsiveness on internal components

packages/components/src/index.ts
```diff
@@ -1,4 +0,0 @@
-export * from "./Filter"
```
mcwinter07, line 1: hmmm just realised this shouldn't be deleted 😬 will fix this up 😆

gyfchong:
Is it possible to have the `Canvas` without the source code? I'd prefer if the code snippets showed `Worfklow` API and not the individual components' API. So whilst it renders just the header, you'd hardcode the snippet to showcase what code looks like for the`<Workflow />` component instead.

packages/components/src/Workflow/_docs/Workflow.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Canvas layout="fullscreen" of={WorkflowHeader.Playground} />
+
+## Tracking progress
+
+The Footer tracks the progress of the form by comparing the current step, indicated by `stepName`, against its position in the `steps` array.
+
+<Canvas layout="fullscreen" of={WorkflowFooter.FirstStep} />
+
+The Footer is agnostic to the JSX elements that are used in the `previousAction` and `nextAction`. We recommend using the `Button` component but an alternative JSX component can be used if you require NextJS Routing.
```
gyfchong, line 63: > if you require NextJS Routing.

[Not entirely true](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3128133092/Solution+Preview+support+for+custom+routing) 😉

mcwinter07: Fair point, I could rephrase it to:
> While we recommend using the Kaizen `Button`, a button-like component can be used in its place to satisfy project-specific requirements

mcwinter07:
> Is it possible to have the Canvas without the source code? I'd prefer if the code snippets showed Worfklow API and not the individual components' API. So whilst it renders just the header, you'd hardcode the snippet to showcase what code looks like for the<Workflow /> component instead.

@gyfchong So just to confirm, are you meaning something like this: 
Screenshot 2023-05-25 at 9 38 07 am

It's a little chonky but I can hard code the canvas output with like this:
```
<Canvas layout="fullscreen" of={WorkflowHeader.MultipleActions} source={{code: `<Workflow
  headerActions={[
    <Button
      key="would-use-uui-1"
      label="Preview"
      icon={VisibleIcon}
      secondary
      iconPosition="start"
    />,
    <WorkflowExit
      key="would-use-uui-2"
      exitLabel="Save and close"
      exitTitle="Before you exit"
      exitDescription="Your content has not yet been saved. Click the button below or discard the changes"
      confirmExitLabel="Close and save"
      dismissExitLabel="Dismiss"
      onExit={(): void => alert("mock example of a save action")}
    />,
  ]}
  nextAction={<Button label="Next" reversed/>}
  previousAction={<Button label="Back" reversed/>}
  status={{
    content: 'Draft',
    variant: 'statusDraft'
  }}
  stepName="Settings"
  steps={[
    'Settings',
    'Questions',
    'Preview',
    'Employees',
    'Schedule'
  ]}
  workflowName="Create a self-reflection cycle"
>
  <MockContent />
</Workflow>`}} />
```

Not sure if there's a way to do this a little cleaner if this is the way forward 🤔 I'd dig in a little more but I'd be concerned that this becomes a bit of a docs rabbit hole and I'm trying to get this into the cycle UI repo for the presentation tomorrow 😬

gyfchong:
More like this, where it focuses directly on the props you're talking about, so it shows what to put into the `Workflow` component rather than a `Header` they cannot see.
```
<Workflow
  headerActions={[
    <Button
      key="would-use-uui-1"
      label="Preview"
      icon={VisibleIcon}
      secondary
      iconPosition="start"
    />,
    <WorkflowExit
      key="would-use-uui-2"
      exitLabel="Save and close"
      exitTitle="Before you exit"
      exitDescription="Your content has not yet been saved. Click the button below or discard the changes"
      confirmExitLabel="Close and save"
      dismissExitLabel="Dismiss"
      onExit={(): void => alert("mock example of a save action")}
    />,
  ]}
  ...
>
  <MockContent />
</Workflow>
```

>  I'm trying to get this into the cycle UI repo for the presentation tomorrow

Fair enough! Give my suggestion above a quick go, and if not, then let's get it merged by the arvo.

One last thing about the nav though, let's put it in `Pages/Workflow` and put it right under the `Components` section 👍

mcwinter07:
> More like this, where it focuses directly on the props you're talking about, so it shows what to put into the `Workflow` component rather than a `Header` they cannot see.
> 
> ```
> <Workflow
>   headerActions={[
>     <Button
>       key="would-use-uui-1"
>       label="Preview"
>       icon={VisibleIcon}
>       secondary
>       iconPosition="start"
>     />,
>     <WorkflowExit
>       key="would-use-uui-2"
>       exitLabel="Save and close"
>       exitTitle="Before you exit"
>       exitDescription="Your content has not yet been saved. Click the button below or discard the changes"
>       confirmExitLabel="Close and save"
>       dismissExitLabel="Dismiss"
>       onExit={(): void => alert("mock example of a save action")}
>     />,
>   ]}
>   ...
> >
>   <MockContent />
> </Workflow>
> ```
> 
> > I'm trying to get this into the cycle UI repo for the presentation tomorrow
> 
> Fair enough! Give my suggestion above a quick go, and if not, then let's get it merged by the arvo.
> 
> One last thing about the nav though, let's put it in `Pages/Workflow` and put it right under the `Components` section 👍

Nice - I like it!

mcwinter07:
Woop, just gonna quickly update the nav to Pages/Workflow then should be good for a review 👍

gyfchong (approved):
Lovely

gyfchong (approved):

PR #3645, opened 2023-05-18
fix(Select): Set background color to white by default [KDS-1449]

christian314159:
## Why
- Current Select doesn't set background color unless hovered or focused; when placed on non-white background this could potentially be an issue (e.g. color contrast)

## What
- Set BG to white for non-reversed variant
image

## What not
- Doesn't address any problems of label or description not meeting color contrast guidelines (as there is no background to be set)

### DIFF
--- a/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
+++ b/packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
@@ -11,7 +11,7 @@
 
 .button {
   @include button;
-
+  background-color: $color-white;
   border-radius: $border-solid-border-radius;
   margin-top: $spacing-xs;
   width: 100%;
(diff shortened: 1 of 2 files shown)


packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss
```diff
@@ -121,6 +121,7 @@
 }
 
 .reversed.disabled {
+  background-color: transparent;
```
HeartSquared, line 124: Should this be transparent or some other explicit colour instead?

christian314159: As far as I could tell, it is currently transparent as the reversed, disabled variant - had to set it explicitly here, because the CSS spilled over.

HeartSquared: I suppose if that's what it's meant to be 😬

HeartSquared (dismissed):

HeartSquared (approved):

PR #3652, opened 2023-05-19
Update selectors in filter multi select test to fix playwright

mcwinter07:
## Why
- CI was failing playwright because accessible names were updated and the .locator could not find them

## What
- updates FMS playwright selector name


gyfchong:
Wow my bad, I should've read the errors closer! Thanks.

gyfchong (approved):

PR #3656, opened 2023-05-22
Use Collapsible title as toggle button accessible label

dougmacknz:
## Why
This toggle button for Collapsible's has a built in label of "Toggle section".
image

The problem with this is that when a screen reader user scans the page's buttons, these don't have the context that would help them.

## What
Use the Collapsible title as a better accessible label for toggling the collapsible, rather than a generic built in "Toggle section" string.

### DIFF
--- a/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.tsx
+++ b/draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.tsx
@@ -130,7 +130,7 @@ export class Collapsible extends React.Component<CollapsibleProps, State> {
           )}
           <div>
             <IconButton
-              label="Toggle section"
+              label={title}
               icon={open ? chevronUp : chevronDown}
               type="button"
               aria-expanded={open}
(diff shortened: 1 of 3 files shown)


draft-packages/collapsible/KaizenDraft/Collapsible/Collapsible.tsx
```diff
@@ -130,7 +130,7 @@ export class Collapsible extends React.Component<CollapsibleProps, State> {
           )}
           <div>
             <IconButton
-              label="Toggle section"
+              label={title}
```
gyfchong, line 133: Would this be a negative experience for someone who is tabbing through the UI instead? ie. it might read the title of the collapsible and then the title again when they get to the button.

Would a `labelledBy` be more appropriate? Or present the same issue?

dougmacknz:
> Would a labelledBy be more appropriate? Or present the same issue?

I think it would be the same experience. They did suggest using `aria-labelledby` but because of `IconButton`'s required `label` prop it will be a bit awkward.

The ideal situation here is that the title itself is the button and we hide the chevron from screen readers. The reason I didn't do that is because it's pretty common for our consumers to do a complex heading with a title and a bunch of other things that we won't want as part of the button. So this was the much quicker solution with the situation we're in.

gyfchong (approved):

PR #3658, opened 2023-05-22
fix: rgb functions allow 3 arguments only

ckychris:
## Why
Got further build errors with the select component. 

## What
- Changed all rgb functions with 4 args into rgba (Sylelint cannot enforce this correctly)
- Changed all alpha values to numbers only (Sylelint cannot enforce this correctly)

### DIFF
--- a/draft-packages/popover/KaizenDraft/Popover/Popover.module.scss
+++ b/draft-packages/popover/KaizenDraft/Popover/Popover.module.scss
@@ -56,7 +56,7 @@ $large-width: 450px;
 %box {
   background: $color-white;
   border: $border-solid-border-width $border-solid-border-style $color-gray-300;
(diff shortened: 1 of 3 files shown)


gyfchong:
Have you yalc'd this to confirm whether or not it resolves your build issues?

ckychris:
> Have you yalc'd this to confirm whether or not it resolves your build issues?

I have done a similar test by directly modifying the values in `node_modules`. I have done those tests for the last PR but the one key difference is that I didn't do a final test after yarn:lint:fix which didn't account for the rgb->rgba conversion. This time I have double-checked that the changes tests are exactly the same. 😌

gyfchong (approved):

gyfchong:
Oops there was no changeset, you'll need to add one to this branch and I'll remerge

PR #3663, opened 2023-05-22
add typescript to renovate ignore

gyfchong:
## Why
- `ttypescript` has no clear typescript v5 compatible replacement, which requires investigation.
- It is also unclear as to whether or not we should upgrade to v5 at the risk of causing incompatibilities with consumer typescript versions

## What

Add `typescript` to the renovate ignore.


renovate.json5
```diff
@@ -4,6 +4,7 @@
   "ignoreDeps": [
     // We can't always stick to latest due to infra and compatibility with other plugins
     "node",
+    "typescript",
```
dougmacknz, line 7: Don't want to do a comment on it like the others have?

gyfchong: I think it's the same as `node`

dougmacknz (approved):

PR #3666, opened 2023-05-22
Fix storybook css modules

gyfchong:
## Why
We are looking to move forward with CSS modules in KAIO, but Storybook only supported SCSS Modules.

## What
- Remove custom webpack style loader config
- Add Storybook integrated styling addon
- Uses the root project postcss config to bundle
  - In prep for KAIO to also use for a single source of truth for bundling our css

To see an example of this working, you can look at the ExampleComponent:
- [Code](https://github.com/cultureamp/kaizen-design-system/pull/3666/files#diff-d51c1bd3f2bdae2ad8dfe2cbe9b6011954188f1c4353a991a49e208971ce4cdcR4)
- [Storybook](https://dev.cultureamp.design/fix-storybook-css-modules/?path=/docs/deprecated-examplecomponent--docs)


storybook/tailwind.css
```diff
@@ -28,7 +26,7 @@ pre code {
   }
 
   .sbdocs-a {
-    color: $color-blue-400;
+    color: --color-blue-400;
```
christian314159, line 29: Might be a bit of a noob question, b/c I never really worked with CSS variables, but shouldn't that be `var(--color-blue-400)`?

PR #3670, opened 2023-05-24
fix(A11y): TitleBlock mobile menu needs focus lock

christian314159:
## Why
- Fixes cultureamp/kaizen-discourse#77


dougmacknz (approved):
🚢

dougmacknz (approved):

PR #3672, opened 2023-05-25
Update ProgressStepper SVGs and responsiveness

mcwinter07:
## Why
- The icon was recently changed in the UI Kit following work on SR rebuild
  - See update UI Kit [here](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%F0%9F%92%9C-UI-Kit%3A-Heart?type=design&node-id=35882%3A93848&t=83zwQRVgTxiP0xly-1) 
- Actioning design feedback on responsive styles on

## What
- Replaces SVG inner content with new SVG design
- Updates progress stepper to use React SVG
- Updates workflow footer to a simplified progress string on mobile,ie "step 2 of 3"
  - The container query will still handle the use case of a user creating too many steps

## Before
Screenshot 2023-05-25 at 11 07 36 am

## After

Screenshot 2023-05-25 at 11 07 05 am


christian314159 (approved):

packages/components/src/Workflow/subcomponents/Footer/components/ProgressStepper/ProgressStepper.tsx
```diff
@@ -17,21 +16,24 @@ const getStepStatus = (
   isComplete: boolean,
   isCurrentStep: boolean,
   step: string
-): { icon: React.SVGAttributes<SVGSymbolElement>; accessibleName: string } => {
+): {
+  icon: React.SVGAttributes<SVGSymbolElement>
```
gyfchong, line 20: Won't solve your render issue but should be a `JSX.Element`...?

packages/components/src/Workflow/subcomponents/Footer/components/ProgressStepper/ProgressStepper.tsx
```diff
@@ -76,11 +78,7 @@ export const ProgressStepper = ({
                 </Paragraph>
                 <div className={styles.stepIndicator}>
                   <span className={styles.stepIcon}>
-                    <Icon
-                      role="presentation"
-                      icon={stepStatus.icon}
-                      inheritSize
-                    />
+                    <>{stepStatus.icon}</>
```
gyfchong, line 81: Maybe it's worth experimenting with `{ accessibleName, icon: Icon } = getStepStatus(...)`

And using `<Icon />`

packages/components/src/Workflow/subcomponents/Footer/components/ProgressStepper/ProgressStepper.tsx
mcwinter07: Oooo yep I should update that type

packages/components/src/Workflow/subcomponents/Footer/components/ProgressStepper/ProgressStepper.tsx
mcwinter07: Gone a little in between since the <Icon /> throws the ts error:
```JSX element type 'Icon' does not have any construct or call signatures.ts(2604)```

PR #3673, opened 2023-05-25
Add OptionalIntlProvider to KaizenProvider

JakePitman:
## Why
We want to add translations to our KZ components. This will require a provider from `@cultureamp/i18n-react-intl` to be present in context to avoid throwing errors. This component checks to see if a provider exists, and if it doesn't, wraps its children in one.
We are adding this to `KaizenProvider` as we will push for `KaizenProvider` to be a standard part of apps that consume KZ components in the future.

## What
- Creates the `OptionalIntlProvider` component
- Adds test coverage for `OptionalIntlProvider`
- Adds the `OptionalIntlProvider` to `KaizenProvider`


packages/components/src/KaizenProvider/OptionalIntlProvider/OptionalIntlProvider.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { IntlProvider } from "react-intl"
+
+jest.mock("@cultureamp/i18n-react-intl", () => ({
+  StaticIntlProvider: ({ children }: { children: React.ReactElement }) => (
+    <div data-testid="mockedIntlProvider">{children}</div>
+  ),
+}))
+
+describe("<OptionalIntlProvider>", () => {
+  it("adds an intl provider to the app's context when IntlContext is not present", async () => {
```
gyfchong, line 13:
```suggestion
  it("wraps the children in a StaticIntlProvider if no other react-intl provider is present", async () => {
```

packages/components/src/KaizenProvider/OptionalIntlProvider/OptionalIntlProvider.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <OptionalIntlProvider locale="en">
+        <div>children</div>
+      </OptionalIntlProvider>
+    )
+    const result = await screen.findByTestId("mockedIntlProvider")
+    expect(result).toBeInTheDocument()
+    const childrenResult = await screen.findByText("children")
+    expect(childrenResult).toBeInTheDocument()
+  })
+  it("does not render a new provider", async () => {
```
gyfchong, line 24:
```suggestion
  it("does not wrap the children in a StaticIntlProvider when a react-intl provider is present", async () => {
```

packages/components/src/KaizenProvider/KaizenProvider.spec.tsx
```diff
@@ -9,6 +9,12 @@ const KaizenProviderWrapper = ({
   <KaizenProvider {...restProps}>{children}</KaizenProvider>
 )
 
+jest.mock("./OptionalIntlProvider", () => ({
```
gyfchong, line 12: Only need to do this because RTL doesn't do shallow renders ie. it renders everything down to the last component in the tree, so in this case it is trying to render the `StaticIntlProvider` from `i18n-react-intl`

.changeset/curvy-steaks-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add OptionalIntlProvider to KaizenProvider
```
gyfchong, line 5: Add a bit of description as to why we're adding it and what it does

packages/components/rollup.config.mjs
```diff
@@ -43,6 +46,9 @@ const getCompiledConfigByModuleType = format => ({
     commonjs(),
     esbuild(),
     image(),
+    jsonPlugin(),
+    // These libraries aren't used in KAIO, but require polyfills in consuming repos
```
gyfchong, line 50: Need to explain a bit more as to why they're being ignored.

packages/components/package.json
```diff
@@ -37,6 +37,7 @@
     "ci:test:react-17": "USE_REACT_17=true yarn test:ci"
   },
   "dependencies": {
+    "@cultureamp/i18n-react-intl": "^1.1.8",
```
gyfchong, line 40: Probably make this a peer-dep

packages/components/package.json
```diff
@@ -84,6 +87,7 @@
     "typescript-transform-paths": "^3.4.6"
   },
   "peerDependencies": {
+    "@cultureamp/i18n-react-intl": "^1.1.8",
```
gyfchong, line 90: Ah probably can't be a peer, just bundle it in for now

JakePitman: nw

.changeset/curvy-steaks-drop.md
```diff
@@ -0,0 +1,8 @@
+---
+"@kaizen/components": minor
+---
+
+Add OptionalIntlProvider to KaizenProvider.
+With OptionalIntlProvider added this way, KaizenProvider will wrap its children 
+in a StaticIntlProvider from @cultureamp/i18n-react-intl if one is not present in
+the app's context already.
```
gyfchong, line 8: Whilst this is technically correct, this is to be read by our consumers, so we need to write what feature this unlocks, and not so much how it works.

We can still say "Add X" so I think in this case we should mention just the home package and then what the feature is.

.changeset/curvy-steaks-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@kaizen/components": minor
+---
+
+KaizenProvider will now add @cultureamp/i18n-react-intl's StaticIntlProvider to 
+the consuming app's context (if one isn't provided already).
```
gyfchong, line 6:
```suggestion
the consuming app (if one isn't provided already).
```

Include what it's for, imagine someone asking why do we need this?

gyfchong (approved):

PR #3685, opened 2023-05-31
feat(filterbar): add FilterBar component

HeartSquared:
## What

- Add a bare minimum FilterBar component (a container that houses static Filter components; no extra functionality)
- Add a compatible FilterSelect component
- Fix FilterSelect so `selectedKey` will work with a number value

## To do
- [x] Style the FilterBar
- [x] Improve the docs
- [x] Add changeset

## Why

[KDS-1427](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1427)
[Figma](https://www.figma.com/file/5Dj1k0f04ezsMftUcNuqc8/%F0%9F%A4%96-Custom-Patterns-and-Components?type=design&node-id=690-87492&t=5OoNrjUMdg8K942f-4)

[KDS-1427]: https://cultureamp.atlassian.net/browse/KDS-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -0,0 +1,139 @@
...
+const [values, setValues] = useState<Partial<ValuesMap>>({
+  filterId1: "default-value"
+})
+
+return (
+  <FilterBar<ValuesMap> values={values} {...rest} />
+)
+```
+
+### `onValuesChange`
```
gyfchong, line 129: Might be good to include a story here to show the output of the on change event, kinda like your validation stories.

packages/components/src/FilterBar/context/FilterBarContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onValuesChange: (values: Partial<ValuesMap>) => void
+}
+
+export const FilterBarProvider = <ValuesMap extends FiltersValues>({
+  children,
+  filters,
+  values,
+  onValuesChange,
+}: FilterBarProviderProps<ValuesMap>): JSX.Element => {
+  const [state, dispatch] = useReducer(
```
gyfchong, line 47: 😅 🤐

packages/components/src/FilterBar/FilterBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const FilterBar = <ValuesMap extends FiltersValues>({
+  filters,
+  classNameOverride,
+  ...providerProps
+}: FilterBarProps<ValuesMap>): JSX.Element => (
+  <FilterBarProvider<ValuesMap> filters={filters} {...providerProps}>
+    {(activeFilters): JSX.Element => (
+      <div className={classnames(styles.filterBar, classNameOverride)}>
+        {Object.values(activeFilters).map(({ id, Component }) => (
+          <React.Fragment key={id}>
```
gyfchong, line 25: I know we expect all the `id` given to be unique, but should we trust that? Would adding a short UUID on the end mess thing up?

HeartSquared: We should trust it. The `id` is used for more than just this key after all.

packages/components/src/FilterBar/context/filtersStateReducer.ts
```diff
@@ -0,0 +1,43 @@
...
+    case "update_values":
+      return Object.keys(state).reduce<FiltersState<ValuesMap>>((acc, key) => {
+        acc[key as keyof ValuesMap] = {
+          ...state[key],
+          value: action.values[key],
+        }
+        return acc
+      }, {} as FiltersState<ValuesMap>)
+
+    case "update_single":
```
gyfchong, line 37:
```suggestion
    case "update_single_filter":
```

Seems a bit ambiguous to imply a `filter` is what's being updated

packages/components/src/FilterBar/context/filtersStateReducer.ts
```diff
@@ -0,0 +1,43 @@
...
+      id: keyof ValuesMap
+      data: Partial<FilterState<keyof ValuesMap, ValuesMap>>
+    }
+
+export const filtersStateReducer = <ValuesMap extends FiltersValues>(
+  state: FiltersState<ValuesMap>,
+  action: Actions<ValuesMap>
+): FiltersState<ValuesMap> => {
+  switch (action.type) {
+    case "update_values":
```
gyfchong, line 28:
```suggestion
    case "update_all_values":
```

If I understood this correctly?

packages/components/src/FilterBar/context/filtersStateReducer.ts
gyfchong: Cause reading the one above on `values` makes me think this is also only for a value

packages/components/src/FilterBar/context/types.ts
```diff
@@ -0,0 +1,13 @@
+export type FiltersValues = Record<string, any>
+
+export type FilterState<Id, Value> = {
```
gyfchong, line 3: Are we trying to avoid using the word `FilterBar` in all of this for any particular reason?

HeartSquared: This is the State of a single Filter within the FilterBar. It would be `FilterBarFilterState` if you want to include it lol.

packages/components/src/FilterBar/subcomponents/FilterSelect/FilterSelect.tsx
```diff
@@ -0,0 +1,48 @@
...
+  id,
+  onSelectionChange,
+  ...props
+}: FilterBarSelectProps<Option>): JSX.Element => {
+  const { getFilterState, toggleOpenFilter, updateValue } =
+    useFilterBarContext<Option["value"]>()
+
+  if (!id) throw Error("Missing `id` prop in FilterBarSelect")
+
+  const filterState = getFilterState(id)
```
gyfchong, line 28: Thoughts on sending the `id` directly to `useFilterBarContext`? Or perhaps rename `useFilterBarContext` to `useFilterState`?

HeartSquared: Renaming the hook isn't a good idea because it's a context, and you don't want anyone confusing it to be similar to a `useState`. More util functions will be coming out of the context later, so even if it looks like it matches now, it won't later.

Sending the `id` to the context hook probably wouldn't work, since it's a context consumer. Well, I suppose it _could_ work, but it means we'd need to have it call all the functions provided by the provider, so essentially it's an almost-duplicate 🤔
The problem is, that therefore ties all the functions to the provided `id`, and if you wanted to get values from another filter (eg. a Dependent Filter wants to check that a Source Filter has a value, and then it will open itself), you can't, unless I also provide you the original functions at the same time 😅

I'm not sure if it's worth the added complexity given the usage isn't that complicated.

gyfchong: Fair enough!

gyfchong (approved):

PR #3687, opened 2023-06-01
Remove workflow exit and update docs

mcwinter07:
## Why
The Workflow Exit was causing a build issue on Next repos. This was determined to be caused by the @kaizen/draft-modal components and its compilation in KAIO.

We have removed this component as it was an optional implementation for our consumers and have instead provided documentation on how to implement this within their solution.

There is also a minor change to the Wrapper to favor composability.

For more details on the issue, see the [KZ thread](https://cultureamp.slack.com/archives/C02NUQ27G56/p1685504632633019) where we discuss the problem

## What
- remove dep on @kaizen/draft-modal
- remove WorkflowExit component from Workflow pkg
  - Update reference in docs and provide implementation example
- Move Workflow Wrapper div into exportable component
  - Move styles into Wrapper folder


packages/components/src/Workflow/_docs/Workflow.mdx
```diff
@@ -74,29 +72,44 @@ While the number of JSX element is not limited, is important to consider the rea
...
+  return (
+    <>
+      <Button
+        label={exitLabel}
+        icon={closeIcon}
+        iconPosition="end"
+        secondary
+        onClick={() => {
+          setShowModal(true)
+        }}
```
HeartSquared, line 92:
```suggestion
        onClick={() => setShowModal(true)}
```

Just a suggestion, if you wanted to one-line it :)

packages/components/src/Workflow/subcomponents/Wrapper/Wrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+  OverrideClassName<HTMLAttributes<HTMLDivElement>>
+>
+
+export const Wrapper = ({
+  children,
+  classNameOverride,
+  ...restProps
+}: WorkflowWrapperProps): JSX.Element => (
+  <div
+    className={classnames([styles.wrapper, classNameOverride])}
```
HeartSquared, line 15:
```suggestion
    className={classnames(styles.wrapper, classNameOverride)}
```
Array not necessary 😉

packages/components/src/Workflow/subcomponents/Wrapper/Wrapper.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { PropsWithChildren, HTMLAttributes } from "react"
+import classnames from "classnames"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./Wrapper.module.scss"
+export type WorkflowWrapperProps = PropsWithChildren<
+  OverrideClassName<HTMLAttributes<HTMLDivElement>>
+>
+
```
mcwinter07, line 8: This update is less around the fix but a small DX improvement/continuity with the Main component we export that I thought we should provide

packages/components/src/Workflow/subcomponents/Wrapper/Wrapper.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { PropsWithChildren, HTMLAttributes } from "react"
+import classnames from "classnames"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./Wrapper.module.scss"
+export type WorkflowWrapperProps = PropsWithChildren<
+  OverrideClassName<HTMLAttributes<HTMLDivElement>>
+>
```
HeartSquared, line 7:
```suggestion
export type WorkflowWrapperProps = OverrideClassName<HTMLAttributes<HTMLDivElement>> & {
  children: React.ReactNode
}
```

I'd probably why away from helper types as they can change between React versions. Better to be explicit (it's why we got rid of React.FC).

mcwinter07: Fair call, better to be explicit and safe

packages/components/src/Workflow/Workflow.tsx
```diff
@@ -29,8 +28,8 @@ const WorkflowComponent = ({
   classNameOverride,
   ...restProps
 }: WorkflowProps): JSX.Element => (
-  <div
-    className={classnames([styles.workflowWrapper, classNameOverride])}
+  <Workflow.Wrapper
+    classNameOverride={classnames({ classNameOverride })}
```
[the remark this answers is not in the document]
mcwinter07: Ah yes, completely unnecessary 😆

packages/components/package.json
```diff
@@ -44,7 +44,6 @@
     "@kaizen/typography": "^2.3.10",
     "@kaizen/draft-tag": "^3.4.16",
     "@kaizen/brand": "^1.5.8",
-    "@kaizen/draft-modal": "^10.6.19",
```
HeartSquared, line 47: Btw why aren't these alphabetical 😛

.changeset/lucky-zebras-serve.md
```diff
@@ -0,0 +1,9 @@
+---
+"@kaizen/components": minor
+---
+
+- remove dep on @kaizen/draft-modal
```
HeartSquared, line 5:
```suggestion
- remove dep on @kaizen/draft-modal as the styles causes compile to fail
```

Something like that so it doesn't sound like we're just removing it to go backwards :P

packages/components/package.json
mcwinter07: It's a secret alphabet... where "t" comes before "b"... Now you're in the inner circle 😈

packages/components/src/Workflow/Workflow.tsx
```diff
@@ -29,8 +28,8 @@ const WorkflowComponent = ({
   classNameOverride,
   ...restProps
 }: WorkflowProps): JSX.Element => (
-  <div
-    className={classnames([styles.workflowWrapper, classNameOverride])}
+  <Workflow.Wrapper
+    classNameOverride={classnames(classNameOverride)}
```
HeartSquared, line 32: Ah, I don't think I was clear before.

Either remove the whole line and have `classNameOverride` just be part of `...restProps`

```
  ...restProps
  }: WorkflowProps): JSX.Element => (
    <Workflow.Wrapper {...restProps}>  	
```

Or if you want to keep it destructured, then remove `classnames`

```
  classNameOverride,
  ...restProps
  }: WorkflowProps): JSX.Element => (
    <Workflow.Wrapper
      classNameOverride={classNameOverride}
      {...restProps}
    >  	
```

.changeset/lucky-zebras-serve.md
mcwinter07: I hummed and had on that one, I'm never really sure how much context to put in a changeset vs what in the PR description. Happy to do so but feel like we need that doc on how best to write a changeset, least for our team

packages/components/src/Workflow/subcomponents/Wrapper/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./Wrapper"
```
HeartSquared, line 1: Thoughts on renaming these `index.tsx` files to `index.ts` while you're here? 😄

.changeset/lucky-zebras-serve.md
HeartSquared: Yep, that definitely came up in WoW as something we need to figure out. I guess I'd say it should list the change made and how it solves the problem? It's a fine line though, so we definitely need some sort of convention 😬

packages/components/src/Workflow/subcomponents/Wrapper/index.tsx
mcwinter07: Could hurt to tackle that in one fell swoop

HeartSquared: 💔

packages/components/src/Workflow/Workflow.tsx
mcwinter07: Right, read you loud and clear now! I think I'll opt for the second one, to make it clear if anyone were to look at the src code

packages/components/src/Workflow/subcomponents/Wrapper/index.tsx
mcwinter07: ** couldn't 😆 Should be good for another review now

packages/components/package.json
mcwinter07: Couldn't unsee it so I've fixed that now 😆

HeartSquared (approved):

PR #3693, opened 2023-06-02
Manual-release

gyfchong:


.buildkite/pipeline.yaml
```diff
@@ -14,7 +14,7 @@ steps:
   - name: ":butterfly: Release"
     command: ".buildkite/scripts/release.sh"
     timeout_in_minutes: 15
-    if: build.branch == "main" && build.author.name == "github-actions[bot]"
+    if: build.branch == "main" && (build.author.name == "github-actions[bot]" || build.pull_request.labels includes "danger-manual-release")
```
HeartSquared, line 17: This seems...._dangerous_.

Side note, you've got conflicts in this PR, and you just added the label? Doesn't that mean it will explode?

gyfchong: Hmm it won't explode, but actually yeah this wouldn't work either... Cause as soon as you merge, it might not know what the PR's label was...

gyfchong: Man I hate BuildKite lol

mcwinter07: Wonder if you can clamp down who can add what label 🤔 The main problem this is trying to solve is that we don't have a good way to trigger a rerelease if the initial one failed

PR #3697, opened 2023-06-02
Docs storybook updates

christian314159:
## Why
- That component list is just soooooooooooooooo long

## What
- Grouped some of the components


gyfchong (approved):

packages/components/src/Filter/_docs/Filter.stories.tsx
```diff
@@ -5,7 +5,8 @@ import { FilterButton, FilterButtonRemovable } from "~components/FilterButton"
 import { Filter, FilterContents } from ".."
 
 export default {
-  title: "Components/Filter",
+  tags: ["autodocs"],
```
HeartSquared, line 8: You need to remove the `tags: ["autodocs"]` from the updated Filter docs (the ones I added `mdx` files for)

gyfchong (approved):

PR #3698, opened 2023-06-02
chore(storybook): add NoClipCanvas component

HeartSquared:
## What

Add a `NoClipCanvas` component; a visual replica of `Canvas` where popover doesn't get hidden.

**Note:** This component lacks the toolbar that appears above on controllable examples, but Controls still works.

- Update FilterPopover to use position `absolute` instead of `fixed`, as it fixes the problem where the contents can get cut off in a browser and become inaccessible as the page will not scroll.
	- All Filter docs have been updated to use `CustomCanvas` to show properly again.
- Fix `FilterMultiSelect` and `Select` docs page not able to open selects (this is due to focus locks from examples that were open by default and unable to be closed)

## Why

Popovers get snipped by the Storybook Canvas component.

|Canvas|CustomCanvas|
|---|---|
|image|image|


storybook/components/CustomCanvas/CustomCanvas.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export type CustomCanvasProps = {
+  of: SourceProps["of"]
+  initialIsOpen?: boolean
+  className?: string
+}
+
+/**
+ * An almost-duplicate of Storybook's <Canvas>.
+ * Only use this when Canvas prevents your component rendering correctly (eg. using a popover).
```
gyfchong, line 15: Does it present a problem in some cases? Why not use this for all components? I'd rather we use one for all so we avoid mixed communication.

storybook/components/CustomCanvas/CustomCanvas.tsx
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const CustomCanvas = ({
```
gyfchong, line 17: Thinking we just name this `Canvas` or if we're worried about being confused with Storybook's `KaizenCanvas`? While it is a custom implementation, I'm not sure if it's important to name it such if it can be used generally.

storybook/components/CustomCanvas/CustomCanvas.tsx
HeartSquared: It doesn't include the toolbar with the zoom options etc 😬 Given that we don't allow people to see individual Stories nowadays through the site, I figured it's best to leave them using the base?

storybook/components/CustomCanvas/CustomCanvas.tsx
HeartSquared: Yeah I just wanted to give it a different name so it's easy to differentiate in the case you want to use both types in your docs somehow.

Happy to name it `KaizenCanvas` :)

storybook/components/CustomCanvas/CustomCanvas.tsx
gyfchong: I don't know how important zoom is though, I'd be happy to make the bet that no one uses it 😄

HeartSquared: We saw Jason using it the other day in a call 😂

storybook/components/CustomCanvas/CustomCanvas.tsx
gyfchong: I guess better to leave the door open than close it!

storybook/components/CustomCanvas/CustomCanvas.tsx
gyfchong: Haha then maybe we call this the `NoClipCanvas` or something, specifically to resolve the popover issue and help someone decide when to use it.

HeartSquared: Smart! Let's go with that :D

gyfchong (approved):

PR #3703, opened 2023-06-05
Improve-test-scripts

gyfchong:
## Why
KAIO wasn't included in the local test run and ci scripts needed to have double `run` definitions to cover KAIO. Making it hard to understand/run the same scripts locally to test.

## What
- Moved all test related CI scripts in package json
- Added KAIO to all test scripts

In summary, I don't think the GH actions are a good place to orchestrate our scripts, rather it's there to just run them


packages/components/package.json
```diff
@@ -21,17 +21,20 @@
...
+    "build:components": "rollup -c",
     "compile": "tsc",
     "dist:clean": "rm ./dist/tailwind.css ./dist/raw-styles.css",
     "dist:combine-styles": "concat-cli -f ./dist/esm/*.css ./dist/*.css -o ./dist/raw-styles.css",
     "dist:copy-tailwind": "cp styles/tailwind.css dist/tailwind.css",
     "dist:postcss": "postcss dist/raw-styles.css --output dist/styles.css",
     "postBuild": "yarn dist:copy-tailwind && yarn dist:combine-styles && yarn dist:postcss",
-    "prepublish": "yarn build:components && yarn postBuild && yarn dist:clean",
-    "test": "yarn jest"
+    "jest:ci": "jest --ci",
```
HeartSquared, line 35: Why not `"test:ci": "yarn test --ci"`?

HeartSquared (dismissed):

packages/components/package.json
gyfchong: Ah can do

HeartSquared (approved):

PR #3704, opened 2023-06-05
Make alt text optional on all illustration components

dougmacknz:
## Why
There's no reason to have this prop required because 99.9% of usages should just have a blank string.

I'm hypothesizing that having this prop required is causing a lot of people to think they should be writing alt text for our illustration components. Devs that aren't super across a11y may not be aware of the validity of a blank string and because this prop is required, think that they should write something for it.

## What
Make the `alt` prop optional on all of our illustration components.


HeartSquared (dismissed):

draft-packages/illustration/docs/IllustrationScene.stories.tsx
```diff
@@ -133,7 +133,7 @@ const IllustrationScenesTemplate: StoryFn<IllustrationScenesTemplateProps> = ({
   sceneComponents,
   ...restArgs
 }: IllustrationScenesTemplateProps) => {
-  const { isAnimated, loop, autoplay, alt = "", ...restProps } = restArgs
+  const { isAnimated, loop, autoplay, alt: _, ...restProps } = restArgs
```
HeartSquared, line 136:
```suggestion
  const { isAnimated, loop, autoplay, ...restProps } = restArgs
```

We want to keep it as part of `restProps` right?

gyfchong, line 136: If it's optional, people would assume it still gets added to the component. I foresee people wondering why it's not being added and tests perhaps failing cause they are looking to the `alt` on the illustration...

dougmacknz: The problem is that there's a discriminated union happening, animated ones don't take alt text. So if I don't pull this out, I end up with a TS error below because it ends up getting added to the animated example.

dougmacknz: It does still get added to the `img`, if that's what you mean? This is just the story.

HeartSquared (approved):

PR #3706, opened 2023-06-06
Add filterbarmultiselect

gyfchong:
## Why
Consumers need a FilterMultiSelect in the FilterBar

## What
Add `FilterBarMultiSelect`


packages/components/src/FilterBar/_docs/FilterBar.stories.tsx
```diff
@@ -16,6 +17,8 @@ const meta = {
 export default meta
 
 const sampleCode = `
+import { FilterMultiSelect } from "@kaizen/select"
```
HeartSquared, line 20: Sadness

packages/components/src/FilterBar/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from "./FilterBar"
+export * from "./subcomponents"
```
HeartSquared, line 2: We have dot notation so that we don't export these directly 😬 What are you trying to export from these? Types?

packages/components/src/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -0,0 +1,77 @@
...
+  RootProps,
+  "isOpen" | "setIsOpen" | "renderTrigger" | "label" | "selectedKey" | "trigger"
+> & {
+  id?: string
+}
+
+// This should technically be handled within the FilterMultiSelect
+// so that it only returns consumable formats instead of Set.
+// Someone fix please.
+export type ConsumableSelection = string | Key[] | undefined
```
HeartSquared, line 17: Given that `FilterBarDatePicker` and `FilterBarDRP` give `Date` and `DateRange` objects respectively, does it actually matter if we provide a `Set` to the consumer? They will probably need to do some conversion for the dates anyway to construct their query right? So asking them to convert the Set is probably fine too?

packages/components/src/FilterBar/index.ts
gyfchong: Oh oops I think I put this in the wrong place

HeartSquared (approved):

HeartSquared (dismissed):

HeartSquared (approved):

PR #3710, opened 2023-06-09
A11y: New animation for live tag [KDS-1523]

christian314159:
## Why
- Fixes cultureamp/kaizen-discourse#91

## What
- Replaced the pulse animation with an icon that cycles through 3 animation loops and then stops

image


gyfchong (dismissed):

jmitsios:
@christian314159 — I noticed a little rough and jagged edges around the icon after a moment of looking at it — so I zoomed in to check it out:
image

Are the 'lighter' lines remaining in position when the darker ones appear over the top of them, or are they disappearing? Is this double-up of shapes causing a harder fuzzy edge rather than a sharp and smooth one?

draft-packages/tag/docs/Tag.stories.tsx
```diff
@@ -164,3 +164,9 @@ StickerSheetReversed.parameters = {
   chromatic: { disable: false },
   controls: { disable: true },
 }
+
+export const LiveTag: StoryFn<typeof Tag> = args => (
```
dougmacknz, line 168: Do we definitely want another story for this? It's covered in the sticker sheet

.changeset/chilly-coats-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/draft-tag": minor
```
dougmacknz, line 2: Do we need a bump for component library as well for the icon being added?

christian314159: Yes, you're right, thanks!

draft-packages/tag/docs/Tag.stories.tsx
christian314159: No, we don't. Added it just for dev convenience.

gyfchong (approved):

PR #3725, opened 2023-06-14
set default month to the past in filterdatepicker stories to prevent chromatic changes

gyfchong:
## Why

DatePicker's current date shifts every day, causing chromatic to take a new snapshot 

## What

Place the default month in the past, so that the date never changes.

### DIFF
--- a/packages/components/src/FilterDatePicker/_docs/FilterDatePicker.stickersheet.stories.tsx
+++ b/packages/components/src/FilterDatePicker/_docs/FilterDatePicker.stickersheet.stories.tsx
@@ -97,6 +97,7 @@ const StickerSheetTemplate: StoryFn<{ textDirection: "ltr" | "rtl" }> = ({
               inputProps={{ labelText: "Date" }}
               locale="en-AU"
               selectedDate={dateValueValidation}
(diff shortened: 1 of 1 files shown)


packages/components/src/FilterDatePicker/_docs/FilterDatePicker.stickersheet.stories.tsx
```diff
@@ -97,6 +97,7 @@ const StickerSheetTemplate: StoryFn<{ textDirection: "ltr" | "rtl" }> = ({
               inputProps={{ labelText: "Date" }}
               locale="en-AU"
               selectedDate={dateValueValidation}
+              defaultMonth={new Date("01-01-2022")}
```
HeartSquared, line 100: Does this work? Usually it's `yyyy-mm-dd`

HeartSquared (approved):

packages/components/src/FilterDatePicker/_docs/FilterDatePicker.stickersheet.stories.tsx
gyfchong: Damn it, always trips me. But yes it does haha

PR #3726, opened 2023-06-14
Comment out OptionalIntlProvider in KaizenProvider

JakePitman:
## Why
We added this to KaizenProvider prematurely. OptionalIntlProvider will break without support for collated translation files.
We're waiting on a webpack plugin from Unified Home to make this possible.

## What
Removes OptionalIntlProvider from KaizenProvider (by temporarily commenting it out).


mcwinter07 (approved):
Looks hunky dory to me 👍 Looking forward to when we comment this back in 🤞

PR #3727, opened 2023-06-14
feat(filter-multi-select): copy FilterMultiSelect from kaizen/select

HeartSquared:
## Why

Using `FilterBar.MultiSelect` is missing icons for consumers.

## What

Add FilterMultiSelect into KAIO


yarn.lock
```diff
@@ -2272,13 +2272,28 @@
     "@formatjs/intl-localematcher" "0.2.32"
     tslib "^2.4.0"
 
+"@formatjs/ecma402-abstract@1.17.0":
```
HeartSquared, line 2275: `yarn.lock` changes are probably due to installing updated versions of the deps that were used by `@kaizen/select`.

Given the tests all still pass, I assume they weren't breaking.

packages/components/src/FilterMultiSelect/subcomponents/Trigger/FilterTriggerButton/FilterTriggerButton.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useRef } from "react"
+import { FilterTriggerRef } from "~components/Filter"
+import { FilterButton } from "~components/FilterButton"
```
gyfchong, line 3: Oh, didn't think to replace internally, nice one

gyfchong (approved):

PR #3740, opened 2023-06-16
feat: update heading to truncate when nowrap

NatalieLoyzaga:
## Why

Fix styles for nowrap to be added on heading and remove wrapping div


.changeset/loud-cars-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/draft-table": minor
```
gyfchong, line 2: Would be a `patch` if we let it through ;P

gyfchong (approved):

PR #3758, opened 2023-06-18
patch: Submit typo

kaydbwithers:
## Why
In the Kaizen Storybook, the button text reads "Sumbit". It should be "Submit".

## What
This PR fixes the Submit typo.

### DIFF
--- a/.changeset/cold-ties-guess.md
+++ b/.changeset/cold-ties-guess.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/button": patch
+---
+
+Submit typo
--- a/.changeset/cold-ties-guess.md
+++ b/.changeset/cold-ties-guess.md
@@ -1,5 +1,0 @@
----
-"@kaizen/button": patch
----
-
-Submit typo
--- a/packages/button/docs/Button.stories.tsx
+++ b/packages/button/docs/Button.stories.tsx
@@ -169,7 +169,7 @@ export const NativeFormButton: StoryFn = () => (
       <input type="text" defaultValue="content" />
     </form>
     <Button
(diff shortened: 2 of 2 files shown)


.changeset/cold-ties-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/button": patch
```
HeartSquared, line 2: No need for a changeset (well rather, make this empty) since it's just fixing a story, therefore nothing to release :)

Thanks for the fix!

.changeset/cold-ties-guess.md
```diff
@@ -0,0 +1,5 @@
+---
```
christian314159, line 1: Hey @kaydbwithers, you can remove this changeset file. The PR only includes changes to a Story, so we wouldn't necessarily want to release a new version. The Storybook deploy should pick this up just fine.

I don't think our CI requires a Changeset file to be present (yet). Otherwise, `yarn changeset --empty` creates an empty one (or delete this one's contents).

christian314159 (changes requested):

.changeset/cold-ties-guess.md
christian314159: Haha, just saw @HeartSquared also replied with the same ;)

.changeset/cold-ties-guess.md
kaydbwithers: Thanks @HeartSquared, that's done now.

HeartSquared (approved):

christian314159 (approved):

PR #3760, opened 2023-06-19
Add role=combobox to Select

dougmacknz:
## Why
Recommendation from Intopia to add this role to better convey what this is to screen readers.

## What
Adds `role="combobox"` to the `Select` button, plus `aria-controls`, linking it with the list box


HeartSquared (approved):
Does this need to apply to `FilterSelect` too?

HeartSquared (approved):

PR #3761, opened 2023-06-19
Update Steps to array of objects

mcwinter07:
## Why
After the initial release we discussed the potential brittleness of use just the current Step's name to index its position in the Progress Stepper and potential lack of scalability in future development. This PR was created off the back of that discussion (see comments the initial [SR PR](https://github.com/cultureamp/performance-review-cycles-ui/pull/257#discussion_r1217544830) to integrate the Workflow).

## What
Creates a future release for workflow
- updates the stepName to `currentStepId`
- updates steps from non empty array of strings to non empty array of `Steps`
  - exports Step and Steps types for convenience in consuming repos
-  refactor Workflow to use the new type and id to reference active steps from array
- adds test coverage to ProgressStepper


gyfchong:
This looks good, but it's technically a breaking change...

mcwinter07:
> This looks good, but it's technically a breaking change...
This is true, though I didn't want to have to force a breaking change to KAIO 😬  I've flagged it with SR and was planning on releasing as a minor and just making sure they're across the change before I merge it in, same with Jared over with engagement

gyfchong:
Perhaps add the new version of Workflow to the future directory, deprecate the existing one and get them all to shift to the future one instead. I know it seems like a lot to go through for 2 customers, but I don't want to get into the habit of sneaking breaking changes into minor bumps, just cause we can manually let people know.

mcwinter07:
> Perhaps add the new version of Workflow to the future directory, deprecate the existing one and get them all to shift to the future one instead. I know it seems like a lot to go through for 2 customers, but I don't want to get into the habit of sneaking breaking changes into minor bumps, just cause we can manually let people know.

Fair enough, gives us a chance to trial out the new future folder too

.changeset/itchy-sloths-fold.md
```diff
@@ -0,0 +1,11 @@
...
+"@kaizen/components": minor
+---
+
+Updates steps to object and stepName to currentStepId
+
+- updates the stepName to currentStepId
+- updates steps from non-empty array of strings to non-empty array of Steps
+  - exports Step and Steps types for convenience in consuming repos
+- refactor Workflow to use the new type and id to reference active steps from array
+- adds test coverage to ProgressStepper and improve docs
```
gyfchong, line 11: I'd update this to include the name of the component at the top and a note about it's location in `future`

packages/components/src/__future__/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./ExampleComponent"
```
gyfchong, line 1: We can delete the `ExampleComponent` now 😄

mcwinter07: Yay deleting things! 🕺

packages/components/src/Workflow/_docs/Workflow.mdx
```diff
@@ -15,11 +16,17 @@ import * as WorkflowHeader from "./WorkflowHeader.stories"
 
 <KaioNotification />
 
+<InlineNotification type="negative" persistent>
+  {`This version of the Workflow is deprecated and will be updated in the next major release.
+  Import from "@kaizen/components/future" for the latest version.`}
```
gyfchong, line 21:
```suggestion
  {`This version of the Workflow is deprecated and will be removed in the next major release.
  Import from "@kaizen/components/future" for the latest version.`}
```

mcwinter07: heheh um'd and ah'd on that exact word 😆

packages/components/src/__future__/Workflow/_docs/Workflow.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+# Workflow
+
+<ResourceLinks
+  sourceCode="https://github.com/cultureamp/kaizen-design-system/tree/main/packages/components/src/Workflow"
+  className="!mb-8"
+/>
+
+<KaioNotification />
+
```
gyfchong, line 17: Might want to pop an info in here to tell people that if they're running [insert older version] of KAIO where to find the docs. Just to make it complete 😄

gyfchong:
You'll need to keep the original `Workflow` files around and add a deprecation notice. The time to delete it is in the major.

mcwinter07:
Hey @gyfchong do you think I should a add the `@deprecated` to the non-future version of workflow? Just want to make sure that's not confusing to the user.

Screenshot 2023-06-20 at 4 46 12 pm

Screenshot 2023-06-20 at 4 46 14 pm

gyfchong:
Yeah, we deprecate the non-future to make sure people upgrade to future

packages/components/src/Workflow/Workflow.tsx
```diff
@@ -45,6 +45,8 @@ const WorkflowComponent = ({
   </Workflow.Wrapper>
 )
 
+/**
+ * @deprecated Use { Workflow } from `@kaizen/components/future` */
```
gyfchong, line 49:
```suggestion
 * @deprecated Use { Workflow } from `@kaizen/components/future`
 */
```

gyfchong: 😅

packages/components/src/__future__/Workflow/_docs/Workflow.mdx
```diff
@@ -0,0 +1,172 @@
...
+<InlineNotification type="cautionary" persistent>
```
gyfchong, line 19:
```suggestion
<InlineNotification type="informative" persistent>
```

I don't want this to blend with the KAIO notification, I think informative would be more appropriate anyways.

packages/components/src/__future__/Workflow/_docs/Workflow.mdx
```diff
@@ -0,0 +1,172 @@
...
+  {`This is the latest version of the Workflow component. If you are using "@kaizen/components" at 1.16.0 or lower, please refer to our legacy documentation.`}
```
gyfchong, line 20:
```suggestion
  {`If you are using "@kaizen/components" at 1.16.0 or lower, please refer to our `} <LinkTo pageId="pages-workflow-legacy">legacy documentation.</ LinkTo>
```
Everything is always "the latest version" 😉

packages/components/src/__future__/Workflow/_docs/Workflow.mdx
mcwinter07: True, the contrast is a little nicer on the eyes as well
Screenshot 2023-06-20 at 5 00 58 pm

packages/components/src/__future__/Workflow/_docs/Workflow.mdx
gyfchong: Also don't forget to import:
```
import { LinkTo } from "../../storybook/components/LinkTo"
```

mcwinter07: Actually had issues with the MDX renderer when I tried to use LinkTo as a child InlineNotification within MDX
Screenshot 2023-06-20 at 5 03 52 pm
I also tried just adding in anchor tag, but doesn't work too well either since the styles are inconsistent.

mcwinter07: That said looking at the KaioNotification it looks fine if you create the component outside of the MDX

gyfchong: Oh interesting, never mind then. I'll make a ticket to create a component.

mcwinter07: Yeah its an odd one, I assume the MDX parser has some issues Storybook's LinkTo component. I also tried the function `linkTo` to see if I could attach it to a button, but didn't have much luck with that either.

gyfchong (approved):

PR #3762, opened 2023-06-19
feat(filter-bar): enable removable filters

HeartSquared:
## What

Enable removable filters.

image

For consumers to use this functionality, they add `isRemovable` to the filter config.

- Removable filters have a Remove button to the right
- Removed filters (inactive) are put into the Add Filters list
- The Add Filters button is disabled when the list is empty
- Removable filters will default render in the Add Filters list UNLESS they have a value, of which they will render within the active filters after the non-removable ones
- Filters will list in the order they are added
- The Add Filters menu options are sorted in the order the engineer has listed the filters to be (not in the order they are removed)

## Why

[KDS-1470](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1470)

[KDS-1470]: https://cultureamp.atlassian.net/browse/KDS-1470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/FilterBar/context/reducer/setupFiltersState.ts
```diff
@@ -0,0 +1,24 @@
+import { Filters } from "../../types"
+import { FiltersState, FiltersValues } from "../types"
+
+export const setupFiltersState = <ValuesMap extends FiltersValues>(
+  filters: Filters<ValuesMap>
+): FiltersState<ValuesMap> =>
+  filters.reduce<FiltersState<ValuesMap>>(
+    (acc, filter) => {
```
gyfchong, line 8: Can we find a better name for `acc`?

packages/components/src/FilterBar/context/reducer/updateSingleFilter.ts
```diff
@@ -0,0 +1,22 @@
...
+export const updateSingleFilter = <ValuesMap extends FiltersValues>(
+  state: FiltersState<ValuesMap>,
+  id: keyof ValuesMap,
+  data: Partial<FilterState<keyof ValuesMap, ValuesMap[keyof ValuesMap]>>
+): FiltersState<ValuesMap> => {
+  const newFilterState = { ...state.filters[id], ...data }
+
+  if (data.isActive !== undefined) {
+    if (data.isActive) {
+      state.activeFilterIds.add(id)
```
gyfchong, line 12: Would it be easier to destructure state?

HeartSquared: I can try. Wasn't sure if it would throw an error if it was undefined (although now that I think about it, it should only care if `data` didn't exist, not `isActive`).

packages/components/src/FilterBar/context/reducer/setupFiltersState.ts
HeartSquared: Lol, but it's an accumulator! I actually like it, since it's a temp variable which I associate with `reduce`.

That being said, what would you prefer it to be?

packages/components/src/FilterBar/context/utils/getInactiveFilters.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+    Component: <div />,
+    isOpen: false,
+    isRemovable: true,
+    isActive: false,
+  },
+} satisfies FiltersState<Values>["filters"]
+
+const state = {
+  filters,
+  activeFilterIds: new Set<keyof Values>(["flavour"]),
```
gyfchong, line 31: Has this `Set` you free?

packages/components/src/FilterBar/FilterBar.spec.tsx
```diff
@@ -91,66 +111,182 @@ const FilterBarWrapper = (
...
-            onClick={() => setValues({ flavour: "honey-milk-tea" })}
-          >
-            Update Flavour to honey-milk-tea
-          </button>
-        </div>
+      const addFiltersButton = getByRole("button", { name: "Add Filters" })
+      await user.click(addFiltersButton)
+
+      const list = getByRole("list")
+      const menuOption = within(list).getByRole("button", { name: "Topping" })
```
gyfchong, line 135: `within` is fancy!

packages/components/src/FilterBar/FilterBar.spec.tsx
```diff
@@ -91,66 +111,182 @@ const FilterBarWrapper = (
...
+      const { getByRole } = render(
+        <FilterBarWrapper<ValuesRemovable>
+          filters={filtersRemovable}
+          defaultValues={{ topping: "pearls" }}
+        />
       )
-    }
+      expect(getByRole("button", { name: "Topping : Pearls" })).toBeVisible()
+      expect(
+        getByRole("button", { name: "Remove filter - Topping" })
```
gyfchong, line 151: Something I picked up in @mcwinter07's PR on Workflows is that he's switching to identifying things by `id` rather than than the labels of steps to avoid having to do i18n string matching when managing the current step.

Perhaps it would be better for us to apply this approach more broadly so that our micro-copy can safely change without the unit tests breaking.

packages/components/src/FilterBar/context/utils/getInactiveFilters.spec.tsx
HeartSquared: No joke, makes things so much simpler 😂

packages/components/src/FilterBar/FilterBar.spec.tsx
HeartSquared: We shouldn't be testing by targeting by `id`. Even if we have i18n implemented, I could just ensure these tests are all testing in English. If your translated text changes, then it makes sense for the test to fail. Changing the baked-in copy should be the same as changing a colour on a component. We should be alerted of it.

dougmacknz: Yeah +1 to selecting things by role and label instead of id, it ensures we're adding accessible names which is really important.

packages/components/src/FilterBar/FilterBar.spec.tsx
```diff
@@ -91,66 +111,182 @@ const FilterBarWrapper = (
...
+      const addFiltersButton = getByRole("button", { name: "Add Filters" })
+      await user.click(addFiltersButton)
+
+      const menuOptionIceLevel = getByRole("button", { name: "Ice Level" })
+      await user.click(menuOptionIceLevel)
+      await user.click(addFiltersButton)
+
+      const menuOptionFlavour = getByRole("button", { name: "Flavour" })
+      await user.click(menuOptionFlavour)
+      await user.click(addFiltersButton)
```
gyfchong, line 224: Looks tedious, I bet someone will want to be able to select multiple filters to bulk add at some point lol

packages/components/src/FilterBar/FilterBar.spec.tsx
```diff
@@ -91,66 +111,182 @@ const FilterBarWrapper = (
...
+              filters={filters}
+              values={values}
+              onValuesChange={setValues}
+            />
+            <button
+              type="button"
+              onClick={() => setValues({ flavour: "honey-milk-tea" })}
+            >
+              Update Flavour to honey-milk-tea
+            </button>
```
gyfchong, line 273: Instead of an actual button here, can't you just trigger `setValues` programmatically?

packages/components/src/FilterBar/FilterBar.spec.tsx
HeartSquared: That's not how the UI works Geoff 😂

packages/components/src/FilterBar/FilterBar.spec.tsx
HeartSquared: Yes, but this is replicating a user's behaviour. Ensuring that an external event works.

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -105,6 +106,16 @@ Available Filter components:
   - `onSelectionChange` remains available as optional
   - **Note:** To use the subcomponents, import `FilterSelect` from `@kaizen/components` (not `@kaizen/select`)
 
+##### `isRemovable`
+
+When this attribute is applied, the Filter will become removable (a remove button will be attached to the trigger button),
+and their value will be cleared upon remove.
```
christian314159, line 112:
```suggestion
and its value will be cleared upon removal.
```
I think the subject here is "the Filter"

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -105,6 +106,16 @@ Available Filter components:
...
+
+Removable Filters will be added to the `Add Filter` menu options.
```
christian314159, line 114: Do you mean that a Filter - once removed - will be available within in the "Add Filter" menu again? If so, this could be clearer ;)

packages/components/src/FilterBar/context/reducer/filtersStateReducer.ts
```diff
@@ -0,0 +1,24 @@
...
+      type: "update_single_filter"
+      id: keyof ValuesMap
+      data: Partial<FilterState<keyof ValuesMap, ValuesMap[keyof ValuesMap]>>
+    }
+
+export const filtersStateReducer = <ValuesMap extends FiltersValues>(
+  state: FiltersState<ValuesMap>,
+  action: Actions<ValuesMap>
+): FiltersState<ValuesMap> => {
+  switch (action.type) {
```
christian314159, line 17: 🤔 Just wondering, do `switch` statements not mandate a `default` option? Maybe that was just a linting rule in the last code base I worked in. Anyway ...

gyfchong:
Don't get too excited, your changeset is missing 😛

packages/components/src/FilterBar/context/reducer/setupFiltersState.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+  { id: "flavour", name: "Flavour", Component: <div /> },
+  {
+    id: "sugarLevel",
+    name: "Sugar Level",
+    Component: <div />,
+    isRemovable: true,
+  },
+] satisfies Filters<Values>
+
+describe("setupFiltersState", () => {
```
christian314159, line 21: Don't think that is yet written down anywhere, but when I last went and cleaned up tests, I tried to establish a convention where the main `describe` block follows this pattern:
- When a component is tested it says `"<Component />"`
- When it's a function it indicates so with `"myFunction()"`

packages/components/src/FilterBar/context/reducer/updateSingleFilter.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+} satisfies FiltersState<Values>["filters"]
+
+describe("filtersStateReducer: update_single_filter", () => {
+  it("updates state of a single filter", () => {
+    const state = {
+      filters: stateFilters,
+      activeFilterIds: new Set<keyof Values>(["flavour"]),
+    } satisfies FiltersState<Values>
+
+    expect(
```
christian314159, line 32: An observation only: I'm sure this test does properly test the functionality, but it's hard for me when reading it do understand really what exactly is going on here and compared.
It's using complex objects, splatting external state in, etc. - not very obvious.

If I'm reading this correctly, it's all about setting `isOpen` to `true` for the `Flavour` filter?

Could this be
- Assert `isOpen` is `false`
- Invoke reducer
- Assert `isOpen` is true
?
I get that it compares a lot more, and thus might be 'safer', but might not be the responsibility of this test either?

packages/components/src/FilterBar/context/reducer/setupFiltersState.spec.tsx
christian314159: Actually noticed you did this later on :)

christian314159 (commented):
Just some nitpicking from the cheap seats.

packages/components/src/FilterBar/_docs/FilterBar.mdx
HeartSquared: Yes, but also by default 😬 I'll see what I can do (otherwise a suggestion would be great!)

packages/components/src/FilterBar/context/reducer/filtersStateReducer.ts
HeartSquared: Not if you have TypeScript :D 
I could put a `default` in, but it's something we'd never reach without TS erroring on us so 🤷🏻

packages/components/src/FilterBar/context/reducer/setupFiltersState.spec.tsx
HeartSquared: Ah, must have missed it! Good spot!

packages/components/src/FilterBar/context/reducer/updateSingleFilter.spec.ts
HeartSquared: Hmm... Yeah I was going for the safer option, thus testing it as a whole.

Unless I add a test that takes in the state with no `data` and ensure it comes back exactly the same? The main thing is I want to ensure nothing goes missing in between.

In the case of the Active State, that does a little more than changing just the `filters` (`activeFilterIds` get updated too).

If there's at least one test that ensures things don't disappear (eg. if I go to update Flavour, I don't accidentally delete SugarLevel by forgetting to spread the rest of the filters back in), then I could simplify the others.

Thoughts?

.changeset/angry-pans-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add removable filters to FilterBar.
```
gyfchong, line 5: Which allows...? 😄

gyfchong (approved):

PR #3765, opened 2023-06-19
Remove gates on running chromatic

gyfchong:
## Why
Needing to add a label to run Chromatic is interrupting contributor workflows. Many PRs are being left to run CI only to return and find out chromatic has not been run and they need to continue to wait.

It also complicates our CI process, needing more than surface level information to understand why chromatic isn't running and how to enable it, rather than the process being _continuous_

## What
Due to the introduction of turbosnap which only runs chromatic for changed snapshots, I am removing the requirement for the `run chromatic` label that triggers the chromatic workflow, which means PRs will now run it automatically.

We will re-introduce this if costs do increase.


.github/workflows/chromatic.yaml
```diff
@@ -5,17 +5,13 @@ name: "chromatic"
...
       - changeset-release/main
   push:
 
 jobs:
   visual-testing:
-    if: >
-      github.ref == 'refs/heads/main' ||
-      github.event.label.name == 'run chromatic' ||
-      startsWith(github.event.head_commit.message, 'Merge branch ''main'' into')
+    if: github.ref != 'refs/heads/main'
```
HeartSquared, line 14: Nah, you need to run it on `main` so that it can automatically accept the new baselines :)

gyfchong: Ah good point

HeartSquared (approved):

PR #3766, opened 2023-06-19
feat(filter-bar): add Clear All button

HeartSquared:
## What

Add a Clear All button to the FilterBar.

All selected values are cleared, and any active removable filters are made inactive and moved into the Add Filters menu.

image

## Why

[KDS-1471](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1471)

[KDS-1471]: https://cultureamp.atlassian.net/browse/KDS-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/FilterBar/FilterBar.tsx
+++ b/packages/components/src/FilterBar/FilterBar.tsx
@@ -13,6 +13,7 @@ import {
   FilterBarSelect,
 } from "./subcomponents"
 import { AddFiltersMenu } from "./subcomponents/AddFiltersMenu"
+import { ClearAllButton } from "./subcomponents/ClearAllButton"
 import styles from "./FilterBar.module.scss"
 
 export type FilterBarProps<ValuesMap extends FiltersValues> = OverrideClassName<
@@ -27,13 +28,19 @@ export const FilterBar = <ValuesMap extends FiltersValues>({
   <FilterBarProvider<ValuesMap> filters={filters} {...providerProps}>
     {(activeFilters): JSX.Element => (
       <div className={classnames(styles.filterBar, classNameOverride)}>
-        {Object.values(activeFilters).map(({ id, Component }) => (
-          // `id` will always be `string`, but keyof ValuesMap transformed it
-          <React.Fragment key={id as string}>
-            {React.cloneElement(Component, { id })}
-          </React.Fragment>
-        ))}
-        <AddFiltersMenu />
+        <div className={styles.filtersContainer}>
+          {Object.values(activeFilters).map(({ id, Component }) => (
+            // `id` will always be `string`, but keyof ValuesMap transformed it
+            <React.Fragment key={id as string}>
+              {React.cloneElement(Component, { id })}
+            </React.Fragment>
+          ))}
+          <AddFiltersMenu />
+        </div>
+
+        <div>
+          <ClearAllButton />
+        </div>
       </div>
     )}
   </FilterBarProvider>
--- a/packages/components/src/FilterBar/context/FilterBarContext.tsx
+++ b/packages/components/src/FilterBar/context/FilterBarContext.tsx
@@ -15,6 +15,7 @@ export type FilterBarContextValue<
   showFilter: (id: keyof ValuesMap) => void
   hideFilter: (id: keyof ValuesMap) => void
   getInactiveFilters: () => Array<FilterState<keyof ValuesMap, Value>>
+  clearAllFilters: () => void
 }
 
 const FilterBarContext = React.createContext<FilterBarContextValue<any> | null>(
@@ -81,6 +82,10 @@ export const FilterBarProvider = <ValuesMap extends FiltersValues>({
       onValuesChange({ ...values, [id]: undefined })
     },
     getInactiveFilters: () => getInactiveFilters<ValuesMap>(state),
+    clearAllFilters: () => {
+      dispatch({ type: "clear_active_filters" })
+      onValuesChange({})
+    },
   } satisfies FilterBarContextValue<any, ValuesMap>
 
   useEffect(() => {
--- a/packages/components/src/FilterBar/context/FilterBarContext.tsx
+++ b/packages/components/src/FilterBar/context/FilterBarContext.tsx
@@ -83,7 +83,7 @@ export const FilterBarProvider = <ValuesMap extends FiltersValues>({
     },
     getInactiveFilters: () => getInactiveFilters<ValuesMap>(state),
     clearAllFilters: () => {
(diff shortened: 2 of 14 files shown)


packages/components/src/FilterBar/context/FilterBarContext.tsx
```diff
@@ -81,6 +82,10 @@ export const FilterBarProvider = <ValuesMap extends FiltersValues>({
       onValuesChange({ ...values, [id]: undefined })
     },
     getInactiveFilters: () => getInactiveFilters<ValuesMap>(state),
+    clearAllFilters: () => {
+      dispatch({ type: "clear_active_filters" })
```
HeartSquared, line 86: Technically this action just turns any active removable filters to inactive, but `deactivate_filters` sounded funny. Open to a clearer name if needed 😂

packages/components/src/FilterBar/context/reducer/clearActiveFilters.ts
```diff
@@ -0,0 +1,16 @@
+import { FiltersState, FiltersValues } from "../types"
+
+export const clearActiveFilters = <ValuesMap extends FiltersValues>(
+  state: FiltersState<ValuesMap>
+): FiltersState<ValuesMap> => {
+  state.activeFilterIds.forEach((id, _, ids) => {
```
HeartSquared, line 6: I do normally prefer writing things in an immutable way, but this seemed straight-forward enough. I'll update it if we ever make our linting yell at anything mutable 😄

packages/components/src/FilterBar/FilterBar.tsx
```diff
@@ -27,13 +28,19 @@ export const FilterBar = <ValuesMap extends FiltersValues>({
...
+            <React.Fragment key={id as string}>
+              {React.cloneElement(Component, { id })}
+            </React.Fragment>
+          ))}
+          <AddFiltersMenu />
+        </div>
+
+        <div>
+          <ClearAllButton />
+        </div>
```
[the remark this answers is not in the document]
HeartSquared: I had no idea what to name it, thus I left it for now 😂 It will likely have styles later 😬

packages/components/src/FilterBar/context/FilterBarContext.tsx
gyfchong: `deactivate_filters` sounds fine to me

packages/components/src/FilterBar/context/reducer/clearActiveFilters.ts
gyfchong: Sometimes `foreach` is more performant too

packages/components/src/FilterBar/context/FilterBarContext.tsx
HeartSquared: Haha alright! Sold!

packages/components/src/FilterBar/subcomponents/ClearAllButton/ClearAllButton.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react"
+import { Button } from "~components/Button"
+import { useFilterBarContext } from "../../context/FilterBarContext"
+import styles from "./ClearAllButton.module.scss"
+
+export const ClearAllButton = (): JSX.Element => {
+  const { clearAllFilters } = useFilterBarContext()
+
+  return (
+    <Button
```
HeartSquared, line 10: @dougmacknz would this need an `aria-controls` since it manipulates the filters?

packages/components/src/FilterBar/FilterBar.tsx
```diff
@@ -27,13 +28,19 @@ export const FilterBar = <ValuesMap extends FiltersValues>({
...
     {(activeFilters): JSX.Element => (
       <div className={classnames(styles.filterBar, classNameOverride)}>
-        {Object.values(activeFilters).map(({ id, Component }) => (
-          // `id` will always be `string`, but keyof ValuesMap transformed it
-          <React.Fragment key={id as string}>
-            {React.cloneElement(Component, { id })}
-          </React.Fragment>
-        ))}
-        <AddFiltersMenu />
+        <div className={styles.filtersContainer}>
```
gyfchong, line 31: Might need to become a live region to announce changes in the bar?

HeartSquared: Hmm... that's an interesting one. A polite one?

gyfchong: On second thought let's not complicate the screen reader experience on a hunch, perhaps the button's "Clear all filters" description is enough.

packages/components/src/FilterBar/context/FilterBarContext.tsx
```diff
@@ -81,6 +82,10 @@ export const FilterBarProvider = <ValuesMap extends FiltersValues>({
       onValuesChange({ ...values, [id]: undefined })
     },
     getInactiveFilters: () => getInactiveFilters<ValuesMap>(state),
+    clearAllFilters: () => {
```
gyfchong, line 85: Why not call it `deactivateFilters`?

HeartSquared: Because it also clears the values on top of deactivating :) It could be `clearFilters` if the `all` feels verbose?

gyfchong: Perhaps the action does too much if it both clears and deactivates?

HeartSquared: Nah, it's the correct action for the UX. It's only being called separately because `values` are held outside of the context 😅

gyfchong: I don't think the action needs to match the UX, we could have a `clearValues` and a `deactivateFilters` action , both of which are triggered in a single button. I think this makes the interactions easier to read and maintain eg. if we want to update how clearing works but not how deactivation works, it's easier to trace.

HeartSquared: Had a discussion, and I'm going to refactor the way the context gets the values. This should simplify feature implementation by having a single source of truth as opposed to two.

HeartSquared:
Superseded by https://github.com/cultureamp/kaizen-design-system/pull/3781

PR #3768, opened 2023-06-19
Add auth token to github packages

gyfchong:
I think the yarn install for changesets check needs extra permissions, hard to tell if this will fix the issue until it hits `main`.

Edit: Seems PR's are fine without this now...? Only merge if it becomes a problem again soon...


HeartSquared (approved):
https://github.com/cultureamp/kaizen-design-system/actions/runs/5317544911/jobs/9628171543?pr=3772

It un-fixed. Let's get this in.

PR #3773, opened 2023-06-20
Remove-lottie-web

gyfchong:
Lottie is heavy for what we use it for and animations aren't necessary within the modals. Removing animations until we find a lighter alternative.


.changeset/quick-kids-collect.md
```diff
@@ -0,0 +1,6 @@
+---
+"@kaizen/draft-illustration": major
+---
+
+- Remove animated Spot illustrations, defaulting to Static illustrations.
+- Deprecated `isAnimated` prop on Spot illustrations
```
gyfchong, line 6: Hmm if it's a major, I guess no point making it deprecated other than telling people why

HeartSquared: Yeah. That's why I said if we deprecate it, it should be minor lol.

HeartSquared (approved):

PR #3780, opened 2023-06-20
refactor(filter-bar): reduce duplication in context state values

HeartSquared:
## What / Why

Clean up the FilterBar context state values to prevent the need for double handling.
New features should therefore require less code 🤞🏻


packages/components/src/FilterBar/context/reducer/filterBarStateReducer.ts
```diff
@@ -0,0 +1,39 @@
...
+        filters: updateSingleFilter(state, action.id, action.data),
+      }
+
+    case "activate_filter":
+      state.activeFilterIds.add(action.id)
+      return { ...state }
+
+    case "activate_filters_with_values":
+      Object.keys(action.values).forEach(id => {
+        if (action.values[id]) state.activeFilterIds.add(id)
```
gyfchong, line 31: Can this dispatch the activate filter action to avoid the repeat?

gyfchong (approved):
Just a nitpick, I think we can probably revisit any other time as part of other work if we find it's an actual issue

PR #3781, opened 2023-06-20
feat(filter-bar): add Clear All button

HeartSquared:
## What

Add a Clear All button to the FilterBar.

All selected values are cleared, and any active removable filters are made inactive and moved into the Add Filters menu.

image

## Why

[KDS-1471](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1471)

[KDS-1471]: https://cultureamp.atlassian.net/browse/KDS-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/FilterBar/context/reducer/filterBarStateReducer.ts
+++ b/packages/components/src/FilterBar/context/reducer/filterBarStateReducer.ts
@@ -1,7 +1,7 @@
 import { FilterBarState, FiltersValues, InternalFilterState } from "../types"
 import { updateSingleFilter } from "./updateSingleFilter"
 
-type Actions<ValuesMap> =
+type Actions<ValuesMap extends FiltersValues> =
   | {
       type: "update_single_filter"
       id: keyof ValuesMap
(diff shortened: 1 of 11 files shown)


packages/components/src/FilterBar/context/reducer/filterBarStateReducer.ts
```diff
@@ -35,5 +39,12 @@ export const filterBarStateReducer = <ValuesMap extends FiltersValues>(
     case "deactivate_filter":
       state.activeFilterIds.delete(action.id)
       return { ...state }
+
+    case "clear_all_filters":
+      state.activeFilterIds.forEach(id => {
+        if (state.filters[id].isRemovable) state.activeFilterIds.delete(id)
```
gyfchong, line 45: Should this logic not also be part of the deactivate action? You shouldn't be able to deactivate non removable filters right?

And then, are you able to then dispatch the deactivate_filter action here rather than repeating? I'm imagining that this action would be better by dispatching a combo of other actions.

gyfchong (approved):
Same with the refactor, I think the actions can be adjusted if we need to, I think it's setup well enough to change if necessary

gyfchong (approved):

PR #3782, opened 2023-06-20
fix(FilterMultiSelect): correct RemovableFilterTrigger border-radius shorthand

BrendanTimmons:
## Why
Border radius when using the removable multi select component had the incorrect shorthand resulting in the wrong corners being rounded / not rounded.
[Ticket](https://cultureamp.atlassian.net/browse/TCP-168)

## What
Fix the shorthand border-radius syntax so that the left 2 borders on the select and the right 2 borders on the remove button are rounded and the "inner" corners where the two components touch are not rounded.

Fixed:
Screenshot 2023-06-21 at 9 44 27 am

Broken:
Screenshot 2023-06-21 at 9 46 24 am


gyfchong (approved):
Cheers!

PR #3786, opened 2023-06-22
update the future release /dist reference path in package.json

mcwinter07:
## Why
The dist reference paths did not align to the outputs and we could not import components from the future folder.
```
{
  "main": "../dist/__future__/index.js",
  "module": "../dist/esm/__future__/index.js",
  "types": "../dist/__future__/index.d.ts"
}
```

This is a small fix to update the paths to the `future.js` file in the esm and cjs folders, and the types pulled from `esm/dts/__future__`

### Before
Screenshot 2023-06-22 at 11 35 18 am

### After
Screenshot 2023-06-22 at 11 34 34 am

## What
- updates `@kaizen/components` future package.json

### DIFF
--- a/packages/components/future/package.json
+++ b/packages/components/future/package.json
@@ -1,5 +1,5 @@
 {
-  "main": "../dist/__future__/index.js",
-  "module": "../dist/esm/__future__/index.js",
-  "types": "../dist/__future__/index.d.ts"
+  "main": "../dist/cjs/future.js",
+  "module": "../dist/esm/future.js",
+  "types": "../dist/esm/dts/__future__/index.d.ts"
 }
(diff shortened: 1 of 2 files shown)


packages/components/future/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "main": "../dist/__future__/index.js",
-  "module": "../dist/esm/__future__/index.js",
-  "types": "../dist/__future__/index.d.ts"
+  "main": "../dist/cjs/future.js",
+  "module": "../dist/esm/future.js",
+  "types": "../dist/esm/dts/__future__/index.d.ts"
```
gyfchong, line 4: This is in a weird place though lol

gyfchong (approved):

packages/components/future/package.json
mcwinter07: Yeah I figured we could take a crack at updating the output path or creating a future.d.ts when we loop back to the tooling health day :)

PR #3788, opened 2023-06-23
feat(filter-bar): add functionality for dependent filters

HeartSquared:
## What

Add functionality for dependent filters.

- New filter attribute `isUsableWhen` (`(state) => boolean`) added for consumers
	- Filters failing to meet their condition will be removed from the FilterBar options (both active/inactive) and their value will be cleared
- `isUsable` added to state
- `values` added to state
	- `(consumer) values` and `state.values` must be able to change each other
		- In the case that the filter is deemed unusable, but the consumer has passed in a value, we need to clear it
		- We still need to allow consumers to change values from an external event

## Why

[KDS-1472](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1472)
Some filters should only be shown in the FilterBar (both active/inactive) when certain conditions are met.

[KDS-1472]: https://cultureamp.atlassian.net/browse/KDS-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/FilterBar/FilterBar.spec.tsx
```diff
@@ -233,6 +258,177 @@ describe("<FilterBar />", () => {
...
+        const menuOption = within(list).getByRole("button", { name: "Topping" })
+
+        await waitFor(() => {
+          expect(menuOption).toBeVisible()
+        })
+      })
+    })
+
+    describe("Condition result change", () => {
+      it("clears the value for an unusable filter", async () => {
```
gyfchong, line 352: Is this saying that you can't set a default for unusable filters?

packages/components/src/FilterBar/context/FilterBarContext.tsx
```diff
@@ -80,33 +77,38 @@ export const FilterBarProvider = <ValuesMap extends FiltersValues>({
...
           dispatch({ type: "deactivate_filter", id })
       })
-      onValuesChange({})
+      dispatch({ type: "update_values", values: {} })
     },
   } satisfies FilterBarContextValue<any, ValuesMap>
 
+  const shouldUpdateValues = filters.some(
+    ({ id }) => state.values[id] !== values[id]
+  )
```
gyfchong, line 102: Maybe make this a bit more pure and pass it the state?

packages/components/src/FilterBar/context/reducer/setupFiltersState.ts
```diff
@@ -1,21 +1,45 @@
...
     },
-    { filters: {}, activeFilterIds: new Set() } as FilterBarState<ValuesMap>
+    {
+      filters: {},
+      activeFilterIds: new Set(),
+      values: {},
+    } as FilterBarState<ValuesMap>
   )
+
+  const filtersState = transformToFiltersState(state, values)
```
gyfchong, line 30: How come this is a two step process? Why not extend the filter states in the first reducer?

HeartSquared: Terrible isn't it?

Since the consumer should be able to make a conditional out of the current state and any of its attributes, I need to combine state and values (which may be changed from outside, so not the one from state) together to then pass into `isUsableWhen` to get an accurate calculation.

Using the result from that, I then have to update the state to actually reflect what it should be (eg. deactivate and clear the value for a filter because it's actually now no longer usable).

That being said, I moved things around so many times that if duplicating everything into a single state makes more sense I could look into it. I'll just need to filter out some attributes to the user (eg. I can't have them calling `isUsableWhen` from `state` in their `isUsableWhen` cos I'll end up in a loop).

To briefly explain why some things have been split out:
- `activeFilterIds` is required to keep the sorting order for rendering. I had opted to calculate the value instead of duplicating the attribute into the state previously to prevent mismatching values.
- `state.values` has been introduced as `isUsableWhen` can now make the context influence a change in the selected values (clearing it when not usable). The idea was that I'd watch for changes to these values, then update the consumer values, but it seems that it doesn't work unless I make the `useEffect` dependency the whole state object. Happy to just move this into attributes, but it does mean that I'd need another `reduce` to convert the `state.filters` object to the correct shape when passing back into `onValuesChange` as opposed to now just taking and passing it in as is.
- `(consumer) values` exists as the consumer may want to manipulate the values from an external event.

packages/components/src/FilterBar/context/FilterBarContext.tsx
HeartSquared: You mean make a util and pass in `state` and `(consumer) values`?

packages/components/src/FilterBar/FilterBar.spec.tsx
HeartSquared: Yes that is correct.

If the consumer somehow has a funny combination where a dependent filter has a value but is failing the requirement to show up, the value will be cleared.

gyfchong: I think this is actually a use case, as [seen in Lisa's PoC](https://dev.cultureamp.design/KZN-1148/add-lisa-filterbar-example/?path=/docs/components-filter-bar-expandable-filter--docs#expandable-filter-conditional-filters), because we haven't solved the "all" problem yet.

packages/components/src/FilterBar/context/FilterBarContext.tsx
gyfchong: Nah, just like
```suggestion
  const shouldUpdateValues = (state) => filters.some(
    ({ id }) => state.values[id] !== values[id]
  )
```

HeartSquared: Lol, but I just want a boolean value :')

packages/components/src/FilterBar/context/reducer/setupFiltersState.ts
gyfchong: Hm, probably not worth looking into shuffling this now then, your test coverage should help optimise it later if it becomes an issue.

packages/components/src/FilterBar/FilterBar.spec.tsx
HeartSquared: That's irrelevant?

This solves for a case like this:
- **Toppings** relies on **Flavour** to be set
	- `toppings: { isUsableWhen: (state) => state.flavour.value !== undefined }`
- A user messes with the query params and we somehow end up with default values `{ toppings: ["pearls"] }`
- **Toppings** is not shown in the FilterBar because **Flavour** is missing a value
	- Therefore, since **Toppings** isn't shown, the value should also not exist

If we should be able to select **Toppings** without **Flavour** having a value, then this is the wrong condition to use.
There are other conditions like `state.flavour.isActive` which can be used instead.

packages/components/src/FilterBar/context/FilterBarContext.tsx
gyfchong: Lol kay

packages/components/src/FilterBar/FilterBar.spec.tsx
gyfchong: That's fine, but the use case is if the unusable Toppings filter becomes usable after Flavour gets a value, they might want Tops filter to be defaulted to pearls.

gyfchong: Oh wait, you mean that this is only if values are changing

gyfchong: Can you confirm if the use case is possible though? Lol

HeartSquared: Yes, only if the values are changing :)

Your case isn't something I've considered, but it seems to be more business logic so it would probably need to be implemented by the consumer. If it's a confirmed thing that they want an example for, we could look into it later.

packages/components/src/FilterBar/context/reducer/setupFiltersState.ts
HeartSquared: Yeah definitely :) And the only types I'm changing around for it is internal. If anyone digs for them and uses them, then they'll need to face the consequences of the "breaking" change.

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -118,44 +164,28 @@ unless there is a default value provided, of which it will be in the active filt
...
-  },
-] satisfies Filters<ValuesMap>
+The `state` arg will provide you with the following attributes for all the filters (mapped to the keys of the `ValuesMap`):
+```ts
+id: keyof ValuesMap
+name: string
+isRemovable?: boolean
+isOpen: boolean
+value?: ValueMap[keyof ValuesMap]
+isActive: boolean
```
HeartSquared, line 180: Writing this out really made me wonder whether a consumer would ever need all these attributes when writing a condition. I can only reasonably think of `isActive` and `value` making sense to use.

Should I revise the types?

gyfchong: Agreed, probably not `isOpen` and `isRemovable`. Maybe the `id` and `name` could be useful if they want to trigger analytics/other funcs that might string match

HeartSquared: Oh true. Didn't think about potentially using it for analytics. Although the condition gets run for every filter on every state update, so I don't know if they'd actually want to do that 😂 Their choice though.

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -189,6 +219,12 @@ A callback (`(values: Partial<ValuesMap>) => void`) for when the active `values`
 
 ## Examples
 
+### Dependent filters
+
+Example where `Topping` and `Drank` are usable only if `Flavour` has a value:
```
gyfchong, line 224: It's a bit hard to look at the live example code and find the exact thing that's driving the example and learn how to apply it, eg. I might have existing config and jump straight to this example hoping to retrofit.

So I think it would be nice to have a succinct code snippet here that highlights the `isUsableWhen` config, maybe just the filters array.

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -118,44 +164,28 @@ unless there is a default value provided, of which it will be in the active filt
...
+isOpen: boolean
+value?: ValueMap[keyof ValuesMap]
+isActive: boolean
 ```
 
+When a dependent filter has not met its condition (not _Usable_):
+- it will be removed from the FilterBar
+- if it has a value, the value will be cleared
+
+If this attribute is not defined, the filter will default to be _Usable_.
```
gyfchong, line 187: Might also be good to include a code snippet of what this looks like in config, hard to imagine where I should put this if I came to the `filters` section and jumped straight to `isUsableWhen`

packages/components/src/FilterBar/_docs/FilterBar.mdx
```diff
@@ -71,19 +71,65 @@ return (
...
-##### `Component`
+#### Filter attributes
+
+Each filter has the following attributes:
+
+- [`id: string`](#id-string)
+- [`name: string`](#name-string)
+- [`Component: JSX.Element`](#component-jsxelement)
+- [`isRemovable?: boolean`](#isremovable-boolean)
+- [`isUsableWhen?: (state: FiltersState<ValuesMap>) => boolean`](#isusablewhen-state-filtersstatevaluesmap--boolean)
```
gyfchong, line 120: To make it easier for consumers to understand what to navigate to these for, mainly around the `isUsableWhen`, I think you need to add the description of the feature to these instead of the typescript defs. Cause I'm not sure it's clear enough why someone would visit the `isUsableWhen` section if it doesn't tell them what the attribute solves.

The top two probably could just have what they're used for and not a section of their own, but maybe play around with the idea.

packages/components/src/FilterBar/_docs/FilterBar.stories.tsx
```diff
@@ -216,6 +216,73 @@ export const OnValuesChange: StoryFn<typeof FilterBar> = () => {
...
+        />
+      ),
+      isUsableWhen: state => state.flavour.value !== undefined,
+    },
+    {
+      id: "drank",
+      name: "Drank",
+      Component: <FilterBar.DatePicker />,
+      isRemovable: true,
+      isUsableWhen: state => state.flavour.value !== undefined,
```
christian314159, line 258: Just an idea here after trying out the Story:
What if we made the `Drank` filter dependent on `Topping`? Would make the example a bit more elaborate and show you can make something dependent on a filter that is dependent on something else?

HeartSquared:
Superseded by https://github.com/cultureamp/kaizen-design-system/pull/3812

PR #3801, opened 2023-06-26
Upgrade-node-to-18

gyfchong:
## Why
Certain dependencies are requiring us to move to node 18

## What
- Update our `package.json` to node 18
- Update Docker to node 18


packages/date-picker/src/DateRangePicker/DateRangePicker.spec.tsx
```diff
@@ -2,7 +2,8 @@ import React, { useEffect, useState } from "react"
 import { render, screen, waitFor } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { DateRange } from "react-day-picker"
-import { DateRangePicker, DateRangePickerProps, formatDateRangeValue } from "./"
+import { formatDateRangeValue } from "./utils/formatDateRangeValue"
```
HeartSquared, line 5: How did this even happen 😅

packages/date-picker/src/DateRangePicker/DateRangePicker.spec.tsx
```diff
@@ -2,7 +2,8 @@ import React, { useEffect, useState } from "react"
...
+import { DateRangePicker, DateRangePickerProps } from "./"
```
HeartSquared, line 6:
```suggestion
import { DateRangePicker, DateRangePickerProps } from "./index"
```

Would this be safer? 🤷🏻

HeartSquared (dismissed):

packages/date-picker/src/DateRangePicker/DateRangePicker.spec.tsx
gyfchong: Tests failed if I didn't do it this way

HeartSquared: Yeah it sounds like the util used to be exported from the index, but no longer does. It's just weird that it wasn't caught earlier.

gyfchong: It is exported from index, but oddly doesn't pick up:
```export * from "./DateRangePicker"
export * from "./utils"
```

HeartSquared (approved):

PR #3804, opened 2023-06-26
Add option for custom icon in tags

yvonneyeh:
Adds option to use any [icon](https://cultureamp.design/storybook/?path=/docs/components-icon--docs) available in [@kaizen/component-library](https://www.npmjs.com/package/@kaizen/component-library) as a custom icon in a tag.

Screenshot 2023-06-26 at 4 48 54 PM

### DIFF
--- a/.changeset/curly-chairs-behave.md
+++ b/.changeset/curly-chairs-behave.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/draft-tag": minor
+---
+
+Tags with custom icon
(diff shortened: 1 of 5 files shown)


gyfchong:
Thanks for this, but I think a better approach would be to create a sentiment that gives you the blue colour without the icon, then pass a ReactNode to the `Tag` as a child which includes your custom icon.

We're trying to avoid adding more props that take icons in preparation for some upcoming changes.

gyfchong:
We are rebuilding the Tag component to incorporate this feedback on allowing any icon. https://github.com/cultureamp/kaizen-design-system/pull/3851

PR #3806, opened 2023-06-28
Update render method for tooltips accessible description

mcwinter07:
## Why
This is part of the accessibility fixes to the use of tooltip across all of our components - specifically this came about from auditing the Rich Text Editor that uses tooltips, which points to elements not in the DOM

## What
Updates Tooltip to pass accessible descriptions only into native or custom semantic components and only when in the DOM
 - aria-describedby should not be attach to non semantic elements, ie: div and spans
 - id for aria-describedby is undefined until the tooltip is in the dom
  - This will stop the a11y error of pointing to an id of an element that does exists
 - Increases test coverage to validate somantic elements of the tooltip with recieve aria-describedby
 - Moves tooltip animation tests to seperate file so mock does not interfer with return value


.changeset/dull-pens-grow.md
```diff
@@ -0,0 +1,11 @@
+---
+"@kaizen/draft-tooltip": patch
+---
+
+Updates tooltip to pass accessible descriptions only into native or custom somantic components and only when in the DOM
```
gyfchong, line 5:
```suggestion
Updates tooltip to pass accessible descriptions only into native or custom semantic components and only when in the DOM
```

mcwinter07: really need to reinstall that spell check plugin 😆

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,82 @@
...
       render(
-        <Tooltip text="Example" animationDuration={1000}>
-          <div />
+        <Tooltip
+          text="Tooltip popup description for div"
+          display="inline"
+          isInitiallyVisible
+          position="below"
+        >
+          <div>Non somantic element</div>
```
HeartSquared, line 55: Or was it on purpose 😂 
```suggestion
          <div>Non semantic element</div>
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +156,39 @@ const TooltipContent = ({
   ) : null
 }
 
+const isSemanticElement = (
+  elementType: string | JSXElementConstructor<any>
```
HeartSquared, line 160: It's not `JSX.Element`?

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -227,9 +272,9 @@ export const Tooltip = ({
           onMouseLeave={(): void => setIsHover(false)}
           onFocusCapture={(): void => setIsFocus(true)}
           onBlurCapture={(): void => setIsFocus(false)}
-          aria-describedby={tooltipId}
         >
-          {children}
+          {/* custom child render should go here */}
```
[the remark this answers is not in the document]
mcwinter07: yep, that'll go 👍

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
mcwinter07: At this point its been narrowed to the `type` property of a `React.ReactPortal | React.ReactElement<unknown, string | React.JSXElementConstructor<any>>` which, assuming TS isn't telling porky pies, should be either and `string` or `JSXElementConstructor<any>`

Screenshot 2023-07-05 at 12 32 03 pm

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
mcwinter07: Nope 😆

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +156,39 @@ const TooltipContent = ({
...
+  if (elementType === "div" || elementType === "span") {
+    return false
+  }
+  return true
+}
+
+/**
+ *  validates implicit or explicity semantic role require to make aria-describedby announce predictably with screen readers
+ */
+const hasSemanticRole = (element: ReactNode): boolean => {
```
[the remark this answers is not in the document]
mcwinter07: We can do that? I've only ever seen it as something simple like: `isSemanticRole = (element: AllPossibleThings): element is string`

Would you try to use the ReturnType utility? Little confused 🤔

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +156,39 @@ const TooltipContent = ({
...
+  elementType: string | JSXElementConstructor<any>
+): boolean => {
+  if (elementType === "div" || elementType === "span") {
+    return false
+  }
+  return true
+}
+
+/**
+ *  validates implicit or explicity semantic role require to make aria-describedby announce predictably with screen readers
```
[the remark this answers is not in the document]
gyfchong:
```suggestion
 *  Validates implicit or explicitly semantic role required to make `aria-describedby` announce predictably with screen readers
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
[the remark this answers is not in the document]
HeartSquared: It's essentially the same boolean condition, except Typescript will now assume the variable in any following code has the type you've said it is :)

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +156,39 @@ const TooltipContent = ({
...
+ */
+const hasSemanticRole = (element: ReactNode): boolean => {
+  if (!React.isValidElement(element)) return false
+  const { props, type } = element
+
+  if (isSemanticElement(type) || "role" in props) return true
+
+  return false
```
HeartSquared, line 177: What if we put all of this into `isSemanticElement` so you can just pass the element in :)

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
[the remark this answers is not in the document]
mcwinter07: Sorry I more so meant how do you type `ValidThing` - a valid element needs to be both pass `SemanticElement(type) || "role" in props` so I'm just a little confused about that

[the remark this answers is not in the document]
mcwinter07: Happy to pair on this if you've got a second :)

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -30,6 +38,9 @@ export type TooltipProps = {
    */
   position?: Position
   text: React.ReactNode
+  /**
+   * This is the interactable element that is being described by the tooltip `text`
+   */
   children?: React.ReactNode
```
HeartSquared, line 44: This probably shouldn't be optional, but I guess that's out of scope 😬

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +156,18 @@ const TooltipContent = ({
...
 }
 
+const renderChildren = (
+  content: ReactNode,
+  tooltipId: string,
+  hasActiveTooltip: boolean
+): ReactNode => {
+  if (!hasSemanticRole(content)) return content
+
+  return cloneElement(content as ReactElement | React.ReactPortal, {
```
HeartSquared, line 166: I think we should be able to get to a state where we don't need to type-cast. Let's pair tomorrow? :)

draft-packages/tooltip/KaizenDraft/Tooltip/hasSemanticRole.spec.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react"
+import { Button, IconButton } from "@kaizen/button"
+import { hasSemanticRole } from "./hasSemanticRole"
+
+describe("hasSemanticRole", () => {
+  it("will return true if provided a native element with a semantic role", () => {
+    expect(
+      hasSemanticRole(
+        <button onClick={jest.fn} type="button">
```
HeartSquared, line 9:
```suggestion
        <button onClick={jest.fn()} type="button">
```

draft-packages/tooltip/KaizenDraft/Tooltip/hasSemanticRole.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+    expect(
+      hasSemanticRole(
+        <span tabIndex={0} role="button" onKeyDown={jest.fn} onClick={jest.fn}>
+          custom semantic el
+        </span>
+      )
+    ).toBe(true)
+    expect(
+      hasSemanticRole(
+        <div tabIndex={0} role="button" onKeyDown={jest.fn} onClick={jest.fn}>
```
HeartSquared, line 34: Do you need the `onKeyDown` and `onClick`s?

mcwinter07: yeah looks a little janky but div with the role button needs a keyboard event listener 😢 otherwise ts will cry

packages/button/src/Button/components/GenericButton.tsx
```diff
@@ -47,6 +47,7 @@ export type GenericProps = {
   onBlur?: (e: FocusEvent<HTMLElement>) => void
   component?: ComponentType<CustomButtonProps>
   classNameOverride?: string
+  "aria-describedby"?: string
```
mcwinter07, line 50: @dougmacknz I decided to add this to the button props so a consumer (or in this case consuming components like the Tooltip) doesn't need to use the custom component rendered to be able to pass in aria-described to the actual button (instead of the span that wraps it)

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { ReactNode } from "react"
+
+// Noting that this is contingent on our displayName for our components -  this something we control
+const allowedDisplayNames = ["Button", "IconButton"]
```
mcwinter07, line 4: @dougmacknz keens to get your thoughts on which components we should allow the Tooltip to describe (aria-describedby). I'm wondering if we should have table header row in here?

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export function extractDisplayName(type: React.FunctionComponent): string {
+  return type.displayName || type.name || "Unknown"
+}
+
+/**
+ *  Validates implicit or explicitly semantic roles required to make `aria-describedby` announce predictably with screen readers
+ */
+export const isSemanticElement = (
```
mcwinter07, line 13: @dougmacknz also keen to get your eyes across this util :) holler if you want me to run you through this

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.tsx
dougmacknz: So I checked the other places in Kaizen that we're using Tooltips:
- ✅ RTE's toggle buttons are fine
- ✅ Table headers are fine. There's an issue there where the header is only a button if the column is sortable, but out of scope for this work.

There's a couple of places in the filter bar and FilterMultiSelect world where it's not working, but it's actually a good thing because the tooltips are just repeating the button's label.

image
image

Except for this Button Group one, which I think is using the same `FilterButtonBase` as the one above.
image

So a bit of a difficult decision to make there, same as the `IconButton` problem.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+import ArrowIcon from "@kaizen/component-library/icons/arrow-right.icon.svg"
 
+import { Tooltip } from "./index"
 describe("<Tooltip />", () => {
-  beforeEach(() => {
-    AnimationProvider.mockReturnValue(<div />)
-  })
-  describe("When no animationDuration prop is given", () => {
-    it("animationDuration prop is passed to AnimationProvider as undefined", () => {
+  describe("Accessible descriptions", () => {
```
dougmacknz, line 8:
```suggestion
  describe("Linking the tooltip to the inner element with aria-describedby", () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+
+      await waitFor(() => {
+        expect(
+          screen.getByRole("button", { name: "More info" })
+        ).toHaveAccessibleDescription(
+          "Tooltip popup description for Kaizen Button"
+        )
+      })
+    })
+    it("will render a IconButton that will have an accessible description", async () => {
```
dougmacknz, line 29: Given you've already tested this functionality above, I think we could do without this one. I don't feel super strongly, but we might want to consider that too many tests may eventually slow down our build.

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+        ).toHaveAccessibleDescription(
+          "Tooltip popup description for Kaizen IconButton"
+        )
+      })
     })
-  })
-  describe("When animationDuration prop is given", () => {
-    it("animationDuration prop is passed to AnimationProvider", () => {
+
+    it("will render a button without an aria-describedby if the tooltip is not active", async () => {
```
dougmacknz, line 50:
```suggestion
    it("doesn't add an accessible description if the tooltip is inactive", async () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+      await waitFor(() => {
+        expect(screen.queryByRole("tooltip")).toBe(null)
+        expect(
+          screen.getByRole("button", { name: "More info" })
+        ).not.toHaveAttribute("aria-describedby")
+      })
+    })
+
+    // Non-semantic elements without roles should not have aria-description on them - they will not read to screen readers and will throw an error
+    it("will render non-semantic elements without accessible descriptions", async () => {
```
dougmacknz, line 70:
```suggestion
    it("doesn't add an accessible description when wrapping a non-semantic element", async () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+      )
+      await waitFor(() => {
+        expect(screen.getByText("Non semantic element")).not.toHaveAttribute(
+          "aria-describedby"
+        )
+      })
+    })
+  })
+
+  it("will render elements with semantic roles with accessible descriptions", async () => {
```
dougmacknz, line 89:
```suggestion
  it("adds an accessible description when wrapping a semantic element", async () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +155,20 @@ const TooltipContent = ({
   ) : null
 }
 
+const renderChildren = (
+  content: ReactNode,
+  tooltipId: string,
+  hasActiveTooltip: boolean
+): ReactNode => {
+  if (isSemanticElement(content)) {
```
dougmacknz, line 163: Idea: we could throw a console warning or something like that if the element passed in here is `isSemanticElement: false`. People don't love/respond well to the Icon warnings we have, but that's probably more because it could have been enforced with TS instead. Might be better than nothing.

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ).toBe(true)
+    expect(isSemanticElement(<a href="/">link</a>)).toBe(true)
+  })
+
+  it("will return false if provided a native element without a semantic role", () => {
+    expect(isSemanticElement(<span>click</span>)).toBe(false)
+    expect(isSemanticElement(<div>link</div>)).toBe(false)
+  })
+
+  it("will return true if provided a custom elements with a semantic role", () => {
```
dougmacknz, line 24:
```suggestion
  it("returns true if provided one of the Kaizen components from the allowed list", () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react"
+import { Button, IconButton } from "@kaizen/button"
+import { Icon } from "@kaizen/component-library"
+import ArrowIcon from "@kaizen/component-library/icons/arrow-right.icon.svg"
+import { isSemanticElement } from "./isSemanticElement"
+
+describe("isSemanticElement", () => {
+  it("will return true if provided a native element with a semantic role", () => {
```
dougmacknz, line 8: Just to distinguish this from the case of adding a role to a non-semantic element
```suggestion
  it("returns true if provided a semantic element", () => {
```

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+      isSemanticElement(
+        <button onClick={jest.fn()} type="button">
+          click
+        </button>
+      )
+    ).toBe(true)
+    expect(isSemanticElement(<a href="/">link</a>)).toBe(true)
+  })
+
+  it("will return false if provided a native element without a semantic role", () => {
```
dougmacknz, line 19:
```suggestion
  it("returns false if provided a non-semantic element", () => {
```

dougmacknz (commented):
Sorry lots of nit picks on test names but otherwise looking pretty good 😄

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
```diff
@@ -1,42 +1,105 @@
...
+
+      await waitFor(() => {
+        expect(screen.queryByRole("tooltip")).toBe(null)
+        expect(
+          screen.getByRole("button", { name: "More info" })
+        ).not.toHaveAttribute("aria-describedby")
+      })
+    })
+
+    // Non-semantic elements without roles should not have aria-description on them - they will not read to screen readers and will throw an error
```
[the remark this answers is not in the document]
mcwinter07: Ah yeah, sorry it was more that the axe tooling in storybook will point this out

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.spec.tsx
mcwinter07: That's true, I suppose we can safely assume that even if we were to change the where the Icon was rendered in the button, it should always be a child of the interactive element

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
mcwinter07: I have no opposition there, might help be a quick way to identify it for them if they're digging in console 👍

draft-packages/tooltip/KaizenDraft/Tooltip/utils/isSemanticElement.tsx
mcwinter07: Digging into the code for the remove button, it looks like most of its accessible title is only derived image title prop on the icon (at least in our example). That said the consumers can pass in aria attributes into the `removeButtonProps`, which means the accessible label description should abel to be different, ie:

```
          <FilterButtonRemovable
            triggerButtonProps={{
              label: "Desserts",
            }}
            removeButtonProps={{
              onClick: () => undefined,
              "aria-label": "Remove desserts",
              tooltipText: "Could be different from the label ideally"
            }}
          />
```

Will produce this (with the FilterButtonBase added to the allow list):

Screenshot 2023-07-19 at 3 07 05 pm


I'd say there's a larger question on how we handle a default accessible label for the button but I'd say we could save that for a separate PR since it relates specifically to the implementation of the `FilterButtonRemovable`

draft-packages/tooltip/KaizenDraft/Tooltip/Tooltip.tsx
```diff
@@ -145,6 +155,24 @@ const TooltipContent = ({
...
+    return cloneElement(content, {
+      "aria-describedby": hasActiveTooltip ? tooltipId : undefined,
+    })
+  }
+  /* eslint no-console: "warn" */
+  // We don't want to block them from this but just provide context for better a11y guidance
+  console.warn(
+    "<Tooltip /> is wrapping a non-semantic element, aria-describedby may not be read as expected"
```
dougmacknz, line 171: Suggestion for making the implications a bit more clear. Plus we can say for sure it won't work because we're now not adding `aria-describedby` if it doesn't pass this condition.

I've also added a mention of how the element should be focusable as well to ensure keyboard accessibility.

```suggestion
    "<Tooltip /> is not directly wrapping a semantic element, screen reader users will not be able to access the tooltip info. To ensure accessibility, Tooltip should be wrapping a semantic and focusable element directly."
```

mcwinter07: Good catch 👍 I'll also make sure to draw attention to it when this gets announces in #updates_design_system

dougmacknz (approved):
Wew

PR #3809, opened 2023-06-29
Fix KAIO bundle

gyfchong:
I'm not entirely sure what happened, but the bundling stopped working in the transition to PNPM and noticed that button styles didn't work due to relying on `component-library`

### DIFF
--- a/.changeset/afraid-dogs-cheer.md
+++ b/.changeset/afraid-dogs-cheer.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Various fixes to style imports and peer deps
--- a/.changeset/afraid-dogs-cheer.md
+++ b/.changeset/afraid-dogs-cheer.md
@@ -2,4 +2,5 @@
 "@kaizen/components": patch
 ---
 
-Various fixes to style imports and peer deps
+**Button**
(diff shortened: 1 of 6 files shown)


packages/components/package.json
```diff
@@ -37,6 +37,7 @@
     "ci:test:react-17": "USE_REACT_17=true yarn test:ci"
   },
   "dependencies": {
+    "@cultureamp/fetch": "^3.0.9",
```
gyfchong, line 40: A peer dep for `@cultureamp/i18n-react-intl`

gyfchong: Which... actually isn't, so KAIO was complaining for nothing...

HeartSquared: What if you peerDep the peerDep 😛

.changeset/afraid-dogs-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Various fixes to style imports and peer deps
```
HeartSquared, line 5: What peer deps? 😏

gyfchong: Touché

HeartSquared (approved):

PR #3810, opened 2023-06-29
[BLOCKED] KDS-1579-switch-from-yarn-to-pnpm

gyfchong:
## Why

Need more speed

## What

Change from Yarn to PNPM

## WARNING
This repo only successfully builds because all scripts are pointed to running ONLY KAIO. [See ticket](https://cultureamp.atlassian.net/browse/KDS-1579) for blocked status.


gyfchong:
Re-do this post R16/17 split

PR #3811, opened 2023-06-30
Kds 1394 i 18 n for filter drp

JakePitman:
## Why
FilterDRP has hard-coded strings that can only be translated from within KZ.

## What
- Adds translations for hard-coded strings in FilterDRP:
   - "Date from"
   - "Date to"
   - "Input format"
- Adds a story demonstrating the degree of translation coverage in FilterDRP
- Mocks the `useIntl` function from `@cultureamp/i18n-react-intl`, as our jest environment doesn't have an intl object in context
- Exports the locale folder from `@kaizen/components`, so they can be compiled with consuming repo's locales
- Upgrades to the latest version of `@cultureamp/i18n-react-intl`


setupTests.ts
```diff
@@ -2,3 +2,12 @@ import "@testing-library/jest-dom/extend-expect"
 
 /** @ts-ignore */
 global.IS_REACT_ACT_ENVIRONMENT = true
+
+// This avoids errors related to our jest environment not having an Intl
+// object present in the app's context
+jest.mock("@cultureamp/i18n-react-intl", () => ({
```
gyfchong, line 8: Nice one!

packages/components/src/FilterDateRangePicker/_docs/FilterDateRangePicker.stories.tsx
```diff
@@ -140,6 +141,42 @@ Playground.args = {
   renderTrigger: "Filter Button",
 }
 
+export const Translated: StoryFn<typeof FilterDateRangePicker> = args => {
```
gyfchong, line 144: Is this supposed to be sent to Chromatic? Probably add it to the stickersheet

gyfchong (approved):

PR #3812, opened 2023-06-30
feat(filter-bar): add functionality for dependent filters

HeartSquared:
## What

Add functionality for dependent filters (take 2; replaces https://github.com/cultureamp/kaizen-design-system/pull/3788).
All dependent filters are checked and have their usable state updated according to the current state of all the filters.

- New filter attribute `isUsableWhen` (`(state) => boolean`) added for consumers
	- Filters failing to meet their condition will be removed from the FilterBar options (both active/inactive) and their value will be cleared
- `isUsable` added to state
- `dependentFilterIds` added to state
- `values` added to state
	- `(consumer) values` and `state.values` must be able to change each other
		- In the case that the filter is deemed unusable, but the consumer has passed in a value, we need to clear it
		- We still need to allow consumers to change values from an external event

## Why

[KDS-1472](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1472)
Some filters should only be shown in the FilterBar (both active/inactive) when certain conditions are met.

[KDS-1472]: https://cultureamp.atlassian.net/browse/KDS-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gyfchong (approved):
Works

PR #3842, opened 2023-07-04
Fix tailwind preset tshirt filtering

mcwinter07:
## Why
We were filtering out all non number values from the tailwind preset after width config was removed from `kaizenTailwindTheme`. This caused issues as the tailwind expected tailwind defaults was not bundled in the final solution (ie: w-1/2, w-2/5, etc)

## What
Update filter method for tailwind spacing tokens to allow default tailwind width values
 - This fixes the !Nan filtered our tailwind percentile and string values
 - Updated stories that use w-100 to w-full since the original interntion was to be compile to width: 100%
 
[Decision / meeting log](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3213164830/04+06+23+-+Tailwind+config+width+class+issue)


packages/tailwind/src/kz-spacing.ts
```diff
@@ -1,14 +1,41 @@
...
+  "xxxl-id",
+  "xxxxl-id",
+  "xxxxxl-id",
+]
+
+/**
+ * checks if a key from design tokens should be filtered from tailwind
+ */
+const isValidToken = (key: string | number): boolean => {
+  if (tokenDenyList.indexOf(`${key}`) < 0) return true
```
christian314159, line 28:
```suggestion
  return tokenDenyList.indexOf(`${key}`) >= 0
```
would work as well ;)

mcwinter07: oooo yes, that's much better 😉

christian314159: Actually, scratch that, it should be
```
return !tokenDenyList.includes(`${key}`)
``

File: packages/tailwind/src/kz-spacing.ts [hunk 2]
```diff
@@ -1,14 +1,41 @@
 import { heartTheme } from "@kaizen/design-tokens"
 
+const tokenDenyList: string[] = [
```
gyfchong, line 3: Let's write something to record why this decision was made 😉

mcwinter07: Will whip up a doc in the confluence section

mcwinter07: Documented [here](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3213164830/04+06+23+-+Tailwind+config+width+class+issue) 👍

packages/tailwind/src/kz-spacing.ts
mcwinter07: Updated 👍 though I've returned ```tokenDenyList.indexOf(`${key}`) < 0``` instead of a >= since this checking to see if the key does not exist in the tokenDenyList so the index should be -1 for valid tokens and return true

mcwinter07: sorry just refresh the page and realised there was an additional comment 😆 I've always leaned on indexOf because it meant to be slightly more performant. Not wedded to it so I can change it to include for readabilities sake

.changeset/twelve-days-hunt.md
```diff
@@ -0,0 +1,8 @@
+---
+"@kaizen/tailwind": patch
+---
+
+Update filter method for tailwind spacing tokens to allow default tailwind width values
+ - This fixes the !Nan filtered our tailwind percentile and string values
+ - Updated stories that use w-100 to w-full since the original interntion was to be compile to width: 100%
```
gyfchong, line 7:
```suggestion
 - Updated stories that use w-100 to w-full since the original intention was to be compile to width: 100%
```

mcwinter07: lul, that was not my interntion 😆 Will fix 👍

HeartSquared (approved):

PR #3847, opened 2023-07-05
fix(filter-bar): enable sibling value dependent filters

HeartSquared:
## What

Enable Select/MultiSelect filters to load items based on a sibling's selected value.

- Selected values for dependent filters that are removed on items change are also removed from context values
- Found issues with the types, so fixed them

## Why

[KDS-1475](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1475)

Common use case for consumers.

[KDS-1475]: https://cultureamp.atlassian.net/browse/KDS-1475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/rotten-icons-complain.md
```diff
@@ -0,0 +1,7 @@
+---
+"@kaizen/components": minor
+---
+
+- Update FilterBarMultiSelect to clear values when items change
```
gyfchong, line 5: Is this more of a fix, cause it didn't work as expected before?

packages/components/src/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -39,18 +41,37 @@ const convertConsumableFormatIntoSelection = (
...
   const filterState = getFilterState(id)
 
+  useEffect(() => {
+    if (!checkArraysMatch(items, propsItems)) {
+      setItems(propsItems)
+    }
+  }, [propsItems])
+
+  useEffect(() => {
+    if (Array.isArray(filterState.value)) {
```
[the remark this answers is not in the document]
mcwinter07: in that they're not filtered out of the state if more than one is selected (for anyone reading this comment that doesn't understand our pairing mumbo jumbo) :P

packages/components/src/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -39,18 +41,38 @@ const convertConsumableFormatIntoSelection = (
...
+      const itemValues = items.map(({ value }) => value)
+      const filteredValues = filterState.value.filter(value =>
+        itemValues.includes(value)
+      )
+
+      if (!checkArraysMatch(filterState.value, filteredValues)) {
```
mcwinter07, line 70: Nice, no more engineers in the building! 😉

Translation: Nice, should be a less brittle way to update those values 🕺

mcwinter07 (approved):
Looks 🦖 🐜

PR #3849, opened 2023-07-06
Jshli/update select with popover

jshli:
## Why
Adding a portal to the Select overlay component because in some views, the Menu can experience clipping

## What
- Render the Select `Overlay` component in a React portal
- Swap the `useOverlay` hook used in the Select `Overlay` component for `usePopover` hook to position the list box

## Notes
In this implementation which uses a popover, the popover is designed to 'flip' the listbox if there's not enough room at the bottom of screen
image

We can force the popover to stay at the bottom, but this will compress the `ListBox` to a tiny height
image
To fix that issue, we could set a min-height to the listbox, but then this might cause further issues where the min-height exceeds shorter lists.


packages/select/src/Select/Select.spec.tsx
```diff
@@ -77,7 +77,7 @@ describe("<Select>", () => {
         render(
           <SelectWrapper
             selectedKey="id-sre"
-            isOpen
+            defaultOpen
```
jshli, line 80: Updated this because the `isOpen` prop was firing the `onOpenChange` callback on mount

packages/select/src/Select/components/Overlay/Overlay.tsx
```diff
@@ -1,37 +1,53 @@
...
   // Handle events that should cause the menu to close,
   // e.g. blur, clicking outside, or pressing the escape key.
-  const overlayRef = React.useRef<HTMLDivElement>(null)
   const { state } = useSelectContext()
-  const { overlayProps } = useOverlay(
-    { isDismissable: true, isOpen: state.isOpen, onClose: state.close },
-    overlayRef
+
+  const popoverRef = React.useRef<HTMLDivElement>(null)
+  const { popoverProps } = usePopover(
```
jshli, line 23: `usePopover` uses `useOverlay` under the hood

packages/select/src/Select/components/Overlay/Overlay.tsx
```diff
@@ -1,37 +1,53 @@
...
+      className={styles.menuPopup}
+    >
       <FocusScope contain autoFocus restoreFocus>
         <DismissButton onDismiss={state.close} />
         {children}
         <DismissButton onDismiss={state.close} />
       </FocusScope>
     </div>
   )
+  return ReactDOM.createPortal(contents, document.body)
```
HeartSquared, line 52: Does it _need_ to be a portal? The rest of our popovers use `usePopper` from `react-popper`. Any reason why we can't use the same pattern?

packages/select/docs/Select/Select.stickersheet.stories.tsx
```diff
@@ -39,9 +39,10 @@ const StickerSheetTemplate: StoryFn<{ isReversed: boolean }> = ({
           items={singleMockItems}
           description="This is a description"
           selectedKey="id-sre"
-          trigger={(triggerProps): JSX.Element => (
+          trigger={(triggerProps, buttonRef): JSX.Element => (
```
HeartSquared, line 42: I'm not a fan of the consumer now being forced to add the ref. Take a look at the `Filter` component where the `ref` is cloned on behalf of the consumer.

packages/select/src/Select/components/Overlay/Overlay.tsx
jshli: I think primarily the ListBox should be rendered in a Portal as we're trying to avoid the clipping issue, which is probably the preferred way of solving this problem. A popover's solution would be to dynamically position the Listbox in a way that avoids the edge of boundary, but conventionally, a Select's Listbox renders below the input. I have left a note though that currently, the Listbox does flip thanks to React Aria's implementation of a popover, but if Kaizen decides we'd like the prevent flipping, we'd need to portal it anyway. 

Additionally, seeing as the Kaizen Select component has been built almost entirely the same way React Aria suggests, I felt like it made sense to just continue the same pattern - with the exception of using their `<Overlay>` component, which pretty much just abstracts the `React.createPortal` function.

gyfchong (changes requested):

.changeset/grumpy-ads-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/select": minor
```
gyfchong, line 2: I think this should be a `patch` since a `portal` isn't a feature anyone can use, rather it solves an implementation issue.

gyfchong:
You'll need to add `padding ` to the container of the "Dropdown Full width" stickersheet example in storybook to give the dropdown more space.

jshli:
> You'll need to add `padding ` to the container of the "Dropdown Full width" stickersheet example in storybook to give the dropdown more space.

I added some additional margin to the stories on top. But just a note: looks like the popover doesn't recalculate the position of the select while the dropdown is open, so this drop down will always bee 'flipped'
image

gyfchong (approved):

PR #3850, opened 2023-07-06
Remove optional intl provider and update docs

JakePitman:
## Why
We've seen bundling issues when including `@cultureamp/i18n-react-intl` in KAIO. More details in [this thread](https://cultureamp.slack.com/archives/C015X1ZJ0MV/p1688363548436119).

## What
- Takes OptionalIntlProvider our of KaizenProvider
- Leaves a comment in OptionalIntlProvider explaining why it's not used, and when it can be put back in
- Updates the Internationalization in Kaizen docs to reflect the fact that consumers will need to include their own IntlProvider


packages/components/src/KaizenProvider/_docs/internationalization-in-kaizen.mdx
```diff
@@ -12,7 +12,7 @@ If you need guidance on translating your app check, the `@cultureamp/i18n-react-
 ## What Kaizen internationalization covers
 
 While most Kaizen components receive display text and accessibility information through props, some have text built in. For some of these cases, Kaizen provides translation support.
```
HeartSquared, line 14: So we do or don't still provide translation support?

HeartSquared (approved):

packages/components/src/KaizenProvider/_docs/internationalization-in-kaizen.mdx
JakePitman: Not any more. Reworded it to reflect that

packages/components/src/KaizenProvider/KaizenProvider.tsx
```diff
@@ -1,21 +1,16 @@
...
 
 export const KaizenProvider = ({
   children,
   themeManager,
-  locale = "en",
 }: KaizenProviderProps): JSX.Element => (
-  <OptionalIntlProvider locale={locale}>
-    <ThemeProvider themeManager={themeManager}>{children}</ThemeProvider>
-  </OptionalIntlProvider>
+  <ThemeProvider themeManager={themeManager}>{children}</ThemeProvider>
```
HeartSquared, line 13: Given you've made a change here, your changeset needs a description so it can release :) Probably just a patch will do

JakePitman: Ah yep

HeartSquared (approved):

PR #3851, opened 2023-07-06
KDS-1610 Add new Tag component to KAIO

maiconmborges:
## Why
- Adding new `Tag` component to KAIO
- Fixes [#97](https://github.com/cultureamp/kaizen-discourse/issues/97)

## Notes
-  I chose to implement the component without a `classNameOverride` prop as this is a bandaid solution for misusing the component.
- I deliberately not implemented an `icon` prop as I believe we should let engineers choose what they want to pass as children to compose their requirements
- I took the liberty of building a `TagDismiss` component to help with the composition of a dismissible `Tag`

## What
Screenshot 2023-07-06 at 1 26 49 pm


packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+  | "green"
+  | "yellow"
+  | "orange"
+  | "red"
+  | "purple"
+
+export interface TagProps {
+  children: ReactNode
+  color?: TagColors
+}
```
gyfchong, line 29: Don't worry about adding the icon prop, we will need to do extra work to make it work anyways.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <span
+      className={classNames(
+        "inline-flex items-center px-8 py-4 gap-4 rounded-[18px] font-family-paragraph text-paragraph-sm font-weight-paragraph leading-paragraph-sm whitespace-nowrap",
+        {
+          [`bg-${color}-100`]: color !== "gray",
+          [`text-${color}-700`]: color !== "gray",
+          ["bg-gray-200"]: color === "gray",
+          ["text-purple-800"]: color === "gray",
+        }
```
gyfchong, line 43: String interpolation doesn't work in tailwind, it only works here because of our storybook setup. We will take a look at making this more closer to production so we don't have this magically working.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+  color?: TagColors
+}
+
+export const Tag: FC<TagProps> = ({ children, color = "gray" }) => {
+  const uuid = useMemo(() => v4(), [])
+
+  return (
+    <span
+      className={classNames(
+        "inline-flex items-center px-8 py-4 gap-4 rounded-[18px] font-family-paragraph text-paragraph-sm font-weight-paragraph leading-paragraph-sm whitespace-nowrap",
```
gyfchong, line 37: The rounded needs to be `100%`

maiconmborges: Our TW preset has the `rounded-full` set to `100%` instead of `9999px`.
Screenshot 2023-07-10 at 12 50 41 pm

Should I use something like `rounded-[100vmax]` instead?
Screenshot 2023-07-10 at 12 53 59 pm

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+  | "orange"
+  | "red"
+  | "purple"
+
+export interface TagProps {
+  children: ReactNode
+  color?: TagColors
+}
+
+export const Tag: FC<TagProps> = ({ children, color = "gray" }) => {
```
gyfchong, line 31: We opted to use explicit return types rather than `FC`

packages/components/src/__future__/Tag/Tag.tsx
gyfchong: Hm I don't want to rely on a viewport unit for this, so I think just go with the height of the pill, `20px` and if we need to make it scale we can change it later.

maiconmborges:
After our chat this morning @gyfchong I implemented the `icon` prop, which will need to be updated once we have the name of the icons and the correct way of importing them into `Tag`.

I also created a `BaseTag` component which handles the style of the tag (note: it doesn't need to be exposed to consumers like it is right now).  Based on the component, I composed `Tag` and `AvatarTag` (also exposed as `Tag.Avatar`), which restrict the usage of `BaseTag`.

Screenshot 2023-07-10 at 2 27 46 pm

packages/components/src/__future__/Tag/AvatarTag.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Avatar, AvatarProps } from "@kaizen/draft-avatar"
+import { BaseTag } from "./BaseTag"
+import { Dismiss, DismissProps } from "./Dismiss"
+
+export type AvatarTagProps = AvatarProps & {
+  children: string
+  onDismiss?: DismissProps["onDismiss"]
+}
+
+export const AvatarTag = ({
```
gyfchong, line 11: Please remove this component as we haven't finished designing for all the possible colour combinations with the Avatar's no image state, so we can't technically release with this.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,60 @@
...
+export type TagColors =
+  | "blue"
+  | "gray"
+  | "green"
+  | "orange"
+  | "purple"
+  | "red"
+  | "yellow"
+
+export interface TagProps {
```
gyfchong, line 21: This will need to extend the span HTMLElement props

packages/components/src/__future__/Tag/BaseTag.tsx
```diff
@@ -0,0 +1,40 @@
...
+  | "orange"
+  | "red"
+  | "purple"
+
+export interface BaseTagProps {
+  children: ReactNode
+  color?: TagColors
+}
+
+export const BaseTag = ({
```
gyfchong, line 18: Still not sure what the value is of using the `BaseTag` as a consumer creating a custom `Tag` when I can simply use `<Tag><CustomElement /></Tag>` rather than importing a completely separate component?

Wouldn't your Avatar example still work if I did this:
```
<Tag>
    <span className="-ms-4 -my-[1px]">
      <Avatar {...avatarProps} size="small" />
    </span>
    {children}
</Tag>
```

packages/components/src/__future__/Tag/Tag.tsx
gyfchong: Have a look at our SVG component for an example:
https://github.com/cultureamp/kaizen-design-system/blob/main/packages/components/src/SVG/SVG.tsx#L11

packages/components/src/__future__/Tag/BaseTag.tsx
maiconmborges: Yes, you are right. I got confused when moving the logic around for the styling.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { ReactNode, HTMLAttributes } from "react"
+import classNames from "classnames"
+import { OverrideClassName } from "@kaizen/component-base"
```
gyfchong, line 3: This should come from within `packages/components` from our alias `~/types`

packages/components/src/__future__/Tag/Dismiss.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react"
+import { Icon } from "@kaizen/component-library/components/Icon"
+import clearIcon from "@kaizen/component-library/icons/clear-white.icon.svg"
+
+export type DismissProps = {
+  onDismiss?: React.MouseEventHandler<HTMLButtonElement>
+}
+
+export const Dismiss = ({ onDismiss }: DismissProps): JSX.Element => (
```
gyfchong, line 9: We're going to have to remove this too since its interactive we'll have to include more tests to support it, and we already have a well tested [ClearButton](https://cultureamp.design/?path=/docs/components-utilities-forms-clear-button--docs) component, just that it's currently too big, so we are weighing our options to resize that component and re-use it here to keep consistency.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -0,0 +1,66 @@
...
+  | "yellow"
+  | "orange"
+  | "red"
+  | "purple"
+
+export interface TagProps
+  extends OverrideClassName<HTMLAttributes<HTMLSpanElement>> {
+  children: ReactNode
+  color?: TagColors
+  icon?: TagIcons
```
gyfchong, line 23: We've settled on a `JSXElement` for the `Icons`. For example, consumers will be passing `<AddIcon />` to this prop.

JakePitman (approved):

PR #3852, opened 2023-07-07
convert FilterDateRangePicker to use a reducer for state management

gyfchong:
## Why
Consumers who use the validation object to read the input value were receiving the previously selected date rather than the currently selected one.

## What
Replace the state management with reducers for more reliable state changes and ordering.


packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,140 @@ export const FilterDateRangePickerField = ({
...
   const handleCalendarSelectRange: CalendarRangeProps["onSelect"] = range => {
-    const newStartDate = validateStartDate(range?.from)
-    const newEndDate = validateEndDate(range?.to)
-    setInputStartDateValue(transformDateToInputValue(newStartDate))
-    setInputEndDateValue(transformDateToInputValue(newEndDate))
+    const inputStartValue = transformDateToInputValue(range?.from)
+    const inputEndValue = transformDateToInputValue(range?.to)
+
+    const newStartDate = validateStartDate(range?.from, inputStartValue)
+    const newEndDate = validateEndDate(range?.to, inputEndValue)
```
gyfchong, line 211: I'm pretty sure this was all that really needed to be changed in order for the validation to get the correct input values. But hey, reducers look nicer lol.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,140 @@ export const FilterDateRangePickerField = ({
...
+
+    dispatch({
+      type: "update_input_start_field",
+      inputValue: inputStartValue,
+    })
+
+    dispatch({
+      type: "update_input_end_field",
+      inputValue: inputEndValue,
+    })
```
HeartSquared, line 231: Shouldn't this already be encapsulated by the above actions?

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/types.ts
```diff
@@ -4,3 +4,43 @@ export type DateRangeFieldValidationMessage = {
   dateStart?: ValidationMessage
   dateEnd?: ValidationMessage
 }
+
+export type DateRangeFieldActions =
```
HeartSquared, line 8: Since the reducer is internal, I don't think it's necessary to be distributing this type. I'd just place it with the reducer instead. Up to you though.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,140 @@ export const FilterDateRangePickerField = ({
...
+  ): Date | undefined =>
     dateEndValidation.validateDate({
       endDate: date,
-      endDateInputValue: inputEndDateValue,
+      endDateInputValue: inputValue,
       startDate: selectedRange?.from,
       startDateFieldLabel: inputStartDateLabel,
     })
 
+  const [state, dispatch] = useReducer(filterDatePickerFieldReducer, {
```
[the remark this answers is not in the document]
gyfchong: Probably, but I like a flatter object structure, can't really say for sure but it feels like flat = faster lol

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,140 @@ export const FilterDateRangePickerField = ({
...
+  })
+
   const inputStartDateHandlers = useDateInputHandlers({
     locale,
     disabledDays,
-    setInputValue: setInputStartDateValue,
+    setInputValue: value => {
+      dispatch({
+        type: "update_input_start_field",
+        inputValue: value as string,
```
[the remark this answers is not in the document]
gyfchong: Yeah TS is weird about this
image

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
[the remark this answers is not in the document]
HeartSquared: Haha sure :) I don't really have an issue with it. Flat is fine :)

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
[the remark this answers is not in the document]
HeartSquared: It's saying that `value` might be a SetStateAction 🤔 Have you set `setInputValue` to have a `Dispatch` type? If it should definitely be in the format of `string => void` then it should just be that type.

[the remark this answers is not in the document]
gyfchong: Oh that fixes it!

packages/date-picker/src/hooks/useDateInputHandlers.ts
```diff
@@ -12,7 +12,7 @@ import { parseDateFromTextFormatValue } from "../utils/parseDateFromTextFormatVa
 export type UseDateInputHandlersArgs = {
   locale: Locale
   disabledDays?: DisabledDays
-  setInputValue: Dispatch<SetStateAction<string>>
+  onValueChange: Dispatch<string>
```
HeartSquared:
```suggestion
  onValueChange: (value: string) => void
```

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,132 @@ export const FilterDateRangePickerField = ({
...
+      selectedRange?.from && !isInvalidDate(selectedRange?.from)
+        ? selectedRange?.from
+        : defaultMonth || new Date(),
+  })
+
   const inputStartDateHandlers = useDateInputHandlers({
     locale,
     disabledDays,
-    setInputValue: setInputStartDateValue,
+    setInputValue: value => {
```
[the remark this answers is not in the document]
gyfchong: So I just tried this change, but it's too far reaching for this PR.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,132 @@ export const FilterDateRangePickerField = ({
...
+      dispatch({
+        type: "update_input_start_field",
+        inputValue: value,
+      })
+    },
     onDateChange: date => {
-      const newDate = validateStartDate(date)
-      const endDate = parseDateFromTextFormatValue(inputEndDateValue, locale)
+      const newDate = validateStartDate(date, state.inputStartValue)
+      const endDate = parseDateFromTextFormatValue(state.inputEndValue, locale)
```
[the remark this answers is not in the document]
gyfchong: `isInvalidDate` needs a `Date`, so this is actually here to make it compatible with the validator lol

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -120,74 +105,132 @@ export const FilterDateRangePickerField = ({
...
+
+      dispatch({
+        type: "update_selected_start_date",
+        date: newDate,
+      })
+
+      dispatch({
+        type: "update_selected_end_date",
+        date: endDate,
+      })
```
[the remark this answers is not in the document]
gyfchong: I thought it was here to validate that the end date + start date are a valid combo?

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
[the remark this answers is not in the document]
HeartSquared: Or this?
```
if (newDate && endDate && !isInvalidDate(endDate)) {
```

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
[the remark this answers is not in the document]
gyfchong: TODO in a second PR

packages/date-picker/src/hooks/useDateInputHandlers.ts
gyfchong: Will do in the second PR

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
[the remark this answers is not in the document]
HeartSquared: I think it was to also validate the end date if the changing of the start date now puts the end date before it (which we show an error message for). We don't actually change the end date though.

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/utils/isValidRange.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { isValidRange } from "./isValidRange"
+
+describe("isValidRange()", () => {
+  it("returns true when the start date is before the end date", () => {
```
HeartSquared, line 4: Add a case for when they are equal

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/hooks/useEndDateValidation.spec.tsx
```diff
@@ -51,7 +51,7 @@ describe("useEndDateValidation()", () => {
           startDate: new Date("2023-05-01"),
           startDateFieldLabel: "Start date",
         })
-        expect(newDate).toBeUndefined()
+        expect(newDate).toEqual(new Date("2023-04-03"))
```
HeartSquared, line 54: You need to update the test description (also if you could put `()` in the describe that would be great)

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -134,74 +120,143 @@ export const FilterDateRangePickerField = ({
...
+      endDateInputValue: inputValue,
       startDate: selectedRange?.from,
       startDateFieldLabel: inputStartDateLabel,
     })
 
+  const [state, dispatch] = useReducer(filterDatePickerFieldReducer, {
+    selectedStartDate: selectedRange?.from,
+    selectedEndDate: selectedRange?.to,
+    inputStartValue: transformDateToInputValue(selectedRange?.from),
+    inputEndValue: transformDateToInputValue(selectedRange?.to) || "",
```
HeartSquared, line 154:
```suggestion
    inputEndValue: transformDateToInputValue(selectedRange?.to),
```

packages/components/src/FilterDateRangePicker/subcomponents/FilterDateRangePickerField/FilterDateRangePickerField.tsx
```diff
@@ -134,74 +120,143 @@ export const FilterDateRangePickerField = ({
...
+    startMonth:
+      selectedRange?.from && !isInvalidDate(selectedRange?.from)
+        ? selectedRange?.from
```
HeartSquared, line 157:
```suggestion
      selectedRange?.from && !isInvalidDate(selectedRange.from)
        ? selectedRange.from
```

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
```diff
@@ -83,5 +98,31 @@ describe("<FilterDateRangePicker />", () => {
...
+      await waitFor(() => {
+        expect(inputEndDate).toBeVisible()
+      })
+
+      await user.clear(inputEndDate)
+      await user.type(inputEndDate, "01/04/2022")
+
+      await user.click(document.body)
+
+      expect(filterButton).toBeVisible()
```
HeartSquared, line 125: The button hasn't been removed from the dom, so this will still be valid even if the text has changed. You want to check the text instead.

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.tsx
```diff
@@ -30,12 +31,14 @@ export const FilterDateRangePicker = ({
...
     renderTrigger={(triggerProps): JSX.Element =>
       renderTrigger({
-        selectedValue: isCompleteDateRange(selectedRange) ? (
-          <DateRangeDisplayLabel
-            dateRange={selectedRange}
-            locale={getLocale(locale)}
-          />
-        ) : undefined,
+        selectedValue:
+          isCompleteDateRange(selectedRange) &&
```
HeartSquared, line 35: We can probably remove this now right? Given `isValidRange` also checks that both values exist.

gyfchong: It's weird cause TS throws an error on `dateRange={selectedRange}` cause it could still be undefined.

HeartSquared: So...this?

```suggestion
          selectedRange &&
```

gyfchong: Also no bueno lol
image

HeartSquared: Ohhhhhh! Because `isCompleteDateRange` is a type guard haha. I guess it's not exactly the same as the validity check 🤔 Okay, keep it I guess lol.

HeartSquared (approved):
Seems legit!

PR #3858, opened 2023-07-10
Revert "Remove optional intl provider and update docs"

JakePitman:
Puts the OptionalIntlProvider back in KaizenProvider

### DIFF
--- a/packages/components/src/KaizenProvider/_docs/internationalization-in-kaizen.mdx
+++ b/packages/components/src/KaizenProvider/_docs/internationalization-in-kaizen.mdx
@@ -12,11 +12,11 @@ If you need guidance on translating your app check, the `@cultureamp/i18n-react-
 ## What Kaizen internationalization covers
 
 While most Kaizen components receive display text and accessibility information through props, some have text built in. For some of these cases, Kaizen provides translation support. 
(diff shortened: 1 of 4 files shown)


packages/components/src/KaizenProvider/_docs/internationalization-in-kaizen.mdx
```diff
@@ -11,8 +11,8 @@ If you need guidance on translating your app check, the `@cultureamp/i18n-react-
 
 ## What Kaizen internationalization covers
 
-While most Kaizen components receive display text and accessibility information through props, some have text built in. For some of these cases, Kaizen provides its own translations. 
-Note: Internationalization in Kaizen _does not_ handle translations for the following:
+While most Kaizen components receive display text and accessibility information through props, some have text built in. For some of these cases, Kaizen provides translation support. 
+Note: `KaizenProvider` will _not_ handle translations for the following:
```
gyfchong, line 15: Just quickly, can you flip this to be an inclusion list rather than exclusion? Since your title is about what it covers, it's a bit counter to have this be a list of exclusions.

JakePitman: Sure thing

gyfchong (approved):

PR #3862, opened 2023-07-12
Replace calendar icons

mcwinter07:
## Why
This is an interim fix to a known issue with consuming the component-library Icons in next / KAIO. This problem will be resolved when the the datepicker is moved to KAIO and uses the new SVG implementation.

## What
- replaces SVG imports from component-library with SVG components


christian314159 (approved):
👍

PR #3864, opened 2023-07-13
update ListBoxSection and SectionDivider semantics

mcwinter07:
## Why
Issue was raised that the FilterMultiSelect was experiencing problems with Keyboard navigation. After looking into the consuming code, this was due to it use adjacent ListBox components where it should have used the ListBoxSection to divide FMS Options.

This also uncovered some semantic issues, including valid FMS children and semantic structure being incorrect. This PR aims to a align as close to the suggested pattern prescribed by the [WCAG guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) and examples in the [react-aria library](https://react-spectrum.adobe.com/react-aria/useListBox.html#sections) that this uses under the hood.

We have accepted some minor accessibility issues with in order to release a fix that will address most of the problems. 

## Known issue
re: `Element has children which are not allowed: [role=presentation]`

This will only appear when using a sectionHeader in the ListBoxSection component. This is because valid children of a `listbox` role are elements with the role `option` or `group`. Despite the WCAG example providing the same guidance on `listbox` heading, this issue remains. 

It is suspected that this is either a limitation on `listbox` roles (meaning we should avoid this pattern in a re-write) or an issue with axe ci used to validate the DOM.

## Video walk through
This is slightly out of date with the current fix but a good overview of context to the problem and `listbox` issues

https://github.com/cultureamp/kaizen-design-system/assets/36558508/a4bcadb7-0daf-4e76-955d-88bd3eb2a751

## What
- adds `sectionHeader` prop to `ListBoxSection`
- updates FMS semantics
- adds story to showcase use of new prop


packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,52 @@
...
 import styles from "./ListBoxSection.module.scss"
-
 export interface ListBoxSectionProps {
   items: MultiSelectItem[]
   /**
    * Becomes an aria-label on the section, informing
    * unsighted users
    */
   sectionName: string
+  /** * If provided, will override the aria-label for this group */
```
gyfchong, line 12:
```suggestion
  /** If provided, will override the aria-label for this group */
```

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,52 @@
...
-  </div>
-)
+  sectionHeader,
+}: ListBoxSectionProps): JSX.Element => {
+  const listSectionId = sectionHeader ? v4() : undefined
+  return (
+    <li role="presentation">
+      <ul
+        className={styles.listBoxSection}
+        aria-label={sectionHeader ? undefined : sectionName}
```
gyfchong, line 28: I don't think this is a good idea, unless you enforce it with types, cause if I put ` sectionName` and a `sectionHeader` it's not obvious I'd lose functionality in one of these props. I can also imagine someone using the `aria-label` for testing...

mcwinter07: This is true, only problem is we don't want to have both aria-label and aria-lablledby - could potentially use aria-labelled and have both in that section with a visually hidden section title

mcwinter07: Though the problem is this adds hidden items into the DOM which increase the listbox counter for how many children it has 🤔

mcwinter07: Think I've found a solution that works, its still isn't read well with the screen reader (at least on osx) since we have to use role presentation, but this should should always have the sectionName as the accessible title for a group, otherwise will combine it with the sectionHeader if a sectionHeader is given.

With sectionHeader
Screenshot 2023-07-19 at 4 17 59 pm

Without sectionHeader
Screenshot 2023-07-19 at 4 25 09 pm

gyfchong (approved):

gyfchong (approved):

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,54 @@
...
   sectionName,
   children,
-}: ListBoxSectionProps): JSX.Element => (
-  <div aria-label={sectionName} className={styles.listBoxSection}>
-    {Array.from(items).map(node => node != undefined && children(node))}
-  </div>
-)
+  sectionHeader,
+}: ListBoxSectionProps): JSX.Element => {
+  const listSectionId = sectionHeader ? v4() : undefined
```
dougmacknz, line 23: This will need to be stored in state to avoid regen on rerender

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,54 @@
...
+  return (
+    <li role="presentation">
+      <ul
+        className={styles.listBoxSection}
+        aria-label={sectionName}
+        aria-labelledby={sectionHeader ? listSectionId : undefined}
```
dougmacknz, line 29: This isn't looking quite right to me yet.

In the story you've added, this ul now has an `aria-labelledby` and `aria-describedby` both pointing to the same thing.

If `sectionName` is required here, can't we always just have that as the label, and then if the header exists, link that up as the description? Which would also mean you could remove the visually hidden section name.

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,54 @@
...
+        aria-describedby={listSectionId || undefined}
+        role="group"
+      >
+        {sectionHeader && (
+          <li
+            className={styles.listBoxSectionHeader}
+            id={listSectionId}
+            role="presentation"
+          >
+            <span className="sr-only">{sectionName}. </span>
```
dougmacknz, line 39: I'm proposing we remove this, but where does this global CSS class even come from? Don't forget we have the `VisuallyHidden` component from `@kaizen/a11y` for this.

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
mcwinter07: oh actually that aria-describedby can go, I meant to remove it as it doesn't actually work in this instance

packages/select/docs/FilterMultiSelect.stories.tsx
```diff
@@ -600,3 +601,114 @@ export const Async: StoryFn<typeof FilterMultiSelect> = args => {
...
+                        persistent
+                        noBottomMargin
+                        headingProps={{
+                          tag: "span",
+                          variant: "heading-5",
+                          children: "Confidentiality protection",
+                        }}
+                      >
+                        Results for these filters are hidden to protect
+                        identities of individuals and small grounds
```
dougmacknz, line 672: Little bit nit picky and it wasn't shown in Kayd's example for some reason, but I'm pretty sure they're going to use this on the list below of `disabledItems`, to make this story a bit more of a real life example

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
mcwinter07: I assume because role presentation should remove it from the accessibility tree and while it still works with aria-labelledby, not so much with aria-describedby

packages/select/src/FilterMultiSelect/components/SectionDivider/SectionDivider.tsx
```diff
@@ -1,8 +1,9 @@
 import React from "react"
 import { Divider } from "@kaizen/draft-divider"
-
 import styles from "./SectionDivider.module.scss"
 
 export const SectionDivider = (): JSX.Element => (
-  <Divider classNameOverride={styles.divider} variant="canvas" />
+  <li className={styles.divider} aria-hidden>
```
dougmacknz, line 6:
```suggestion
  <li className={styles.divider} aria-hidden="true">
```

packages/select/docs/FilterMultiSelect.stories.tsx
mcwinter07: Makes sense to me 👍 will update

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
mcwinter07: Actually now I'm thinking about it, do think we should use the useId hook from react?

gyfchong: ❌  We can't use React 18 features yet

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
dougmacknz: Ah ok, that sounds alright then.

packages/select/docs/FilterMultiSelect.stories.tsx
mcwinter07: Only problem is that disabled items aren't navigate-able in this implementation so you can't key over them, which mean the header will not read 🤔 Part of the opinion of [react aria
](https://react-spectrum.adobe.com/react-aria/useListBox.html#disabled-items) I suspect
Screenshot 2023-07-20 at 11 12 09 am

packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -1,24 +1,61 @@
...
-  </div>
-)
+  sectionHeader,
+}: ListBoxSectionProps): JSX.Element => {
+  const [listSectionId, setListSectionId] = useState<string | undefined>(
+    undefined
+  )
+
+  useEffect(() => {
+    setListSectionId(v4())
```
dougmacknz, line 29: You can just do this in the initial state declaration, no need for the useEffect

mcwinter07: Ah right, you mean like:
```  const [listSectionId] = useState<string>(v4())```

dougmacknz: Yup, you should see that used around Kaizen here and there

dougmacknz (approved):
Noiceeee

PR #3865, opened 2023-07-14
add fix for inputType issue in TextField

mcwinter07:
## Why
InputType prop was always superseding the type prop because it was passed in with a default value, ie:
```
...
  type = "text",
  inputType = "text",
...
<Input
  {...otherProps}
  type={inputType || type}
/>
```
This would mean that no matter what `type` passed in, the `inputType` would remain the same.

## What
- remove default value from `inputType` prop so it can be undefined
  - This means `type` can be passed without defaulting to the `inputType` value
- Add test coverage to ensure TextField components have the correct type
- Update stories to use `type` instead of `inputType`

### DIFF
--- a/draft-packages/form/KaizenDraft/Form/TextField/TextField.spec.tsx
+++ b/draft-packages/form/KaizenDraft/Form/TextField/TextField.spec.tsx
@@ -92,34 +92,16 @@ describe("<TextField />", () => {
     expect(input).toBeInTheDocument()
   })
 
-  it("renders a TextField as a password field", () => {
-    render(<TextField {...defaultProps} labelText="Password" type="password" />)
-    const input = screen.getByLabelText("Password")
-    expect(input).toHaveAttribute("type", "password")
-  })
-
-  it("renders a TextField as an email field", () => {
-    render(<TextField {...defaultProps} labelText="Email" type="email" />)
-    const input = screen.getByLabelText("Email")
-    expect(input).toHaveAttribute("type", "email")
-  })
-
-  it("renders a TextField as an text field if no type is provided", () => {
-    render(<TextField {...defaultProps} labelText="Text" />)
-    const input = screen.getByLabelText("Text")
-    expect(input).toHaveAttribute("type", "text")
-  })
-
-  it("supersedes the deprecated inputType when `type` is passed in", () => {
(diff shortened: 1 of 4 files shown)


dougmacknz (approved):
Nice

draft-packages/form/KaizenDraft/Form/TextField/TextField.spec.tsx
```diff
@@ -91,4 +91,35 @@ describe("<TextField />", () => {
 
     expect(input).toBeInTheDocument()
   })
+
```
christian314159, line 94: Not sure if it's covered by the test outside of your change, but to be really thorough we should have a test for each of the 6 scenarios:
- type/inputType * text/email/password
- plus 1 that type supersedes inputType
Should be quick and easy to set up.

gyfchong: > type/inputType * text/email/password

I think the amount of tests here is a bit overkill, just testing that the `type` is passed correctly should cover all the scenarios...?

christian314159: I mean I've done more overkill in my past 😬 I understand what you mean, and it's probably enough confidence. But copy/pasting what is essentially a 2 line test 6 times and changing the value is ok-ish, too, in my book.

Just to be clear: if you @gyfchong or @mcwinter07 think you have the confidence you need, that is definitely all good with me!

mcwinter07: This is true, I suppose that two test technically cover the situation we need. One to verify that `type` passes the input type into the `TextField`, the other to verify that `type` will supersede the deprecated `inputType`.

mcwinter07: I'll just make sure that there are password stories for the Textfield to feed to chromatic to get the visual coverage

mcwinter07: Updated and ready to roll 👍

gyfchong (approved):

PR #3888, opened 2023-07-20
fix(filter-date-picker): only close popover when submitting value

HeartSquared:
## Why

[KDS-1621](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1621)

Team Pathfinder reported that the popover closed when attempting to navigate months after having selected a value.

After investigation, it was actually found to be due to the blur event of the input when a valid date exists.

## What

- Fix the logic so that the popover only closes when submitting a value.
- Add more tests
- Move `useDateInputHandlers` from `@kaizen/date-picker` to KAIO as the only usages of it are within KAIO (to prevent needing a release for date-picker for this fix)

[KDS-1621]: https://cultureamp.atlassian.net/browse/KDS-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/metal-beers-reflect.md
+++ b/.changeset/metal-beers-reflect.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix bug in `FilterDatePicker` where popover would close when blurring input as soon as a value exists.
--- a/packages/components/src/FilterDatePicker/FilterDatePicker.spec.tsx
+++ b/packages/components/src/FilterDatePicker/FilterDatePicker.spec.tsx
@@ -164,6 +164,31 @@ describe("<FilterDatePicker />", () => {
       expect(dialog).toBeInTheDocument()
     })
   })
+  it("does not close the popover when there is a selected date and the user navigates months", async () => {
+    const { getByRole } = render(
+      <FilterDatePickerWrapper selectedDate={new Date("01/01/2022")} />
+    )
+    const triggerButton = getByRole("button", {
+      name: "Drank : 1 Jan 2022",
+    })
+
+    await user.click(triggerButton)
+    const dialog = getByRole("dialog")
+
+    await waitFor(() => {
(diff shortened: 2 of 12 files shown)


.changeset/metal-beers-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix bug where popover would close when blurring input as soon as a value exists.
```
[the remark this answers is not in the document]
HeartSquared: Lol damnit! I always forget the scope isn't included in line 2 :')

packages/components/src/FilterDatePicker/FilterDatePicker.spec.tsx
```diff
@@ -86,81 +86,136 @@ describe("<FilterDatePicker />", () => {
...
+      await waitFor(() => {
+        expect(getByText("32/13/2022 is an invalid date")).toBeInTheDocument()
+      })
     })
 
-    const inputDate = getByLabelText("Date")
-    await user.clear(inputDate)
-    await user.type(inputDate, "07/06/2022")
-    await user.tab()
+    describe("Pressing Enter in the text input", () => {
```
gyfchong, line 115: Was this a new bug you found and fixed?

gyfchong (approved):

packages/components/src/FilterDatePicker/FilterDatePicker.spec.tsx
HeartSquared: Yes and no. You had tests for it, but because the "on submit" was in the wrong place, the test was inaccurate. Moreso that I split it.

PR #3889, opened 2023-07-20
test(filter-drp): refactor tests to hopefully not timeout

HeartSquared:
## What

Refactor FilterDRP tests to not use `screen` in hopes that they run faster and don't time out.
Edit: ^ didn't work, so just extended timeout.

## Why

The timeouts are driving me mad.


.changeset/weak-humans-film.md
```diff
@@ -0,0 +1,2 @@
+---
```
mcwinter07, line 1: Did you mean to make an empty changeset? (also amazing generative name)

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
```diff
@@ -35,95 +35,88 @@ const FilterDateRangePickerWrapper = ({
...
-    render(<FilterDateRangePickerWrapper />)
-    expect(screen.queryByRole("dialog")).not.toBeInTheDocument()
+    const { queryByRole } = render(<FilterDateRangePickerWrapper />)
+    expect(queryByRole("dialog")).not.toBeInTheDocument()
   })
 
   describe("Filter button", () => {
     it("should show the selected range in the button", () => {
-      render(
+      const { getByRole } = render(
```
mcwinter07, line 44: I've always opted for `screen.{selectorType}`, just kind of curious on the restructured approach here. Is is to reduce verbosity or something that contributes to the timeouts?

mcwinter07 (commented):
So on board with this PR 😆

.changeset/weak-humans-film.md
mcwinter07: nvm, covered in standup 👍

mcwinter07 (approved):

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
HeartSquared: It was an attempt to see if it was related (since I normally don't use `screen` and haven't had any timeouts recently.
Then it turns out it failed anyway, so I guess the performance is the same. But the changes were already done so I left it lol.

PR #3891, opened 2023-07-21
update packages and test label name

mcwinter07:
## Why
This was uncovered when bumping to the latest version of react-aria/select and all its dependencies. The accessible name has changed its order, where before it read: {select label} {select value}, now it reads as {select value} {select label}. 

## Video explainer

https://github.com/cultureamp/kaizen-design-system/assets/36558508/f9857b58-d982-406e-ba9d-0a8bfc8e0a61

## More context
For more context see [slack thread discovery](https://cultureamp.slack.com/archives/C02NUQ27G56/p1689124458661699) between  and  

## Potential cause
Likely due to the update that came in this [PR](https://github.com/adobe/react-spectrum/pull/4191/files) 
> useLabels/useLabel consider reordering IDREFs so that aria-label comes before aria-labelledby

## Underlying / larger issue

React-aria seems to determine its accessible name,  pushing the value’s id into aria-describedby. This causes our tests to break because we are forced to add the value into our queries that target accessible names, ie:  getByRole("combobox", {name: {labelName} {labelValue}).

The value ideally should not be passed to the aria-labeled by and should simply be the value of the combobox (see the [WCAG example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) implementation of this)

## Intended change

Hijack the aria-labelledby and pass only the label id into the prop. Let the value be natively interpreted as a child element as done in the WCAG example


mcwinter07:
This has been resolved

PR #3892, opened 2023-07-21
Enable-anyone-launch-release

gyfchong:
## Why

When Changesets opens a PR it does not trigger GitHub actions, because its token does not have permission to. This is a hack to allow one of us to trigger GitHub actions for a changesets PR by enabling automerge on the PR.

## What

Updated Github Action for ci to listen for `auto merge` trigger which will activate the ci checks


.github/workflows/changeset.yaml
```diff
@@ -5,6 +5,18 @@ on:
...
 
+  # When Changesets opens a PR it does not trigger GitHub actions,
+  # because its token does not have permission to. This is a hack
+  # to allow one of us to trigger GitHub actions for a changesets PR
+  # by enabling automerge on the PR.
+  pull_request_target:
+    types:
+      - auto_merge_enabled
+    branches:
+      - master # the target branch of the PR
```
HeartSquared, line 16:
```suggestion
      - main # the target branch of the PR
```

HeartSquared:
So I'm guessing the idea of this is that it doesn't actually auto-merge unless it passes CI yeah? But I suppose if it made its way to the changeset PR, it should have passed CI so it shouldn't ever fail.

.github/workflows/changeset.yaml
gyfchong: Oops copy pasta

gyfchong:
> So I'm guessing the idea of this is that it doesn't actually auto-merge unless it passes CI yeah? But I suppose if it made its way to the changeset PR, it should have passed CI so it shouldn't ever fail.

@HeartSquared pretty much, this is what was recommended and used in FEF, so it's tested.

.github/workflows/changeset.yaml
```diff
@@ -5,6 +5,18 @@ on:
     branches:
       - "main"
 
+  # When Changesets opens a PR it does not trigger GitHub actions,
+  # because its token does not have permission to. This is a hack
+  # to allow one of us to trigger GitHub actions for a changesets PR
+  # by enabling automerge on the PR.
+  pull_request_target:
```
HeartSquared, line 12: I think you might need a new workflow file, because currently it looks like this workflow only runs _on_ `main`.

You'll need another one which runs on the `changeset-release/main` branch instead.

gyfchong: Oh you're right, they use it on their CI which monitors all branches

.github/workflows/ci.yml
```diff
@@ -5,6 +5,18 @@ on:
...
 
+  # When Changesets opens a PR it does not trigger GitHub actions,
+  # because its token does not have permission to. This is a hack
+  # to allow one of us to trigger GitHub actions for a changesets PR
+  # by enabling automerge on the PR.
+  pull_request_target:
+    types:
+      - auto_merge_enabled
+    branches:
+      - master # the target branch of the PR
```
HeartSquared, line 16:
```suggestion
      - main # the target branch of the PR
```

.github/workflows/ci.yml
```diff
@@ -5,6 +5,18 @@ on:
     branches-ignore:
       - main
 
+  # When Changesets opens a PR it does not trigger GitHub actions,
```
HeartSquared, line 8: Hmm... Maybe we need to add this to the Storybook tests flow too? 🤔 It will get skipped, but the workflow still needs to be triggered.

Alternatively we could create a label? Although I guess this saves you from multiple steps.

gyfchong: Oh, cause now it's separated, yeah.

HeartSquared (approved):

PR #3900, opened 2023-07-24
Issue 98: add ErrorPage component

lisatassone:
## Why
This PR moves much of what was in https://github.com/cultureamp/error-pages into Kaizen as dedicated error-handling components.

This is one of the improvement opportunities mentioned [here](https://github.com/cultureamp/engineering-central/discussions/21).

From here both the `error-pages` repo and `next-template` can be updated.

https://github.com/cultureamp/kaizen-discourse/issues/98
SR ticket: https://cultureamp.atlassian.net/browse/SR2-514


packages/components/src/ErrorPage/_docs/ErrorPage.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  args: {
+    callToAction: {
+      onContactSupport: () => alert("Custom handler"),
+    },
+  },
+} satisfies Meta<typeof ErrorPage>
+
+export default meta
+
+export const Error400: StoryObj<typeof meta> = {
```
lisatassone, line 35: Perhaps these should be in a stickersheet story? I had originally removed that file because I didn't have much more to show apart from the main component story. Let me know if you want them moved.

packages/components/src/ErrorPage/subcomponents/BaseErrorPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Paragraph } from "@kaizen/typography"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./BaseErrorPage.module.scss"
+
+const getMailToHref = (code: number): string => {
+  const supportEmail = "support@cultureamp.com"
+  const subject = "Houston we have a problem"
+  const body = `Hi there,\n\nI received a ${code} error page while I was trying to...`
+  return encodeURI(`mailto:${supportEmail}?subject=${subject}&amp;body=${body}`)
+}
```
lisatassone, line 17: I wasn't sure whether to I18n these so I've left them as is for now.

packages/components/src/ErrorPage/subcomponents/BaseErrorPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+  code: number
+  title: string
+  message: React.ReactNode | string
+  callToAction?: {
+    onContactSupport: () => void
+    homeHref?: string
+  }
+} & OverrideClassName<HTMLAttributes<HTMLDivElement>>
+
+export const BaseErrorPage = ({
```
gyfchong, line 31: Looking at the props for this `Base` it doesn't look like there's much room for customisation if the consumers were to use this directly, only being able to alter the text of two props + `callToAction`.
Also each Error Page doesn't seem to need to be customised further than updating the `title` and `message`, which leads me into thinking that this multi-component approach is a bit too verbose.

What if we consider the `title` and `message` as optional props with defaults that get filled in based on the code. So something like:

```
const ErrorPage = ({ code, message }) => {
  const messages = { 
    404: " ... "
  }

  return (<> {message ?? messages[code]} </>
}
```
This to me achieves basically the same amount of customisation and preset defaults as it does now without needing multiple individual components. We also won't need to create a component for every new error code, just add some new messages to the object.

packages/components/src/ErrorPage/_docs/ErrorPage.stories.tsx
```diff
@@ -0,0 +1,79 @@
+import { Meta, StoryObj } from "@storybook/react"
+import { ComponentDocsTemplate } from "../../../../../storybook/components/DocsContainer"
+import { ErrorPage } from "../index"
+
+const meta = {
+  tags: ["autodocs"],
+  title: "Components/ErrorPage",
```
gyfchong, line 7: Pop these into `Pages/Error Page` 😄

packages/components/src/ErrorPage/subcomponents/BaseErrorPage.tsx
lisatassone: > Looking at the props for this Base it doesn't look like there's much room for customisation
You can change all the internals of what this component does with these props so I don't know what you mean here.

> What if we consider the title and message as optional props with defaults that get filled in based on the code. So something like

Sure it could be done like that.

packages/components/src/SVG/icons/EmailIcon.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import { SVG, SVGProps } from "~components/SVG"
+
+export const EmailIcon = (props: Omit<SVGProps, "children">): JSX.Element => (
```
lisatassone, line 4: I followed the guide but the icons don't render. May be to do with the component they are used in. Didn't have time to investigate. Any immediate ideas?

gyfchong: Ah you're right, we have to make a compatible copy of BrandMoment. I'll hop in and help you with that.

packages/components/src/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const supportEmail = "support@cultureamp.com"
+  const subject = "Houston we have a problem"
+  const body = `Hi there,\n\nI received a ${code} error page while I was trying to...`
+  return encodeURI(`mailto:${supportEmail}?subject=${subject}&amp;body=${body}`)
+}
+
+const HOME_HREF = "/app/home"
+
+export type ErrorPageProps = {
+  code: number
```
gyfchong, line 23: Could we use `400 | 404 | ... | number` as the type here?

packages/components/src/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const ErrorPage = ({
+  code,
+  title,
+  message,
+  callToAction,
+  classNameOverride,
+}: ErrorPageProps): JSX.Element => {
+  const { formatMessage } = useIntl()
+  const translations = useMessages(code)
```
gyfchong, line 40: I think the naming of `useMessages` is too close to the `i18n-react-intl` package, and I feel like using the word `content` here better describes what it's getting used as.

```suggestion
  const content = useErrorMessages(code)
```

packages/components/src/ErrorPage/_docs/ErrorPage.stories.tsx
lisatassone: I think the best tests for this are the visual snapshot tests, so that's why I've added these stories.

packages/components/src/ErrorPage/subcomponents/BaseErrorPage.tsx
lisatassone: That was my thinking 👍

packages/components/src/ErrorPage/ErrorPage.tsx
lisatassone: Typescript will squash the type down to just number if we do that, so we'd lose any benefit of listing them.

packages/components/src/ErrorPage/ErrorPage.tsx
lisatassone: done

packages/components/src/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+const HOME_HREF = "/app/home"
+
+export type ErrorPageProps = {
+  code: number
+  title?: string
+  message?: React.ReactNode | string
+  callToAction?: {
+    onContactSupport: () => void
+    homeHref?: string
+  }
```
lisatassone, line 29: Normally I would steer away from defining props in this manner and instead allow the consumer to pass through a render function.  But we do want this component to have an opinion on how our error pages should look and what they should do, so being able to change just these things may be okay.

It may also not be clear that the call to action will always appear even if you don't set anything, but again, maybe that's not a biggie. Its intention is to be an override.

packages/components/src/ErrorPage/ErrorPage.tsx
lisatassone: I don't know of a way to get the type to list the statuses that we DO cover but also allow any number as well.

Perhaps we can make the statuses template literals but also allow numbers as well?

So `400` | `422` ... | number

gyfchong: Yeah I think it's important they get auto complete. Do you have any ideas here @HeartSquared? Otherwise I'm fine with the string approach.

packages/components/src/ErrorPage/_docs/ErrorPage.stories.tsx
gyfchong: This is a bit awkward, we do want consumers to be able to see the outcome for each code, but at the same time this seems expensive.

Let's do a sitckersheet, but for the docs can we simply output the mapping in a table of some sort? Just to represent what to expect when providing the error code

packages/components/src/ErrorPage/ErrorPage.tsx
HeartSquared: image

I just tested whether it would work if I made a `NumberSuggestions` and sadly it doesn't. Typing the quotes is what triggers the auto-complete, and braces do not (which is what numbers use).

You could definitely use `StringSuggestions` and therefore accept only strings, but if you want to accept any number, there's no way to offer suggestions (`400 | number` will end up just as `number`).

The other option is to have your type be a subset of numbers and accept a different type (eg. string) as the flexible type. But accepting multiple types is more gross imo so I wouldn't recommend it.

gyfchong: @lisatassone is it likely that there will be a custom one? I'm afraid of optimising for an edge case.

I can only think of returning to the `BaseErrorPage` to accept any number, essentially boiling it down to the `ErrorPage` being a strict wrapper with context and passing it through.

```
const ErrorPage = ({ code }) => {
  const content = useErrorMessages(code)
  <BaseErrorPage code={code} {...content} />
}
```

Which doesn't look too bad?

packages/components/src/ErrorPage/ErrorPage.tsx
gyfchong: Yeah, a bit confusing, but for now should be okay.

packages/components/src/ErrorPage/ErrorPage.tsx
lisatassone: @gyfchong I'm not sure. It doesn't have to be a number that's passed so if you want to do what @HeartSquared is suggesting I can add that even though the signature is a bit weird, or we just allow the statuses we cover.

If you can let me know which option you want before midday that would be great. Otherwise feel free to jump in and edit what is there.

gyfchong: Let's just allow the statuses we cover for now.

gyfchong (approved):

PR #3901, opened 2023-07-24
fix: colour contrast of draft select placeholder

caitlinmwillington:
## Why

The placeholder text of the draft-select does not meet colour contrast requirements (issue raised in audit). We still rely on this version of the draft-select as we need the filterable and clearable functionality. This PR brings the contrast inline with the latest version of the Select component which uses `color: rgba($color-purple-800-rgb, 0.7);` from `packages/select/src/Select/components/TriggerButton/TriggerButton.module.scss`

### Before (draft select)
Screenshot 2023-07-24 at 2 08 48 pm

### New select version
Screenshot 2023-07-24 at 2 10 39 pm

### After (draft select)
Screenshot 2023-07-24 at 2 10 56 pm


dougmacknz:
Interesting one, cause Intopia also said they wouldn't recommend increasing the contrast either:
image

Would it be possible for you to rework your usages to not use a placeholder at all?

PR #3902, opened 2023-07-24
removed unused github ssh key from buildkite release

gyfchong:
Doesn't seem like the Github SSH was necessary for the flow to work. I think it was a relic of Lerna


HeartSquared (approved):
You can try this, but isn't it needed so you can checkout main to pull in the changes for release?

gyfchong:
> You can try this, but isn't it needed so you can checkout main to pull in the changes for release?

Should be the GH Token that does this... Hopefully...

PR #3907, opened 2023-07-25
Migrate-brandmoment-kaio

gyfchong:
## Why
We need a React icon compatible BrandMoment for Error Pages

## What
Replaced `@kaizen/button` with internal KAIO `Button`


packages/components/src/BrandMoment/components/Box/Box.tsx
dougmacknz, line 1: Oof, do you want me to quickly get rid of the Box usages in this component? 😅

gyfchong: Oh sure, that would be great

packages/components/src/BrandMoment/BrandMoment.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+  beforeEach(() => {
+    window.HTMLMediaElement.prototype.load = mockLoad
+    window.HTMLMediaElement.prototype.play = mockPlay
+    window.HTMLMediaElement.prototype.pause = mockPause
+    Object.defineProperty(HTMLMediaElement.prototype, "muted", {
+      set: jest.fn(),
+    })
+  })
+
+  it("matches the snapshot", () => {
```
HeartSquared, line 33: Just kill this file. It's only a snapshot test, and you didn't even include it lol

PR #3910, opened 2023-07-25
chore: use Storybook for interaction tests instead of Playwright

HeartSquared:
## Why

[KDS-1568](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1568)

Playwright tests rely on a successfully deployed Storybook, but we have found that our branch previews deployed to AWS have at time been flaky (the contents reference will be incorrect, therefore have nothing loaded).

## What

- Install Storybook testing suite and `storybook/test-runner`
- Convert our Playwright tests to either Jest or Storybook interaction tests (essentially Playwright tests under the hood)
- Extend our Chromatic workflow to run our interaction tests based on the deployed Chromatic branch
	- While it appears that the new workflow takes a lot longer than the old Playwright flow, it is because we now test all 400+ stories, whereas Playwright used to only have 19 tests
- Separate the PR and `main` branch Chromatic workflows (they have different intended workflows)
- Remove Playwright and associated config

[KDS-1568]: https://cultureamp.atlassian.net/browse/KDS-1568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/chromatic-main.yaml
```diff
@@ -1,21 +1,14 @@
 # Chromatic action from https://www.chromatic.com/docs/github-actions
 
-name: "chromatic"
```
gyfchong, line 3: Do we need to distinguish the naming? These aren't going to run in the same check list right?

.github/workflows/chromatic-pr.yaml
```diff
@@ -0,0 +1,74 @@
+# Chromatic action from https://www.chromatic.com/docs/github-actions
+
+name: "chromatic and storybook tests (PR)"
```
gyfchong, line 3:
```suggestion
name: "test storybook"
```

Should cover both cases

.github/workflows/chromatic-main.yaml
HeartSquared: Yeah I was wondering about that. Technically they shouldn't, but just in case? 😬 Also just to be explicit that this one is only for `main`, although I guess I could add a description instead 🤔

.github/workflows/chromatic-pr.yaml
HeartSquared: Can do. I'll rename the file too.

.github/workflows/test-stories.yaml
```diff
@@ -0,0 +1,80 @@
+# Chromatic action from https://www.chromatic.com/docs/github-actions
+
+name: test-stories
```
gyfchong, line 3:
```suggestion
name: test stories
```
Consistency 😄

package.json
```diff
@@ -20,8 +20,8 @@
     "storybook:clean": "rm -rf storybook/public",
     "storybook:build": "storybook build -c storybook -o storybook/public --docs",
     "storybook:build-deps": "yarn workspace @kaizen/design-tokens prepublish && yarn workspace @kaizen/tailwind prepublish",
-    "playwright": "IS_DEV=true npx playwright test",
-    "playwright:debug": "IS_DEV=true PWDEBUG=1 npx playwright test",
+    "test:storybook": "STORIES=ignoreBrandMoment test-storybook --config-dir storybook",
```
gyfchong, line 23:
```suggestion
    "test:storybook": "STORIES=whitelist test-storybook --config-dir storybook",
```

More generic name in case we need to ignore more?

HeartSquared: Ideally we shouldn't be ignoring anything. I named it this to call out that it's super strange and ideally we'd fix it. It's actually the opposite of a whitelist hahaha.

gyfchong (approved):

PR #3917, opened 2023-07-28
Fix: Modal actions rendering bug

jakubriedl:
## Why
When multiple action buttons happened to have the same label it was causing react to render broken UI because it was rendering 2 different components with same key. And this broken UI stayed even after rerender with different labels.

It is expected to have the same labels `" "` during load time when user locale is being loaded.

## What
change the key


.changeset/green-donkeys-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/draft-modal": patch
+---
+
+Fix: A rendering bug when labels for actions were the same (even for short period of time)
```
gyfchong, line 5: Seems like a flag if you're giving two buttons the same label, what's the exact scenario you're trying to fix?

jakubriedl: It's there just for split second and breaks react.
`i18n-react-intl` shows during locale loading all strings as `" "` which means they are the same. When the right labels are loaded the components get rerendered. But by then it's already broken.

gyfchong (approved):

PR #3918, opened 2023-07-28
update changeset gh action to trigger release

gyfchong:
Fix that release pipeline


HeartSquared (dismissed):
Aww we lost the Slack notifications. But hopefully it works! 🤞🏻

HeartSquared (approved):
Let's try it

PR #3926, opened 2023-07-31
add slack notifications

gyfchong:
## Why
To know when our publish fails or succeeds

## What

Add a slack notification action for both success and fail after publish occurs.

### DIFF
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -15,6 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
       - name: Create Release Pull Request or Publish to npm
+        id: changesets
         run: |
           npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
       - uses: changesets/action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/changeset.yaml
```diff
@@ -23,3 +24,21 @@ jobs:
...
+          payload-file-path: "../payload/slack-notify-release.json"
+      - name: Notify Slack (failure)
+        if: steps.changesets.outputs.published == 'false'
+        uses: slackapi/slack-github-action@v1.24.0
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+        with:
+          channel-id: "wol_kaizen"
+          payload-file-path: "../payload/slack-notify-release.json"
```
HeartSquared, line 44: Since there's no difference between success and failure, doesn't it mean you can just get rid of the condition and just run the steps?

gyfchong: Hmm I think the output value only exists when publish is run, so having the conditions here means that it won't notify us of a new Changeset PR.

HeartSquared: I wonder if we could check whether it's defined 🤔

HeartSquared (approved):

PR #3936, opened 2023-08-01
ci: run and skip workflows on release branch automerge

HeartSquared:
## Why

Allow the whole team to do releases.

## What

Update required CI workflows to run but skip all steps on the release branch.

### DIFF
--- a/.github/workflows/check-files.yml
+++ b/.github/workflows/check-files.yml
@@ -4,10 +4,22 @@ on:
   push:
     branches-ignore:
       - "main"
-      - "changeset-release/*"
+
+  # When Changesets opens a PR it does not trigger GitHub actions,
+  # because its token does not have permission to. This is a hack
+  # to allow one of us to trigger GitHub actions for a changesets PR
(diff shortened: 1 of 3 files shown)


gyfchong (approved):

.github/workflows/check-files.yml
```diff
@@ -4,10 +4,22 @@ on:
...
+    types:
+      - auto_merge_enabled
+    branches:
+      - main # the target branch of the PR
+    paths:
+      - "**/CHANGELOG.md" # only changesets releases touch changelogs
 
 jobs:
   changesets-exists:
+    if: github.head_ref != 'changeset-release/main'
```
gyfchong, line 22: Wait, does this not work if you add it to Line 6?

HeartSquared: Yeah, in line 6 you'll skip running the whole workflow, which therefore doesn't tick off the requirement from the GH settings. If you put it here, you run the job, but skip the actual steps.

PR #3938, opened 2023-08-01
add push tags

gyfchong:
We need to manually push tags for our published packages.

### DIFF
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -24,8 +24,19 @@ jobs:
           commit: "version packages"
           version: yarn ci:version
           publish: yarn ci:publish
+  push-tags:
+    runs-on: ubuntu-latest
+    needs: version
+    if: needs.version.outputs.published == 'true'
+    steps:
+      - uses: actions/checkout@v3
+      - run: git push --follow-tags
+  slack-notify:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mcwinter07 (commented):
Was this just a step we were missing before with changesets or has a recent changes necessitated this update?

gyfchong:
> Was this just a step we were missing before with changesets or has a recent changes necessitated this update?

It's a step that was included in the Buildkite scripts that was missed when we moved to GH Actions 😅

mcwinter07:
> > Was this just a step we were missing before with changesets or has a recent changes necessitated this update?
> 
> It's a step that was included in the Buildkite scripts that was missed when we moved to GH Actions 😅

Ah gotcha, nice catch! 👍 Will give it an ol' tickeroo ✅

mcwinter07 (approved):

.github/workflows/changeset.yaml
```diff
@@ -24,8 +24,19 @@ jobs:
           commit: "version packages"
           version: yarn ci:version
           publish: yarn ci:publish
+  push-tags:
+    runs-on: ubuntu-latest
+    needs: version
+    if: needs.version.outputs.published == 'true'
```
HeartSquared, line 30: You may need to actually set up the outputs within `version`

PR #3939, opened 2023-08-01
remove clean from turbo build

gyfchong:
## Why
Cleaning before every build defeats the benefits of turbo's cache

## What
- Remove the dependency on running `clean` whenever a `build` is run.
- Add clean to the individual package's `build` commands.

## Notes
How Turbo works is if it detects any changes/is a fresh local build, it will run the `build` commands in the individual packages. If it doesn't, then it skips running the commands.

So in essence, we still get the benefit of cleaning before building, but it only happens if the files change, versus the old way where it would clean every time regardless of if the files had changed -- triggering the miss in cache and having to build it all from scratch every time.


.changeset/sixty-plants-wonder.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
gyfchong, line 2: Nothing has changed in the components' code, so leaving this empty.

HeartSquared (approved):

PR #3942, opened 2023-08-02
add canary release workflow

mcwinter07:
## Why
Since the switch to changeset we have not had a release strategy for canary's and relied on yalc to test new release features in consuming repos. This PR aims to adopt the [Front Ops canary release workflow
](https://github.com/cultureamp/frontend-ops/blob/master/.github/workflows/release.yml) tailored to our environments.

## What
- adds a canary release workflow

### DIFF
--- a/.github/workflows/canary-release.yml
+++ b/.github/workflows/canary-release.yml
@@ -17,14 +17,11 @@ jobs:
       LABEL: ${{ github.event.inputs.label }}
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version-file: ".nvmrc"
-          cache: yarn
-      - run: |
-          npm config set "//npm.pkg.github.com/:_authToken" "$PACKAGE_REGISTRY_ACCESS_TOKEN"
-          yarn install --frozen-lockfile --ignore-scripts
-          npm config set "//npm.pkg.github.com/:_authToken" "$GITHUB_TOKEN"
+      - uses: ./.github/actions/setup
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        run: |
+          npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
       - run: |
           git config --global user.email "cultureamp-ci@cultureamp.com"
           git config --global user.name "cultureamp-ci"
--- a/.github/workflows/canary-release.yml
+++ b/.github/workflows/canary-release.yml
@@ -1,0 +1,34 @@
+name: "Canary release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      label:
+        description: "Pick a label for the release, to be included in the version number — e.g. '1.0.0-\\${label}.x' (note: can't contain slashes!)"
+        required: true
+        type: string
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ github.token }}
+      PACKAGE_REGISTRY_ACCESS_TOKEN: ${{ secrets.PACKAGE_REGISTRY_ACCESS_TOKEN }}
+      LABEL: ${{ github.event.inputs.label }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: ".nvmrc"
+          cache: yarn
+      - run: |
+          npm config set "//npm.pkg.github.com/:_authToken" "$PACKAGE_REGISTRY_ACCESS_TOKEN"
+          yarn install --frozen-lockfile --ignore-scripts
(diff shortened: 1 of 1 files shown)


.github/workflows/canary-release.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: ./.github/actions/setup
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        run: |
+          npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
+      - run: |
+          git config --global user.email "cultureamp-ci@cultureamp.com"
+          git config --global user.name "cultureamp-ci"
+      - run: |
+          yarn compile
```
mcwinter07, line 29: General wondering if we need/want to use turbo here? I ask mainly because ci:version and ci:publish use turbo prepublish so could potentially be:
```yarn turbo prepublish && yarn changeset version --snapshot "$LABEL"```

File: .github/workflows/canary-release.yml [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        run: |
+          npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
+      - run: |
+          git config --global user.email "cultureamp-ci@cultureamp.com"
+          git config --global user.name "cultureamp-ci"
```
gyfchong, line 27: I wonder if we really need these, cause our changeset workflow doesn't 🤔

mcwinter07: Fair question, I assumed it just changes who the canary is authored by?

.github/workflows/canary-release.yml
gyfchong: `yarn ci:version --snapshot "$LABEL"` might work?

.github/workflows/canary-release.yml
gyfchong: Let's give it a go without. At least this workflow is triggered lol

gyfchong:
I think it's worth taking a look at adding a comment that includes the canary version number to the PR that launched the canary.

.github/workflows/canary-release.yml
mcwinter07: Worth a shot!

.github/workflows/canary-release.yml
```diff
@@ -0,0 +1,31 @@
+name: "Canary release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      label:
+        description: "Pick a label for the release, to be included in the version number — e.g. '1.0.0-\\${label}.x' (note: can't contain slashes!)"
```
gyfchong, line 7:
```suggestion
        description: "Pick a label for the release, to be included in the version number"
```
The example confused more than helped, I think if you are able to get a comment with the version, we won't need it.

.github/workflows/canary-release.yml
```diff
@@ -0,0 +1,31 @@
...
+        id: changesets
+        run: |
+          npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
+      - run: |
+          git config --global user.email "cultureamp-ci@cultureamp.com"
+          git config --global user.name "cultureamp-ci"
+      - run: |
+          yarn compile
+          yarn changeset version --snapshot "$LABEL"
+          yarn changeset publish --tag "$LABEL"
```
gyfchong, line 31: Now that I think about it, I think either of these even have an output, so making a comment doesn't seem possible...

mcwinter07: We could just have it post to slack after with the canary name if we want to track it somewhere?

gyfchong: Hm, take a look at the changesets gh action, maybe instead of CLI we can use that. Cause that one outputs the versions.

I don't think having the version somewhere is overly useful if it were manually done, the value to me is in the automation. Workflow wise if I had to check for the version, copy then paste it to somewhere in slack/on this PR, then paste it to the consumer who needed it -- overall adding a step that isn't all too helpful.

mcwinter07: If we're going to use `changesets/action@v1` it looks like its going to be near identical with our changeset yaml? just with a label passed in the extra flags at the end for the snapshot and tag

gyfchong (approved):
Give this a go first, if it works, then we can add a comment like chromatic does!

.github/workflows/canary-release.yml
gyfchong: Should be fine, tempting to consolidate but I think people external to the team might make canary's, so I don't want all the notifications setup for package publishing to fire when a canary gets released 😅

PR #3943, opened 2023-08-02
Skip running chromatic flow again on reopen of PR

gyfchong:
## Why
Weird edge case where if your chromatic ci step has previously built and passed all tests, and you close/re-open your branch, it will skip the publishing of chromatic (since nothing has changed) causing the subsequent test steps to fail

## What
Remove the "reopened" type of pull request event


.github/workflows/test-stories.yaml
```diff
@@ -7,7 +7,6 @@ on:
   pull_request:
     types:
       - opened
-      - reopened
```
HeartSquared, line 10: Hahaha, I suppose this is another option :P

HeartSquared (approved):

PR #3944, opened 2023-08-02
explicitly set changeset createGithubReleases

gyfchong:


.github/workflows/changeset.yaml
```diff
@@ -26,6 +26,7 @@ jobs:
           commit: "version packages"
           version: yarn ci:version
           publish: yarn ci:publish
+          createGithubReleases: true
```
HeartSquared, line 29: Does this push tags for us? :o

HeartSquared (approved):

.github/workflows/changeset.yaml
gyfchong: Unfortunately no lol

PR #3945, opened 2023-08-03
Make id props optional everywhere

dougmacknz:
## Why
Having ids mandatory creates more work/code for people unnecessarily, and makes us more susceptible to duplicate id issues.

## What
In a bunch of our older components, we have a mandatory `id` prop. Make these optional and use `uuid` to generate one if one isn't passed in.

### DIFF
--- a/.changeset/pretty-mugs-peel.md
+++ b/.changeset/pretty-mugs-peel.md
@@ -1,0 +1,9 @@
+---
+"@kaizen/draft-collapsible": patch
+"@kaizen/date-picker": patch
+"@kaizen/draft-form": patch
+"@kaizen/components": patch
+"@kaizen/select": patch
+---
+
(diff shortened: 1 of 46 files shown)


gyfchong:
Are the visual changes expected?

dougmacknz:
> Are the visual changes expected?

Yep, the radio and checkbox stickersheets had a `noBottomMargin` column that weren't actually using the `noBottomMargin` prop so I stuck a fix in.

gyfchong (dismissed):

dougmacknz:
@gyfchong I ended up changing this a bit sorry.

I changed my mind on primitives, I think `id` should be required there because it should be getting paired up with a label.

Similar with subcomponents, we're hopefully the only ones using those directly, so they're fine to stay is.

So just top level components that will get this change now.

gyfchong (approved):

PR #3948, opened 2023-08-03
remove deprecated prepublish + add clean back to build

gyfchong:
## Why

For local development we want to clean whenever we generate a new build to ensure that any renamed/deleted files are not included in the dist

For CI, the dist files do not exist yet, so it is safe to create without cleaning

Also replace deprecated`prepublish` command

## What
- `build` will clean and execute `prepublishOnly`
- `prepublishOnly` is to be used by CI when publishing

**prepublishOnly**
Runs BEFORE the package is prepared and packed, ONLY on `npm publish`.

Versus `prepublish` which ran on `npm install` and `npm publish`

Any commands that need packages built for local purposes should use the `build` command.


mcwinter07 (approved):
Looks hunky dory to me 👍 Was a little confused by the turbo config (mainly the "^") but the docs cleared that up 😉

PR #3949, opened 2023-08-03
Fix push tags after publish

gyfchong:
Didn't work as a new job, but also unnecessary to have it push the tags in a separate job, simplified the command.


mcwinter07 (approved):
Looks good 👍  so something like this for the canary release workflow should work too

PR #3953, opened 2023-08-03
chore: set up vitest for React-18

HeartSquared:
## What

Set up vitest to replace jest for React-18 (excludes RTE).

RTE and React 16/17 remain using only jest (see **Limitations** 1 & 2).

For all `*.spec.ts(x)` tests, use the global `vi` for functions like `vi.fn()`.
These are aliased when running jest.
However, if using `mock` functions (see **Limitations** 4), you must split the tests into `*.spec.vi.ts(x)` (using `vi.mock`) and `*.spec.jest.ts(x)` (using `jest.mock`).

Also found the jest config in KAIO to be redundant, so removed it and made the global one handle it instead :)

## Limitations

Unfortunately there are a few caveats which prevent vitest from fully replacing jest.

### 1. Cannot support multiple React versions

Vitest does not have the ability to modify code within `node_modules`, therefore we cannot tell it to use `react-16/17` and `react-dom-16/17` (our aliases from `package.json`).
While it does offer alias support, it is only for the components within our repo and not for dependencies.

The following error occurred when applying the following config mapping `react(-dom)` with suffix `-16`:
```
  config.resolve = {
    ...config.resolve,
    alias: [
      { find: /^react-dom((\/.*)?)$/, replacement: "react-dom-16$1" },
      { find: /^react((\/.*)?)$/, replacement: "react-16$1" },
      { find: /^@testing-library\/react((\/.*)?)$/, replacement: "@testing-library/react-12$1" },
      { find: /^react-test-renderer((\/.*)?)$/, replacement: "react-test-renderer-17$1" },
    ],
  }
```
image

Despite the alias working on the surface (since we see `react-16`), it did not apply within `react-16` itself, which was referencing `react-dom` (v18).

### 2. Cannot support external deps having deps on our repo

Much like why we cannot support multiple React versions, any dependencies we consume that in turn consume our repo also fail to compile as the alias-ing does not apply.

This only happens for the `@kaizen/rich-text-editor` package as it consumes `@cultureamp/rich-text-toolkit` which consumes `@kaizen/button` and some others.

This circular-ish dependency (which we have flagged should be fixed anyway) is not compatible with vitest.

image

### 3. Fake timers get stuck within waitFor

When using `useFakeTimers`, `testing-library`'s `waitFor` has problems resolving, therefore getting stuck.
A fix has technically been made within the library, but it references the `jest` global, which means it will not work naturally if you use vitest.

To resolve this, we alias any `vi` time-related functions in `vitest.setup` to the `jest` namespace.

### 4. Cannot alias `.mock`

`vi.mock` and `jest.mock` are too different under the hood, and cannot be aliased to each other.

If you find yourself needing mocking, you will need to write duplicate tests (or even if just one, you still need to scope the extension) as `*.spec.vi.ts(x)` and `*.spec.jest.ts(x)`.

### 5. Snapshot tests

We do not encourage writing snapshot tests (as you should use either better assertions or Chromatic instead), but where they existed, I have made them only apply to jest.

This is due to the snapshots updating with a comment naming the last runner, so it was changing between `Vitest` and `Jest`, which we shouldn't really care about.


HeartSquared:
Figure it out again if and when it's wanted

PR #3955, opened 2023-08-03
Fix-circular-dependencies

gyfchong:
## Why
We shouldn't have circular deps

## What
Switched importing of components and types to rely on the source which exports it, rather than the root level `index` which also exports the same things.

## Note
Couldn't figure out the `FilterBar` circular types dep :(


packages/components/src/Workflow/subcomponents/Header/Header.tsx
```diff
@@ -1,6 +1,12 @@
...
-import { HeaderRoot, Branding, Titles, Actions } from "./components"
-import { WorkflowActionsProps, WorkflowTitlesProps } from "./"
+import {
+  HeaderRoot,
+  Branding,
+  Titles,
+  Actions,
+  WorkflowTitlesProps,
+  WorkflowActionsProps,
+} from "./components"
```
mcwinter07, line 9: We'll need to do this for the future release Workflow too 😉

packages/components/src/Button/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./Button"
+export * from "./components/GenericButton"
```
mcwinter07, line 2: Does this mean we're now exporting the GenericButton? Makes sense for exporting the types but I wonder if it mean peeps will start trying to consume this / or if it was possible to access this before 🤔

packages/components/src/Workflow/subcomponents/Header/Header.tsx
gyfchong: Odd, the Future one didn't complain

packages/components/src/Button/index.ts
gyfchong: Doesn't look like `GenericButton` was available before, but this approach aligns with the new development pattern where we export internals to allow for custom builds.

If consumers actually need a custom button, then they should be able to create one. Not that the GenericButton is particularly good lol

mcwinter07: True, wasn't sure if we wanted to offer it but certainly not apposed 👍

mcwinter07 (approved):

PR #3957, opened 2023-08-03
add canary prefix, push tags and clean up steps

mcwinter07:
## Why
Some minor tweaks to the canary workflow just to clean up the build logs, prefix the release with `canary` and push the tags 

## What
- updates canary-release.yml

### DIFF
--- a/.github/workflows/canary-release.yml
+++ b/.github/workflows/canary-release.yml
@@ -11,8 +11,6 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    outputs:
-      published: ${{ steps.changesets.outputs.published }}
     env:
       GITHUB_TOKEN: ${{ github.token }}
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/canary-release.yml
```diff
@@ -4,26 +4,33 @@ on:
...
-        description: 'Pick a label for the release, to be included in the version number — e.g. "fix-button-focus" will released as "@kaizen/button@1.0.0-fix-button-focus-20230719002814"'
+        description: 'Pick a label for the release, to be included in the version number — e.g. "fix-button-focus" will released as "@kaizen/button@0.0.0-canary-fix-button-focus-20230719002814"'
         required: true
         type: string
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
+      published: ${{ steps.changesets.outputs.published }}
```
gyfchong, line 15: This won't work because `changesets` isn't a step in this job. We'll try use the pushed tags in a different job to launch the notification

mcwinter07: Right, I suppose this isn't being used right now so it can be removed

gyfchong (approved):

PR #3958, opened 2023-08-03
fix(AsyncSelect): add ClearIndicator to AsyncSelect when isClearable is true

BrendanTimmons:
## Why
Add the `ClearIndicator` component to `AsyncSelect` so that `isClearable` (an accepted prop) will actually render the clear button.

## What
Before: 
Screenshot 2023-08-03 at 4 48 03 pm

After:
Screenshot 2023-08-03 at 4 48 09 pm

### DIFF
--- a/.changeset/nervous-ghosts-carry.md
+++ b/.changeset/nervous-ghosts-carry.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/draft-select": patch
+---
+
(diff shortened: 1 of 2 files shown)


gyfchong:
Can you include unit test that demonstrates this working?

gyfchong:
Apologies, you'll need to move this fix to [Kaizen Legacy](https://github.com/cultureamp/kaizen-legacy)

PR #3959, opened 2023-08-04
Use discriminated union on new SVG api

dougmacknz:
## Why
Fixes https://github.com/cultureamp/kaizen-discourse/issues/38

## What
Changes the props API on KAIO icon components to better ensure accessibility:
- `role` is now a required prop, prompting engineers to consciously choose between meaningful (`role="img"`) or decorative (`role="presentation"`) on each usage. Previous default was `role="img"`.
- The `title` prop has been renamed to `aria-label` so that there's less confusion with the native `title` HTML attribute, and to be in line with how this is implemented.
- `aria-label` is now required when the icon is meaningful (`role="img"`). Previously this would result in a console warning.
- Removes the ability to provide a `desc`. We can't see any reasons why we'd need to provide a description for an icon, and browser/screen reader support may be patchy anyway. If there's anything useful in a description (I didn't see any cases), they can just go into the aria-label.


packages/components/src/SVG/SVG.tsx
```diff
@@ -1,79 +1,65 @@
...
-  | "presentation" // decorative, should be silent to users who can't see it
-
-export interface SVGProps extends OverrideClassName<SVGAttributes<SVGElement>> {
-  children: ReactNode
+export interface BaseSVGProps
+  extends OverrideClassName<Omit<SVGAttributes<SVGElement>, "role">> {
   inheritSize?: boolean
-  role?: RolesType
-  title?: string
+  children: ReactNode
```
HeartSquared, line 9: You need to remove this :)

.changeset/spicy-schools-sparkle.md
```diff
@@ -2,4 +2,8 @@
 "@kaizen/components": major
 ---
 
-Changes the icon API to require the role prop, then use a discriminated union to require a label when role="img"
+Changes the props API on KAIO icon components to better ensure accessibility:
+- `role` is now a required prop, prompting engineers to consciously choose between meaningful (`role="img"`) or decorative (`role="presentation"`) on each usage. Previous default was `role="img"`.
+- The `title` prop has been renamed to `aria-label` so that there's less confusion with the native `title` HTML attribute, and to be in line with how this is implemented.
+- `aria-label` is now required when the icon is meaningful (`role="img"`). Previously this would result in a console warning.
+- Removes the ability to provide a `desc`. We can't see any reasons why we'd need to provide a description for an icon, and browser/screen reader support may be patchy anyway. If you run into any usages of `desc`, check if there's anything useful in there that's not already in the `aria-label`. If so, just copy it into the `aria-label`.
```
gyfchong, line 9: Because technically this is a new component in a new package, I would actually write this as if it's a new component, so consider this a feature not a major breaking change.

dougmacknz: Alright, you still want a major bump here though right?

gyfchong: Nah, `minor` -- we're already making consumers change their code when they migrate to KAIO, so having to update the API at the same time is reasonable.

gyfchong (approved):

packages/components/src/FilterButton/FilterButtonRemovable/FilterButtonRemovable.tsx
```diff
@@ -39,7 +39,7 @@ export const FilterButtonRemovable = forwardRef<
       <FilterButton ref={ref} {...triggerButtonProps} />
       <Tooltip text={removeButtonLabel} display="inline-block" position="below">
         <FilterButtonBase ref={removeButtonRef} {...removeButtonProps}>
-          <ClearIcon title={removeButtonLabel} />
+          <ClearIcon role="img" aria-label={removeButtonLabel} />
         </FilterButtonBase>
```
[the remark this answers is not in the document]
dougmacknz: Either will work fine, matter of preference I guess

packages/components/src/SVG/SVG.tsx
```diff
@@ -1,79 +1,60 @@
...
 }
 
 /**
- * {@link https://cultureamp.design/components/icon/ Guidance} |
- * {@link https://cultureamp.design/storybook/?path=/docs/components-icon--meaningful-kaizen-site-demo Storybook}
+ * {@link https://cultureamp.design/?path=/docs/components-icon--meaningful-kaizen-site-demo Storybook}
  */
 export const SVG = ({
   inheritSize = false,
   role = "img",
```
HeartSquared, line 30:
```suggestion
  role,
```

packages/components/src/SVG/SVG.tsx
```diff
@@ -1,79 +1,60 @@
...
-  | "img" // meaningful, title should be read aloud to users who can't see it
-  | "presentation" // decorative, should be silent to users who can't see it
-
-export type SVGProps = {
-  children: ReactNode
+type BaseSVGProps = {
   inheritSize?: boolean
-  role?: RolesType
-  title?: string
-  desc?: string
```
HeartSquared, line 16: So we don't want these anymore right? Probably worth mentioning in the changeset that they are removed.

packages/components/src/SVG/SVG.tsx
```diff
@@ -1,79 +1,60 @@
...
 } & OverrideClassName<SVGAttributes<SVGElement>>
 
+type MeaningfulIcon = {
+  role: "img"
+  "aria-label": string
+} & OverrideClassName<SVGAttributes<SVGElement>>
```
HeartSquared, line 13: Only goes on the Base :)

```suggestion
}
```

dougmacknz: Oops, screwed it up on the rebase

packages/components/src/SVG/SVG.tsx
dougmacknz: I did originally but G wanted to me to treat this like a new component rather than a breaking change for the changeset https://github.com/cultureamp/kaizen-design-system/pull/3959#discussion_r1306970644

HeartSquared: Oh. Okay sure.

HeartSquared (approved):

PR #3970, opened 2023-08-08
ListBoxSection header acessible title duplication fix

mcwinter07:
## Why
With the addition of `sectionHeader` to the `ListBoxSection` teams found that there was a double up on the accessible title if the `sectionName` was the same as the `sectionHeader`. This was because the `sectionName` was required and be passed to a `VisuallyHidden` component to create an accessible title. While this solved the issue of using inline notifications, if the user wanted to just have a visual header to describe the group this then caused a double up in the description.

Now we have updated the prop for `sectionName` to be optional if a `sectionHeader` is provided and used a discriminated union to provide some type safety.

The 3 scenarios this will cover now is:
- The `ListBoxSection` has no `sectionHeader` but has an accessible title derived from the `sectionName` (pass to an aria label)
- The `ListBoxSection` has no `sectionName` but has a `sectionHeader` that provides its accessible title
- The `ListBoxSection` has `sectionName` and a `sectionHeader`, which combines the two text nodes to give it an accessible title
  - ie:  a hidden sectionName with a inline notification that provided additional context
  

## What
- updates sectionName to be optional if sectionHeader is provided
  - This will solve the issue of sectionName and sectionHeader being read twice when they are the same
- Minor style change to ensure hide bullet lists as filtering 
- Minor style changes to allow for default text styles for section headings with just text
- add conditional check to render the sectionName only if provided
- add tests to validate type accessible names are constructed as expected


packages/select/src/FilterMultiSelect/components/ListBoxSection/ListBoxSection.tsx
```diff
@@ -3,41 +3,57 @@ import { v4 } from "uuid"
...
+  /**
+   * Becomes an aria-label on the section, informing
+   * unsighted users
+   */
+  sectionName?: string
+  /**
+   * Can be used for a visual title of the ListBoxSection or to provide addition information in a React node.
+   * If this is the same title as sectionName, you should only pass in a sectionHeader to avoid duplicate descriptions.
+   */
+  sectionHeader: ReactNode
```
gyfchong, line 25: Do we really want to have `ReactNode` instead of `string`?

mcwinter07: It's so they can provide a InlineNotification

gyfchong (approved):

mcwinter07:
Hold on merging, may have found an issue

mcwinter07:
I've fixed the stories to showcase how to use the `sectionHeader` and fix the no search results showing empty lists

gyfchong (approved):

mcwinter07:
Closing this as the changes was split into legacy and superseded by this [PR](https://github.com/cultureamp/kaizen-design-system/pull/3996)

PR #3971, opened 2023-08-09
ci(release): add slack notification

gyfchong:
This should _technically_ work... Only doing a proper release can tell.


.github/workflows/changeset.yaml
```diff
@@ -29,27 +29,23 @@ jobs:
...
+        uses: slackapi/slack-github-action@v1.24.0
         with:
-          url: ${{ secrets.SLACK_WEBHOOK_URL }}
-          channel: "#wol_kaizen"
-          username: "Release"
-          text: Complete
-          iconEmoji: ":squid:"
-      - name: Notify Slack (failure)
+          channel-id: "C02NUQ27G56"
+          slack-message: "Packages have been published"
```
HeartSquared, line 41: Any possibility of a ternary for this message? Or a pre-step which sets it as a variable?

HeartSquared (approved):

.github/workflows/changeset.yaml
gyfchong: Oh I'm not sure there's conditional logic in here, just strings. Their examples are more geared to updating an existing message with the status rather than adding conditional logic.

PR #3972, opened 2023-08-09
fix: remove animation from InlineNotification when persistent

kaydbwithers:
## Why
This PR removes the animation from InlineNotificaton if persistent is enabled.

## What
The Survey Reporting team has a use case for an InlineNotification inside a select. When the Select Popover opens, the InlineNotification animates down and does not look very good.
[KDS-1672](https://cultureamp.atlassian.net/browse/KDS-1672).

image

[KDS-1672]: https://cultureamp.atlassian.net/browse/KDS-1672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gyfchong:
@kaydbwithers feel free to remove the failing snapshots 😄

packages/notification/src/__snapshots__/InlineNotification.spec.tsx.snap
```diff
@@ -55,7 +55,7 @@ exports[`<InlineNotification /> renders a basic notification correctly 1`] = `
 
 exports[`<InlineNotification /> renders a persistent notifications 1`] = `
```
gyfchong, line 56: We'd prefer if you remove the snapshot test rather than fix them, as we've moved away from this pattern

gyfchong (approved):

PR #3988, opened 2023-08-15
Delete-all-the-things

gyfchong:
## Why

Legacy begone

## What

Deleted.


.changeset/young-owls-wash.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
gyfchong, line 2: Note Empty changeset

PR #3990, opened 2023-08-17
notification: Upgrade to React 18 render API

sentience:
## Why

`ToastNotificationManager` creates a new React root to render toast notifications at the top level of the document (where they can overlay all other page elements).

In React 18, the previous API for creating React roots has been deprecated, and now generates [noisy console warnings](https://cultureamp.atlassian.net/browse/SR2-656?atlOrigin=eyJpIjoiZWQ1NzFiNTFiNjRjNGY1ZGFmYjNmMWFkYzM5Yjk1MjMiLCJwIjoiaiJ9). To eliminate this console noise, projects must [upgrade to the new createRoot API](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis).

## What

Upgrades to the new `createRoot` API, breaking compatibility with React 17 and earlier.


.changeset/gold-tips-deliver.md
```diff
@@ -0,0 +1,9 @@
+---
+"@kaizen/notification": major
```
sentience, line 2: Should include @kaizen/rich-text-editor (which depends on @kaizen/notification) too, since it will also require React 18 with this change.

sentience:
Closing this PR as we do not want to release non-KAIO packages from this repo. Will await the migration of the components into KAIO and then this fix can be applied to that version of the components. See [discussion on Slack](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1692246571406569).

PR #3992, opened 2023-08-18
KDS-1579 converted yarn to pnpm

gyfchong:
## Why
We have a need, a need for speed.

## What

All of `yarn` converted to `pnpm`


README.md
```diff
@@ -68,16 +68,16 @@ It's dangerous to go alone! Take these:
...
-`yarn compile` | Run all typechecks
-`yarn lint` | Run all linters
-`yarn lint:fix` | Run all linters, fixing violations
-`yarn plop` | Add a new component/subcomponent
-`yarn test:storybook` | Run tests on all Storybook stories
-`yarn test` | Run all [Jest](https://jestjs.io/) tests
-`yarn reset` | Reinstall all dependencies
+`pnpm storybook` | Develop components locally using Storybook
+`STORIES=path/to/package pnpm storybook` | Develop just one package at a time using Storybook (builds faster!)
+`pnpm commit` | Use commitizen to help you write your conventional commits
```
HeartSquared, line 73: Lol oops this isn't correct :D
Maybe you should change the script name to just be `changeset` (or just nuke this one)

gyfchong: Can never seem to get rid of it haha

packages/components/package.json
```diff
@@ -82,6 +80,10 @@
     "@rollup/plugin-json": "^6.0.0",
     "@rollup/plugin-node-resolve": "^15.1.0",
     "@rollup/plugin-typescript": "^11.1.2",
+    "@storybook/addon-actions": "^7.3.1",
+    "@storybook/types": "^7.3.1",
+    "@testing-library/jest-dom": "^6.0.0",
+    "@types/testing-library__jest-dom": "^6.0.0",
```
HeartSquared, line 86: These look sus 👀 They should probably go in the root package.json?

packages/design-tokens/package.json
```diff
@@ -21,14 +21,14 @@
   "license": "MIT",
   "sideEffects": false,
   "scripts": {
-    "test": "cd ../../ && yarn jest ./packages/design-tokens",
-    "build": "yarn clean && yarn prepublishOnly",
-    "build:json": "yarn ts-node --files ./bin/buildCssVarTokens.ts",
+    "test": "cd ../../ && pnpm jest ./packages/design-tokens",
```
HeartSquared, line 24: Weird, why is this here? Does it not get run through the root command?

packages/components/package.json
gyfchong: Ah these were here to fix a type issue, but I just removed them and everything is fine

packages/design-tokens/package.json
gyfchong: Lol I don't know, it definitely does

packages/components/package.json
gyfchong: So this was the problem:
https://github.com/cultureamp/kaizen-design-system/actions/runs/5899311603/job/16001624423?pr=3992#step:4:3097

`pnpm` seems much stricter than `yarn` when it comes to workspaces, in that root level packages aren't "shared"

.github/actions/setup/action.yml
```diff
@@ -7,30 +7,24 @@ description: Sets this repository up to allow efficient run of various github ac
...
+        run_install: false
     # setup node
     - uses: actions/setup-node@v3
       with:
         node-version-file: .nvmrc
-        cache: "yarn"
-        registry-url: https://npm.pkg.github.com/
+        registry-url: https://npm.pkg.github.com
+        cache: "pnpm"
+        node-version: 18
```
HeartSquared, line 22: Aren't these two doing the same thing? 🤔

gyfchong: In what way? The first Line 10 installs the `pnpm` binary so that we can use the command in other scripts. Line 17+ sets up node.

gyfchong:
Storybook seems to be unable to find dependencies, possibly due to the fact that `pnpm` resolves dependencies different to how `yarn` did. We may need to encapsulate Storybook into it's own workspace before we can proceed

PR #3993, opened 2023-08-18
Add axe to storybook test CI step

dougmacknz:
## Why
Adds another layer of catching accessibility issues.

## What
* Adds a check for no a11y violations to our CI pipeline, on the storybook tests step.
* Adds some global ignores for rules we don't care about
* Adds a couple `FIXME` ignores for actual issues
* Fixes a couple issues
  * Adds role=presentation to icon component usages that didn't have it
  * Replaces 2 input usages with no label with our TextField component

See https://storybook.js.org/docs/react/writing-tests/accessibility-testing for documentation on this.


packages/components/src/ButtonGroup/_docs/ButtonGroup.stories.tsx
```diff
@@ -15,6 +16,18 @@ export default {
   component: ButtonGroup,
   argTypes: { ...classNameOverrideArgType },
   parameters: {
+    a11y: {
+      config: {
+        rules: [
+          ...globalA11yRules,
```
dougmacknz, line 22: Had to do this annoying thing because as soon as you set this config on the story level, it overrides the global one 😕

gyfchong:
Update the plop template, and I think it should be good

packages/components/src/Filter/_docs/Filter.stickersheet.stories.tsx
```diff
@@ -10,6 +10,17 @@ const IS_CHROMATIC = isChromatic()
 export default {
   title: "Components/Filter Base",
   parameters: {
+    a11y: {
+      config: {
+        rules: [
+          {
```
gyfchong, line 16: Does this also need the `globalA11yRules`?

dougmacknz: Yep

dougmacknz:
> Update the plop template, and I think it should be good

Don't think we need it there cause you only need to add it if there's a rule to ignore

packages/components/src/Button/_docs/Button.stories.tsx
```diff
@@ -165,7 +170,7 @@ export const Loading: StoryFn = () => <LoadingInput isAnimated width={13} />
 export const NativeFormButton: StoryFn = () => (
   <>
     <form className="mb-6" id="unique-form-id">
-      <input type="text" defaultValue="content" />
+      <TextField labelText="Sample text field" defaultValue="content" />
```
HeartSquared, line 173: Given this is a story for Button, I'd actually say it's better not to import the `draft-form` package because it creates more irrelevant maintenance (especially since it's not even in the same repo anymore).

Open to other perspectives though :)

storybook/test-runner.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const postRender: TestHook = async (page, context) => {
+  const { parameters } = await getStoryContext(page, context)
+
+  if (parameters?.a11y?.disable) {
+    return
+  }
+
+  await configureAxe(page, parameters?.a11y?.config ?? {})
```
HeartSquared, line 20: You could potentially put the global rules somewhere in this function and have it combine with any that you retrieve from the story context (custom rules). That way you should be able to not need to import the global ones in your individual stories where you're trying to extend rules :)

gyfchong:
> add it if there's a rule to ignore

I sense some docs needed for this condition lol

packages/components/src/Button/_docs/Button.stories.tsx
gyfchong: It's coming soon™, but also if it's a `devDep`, should be fine. Just need to replace it later.

dougmacknz: It was already in deps, I didn't have to add it.

This text field & story does seem pointless though so I'd be in favour of just removing it if others are

HeartSquared: Yeah kill it 🔥

dougmacknz: I just realised that the text field is linked up with the button, so it's not actually as bad as I thought

storybook/test-runner.ts
dougmacknz: Great idea, I've implemented that now

dougmacknz:
> I sense some docs needed for this condition lol

Keep in mind that most teams have this flow in their own repos now, so I don't think it's that different from something like needing to deal with a chromatic story failure and other potential build issues.

Happy to document it somewhere if you feel strongly though, CONTRIBUTING.md?

gyfchong:
> Happy to document it somewhere if you feel strongly though, CONTRIBUTING.md?

Your fix to the `globalA11y` settings means we don't need to document any caveats! So all good.

dougmacknz:
> Your fix to the globalA11y settings means we don't need to document any caveats! So all good.

Yew, are we all happy here then? @gyfchong @HeartSquared

gyfchong (approved):

HeartSquared (approved):

PR #3997, opened 2023-08-21
Kds 1669 create a script which converts raw sv gs into react components

JakePitman:
## Why
- We have a lot of raw svgs, and we expect more to be added in future. Converting these into react components manually would be time consuming, prone to human error, and lead to inconsistencies between icons

## What
- Sets up an svgo config with plugins to:
  - Strip out the original `<svg>` tag
  - Replace html attrs with react equivalents (fill-rule -> fillRule etc.)
  - Replace black colors with `currentColor`
  - Strip out anything unnecessary, like comments or unused groups
- A node script to:
  - Inject the cleaned-up svg data into a React Icon component template (one for each svg)
  - Inject unique ids for svg data that use `href` and `<use>`
  - Create an index.ts script to export all Icon components
- A shell script that runs svgo, then the node script, then cleans the final files with prettier

This PR also adds the new Icon components themselves, a stickersheet that loops over the exported Icon components (so it stays up to date), and contributing docs.


.changeset/fuzzy-emus-rush.md
```diff
@@ -0,0 +1,12 @@
+---
+"@kaizen/components": major
```
gyfchong, line 2: I think this is a `minor` we're introducing new Icon components and the original `<Icon />` component isn't being updated. (or rather non existent here)

.changeset/fuzzy-emus-rush.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Export Icon components.
+
+Exports the react-icon-component version of each of our svg files.
+For each of our existing raw svg files, users can now import them as a React component from "@kaizen/components".
+
+BREAKING CHANGE: Raw svg file location has moved from `@kaizen/components/icons/<SVG_FILE>` to `@kaizen/components/assets/svgs/icons/<SVG_FILE>` 
+BREAKING CHANGE: The SVG component has moved from `@kaizen/components/src/SVG/SVG.tsx` to `@kaizen/components/src/Icons/subComponents/SVG/SVG.tsx`
+BREAKING CHANGE: Icon import paths have changed from `import {<ICON_NAME>} from @kaizen/components/SVG/icons/<ICON_NAME>` to `import {<ICON_NAME>} @kaizen/components`
```
gyfchong, line 12: Generally I don't think folder changes are considered a breaking change, as they don't modify an API. Also, try write this as if you're adding a new feature, since the original Icon component isn't changing here.

packages/components/src/Icons/AcademyIcon.tsx
```diff
@@ -0,0 +1,19 @@
+// This file is autogenerated by wrapSVGs.ts
+// Changes to this file will be overwritten
+
+import React from "react"
+import { v4 as uuidv4 } from "uuid"
+import { SVG, SVGProps } from "~components/Icons/subComponents/SVG"
+
+const uniqueId = uuidv4()
```
gyfchong, line 8: I think this is the cause for your a11y errors since this might not actually run every time the icon is rendered, try placing this inside the component's function.

PR #3999, opened 2023-08-22
KDS-1717 Fix notify slack on release

gyfchong:
Trying something I saw in another example.


mcwinter07 (approved):
Would assume we wouldn't need the `${{ }}` to evaluate the truthiness based on the [changeset release action example](https://github.com/changesets/action) but sounds like a good idea to try if that wasn't working 🤞

PR #4003, opened 2023-08-22
Fix-published-flag

gyfchong:
## Why

So that we only get notified when changesets starts to publish.

## What

Using the outputs of changesets action we detect if there's a changeset, if there isn't then we can assume that this is the publishing action, and if the published status is`true`, then we are successful.

### DIFF
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -9,7 +9,9 @@ jobs:
   version:
     runs-on: ubuntu-latest
     outputs:
-      published: ${{ steps.changesets.outputs.published }}
+      # Determine if there's a changeset present to infer that a publish is happening
+      # since changesets itself could either be creating a PR or publishing.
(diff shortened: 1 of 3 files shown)


.github/workflows/changeset.yaml
```diff
@@ -26,16 +28,22 @@ jobs:
...
           publish: yarn ci:publish
+  notify-slack:
+    runs-on: ubuntu-latest
+    needs: version
+    env:
+      PUBLISHED: ${{ needs.version.outputs.hasPublished }}
+    steps:
       - name: Publish successful
-        if: steps.changesets.outputs.published == 'true'
+        if: ${{ env.PUBLISHED }}
```
HeartSquared, line 38:
```suggestion
        if: ${{ env.PUBLISHED == 'true' }}
```
In case it's some strange truthy value?

gyfchong: Yeah fair

gyfchong: Weird Github is weird

HeartSquared (approved):

PR #4005, opened 2023-08-22
Fix-published-flag

gyfchong:
## Why
Boolean logic hard

## What
Fix boolean logic


.github/workflows/changeset.yaml
```diff
@@ -31,22 +32,15 @@ jobs:
...
-          slack-message: "Packages have been published"
-        env:
-          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
-      - name: Publish failed
-        if: ${{ env.PUBLISHED == 'false' }}
-        uses: slackapi/slack-github-action@v1.24.0
-        with:
-          channel-id: "C02NUQ27G56"
-          slack-message: "Publish failed"
+          slack-message: ${{ env.PUBLISHED == 'true' && 'Packages have been published' || 'Publishing failed' }}
```
HeartSquared, line 44: This is interesting... Is this the equivalent of a ternary? 😂

gyfchong: Yes! Haha it's super weird

.github/workflows/changeset.yaml
```diff
@@ -30,23 +29,17 @@ jobs:
           publish: yarn ci:publish
   notify-slack:
     runs-on: ubuntu-latest
-    needs: version
+    needs:
+      version
+      # We only trigger this flow if the publish is happening, inferred by the presence of a changeset.
```
HeartSquared, line 34:
```suggestion
      # We only trigger this flow if the publish is happening, inferred by not having a changeset.
```

.changeset/stupid-pumpkins-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Tester publish
```
HeartSquared, line 5: Do you still want to keep this?

gyfchong: Yeah, still want to see if the ternary works

HeartSquared (approved):

PR #4009, opened 2023-08-28
refactor: clean up KAIO

HeartSquared:
## Why

Housekeeping.

## What

- Remove legacy templates from docs and plop templates
- Update plop templates to have working stories in new MDX format
- Update stickersheets to have RTL stories
- Update textDirection story decorator to use `parameters` instead of `args`
	- Follows pattern of other plugins and prevents confusion with props
- Rename directories which did not conform to our agreed conventions (eg. `components` instead of `subcomponents`)
- Enable TS for `docs` folder and fix the errors
- Fix `BrandMoment` stories
- Update axe testing config to allow custom timeout for stories


docs/Systems/LayoutAndSpacing/DemoComponents/SpacingTokens.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react"
-import { kzSpacing } from "@kaizen/button"
+import { kzSpacing } from "@kaizen/tailwind/src/kz-spacing"
```
gyfchong, line 2: Weird that you have to dig into the `src`?

docs/basic-component.md
```diff
@@ -4,20 +4,24 @@ Recipe for creating a basic component.
...
+  - [Types](#types)
 - [Component.tsx template](#componenttsx-template)
   - [Props](#props)
   - [The component](#the-component)
   - [Display name](#display-name)
 
 ## Intro
 
-To generate a new component and package, new component within an existing package, or a subcomponent,
+All components are to be added to `packages/components`.
```
gyfchong, line 22: Would be good to include a definition of the `future` components here since the plop prompts mention it

docs/pages/app-starter.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta, Unstyled } from '@storybook/blocks'
 import { InlineNotification } from "@kaizen/notification"
-import { LinkTo } from "../../storybook/components/LinkTo"
+import { LinkTo } from "~storybook/components/LinkTo"
```
gyfchong, line 3: Ooo we can use the alias here now, game changer!

docs/Systems/LayoutAndSpacing/DemoComponents/SpacingTokens.tsx
HeartSquared: As discussed, made a new ticket to pull it from `design-tokens` instead.

docs/basic-component.md
HeartSquared: I was afraid you'd ask 😛 Care to help me write something?

gyfchong:
```suggestion
There are 2 types of components that live in `@kaizen/components`, `src` and `future`, both are **production safe**

### src components
These are components which are essentially the "main" branch of components and only receive non-breaking changes. **Any brand new components should go here.**

### future components
Components which have an existing `src` component equivalent, but have any breaking changes to their API. Eventually `future` components will replace their `src` equivalent in the next major release.
```

.changeset/slimy-icons-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Rename directories to match convention (clean up only).
```
gyfchong, line 5: Do we need to release?

HeartSquared: I was thinking it's the equivalent to any other refactor, since I renamed some directories. Just in case something breaks.
I can get rid of it though. Not too fussed.

gyfchong: Hm doesn't hurt I guess

packages/components/src/BrandMoment/_docs/BrandMoment.stories.tsx
```diff
@@ -1,102 +1,132 @@
...
-import arrowLeftIcon from "../../../icons/arrow-left.icon.svg"
-import arrowRightIcon from "../../../icons/arrow-right.icon.svg"
-import mailIcon from "../../../icons/email.icon.svg"
-import feedbackClassifyIcon from "../../../icons/feedback-classify.icon.svg"
-import securityTipIcon from "../../../icons/security-tip.icon.svg"
+// import { ArrowLeftIcon } from "~icons/ArrowLeftIcon"
+import { ArrowRightIcon } from "~icons/ArrowRightIcon"
+import { EmailIcon } from "~icons/EmailIcon"
+// import feedbackClassifyIcon from "../../../icons/feedback-classify.icon.svg"
+// import securityTipIcon from "../../../icons/security-tip.icon.svg"
```
HeartSquared, line 16: These can be fixed in a separate PR after the icons have been added, or if the PR adding the icons is merged first, I'll fix it here.

packages/components/src/ErrorPage/ErrorPage.tsx
```diff
@@ -66,7 +75,7 @@ export function ErrorPage({
             </Paragraph>
           </>
         }
-        illustration={<BrandMomentError isAnimated loop />}
```
gyfchong, line 69: I think we can make a call as to whether or not `ErrorPage` needs to animate at all. Given their relatively low exposure, it seems unnecessary to provide an API for this.

.changeset/slimy-icons-grin.md
```diff
@@ -0,0 +1,8 @@
+---
+"@kaizen/components": patch
+---
+
+Clean up @kaizen/components.
+
+- Add `isAnimated` prop to `ErrorPage` for Chromatic compatibility
+- Rename directories to match convention
```
gyfchong, line 8: The build would fail if you messed any of the naming/moving up. So I'm comfortable with killing the changeset

packages/components/src/ErrorPage/ErrorPage.tsx
HeartSquared: It's actually animated by default 😅 But the animation doesn't show in Chromatic, so I wanted a way to stop it.

.changeset/slimy-icons-grin.md
HeartSquared: Well the changeset is needed for the other prop now :P Or maybe still no since it only exists for our testing? 🤔

packages/components/src/ErrorPage/ErrorPage.tsx
gyfchong: But what I'm saying is set this to `isAnimated={false}`, since I don't think we need an option to animate it in the first place.

gyfchong (approved):

packages/components/src/BrandMoment/_docs/BrandMoment.stories.tsx
```diff
@@ -1,102 +1,132 @@
...
+const IconRTLTemplate: Pick<Story, "render" | "parameters"> = {
+  render: ({ primaryAction, ...args }, { globals }) => (
+    <BrandMoment
+      {...args}
+      primaryAction={{
+        label: "Primary action label",
+        ...primaryAction,
+        icon:
+          globals.textDirection === "ltr" ? (
+            <ArrowRightIcon />
```
dougmacknz, line 53: Naughty 😛 

Interesting that this made it past axe, maybe because it's put inside a button

PR #4010, opened 2023-08-30
refactor: use type instead of interface for props

HeartSquared:
## What

Update props to use `type` instead of `interface`.

## Why

While `type` and `interface` can both be used for props, given the shape of the props should never change, `type` is more suitable as it is definitive.
Both can be extended to create new types/interfaces.

The main differences:
- `interface` allows declaration merging, which means that if you declare another `interface` of the same name, TypeScript will merge the properties together
- `type` is definitive - if there are any `type`s in the same file with the same name, TypeScript will throw an error


JakePitman (approved):
Nice

PR #4013, opened 2023-08-31
Add script to convert raw svgs to React Icon components.

JakePitman:
## Why
- We have a lot of raw svgs, and we expect more to be added in future. Converting these into react components manually would be time consuming, prone to human error, and lead to inconsistencies between icons

## What
- Sets up an svgo config with plugins to:
  - Strip out the original `<svg>` tag
  - Replace html attrs with react equivalents (fill-rule -> fillRule etc.)
  - Replace black colors with `currentColor`
  - Strip out anything unnecessary, like comments or unused groups
- A node script to:
  - Inject the cleaned-up svg data into a React Icon component template (one for each svg)
  - Inject unique ids for svg data that use `href` and `<use>`
  - Create an index.ts script to export all Icon components
- A shell script that runs svgo, then the node script, then cleans the final files with prettier

This PR also adds the new Icon components themselves, a stickersheet that loops over the exported Icon components (so it stays up to date), and contributing docs.


CONTRIBUTING.md
```diff
@@ -200,3 +199,16 @@ run the following command and follow the prompts:
 ```
 yarn plop
 ```
+
+## Adding icons
+
+1. Find the new svg in our [figma file](https://www.figma.com/file/eZKEE5kXbEMY3lx84oz8iN/%F0%9F%92%9C-UI-Kit%3A-Heart?type=design&node-id=35882-93662&mode=design&t=X18YN2TWqlxEIFc3-0)
+2. Export the svg in kebab-case (eg. `zoom-out.svg`), and save it in `/packages/components/icons`
+3. In the root of Kaizen, run `yarn pkg:aio update-icons`
```
gyfchong, line 207: Can we make this command simpler? `yarn update-icons` and have the underlying script do the routing to the workspace.

packages/components/src/BrandMoment/_docs/BrandMoment.stories.tsx
```diff
@@ -52,7 +54,7 @@ const IconRTLTemplate: Pick<Story, "render" | "parameters"> = {
           globals.textDirection === "ltr" ? (
             <ArrowRightIcon role="presentation" />
           ) : (
-            <div>Boo!</div>
+            <ArrowLeftIcon role="presentation" />
```
HeartSquared, line 57: 👻 ❤️

packages/components/src/BrandMoment/_docs/ExampleHeaders.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react"
 import { Button } from "@kaizen/button"
 import { Select } from "@kaizen/draft-select"
 import { assetUrl } from "@kaizen/hosted-assets"
-import closeIcon from "../../../icons/close.icon.svg"
+import closeIcon from "../../../assets/svgs/icons/close.icon.svg"
```
HeartSquared, line 5: Maybe update the usage of `Button` and then use your new icon instead :)

packages/components/src/ClearButton/ClearButton.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { ButtonHTMLAttributes } from "react"
 import classnames from "classnames"
-import { ClearIcon } from "~components/SVG/icons/ClearIcon"
+import { ClearIcon } from "~icons/ClearIcon"
```
HeartSquared, line 3: So is our convention to use `~components/Icons` or `~icons/*Icon`? I think it's better to just stick to one.

packages/components/src/Icons/bin/wrapSVGUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+  return split
+    .map((word: string) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join("")
+    .replace(".icon", "Icon")
+}
+
+// eslint-disable-next-line no-template-curly-in-string
+const hrefReplacement = "href={`#${uniqueId}`}"
+
+// @ts-ignore: Redeclared module error
```
HeartSquared, line 13: Would it work if you don't use `module.exports` but instead export the function directly?
If it definitely doesn't, use `@ts-expect-error` instead :)

packages/components/src/Icons/bin/wrapSVGUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+      .replace('"UNIQUE_ID"', "{uniqueId}")
+      .replace('href="#UNIQUE_ID"', hrefReplacement)
+  } else {
+    // Some raw svgs do not use the use + href pattern.
+    // This removes unused uuids from the template to prevent lint errors.
+    return completedTemplate
+      .replace("const uniqueId = uuidv4()", "")
+      .replace('import { v4 as uuidv4 } from "uuid"', "")
+  }
+}
```
HeartSquared, line 34: Since you already have a return within your `if`, no need for an `else` :)
```suggestion
  }
  
    // Some raw svgs do not use the use + href pattern.
    // This removes unused uuids from the template to prevent lint errors.
    return completedTemplate
      .replace("const uniqueId = uuidv4()", "")
      .replace('import { v4 as uuidv4 } from "uuid"', "")
}
```

packages/components/src/Icons/bin/wrapSVGs.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+
+  it("replaces SVG_CONTENT with the contents of the svg", () => {
+    expect(result.includes("SVG_CONTENT")).toBe(false)
+    expect(result.includes(svgContent)).toBe(true)
+  })
+  it("replaces SVG_CONTENT with the contents of the svg", () => {
+    expect(result.includes("SVG_CONTENT")).toBe(false)
+    expect(result.includes(svgContent)).toBe(true)
+  })
```
HeartSquared, line 42: I think a better test would be to have a single expect that has the whole formed component.
If possible, read from the template file as well, so we can ensure this is an accurate test.
Something like this:

```
const reactTemplate = templateStringFromFile

it("correctly generates component from template", () => {
  const result = insertSvgData(reactTemplate, "MyIcon", "<use></use>")
  const expected = `
    import React from "react"
    import { SVG, SVGProps } from "~components/Icons/subComponents/SVG"
    export const MyIcon = (props: Omit<SVGProps, "children">): JSX.Element => (
      <SVG {...props}><use></use></SVG>
    )
  `
  expect(result).toBe(expected)
})
```

packages/components/src/index.ts
```diff
@@ -14,3 +14,8 @@ export * from "./FilterSelect"
 export * from "./InputSearch"
 export * from "./KaizenProvider"
 export * from "./Workflow"
+export * from "./Icons"
+
+// TODO: Remove when Date Picker is migrated to KAIO
+export * from "./types/DatePicker"
+export * from "./ErrorPage"
```
HeartSquared, line 21: These have been moved up in the file, so delete these dupes.

.gitignore
```diff
@@ -7,3 +7,5 @@ dist
 .yalc
 yalc.loc
 .turbo
+/packages/components/src/Icons/subComponents/built-svgs/
+!/packages/components/svgoUtils.js
```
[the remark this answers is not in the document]
JakePitman: Yeah can get rid of this. Can't remember why I needed this before 🤔

packages/components/src/Icons/bin/wrapSVGUtils.ts
JakePitman: Yeah it leads to this error sadly. Couldn't find a way around it.

```
export var svgToComponentTitle = function (fileName) {
^^^^^^

SyntaxError: Unexpected token 'export'
```

I might have to stick with `@ts-ignore` for this one, because the error only gets thrown in the tests, and not when executing the script. So `@ts-expect-error` throws an `Unused @ts-expect-error directive` error when executing the script

CONTRIBUTING.md
gyfchong:
```suggestion
3. In the root of Kaizen, run `yarn update-icons`
```

packages/components/src/ClearButton/ClearButton.tsx
gyfchong: Format still seems inconsistent across the files, which did we land on?

gyfchong (approved):

PR #4041, opened 2023-09-04
feat(multi-select): add Checkbox subcomponent

HeartSquared:
## What

Checkbox is required for MultiSelect component.
Partial component migration of Checkbox required for it to be consumed in KAIO due to icons being used within.
However, we ended up refactoring it when we came across stylistic limitations and a11y issues coupled with api confusion.

## Why

Issues found (from [Confluence page](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3267756912/3.+Checkbox+-+subcomponent)):

- The styles were tied closely to the dom structure, having the hidden input state create styles for the rendered icon. This unfortunately meant it would be extremely hard for us to target and remove the focus ring.
	- **Result:** We refactored the styles structure (including the focus ring), where we now only need to target the container. However, we did remove all styles we do not need for our use case (eg. reversed).
- The component appeared to have an uncontrolled state, but it does not work.
	- **Result:** We have made some props required so it can only be in a controlled state. An uncontrolled state does not work when you have the possibility of using indeterminate, so this should not be a problem - especially since no one has ever asked for it to be fixed.
- Native HTML attributes being aliased in props. This was likely due to old implementation, and unlikely to be intentional.
	- **Result:** We have opted to use native HTML attributes, and removed the aliased props (eg. `onCheck` is now `onChange`).
- Inaccurate usage of the input checked HTML attribute.
	- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checking_boxes_by_default
	- The `checked` attribute should only be applied on page load to declare whether it should be default checked or not, and it should not change after loading.
	- This is different to the `checked` attribute you will find when checking the element properties in JS, which shows the current checked value.
	- For a screen-reader, this attribute will override the internal JS attribute, therefore should only be used in an uncontrolled implementation.
	- **Result:** Given we have opted to only have a controlled state, we have removed the attribute.
- Unnecessary fallback for input `value` attribute.
	- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value
	- The value attribute can override the submitted form value for the checkbox if it exists, otherwise it naturally falls back onto the checked state of the checkbox.
	- **Result:** Let the fallback happen naturally by removing the prop and having it as part of the spread props.
- `readOnly` is set in a magical way (when onChange is undefined).
	- While this meets accessibility, it isn’t an intuitive experience for consumers of the component as it does not work without an onChange, and does not provide guidance that it is required.
	- **Result:** Used a discriminated union so that instead of setting readOnly for the consumer, we require them to set it if they do not want an onChange - otherwise onChange will be required.


packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useEffect, useRef, InputHTMLAttributes } from "react"
+import classnames from "classnames"
+import { CheckIcon } from "~icons/CheckIcon"
+import { MinusIcon } from "~icons/MinusIcon"
```
JakePitman, line 4:
```suggestion
import { CheckIcon, MinusIcon } from "~components/Icons"
```

JakePitman (dismissed):

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onChange?: never
+}
+
+type Controlled = {
+  readOnly?: false
+  onChange: Required<InputHTMLAttributes<HTMLInputElement>["onChange"]>
+}
+
+export type CheckboxProps = {
+  checkedStatus: CheckedStatus
```
dougmacknz, line 20: Do we want this to match the native attribute? As spoken about recently with icons.

If we're supporting mixed I guess it would be `aria-checked`.

I question whether we should be supporting indeterminate states though.

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  value?: string
```
dougmacknz, line 21: This doesn't need to typed explicitly by the looks

mcwinter07: Nice catch

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export type CheckedStatus = "checked" | "unchecked" | "indeterminate"
+
+type ReadOnly = {
+  readOnly: true
+  onChange?: never
+}
+
+type Controlled = {
+  readOnly?: false
```
dougmacknz, line 15: This is feeling quite influenced by what the old one was doing. Do you think you would do it this way if you were looking at creating a fresh Checkbox component?

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
mcwinter07: We wanted to avoid `aria-checked` since we're leveraging as much of the native `input` as possible, so felt a little like a hat on a hat. We did validate that setting indeterminate in ln55 does get read to the screen reader as expected if that helps for context
Screenshot 2023-09-04 at 3 05 55 pm

gyfchong:
Can we do this folder structure? I don't like how the docs are removed from the subcomponent itself
```
MultiSelect/
├─ subcomponents/
│  ├─ Checkbox/
│  │  ├─ _docs/
│  │  │  ├─ Checkbox.stories.tsx
│  │  ├─ Checkbox.tsx
├─ MultiSelect.tsx
```

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
mcwinter07: In terms of feature parity for the current checkbox and how its used in the platform, I think we do have indeterminate use cases (ie: the goals UI example recently shared). But worth asking our design fellows whether this is something we want to cater for

dougmacknz: Yeah I kinda just wonder if it's one of those things like `<select multiple>` where yes it's (kinda) supported by HTML but how many people actually know what it means when there's a minus in a checkbox?

Plus slight concern for support with different screen reader + browser combos. Bit out of date and doesn't cover JAWS but https://github.com/mui/material-ui/issues/20476#issuecomment-611468844

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
HeartSquared: It was also so that for our stories we wouldn't need to write `onChange={() => undefined` constantly, given we want it to be required.
Doing this fresh, we would probably not have considered `readOnly` until we had designs for it, but knowing about it just made it a little nicer for the consuming api.

packages/components/src/MultiSelect/subcomponents/Checkbox/Checkbox.tsx
HeartSquared: `indeterminate` has been around for a long time, and I think it might have more visual association than you'd guess.

Nonetheless, you're probably right about screen reader + browser combos being a concern, but perhaps that's something that the product teams can alleviate with more aria attributes when they're actually building something? 🤔 

It's not something we need immediately, but I think it's commoncase enough that we can't just completely ignore it unless there is a recommended better solution for it.

gyfchong: There might be use cases in auth as well. Regardless, this component will go through a bit of a review before being released as a `future` component to replace existing `Checkbox` so I think it's okay for this API to be somewhat beta/focussed on solving `MultiSelect` specifically.

mcwinter07 (approved):
🦖 🐜 !

PR #4043, opened 2023-09-06
KDS-1745-create-staging-docs-for-kaio

gyfchong:
## Why
We want to hide the docs for components being migrated to KAIO before releasing

## What
Script to inject styles that hide all docs placed in the KAIO-staging/ path from production environment.


storybook/manager-head.html
```diff
@@ -15,8 +15,22 @@
   #storybook-explorer-searchfield {
     background-color: white;
   }
+</style>
+
+<script>
+  const storybookDocPath = "kaio-staging"
+  const production = new URLPattern("http{s}?://cultureamp.design");
```
mcwinter07, line 22: Might need to do a regex here since the [URLPattern API](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) doesn't seem to have great support in firefox and safari

gyfchong: Nice one thanks!

mcwinter07: No worries, I mainly found out cause I had no idea what it was 😆 Looks very handy though! Will be cool when it gets more support

storybook/manager-head.html
```diff
@@ -15,8 +15,22 @@
...
+  currentUrl = window.location.href;
 
-  data-title["Legacy"] {
-    background-color: red;
+  const shouldHideKAIOStories = production.test(currentUrl);
```
HeartSquared, line 25: Shouldn't we just check the env variable? 🤔

gyfchong: Hm, I didn't think I could access that here

HeartSquared: https://storybook.js.org/docs/react/configure/environment-variables :)

storybook/manager-head.html
HeartSquared: +1 to Winter upskilling without asking 😄

storybook/manager-head.html
gyfchong: Hm, it's undefined here. My guess is that Storybook doesn't actually bundle this code, it just uses it as a template.

HeartSquared (approved):
Let's try :)

PR #4046, opened 2023-09-07
Migrate typography components to KAIO

dougmacknz:
Migrate the Heading and Paragraph components from kaizen-legacy to KAIO.

Changes from the kaizen-legacy version:
* Paragraph component renamed to Text
* The sass files are no longer loading the `fonts.scss` file. The fonts are being loaded by ThemeProvider so I don't think we need this anymore.
* Documentation rewritten to be more succinct and not repeat the info from the props table.


.changeset/forty-spiders-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add Heading and Paragraph components from kaizen-legacy
```
gyfchong, line 5: This can be empty, we aren't planning on publishing these until later

dougmacknz: Ah, sorted

gyfchong (approved):

PR #4048, opened 2023-09-08
Remove Storybook body transition

dougmacknz:
## Why
This transition causes false positive a11y issues for colour contrast.

## What
Remove the built in background transition on Storybook.


gyfchong (approved):
Genius

PR #4060, opened 2023-09-12
KDS-1777 Lock esbuild version

gyfchong:
## Why
Upgrading to `0.19.x` causes the plugin `rollup-plugin-esbuild` to error when imported into the rollup config.

## What
Lock the version to the last known working version `0.18.20` for now as the solution is not clear.

We also may not be sticking with Rollup in the mid-long term.


.changeset/tasty-oranges-fix.md
```diff
@@ -0,0 +1,2 @@
+---
```
dougmacknz, line 1: Got 2 changesets?

dougmacknz (approved):

PR #4065, opened 2023-09-13
Housekeeping: remove data-automation-id

gyfchong:
## Why
No longer support `data-automation-id`

## What
Remove tests etc. that mention it.


packages/components/src/Heading/Heading.spec.tsx
```diff
@@ -42,14 +42,12 @@ describe("<Heading />", () => {
...
-      <Heading variant="display-0" data-automation-id="test-id">
+    const { getByTestId } = render(
+      <Heading variant="display-0" data-testid="test-id">
         Example
       </Heading>
     )
-    expect(
-      container.querySelector('[data-automation-id="test-id"]')
-    ).not.toBeNull()
+    expect(getByTestId("test-id")).not.toBeNull()
```
HeartSquared, line 50:
```suggestion
    expect(getByTestId("test-id")).toBeInTheDocument()
```

packages/components/src/Text/Text.spec.tsx
```diff
@@ -70,14 +70,12 @@ describe("<Text />", () => {
...
-      <Text variant="intro-lede" data-automation-id="test-id">
+    const { getByTestId } = render(
+      <Text variant="intro-lede" data-testid="test-id">
         Example
       </Text>
     )
-    expect(
-      container.querySelector('[data-automation-id="test-id"]')
-    ).not.toBeNull()
+    expect(getByTestId("test-id")).not.toBeNull()
```
HeartSquared, line 78:
```suggestion
    expect(getByTestId("test-id")).toBeInTheDocument()
```

packages/design-tokens/docs/DocsComponents.tsx
```diff
@@ -70,7 +70,6 @@ const TabbedCodeBlocks = ({
               // eslint-disable-next-line jsx-a11y/anchor-is-valid
               <a
                 style={{ flexShrink: 0 }}
-                data-automation-id={tab.automationId}
```
HeartSquared, line 73: Do we need to remove the `automationId` prop then? Or put it into a `data-testid`?

gyfchong: Definitely probably don't need this

HeartSquared (approved):

PR #4088, opened 2023-09-18
Kds 1779 housekeeping tag component

JakePitman:
## Why
We want to make it easier for people outside KZ to make contributions. One way we do this is by allowing them to contribute in TW if they prefer, and we convert to scss ourselves. We'll also handle stuff like formatting docs in line with our standards.

## What
Housekeeping for [this PR](https://github.com/cultureamp/kaizen-design-system/pull/3851)
- Converts TW to SCSS
- Converts the docs to our format


packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -26,33 +27,20 @@ export const Tag = ({
...
-        ["bg-orange-100 text-orange-700"]: color === "orange",
-        ["bg-red-100 text-red-700"]: color === "red",
-        ["bg-purple-100 text-purple-700"]: color === "purple",
+        [styles.gray]: color === "gray",
+        [styles.blue]: color === "blue",
+        [styles.green]: color === "green",
+        [styles.yellow]: color === "yellow",
+        [styles.orange]: color === "orange",
+        [styles.red]: color === "red",
+        [styles.purple]: color === "purple",
```
gyfchong, line 38: I think you can do:
```
classNames(styles[color])
```

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,66 @@
...
+.gray {
+  background-color: $color-gray-200;
+  color: $color-purple-800;
+}
+.blue {
+  background-color: $color-blue-100;
+  color: $color-blue-700;
+  svg {
+    color: $color-blue-500;
+  }
```
gyfchong, line 26: I think this would affect the icon in the dismiss button, right? Might be best to separate the icon styles.

packages/components/src/__future__/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <Tag color="blue">Blue</Tag>
+          <Tag color="green">Green</Tag>
+          <Tag color="yellow">Yellow</Tag>
+          <Tag color="orange">Orange</Tag>
+          <Tag color="red">Red</Tag>
+          <Tag color="purple">Purple</Tag>
+        </List>
+        <List>
+          <Tag Icon={<TagIcon role="presentation" />}>Gray</Tag>
+          <Tag color="blue" Icon={<TagIcon role="presentation" />}>
```
gyfchong, line 38: Oof I didn't notice this before, but the icon prop should be lowercase. Don't worry about this being a breaking change, we haven't released it yet 😛

packages/components/src/__future__/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+type ListProps = {
+  children: React.ReactNode
+}
+const List = ({ children }: ListProps): JSX.Element => (
+  <div className="flex flex-col gap-12 items-start">{children}</div>
+)
+
+export default {
+  title: "KAIO-staging/Tag",
```
gyfchong, line 15:
```suggestion
  title: "Components/Tag",
```
We can fully release this

packages/components/src/__future__/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,33 @@
+import { Canvas, Controls, Meta } from "@storybook/blocks"
+import { ResourceLinks, KaioNotification, Installation } from "~storybook/components"
+import * as TagStories from "./Tag.stories"
+
+<Meta of={TagStories} />
+
+# Avatar
```
gyfchong, line 7:
```suggestion
# Tag
```

packages/components/src/__future__/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+<Installation
+  installCommand="yarn add @kaizen/components"
+  importStatement="import { Tag } from `@kaizen/components/future`"
+/>
+
+## Overview
+
+Tags helps users quickly recognize important information about items that organize and categorize them.
+
```
gyfchong, line 26: We'll need to add a note here that this `Tag` is not yet feature complete and consumers will need to use the Legacy `Tag` if they still need the legacy features.

The goal of this `Tag` was to first provide a flexible option compared to the Legacy one, and then we will slowly add more functionality until we can completely replace the Legacy version.

.changeset/gorgeous-rice-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
gyfchong, line 2:
```suggestion
"@kaizen/components": patch
```
Nitpick: since the `Tag` is already published as a `minor`, this would technically be a `patch` that introduces no new functionality.

packages/components/src/__future__/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+    chromatic: { disable: false },
+    controls: { disable: true },
+  },
+} satisfies Meta
+
+const StickerSheetTemplate: StoryFn = () => (
+  <StickerSheet heading="Tag">
+    <StickerSheet.Header headings={["Label Only", "Icon"]} />
+    <StickerSheet.Body>
+      <StickerSheet.Row>
```
gyfchong, line 26: Nitpick: You could iterate over an array of colours, creating a row for each colour.

```
<StickerSheet.row rowTitle={colour}>
  <Tag color={colour}> ...
  <Tag color={colour} Icon={<TagIcon role="presentation" />}> ...
</StickerSheet.row>
```

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,66 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/typography";
+
+.container {
+  display: inline-flex;
+  align-items: center;
+  padding: 5px 0.5rem 5px 0.5rem;
+  gap: 0.25rem;
```
HeartSquared, line 8: It's not a good idea to mix measurements, because they won't scale properly.

Use the spacing tokens where possible, and then use a rem equivalent of any px value if it doesn't exist as a token.

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  background-color: $color-blue-100;
+  color: $color-blue-700;
+  svg {
+    color: $color-blue-500;
+  }
+}
+.green {
+  background-color: $color-green-100;
+  color: $color-green-700;
+  svg {
```
HeartSquared, line 31: Target the icon container instead of the svg element.

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  gap: 0.25rem;
+  border-radius: 28px;
+  font-family: $typography-paragraph-intro-lede-font-family;
+  font-size: $typography-paragraph-small-font-size;
+  font-weight: $typography-paragraph-small-font-weight;
+  line-height: $typography-paragraph-small-line-height;
+  white-space: nowrap;
+}
+
+.gray {
```
HeartSquared, line 17: Does the grey variant not have a different colour for the icon?

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -26,33 +27,20 @@ export const Tag = ({
...
-          ["text-yellow-500"]: color === "yellow",
-          ["text-orange-500"]: color === "orange",
-          ["text-red-500"]: color === "red",
-          ["text-purple-500"]: color === "purple",
-        })}
-      >
-        {Icon}
-      </span>
-    )}
+    {Icon && <span className={styles.iconContainer}>{Icon}</span>}
```
HeartSquared, line 43: Suggestion: If the styles on the `iconContainer` are inconsequential, you can actually clone the passed in Icon component and use `classNameOverride` to add an `icon` class instead.

Only if you're keen to try though. The container method works fine too.

packages/components/src/__future__/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,33 @@
...
+<Canvas of={TagStories.Playground} />
+<Controls of={TagStories.Playground} />
+
+## API
+
+Coming Soon
```
HeartSquared, line 32: This is a good time to fill in the api section! Create stories for the props and give them some descriptions :)
You can take a look at the Filter(Bar) ones for inspiration - it's likely much simpler for Tag.

packages/components/src/__future__/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+  controls: { disable: true },
+}
+
+export const StickerSheetRTL = StickerSheetTemplate.bind({})
+StickerSheetRTL.storyName = "Sticker Sheet (RTL)"
+StickerSheetRTL.parameters = {
+  chromatic: { disable: false },
+  controls: { disable: true },
+  textDirection: "rtl",
+}
```
HeartSquared, line 75: Change these to the new `Story` object format (see any other stickersheet)

packages/components/src/__future__/Tag/_docs/Tag.stickersheet.stories.tsx
HeartSquared: Agreed with Geoff. Then you can get rid of that `List` component at the top of the file.

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,12 @@
...
-        "https://github.com/cultureamp/kaizen-design-system/tree/main/packages/components/src/__future__/Tag",
-      figma:
-        "https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%A7%AA-Kaizen-Labs?type=design&node-id=28588-183332&mode=design&t=VHZDCTJUpaon8PXg-0",
-    },
-  },
 } satisfies Meta<typeof Tag>
 
 export const Playground: StoryFn<typeof Tag> = args => (
   <Tag {...args}>My tag</Tag>
 )
```
HeartSquared, line 12: Use the `Story` object format. See other existing stories.

packages/components/src/__future__/Tag/_docs/Tag.mdx
gyfchong: Or `Button` 😄

packages/components/src/__future__/Tag/Tag.module.scss
JakePitman: No
https://github.com/cultureamp/kaizen-design-system/pull/3851/files#diff-35c543e702bd65369573bd34a48291a45ec1b4c5c64464f72bc5efaa63805110R43

HeartSquared: You actually just made me look at the Figma designs. Unless I'm looking at the wrong link, the text shouldn't even be changing between the Tags. Might be worth reaching out to @jmitsios to check.

https://www.figma.com/file/ZRfnoNUXbGZv4eVWLbF4Az/%F0%9F%96%BC%EF%B8%8F-Component-Gallery?type=design&node-id=9-37843&mode=design&t=zhd1HLgFFOVjz4yR-0

HeartSquared: Brought it up with the team, and this component is based on different designs
https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%A7%AA-Kaizen-Labs?node-id=28588%3A183332&mode=dev

@gyfchong doesn't want to expose the link to the Figma though, so you'll want to remove the link in the docs.

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,79 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/typography";
+
+.container {
+  display: inline-flex;
+  align-items: center;
+  padding: 0.3rem $spacing-8 0.3rem $spacing-8;
+  gap: 0.25rem;
```
HeartSquared, line 9: Spacing token available :)

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -1,58 +1,37 @@
...
+  "gray",
+  "blue",
+  "green",
+  "yellow",
+  "orange",
+  "red",
+  "purple",
+] as const
+
+export type TagColors = (typeof TagColorKeys)[number]
```
HeartSquared, line 16: To prevent confusion for our consumers, we don't want to export `TagColorKeys` from the entrypoint.
Create a `Tag/types.ts`, and move `TagColorKeys` and `TagColors` there, then in `Tag/index.ts` export only `TagColors` from the new file.

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
-import { TagIcon } from "~components/Icons"
-import { ComponentDocsTemplate } from "~storybook/components/DocsContainer"
-import { StickerSheet } from "~storybook/components/StickerSheet"
-import { Tag } from ".."
+import { Meta, StoryObj } from "@storybook/react"
+import { AcademyIcon, ActionOffIcon, AddIcon, TagIcon } from "~components/Icons"
+import { Tag, TagColorKeys } from ".."
+import styles from "./TagStories.module.scss"
 
 export default {
```
HeartSquared, line 7: Create a `Story` type within this file (eg. in `WorkflowFooter`), then add any default args for your stories into the `args` prop in the `meta`.

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
-    },
-  },
 } satisfies Meta<typeof Tag>
 
-export const Playground: StoryFn<typeof Tag> = args => (
-  <Tag {...args}>My tag</Tag>
-)
+export const Playground: StoryObj = {
+  render: args => <Tag {...args}>My tag</Tag>,
+}
```
HeartSquared, line 14: We usually like to have the Canvas open for Playground stories, thus the extra params.

```suggestion
export const Playground: Story = {
  parameters: {
    docs: {
      canvas: {
        sourceState: "shown",
      },
    },
  },
}
```

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
+
+export const SimpleTag: StoryObj = {
+  render: args => <Tag {...args}>My Tag</Tag>,
+}
```
HeartSquared, line 18:
```suggestion
export const SimpleTag: StoryObj = {}
```

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
 
-type ListProps = {
-  children: React.ReactNode
+export const Children: StoryObj = {
+  render: args => <Tag {...args}>This text is the children</Tag>,
 }
```
HeartSquared, line 22: This is the format for overriding any default args (extends the ones defined in meta).
`render` has a default definition of (where `Component` is what you have defined in `meta.component`) `args => <Component {...args />`, thus you do not need to define it unless it does not suit your needs.

```suggestion
export const Children: StoryObj = {
  args: {
    children: "This text is the children",
  }
}
```

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
-          <Tag color="purple" Icon={<TagIcon role="presentation" />}>
-            Purple
-          </Tag>
-        </List>
-      </StickerSheet.Row>
-    </StickerSheet.Body>
-  </StickerSheet>
-)
+export const Color: StoryObj = {
+  render: _args => (
```
HeartSquared, line 25:
```suggestion
  render: () => (
```

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
+  ),
+}
 
-export const StickerSheetDefault = StickerSheetTemplate.bind({})
-StickerSheetDefault.storyName = "Sticker Sheet (Default)"
-StickerSheetDefault.parameters = {
-  chromatic: { disable: false },
-  controls: { disable: true },
+export const Icon: StoryObj = {
+  render: _args => (
```
HeartSquared, line 37:
```suggestion
  render: () => (
```

packages/components/src/__future__/Tag/_docs/Tag.stories.tsx
```diff
@@ -1,94 +1,52 @@
...
 }
 
-export const StickerSheetRTL = StickerSheetTemplate.bind({})
-StickerSheetRTL.storyName = "Sticker Sheet (RTL)"
-StickerSheetRTL.parameters = {
-  chromatic: { disable: false },
-  controls: { disable: true },
-  textDirection: "rtl",
+export const ClassNameOverride: StoryObj = {
+  render: _args => <Tag classNameOverride={styles.myCustomClass}>Hello</Tag>,
```
HeartSquared, line 51: Use `args` key instead :)

Given this is documentation only, you can use Tailwind instead of defining a scss class.

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -1,60 +1,30 @@
...
+import { TagColors } from "./types"
+import styles from "./Tag.module.scss"
 
 export interface TagProps
   extends OverrideClassName<HTMLAttributes<HTMLSpanElement>> {
   children: ReactNode
   color?: TagColors
-  Icon?: React.ReactElement
+  icon?: React.ReactElement
 }
```
HeartSquared, line 12: Ah foo, I missed this earlier. My bad.

```
export type TagProps {
  children: ReactNode
  color?: TagColors
  Icon?: React.ReactElement
} & OverrideClassName<HTMLAttributes<HTMLSpanElement>>
```

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -1,60 +1,30 @@
...
-        ["bg-blue-100 text-blue-700"]: color === "blue",
-        ["bg-green-100 text-green-700"]: color === "green",
-        ["bg-yellow-100 text-yellow-700"]: color === "yellow",
-        ["bg-orange-100 text-orange-700"]: color === "orange",
-        ["bg-red-100 text-red-700"]: color === "red",
-        ["bg-purple-100 text-purple-700"]: color === "purple",
-      },
-      classNameOverride
-    )}
+    className={classNames(styles.container, styles[color], classNameOverride)}
```
HeartSquared, line 21: Usually we'd use the component name for the main classname :) It's part of the template, but I suppose it was removed in the contribution 😅 

```suggestion
    className={classNames(styles.tag, styles[color], classNameOverride)}
```

packages/components/src/__future__/Tag/Tag.tsx
```diff
@@ -1,60 +1,30 @@
...
-      </span>
-    )}
+    {icon && <span className={styles.iconContainer}>{icon}</span>}
     {children}
   </span>
 )
 
 Tag.displayName = "Tag"
+
+export { TagColors }
```
HeartSquared, line 30: Close! You actually want to add this export to the `index.ts`, not through this file.

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,79 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/typography";
+
+.container {
+  display: inline-flex;
+  align-items: center;
+  padding: 0.3rem $spacing-8 0.3rem $spacing-8;
```
HeartSquared, line 8: Just checked the designs, and it should actually be 4px on the y-axis

```suggestion
  padding: $spacing-4 $spacing-8;
```

https://www.figma.com/file/hANEprZKom5Br1IBXhE8mr/%F0%9F%A7%AA-Kaizen-Labs?node-id=28615%3A183611&mode=dev

packages/components/src/__future__/Tag/Tag.module.scss
```diff
@@ -0,0 +1,79 @@
...
+  line-height: $typography-paragraph-small-line-height;
+  white-space: nowrap;
+}
+
+.iconContainer {
+  display: inline-flex;
+}
+
+.gray {
+  background-color: $color-gray-200;
```
HeartSquared, line 23: Spoke to @jmitsios and this was a mistake in the Figma. It has been fixed to be `$color-gray-300`, so this needs to be updated to match :)

HeartSquared (approved):

PR #4101, opened 2023-09-24
KDS-1752 Radio, RadioGroup, RadioField

gyfchong:
## Why
Migrationnnnn, Migration.

## What
Radio, RadioGroup, RadioField, Label


packages/components/styles/_forms-variables.scss
```diff
@@ -0,0 +1,11 @@
+// General purpose tokens
```
mcwinter07, line 1: It'll be nice to have these for the other form components we port over / make in the future

packages/components/styles/_forms-variables.scss
```diff
@@ -0,0 +1,11 @@
...
+$dt-color-form-border-color: $color-gray-500;
+$dt-color-form-border-color-focus: $color-blue-500;
```
mcwinter07, line 3: Probably don't need to tackle in this PR but we could potentially create a more global stylesheet to store focus ring variables since they shouldn't differ from component to component

packages/components/styles/_forms.scss
```diff
@@ -1,9 +1,9 @@
+@import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/design-tokens/sass/border";
 @import "~@kaizen/design-tokens/sass/typography";
 @import "./legacy/type";
```
mcwinter07, line 4: Is there anything that is used from `legacy/type` in this stylesheet? Most of those are prefixed by `ca-` 🤔 would be awesome if we didn't need it

packages/components/src/Radio/RadioGroup/RadioGroup.module.scss
```diff
@@ -0,0 +1,27 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+
+.radioGroupLabel {
+  margin-bottom: $spacing-sm;
+
+  label {
+    :global(.ideal-sans) & {
+      // This is to override bootstrap styles. Remove when appropriate
```
mcwinter07, line 9: Hopefully we don't need this any more... then again... 😬

packages/components/src/Radio/RadioGroup/RadioGroup.module.scss
```diff
@@ -0,0 +1,27 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+
+.radioGroupLabel {
+  margin-bottom: $spacing-sm;
```
mcwinter07, line 5: Should we update all spacing tokens to use `$spacing-{number}`equivalent in these PRs?

packages/components/src/Radio/Radio/Radio.module.scss
```diff
@@ -0,0 +1,101 @@
...
+@import "~@kaizen/design-tokens/sass/color";
+@import "../../../styles/forms";
+@import "../../../styles/forms-variables";
+
+$radio-size: 24px;
+$icon-size: 10px;
+$icon-offset: 5px;
+$focus-ring-offset: 2px;
+
+$dt-color-radio-background-color-hover: $color-gray-100;
```
mcwinter07, line 11: I think background hover color for our inputs are gray-200 now

Screenshot 2023-09-27 at 2 46 46 pm

Screenshot 2023-09-27 at 2 47 47 pm

packages/components/src/Radio/Radio/_docs/Radio.mdx
```diff
@@ -0,0 +1,26 @@
...
+<KaioNotification />
+
+<Installation
+  installCommand="yarn add @kaizen/components"
+  importStatement='import { Radio } from "@kaizen/components"'
+/>
+
+## Overview
+
+Either one or the other. Commonly used via [RadioField](https://cultureamp.design/?path=/docs/components-radio-controls-radio-field--docs) or [RadioGroup](https://cultureamp.design/?path=/docs/components-radio-controls-radio-group--docs), proceed with caution.
```
mcwinter07, line 23: > proceed with caution

😆 For safety we could call out that this is a primitive and should ideally not be used

packages/components/src/Radio/RadioGroup/RadioGroup.module.scss
gyfchong: Hm, nah leave them. We can fix them later when we transition to CSS Modules or something.

packages/components/src/Radio/Radio/Radio.module.scss
gyfchong: Nice catch

packages/components/styles/_forms.scss
gyfchong: Nice catch! Nothing. Removed.

packages/components/styles/_forms-variables.scss
gyfchong: Indeed a good one, pop a ticket in the backlog

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <Radio id="radio" name="radio" value="radio" reversed={isReversed} />
+        </StickerSheet.Row>
+        <StickerSheet.Row>
+          <Radio
+            id="radio"
+            name="radio"
+            value="radio"
+            reversed={isReversed}
+            selectedStatus
+          />
```
mcwinter07, line 42: We should also add the pseudo stories in the [new format](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3230040105/2.3+Testing#Testing-pseudo-states) Cass has whipped up, which should look like this:

```jsx
  render: () => (
    <Radio
      {...radioProps}
      data-sb-pseudo-styles="hover"
    />
  ),
  parameters: {
    pseudo: {
      hover: '[data-sb-pseudo-styles="hover"]',
   }
```

gyfchong: Unfortunately doesn't work cause neither the `classNameOverride` or the `data-attr` reach the actual child element that triggers the hover state, but focus works!

packages/components/src/Radio/Radio/_docs/Radio.mdx
```diff
@@ -0,0 +1,28 @@
...
+<Installation
+  installCommand="yarn add @kaizen/components"
+  importStatement='import { Radio } from "@kaizen/components"'
+/>
+
+## Overview
+
+A primitive, not to be used alone.
+
+Commonly used via [RadioField](https://cultureamp.design/?path=/docs/components-radio-controls-radio-field--docs) or [RadioGroup](https://cultureamp.design/?path=/docs/components-radio-controls-radio-group--docs).
```
HeartSquared, line 25: Nice doc. Try using `<LinkTo` for relative links :)

```suggestion
import { LinkTo } from "~storybook/components/LinkTo"

Commonly used via <LinkTo pageId="components-radio-controls-radio-field">RadioField</LinkTo> or <LinkTo pageId="components-radio-controls-radio-group">RadioGroup</LinkTo>.
```

packages/components/src/Radio/RadioField/RadioField.module.scss
```diff
@@ -0,0 +1,51 @@
...
+    -webkit-tap-highlight-color: transparent;
+
+    :global(.ideal-sans) & {
+      // This is to override bootstrap styles. Remove when appropriate
+      font-size: $typography-paragraph-body-font-size;
+      padding-top: 0;
+    }
+  }
+
+  @media (hover: hover) and (pointer: fine) {
```
HeartSquared, line 25: The label appears to change colour? 😐 I suppose maybe you should have a stickersheet for this then? 😂

packages/components/src/Radio/Radio/Radio.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { InputHTMLAttributes } from "react"
+import classnames from "classnames"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./Radio.module.scss"
+
+export type RadioProps = OverrideClassName<
+  Omit<InputHTMLAttributes<HTMLInputElement>, "type" | "checked">
+> & {
+  id: string
```
HeartSquared, line 9: Why is `id` required when we do nothing with it? Noticed this when I saw RadioField made it optional.
Maybe just make it optional here (it won't break anything), and everyone can be happy.

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <Radio
+            id="radio4"
+            name="radio4"
+            value="radio4"
+            reversed={isReversed}
+            data-sb-pseudo-styles="hover"
+          />
+        </StickerSheet.Row>
+        <StickerSheet.Row>
```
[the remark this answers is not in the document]
gyfchong: Added `selectedStatus`

packages/components/src/Radio/Radio/Radio.tsx
mcwinter07: Isn't that because we have a generate an id in the RadioField if one is not provided, whereas a Radio should always and id that the a label will point to?

packages/components/src/Radio/RadioField/RadioField.module.scss
gyfchong: Oh, this isn't the `label` changing colour, it's so that hovering the label changes the radio's colour...

packages/components/src/Radio/Radio/Radio.tsx
gyfchong: Yeah `Radio` itself would always need an `id` to link it to a `label`

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+        rules: [
+          {
+            // Built with no label on purpose, to be used within `RadioField` where label is present
+            id: "label",
+            enabled: false,
+          },
+        ],
+      },
+    },
+  },
```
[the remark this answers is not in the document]
gyfchong: This might be a good way to document something that's a feature not a bug? If we add `aria-label` it makes it feel like something that is a legit pattern to follow and not something we do just for stickersheets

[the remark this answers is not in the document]
HeartSquared: Good point! Yeah I'm on board with this approach :) Worth documenting.

packages/components/src/Radio/RadioField/RadioField.module.scss
HeartSquared: 🫠

packages/components/src/Radio/Radio/Radio.tsx
HeartSquared: I suppose so. I wonder if this is the right place to enforce that though (maybe it is).
Maybe just add a comment above to explain why then (which should flow onto the docs).

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+} satisfies Meta
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed }) => (
+    <StickerSheet isReversed={isReversed}>
+      <StickerSheet.Header
+        headings={["Default", "Disabled", "Focus", "Hover"]}
```
HeartSquared, line 32:
```suggestion
        headings={["Default", "Disabled", "Focus", "Hover"]}
        hasVerticalHeadings
```

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+            value="radio33"
+            reversed={isReversed}
+            selectedStatus
+            data-sb-pseudo-styles="focus"
+          />
+          <Radio
+            id="radio44"
+            name="radio44"
+            value="radio44"
+            reversed={isReversed}
```
HeartSquared, line 92:
```suggestion
            reversed={isReversed}
            selectedStatus
```

packages/components/src/Radio/Radio/_docs/Radio.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+            value="radio1"
+            reversed={isReversed}
+          />
+          <Radio
+            id="radio2"
+            name="radio2"
+            value="radio2"
+            reversed={isReversed}
+            disabled
+          />
```
HeartSquared, line 48: Does this actually house the disabled styles? The stickersheet doesn't seem to show it looking different

HeartSquared (approved):
🔘

PR #4102, opened 2023-09-24
KDS-1802  add Label to KAIO

gyfchong:
## Why
Pretty atomic and re-used component. So deserves it's own special PR.


packages/components/src/Label/_docs/Label.stickersheet.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    controls: { disable: true },
+  },
+} satisfies Meta
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed }) => (
+    <StickerSheet isReversed={isReversed}>
+      <StickerSheet.Header headings={["Default", "Disabled", "Prominent"]} />
+      <StickerSheet.Body>
+        <StickerSheet.Row rowTitle="Base">
```
HeartSquared, line 22: image

Either this or get the headings to shift across one
```suggestion
        <StickerSheet.Row>
```

packages/components/styles/_form-variables.scss
```diff
@@ -0,0 +1,11 @@
+// General purpose tokens
+$dt-color-form-border-color: $color-gray-500;
```
HeartSquared, line 2: Shouldn't we import the colour tokens into this file?

packages/components/src/Label/_docs/Label.stickersheet.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  name: "Sticker Sheet (Reversed)",
+  parameters: { backgrounds: { default: "Purple 700" } },
+  args: { isReversed: true },
+}
+
+export const StickerSheetRTL: StickerSheetStory = {
+  ...StickerSheetTemplate,
+  name: "Sticker Sheet (RTL)",
+  parameters: { textDirection: "rtl" },
+}
```
HeartSquared, line 52: To make use of this, I think you need to add a few more examples using `children` and `labelText` in combination (which, if I look at it, seems like the `children` should be the main text and `labelText` should be subtext?)

packages/components/src/Label/_docs/Label.stickersheet.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <StickerSheet.Header headings={["Default", "Disabled", "Prominent"]} />
+      <StickerSheet.Body>
+        <StickerSheet.Row rowTitle="Base">
+          <Label labelText="Label Text" reversed={isReversed}></Label>
+          <Label labelText="Label Text" reversed={isReversed} disabled></Label>
+          <Label
+            labelText="Label Text"
+            reversed={isReversed}
+            variant="prominent"
+          ></Label>
```
HeartSquared, line 29:
```suggestion
          <Label reversed={isReversed}>Label Text</Label>
          <Label reversed={isReversed} disabled>Label Text</Label>
          <Label reversed={isReversed} variant="prominent">Label Text</Label>
```

packages/components/src/Label/_docs/Label.mdx
```diff
@@ -0,0 +1,26 @@
...
+/>
+
+<KaioNotification />
+
+<Installation
+  installCommand="yarn add @kaizen/components"
+  importStatement='import { Label } from "@kaizen/components"'
+/>
+
+## Overview
```
HeartSquared, line 21: Add some extra docs for the missing API

packages/components/src/Label/Label.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+      container.querySelector(`[id="${defaultLabelProps.id}"]`)
+    ).toBeTruthy()
+  })
+
+  it("renders a reversed label", () => {
+    const { container } = renderLabel({ reversed: true })
+    expect(container.querySelector(".reversed")).toBeTruthy()
+  })
+
+  describe("label type", () => {
```
HeartSquared, line 41: Just delete this file and add the test cases to the stickersheet instead. These are either meant to be visual or are just testing that we spread the rest of the attributes into the component.

packages/components/src/Label/Label.tsx
```diff
@@ -0,0 +1,53 @@
...
+import classnames from "classnames"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./Label.module.scss"
+type LabelType = "text" | "checkbox" | "toggle" | "radio"
+
+export type LabelProps = OverrideClassName<
+  LabelHTMLAttributes<HTMLLabelElement>
+> & {
+  children?: React.ReactNode
+  labelText?: string | React.ReactNode
```
HeartSquared, line 11: This is honestly extremely confusing API because both these props do the same thing except when they are next to something like a checkbox (which still, technically could've just added classes and remained using `children` - less you'd have text to both the left and the right of the checkbox at the same time).

Give it some thought on whether you want to clean it up a little or not, given now is safer to do so as opposed to after teams start to use it (that being said, this component is too complex for what it needs to be anyway, so you can choose to just say it gets simplified in a rewrite).

packages/components/src/Label/Label.module.scss
```diff
@@ -0,0 +1,120 @@
...
+  [dir="rtl"] & {
+    text-align: right; // fallback for IE and Edge
+    text-align: start;
+  }
+}
+
+.reversed {
+  color: $color-white;
+
+  a {
```
HeartSquared, line 28: I don't know why we override anchor styles here, but I suppose it means it should be added to the stickersheet.

packages/components/src/Label/Label.tsx
gyfchong: Not on the fly 😅

packages/components/src/Label/Label.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { LabelHTMLAttributes } from "react"
+import classnames from "classnames"
+import { OverrideClassName } from "~types/OverrideClassName"
+import styles from "./Label.module.scss"
+type LabelType = "text" | "checkbox" | "toggle" | "radio"
+
+export type LabelProps = OverrideClassName<
+  LabelHTMLAttributes<HTMLLabelElement>
+> & {
```
HeartSquared, line 9: Just a suggestion, but if you put `OverrideClassName<LabelHTMLAttributes<HTMLLabelElement>>` after the `{}`, I think it doesn't break into multiline.

gyfchong: Hm, looks fine to me 😄

packages/components/src/Label/_docs/Label.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+ * By default, your label text will always sit to the right of the inline form control (ie. in RTL terms, the "end").
+ * Using the `labelPosition` you can flip the label text to sit to the left (ie. in RTL terms, the "start").
+ */
+
+export const Position: Story = {
+  render: () => (
+    <Label labelText="Checkbox" labelType="checkbox" labelPosition="start">
+      <InlineFormControl />
+    </Label>
+  ),
```
HeartSquared, line 58: You can use this format for all the stories which only have a single Label :)

```suggestion
  args: {
    labelText: "Checkbox",
    labelType: "checkbox",
    labelPosition: "start",
    children: <InlineFormControl />
  },
```

gyfchong: You reminded me, this needed two examples to showcase difference between start and end

HeartSquared (approved):

PR #4103, opened 2023-09-26
Migrate FieldMessage to KAIO

mcwinter07:
## Why
I'm splitting up this PR since the FieldMessage is a primitive used in forms so trying to get this in first before it blocks @gyfchong with the form components

## What
- Migrates the FieldMessage component
  - removes deprecated automationid
  - updates to CSF
  - adds documentation to MDX


packages/components/src/FieldMessage/FieldMessage.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+  })
+
+  it("renders a warning icon with an error status", () => {
+    const { container, getByLabelText } = render(
+      <FieldMessageWrapper status="caution" />
+    )
+
+    expect(container.querySelector(".warningIcon")).toBeTruthy()
+    expect(getByLabelText("caution message")).toBeInTheDocument()
+  })
```
gyfchong, line 75: Can remove these in favour of the Stickersheets

packages/components/src/FieldMessage/_docs/FieldMessage.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+<Meta of={FieldMessageStories} />
+
+# FieldMessage
+
+{/*
+  * @todo:
+  * - Adjust Github link (if needed)
+  * - Add Figma and Design Guidelines (Confluence) link
+  */}
```
gyfchong, line 13: Remove

packages/components/src/FieldMessage/_docs/FieldMessage.stickersheet.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+  },
+  args: {
+    message:
+      "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullasemper odio vitae sem gravida rutrum.",
+  },
+} satisfies Meta<typeof FieldMessage>
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed, ...otherProps }) => (
+    /** @note: If you have multiple StickerSheets to display, you can add a `heading` */
```
gyfchong, line 23: Remove

packages/components/src/FieldMessage/_docs/FieldMessage.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const Reversed: Story = {
+  args: { reversed: true, status: "error" },
+  parameters: { backgrounds: { default: "Purple 700" } },
+  name: "reversed",
+}
+
+export const Position: Story = {
+  args: { position: "top" },
+}
```
gyfchong, line 68: I'd probably render two `FieldMessage` side by side to show the default vs the top versionss

packages/components/src/FieldMessage/_docs/FieldMessage.mdx
```diff
@@ -0,0 +1,68 @@
...
+## Overview
+
+FieldMessage is used and consumed in Kaizen form elements to display either form field descriptions or validation messages.
+
+<Canvas of={FieldMessageStories.Playground} />
+<Controls of={FieldMessageStories.Playground} />
+
+## API
+
+### message
```
gyfchong, line 37:
```suggestion
### Message
```
Title case please

packages/components/src/FieldMessage/_docs/FieldMessage.mdx
mcwinter07: Ah yep, nice catch, will axe the boilerplate comments 😉

packages/components/src/FieldMessage/FieldMessage.spec.tsx
mcwinter07: Could probably get rid of the SVG query selector but do we want to keep the test in that checks the accessible label is constructed, ie: `getByLabelText("caution message")`?

Actually also on the fence whether this icon needs an actual img role? Didn't want to change to much in case our consumer rely on this accessible title but does beg the question if this provides much context to the end user?

mcwinter07: So I was thinking something more like this if we want to test the accessible title that gets constructed:
```
  it("renders the error icon with an accessible title", () => {
    const { getByLabelText } = render(<FieldMessageWrapper status="error" />)

    expect(getByLabelText("error message")).toBeInTheDocument()
  })
```

gyfchong: Ah I see, makes sense!

packages/components/src/FieldMessage/_docs/FieldMessage.stories.tsx
mcwinter07: Good idea, I've updated the story to have them side by side. I used a cautionary status since its visually easier to see than just whitespace

gyfchong (approved):

PR #4105, opened 2023-09-26
Ignore-color-contrast-global-config

gyfchong:
## Why
Some elements will fail colour contrast on purpose, eg. disabled styles

## What
A universally available data attrib that can disable the color-contrast rule.


HeartSquared (dismissed):

storybook/global-a11y-rules.ts
```diff
@@ -20,7 +20,8 @@ export const globalA11yRules = [
     enabled: false,
   },
   {
+    // add a data-attribute for exceptions such as pseudo states
     id: "color-contrast",
-    selector: "*:not([data-sb-a11y-color-contrast-disable])",
+    selector: ":not([data-a11y-color-contrast-disable])",
```
gyfchong, line 25: I like the `sb` to signify it's for storybook

HeartSquared: Oh! I missed that my bad. I unstashed from my other branch and thought the only change was the `*` 😅 We can put it back.

HeartSquared (approved):

PR #4106, opened 2023-09-26
refactor: use storybook pseudo states addon

HeartSquared:
## Why

The [Storybook pseudo states addon](https://storybook.js.org/addons/storybook-addon-pseudo-states) has been updated to allow targeting elements as opposed to applying to the story as a whole.
This means we can clean up our custom classnames for emulating pseudo states :)

## What

- Add devDep `storybook-addon-pseudo-states`
- Update existing stickersheets and clean up emulated pseudo states
- Add stickersheet for `ClearButton` and `InputSearch` as they were missing
- Update plop template to include pseudo state emulation
- Fix Stickersheet `"empty-table-header"` a11y issue


packages/components/src/FilterButton/_docs/filter-buttons.stickersheet.stories.tsx
```diff
@@ -79,6 +79,14 @@ const StickerSheetTemplate: StickerSheetStory = {
       </StickerSheet>
     </>
   ),
+  parameters: {
+    pseudo: {
+      hover: ".story__filter-button-base--hover",
+      active: ".story__filter-button-base--active",
+      focus: ".story__filter-button-base--focus",
+      focusVisible: ".story__filter-button-base--focus",
```
HeartSquared, line 87: Maybe I should just use `id`s. As long as we never render multiple stickersheets at the same time, there shouldn't be a conflict.

gyfchong:
Can we leave this till after we migrate everything? Then we can update all pseudo states in one go?

HeartSquared:
> Can we leave this till after we migrate everything? Then we can update all pseudo states in one go?

No need to conform to this format just yet - we can clean them up after the migration if they were pre-existing. This just makes it easier for any newer pseudo states we need to emulate (eg. new components/back-filling for migrations).

packages/components/src/Button/Button/_docs/Button.stickersheet.stories.tsx
```diff
@@ -251,6 +251,14 @@ const StickerSheetTemplate: StickerSheetStory = {
       </>
     )
   },
+  parameters: {
+    pseudo: {
+      hover: ".story__button--hover",
+      active: ".story__button--active",
+      focus: ".story__button--focus",
+      focusVisible: ".story__button--focus",
```
gyfchong, line 259: Would these work as data attributes at a global level?

HeartSquared: These are all temporary styles which get added to the stylesheet when visiting the story. They don't actually exist.
But yes I suppose we could probably use a data attribute as the target (haven't tried).

I think they do become global though, which I'm not sure if it will start meshing the styles as you navigate through stories (they only reset on refresh), which is why I made them unique per stickersheet to ensure nothing weird happens.

HeartSquared: As discussed, didn't apply it at a global level to avoid confusion with magic data attributes (at least for now), especially since the structure of some of our components pass data attributes into an inner element as opposed to the container (classNameOverride has been put here in most cases though).

Have updated to use data attributes where possible though :)

packages/components/src/ButtonGroup/_docs/ButtonGroup.stickersheet.stories.tsx
```diff
@@ -86,7 +86,7 @@ const StickerSheetTemplate: StickerSheetStory = {
             <ButtonGroup>
               <FilterButtonBase>First</FilterButtonBase>
               <FilterButtonBase
-                classNameOverride="story__filter-button-base--hover"
+                data-pseudo-styles="hover"
```
gyfchong, line 89:
```suggestion
                data-sb-pseudo-styles="hover"
```
 To match the a11y disable attribute?

gyfchong (approved):

PR #4107, opened 2023-09-26
Update plop actions for subcomponent generation

mcwinter07:
## Why

[KDS-1864](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1864)

This is a small quality of life improvement for something Cass and I wanted to implement to ensure subcomponents are correctly folder-ed

## What
- updates plopfile config
  -  adds sub-components to a folder and generates nested _docs

[KDS-1864]: https://cultureamp.atlassian.net/browse/KDS-1864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plopfile.js
```diff
@@ -38,10 +38,17 @@ module.exports = (
...
-            destination: `packages/components/${src}/{{pascalCase parentComponentName}}/subcomponents`,
+            destination: `packages/components/${src}/{{pascalCase parentComponentName}}/subcomponents/{{pascalCase componentName}}`,
             base: "plop-templates/basic-component/src",
             templateFiles: "plop-templates/basic-component/src/**/*.hbs",
           },
+          {
+            type: "addMany",
+            destination: `packages/components/${src}/{{pascalCase parentComponentName}}/subcomponents/{{pascalCase componentName}}/_docs`,
+            base: "plop-templates/basic-component/docs",
+            templateFiles:
```
HeartSquared, line 49: Nice add! However, this means that the `title` in the meta will be incorrect, so maybe we should also add a slight change to the template?

Example here https://github.com/cultureamp/kaizen-design-system/pull/4009/files#diff-a14030f4b4431c8cc6e6b098367fcc4173ef6413c59aecbd6594196ed1a367f1L60-L62

mcwinter07: Nice catch, I'll make that amendment and we should be hunky dory 🕺

HeartSquared: I've made the change while you're away, so we can use it for migrations :)

gyfchong (approved):

PR #4110, opened 2023-09-28
feat(select): migrate Select component

HeartSquared:
## Why

Migration.

## What

- Add `Select` component as a `future` component
	- Refactors made to reuse subcomponents and utils from `FilterSelect`
	- Is now generic
	- Uses `MultiSelect/Popover`
		- Slight refactor to give it a generic as usage required it (nothing breaks)
- Subcomponents and utils moved from `FilterSelect` to `future/Select`
- Add `StickerSheetCell` to be able to control the TD in a stickersheet (previous behaviour remains as fallback)


packages/components/src/FilterSelect/FilterSelect.tsx
```diff
@@ -7,18 +7,18 @@ import {
...
 import { Filter, FilterContents } from "~components/Filter"
 import { FilterButtonProps } from "~components/FilterButton"
+import { SelectProvider } from "~components/__future__/Select/context"
+import { ListBox } from "~components/__future__/Select/subcomponents/ListBox"
+import { ListBoxSection } from "~components/__future__/Select/subcomponents/ListBoxSection"
+import { ListItem } from "~components/__future__/Select/subcomponents/ListItem"
+import { ListItems } from "~components/__future__/Select/subcomponents/ListItems"
+import { Option } from "~components/__future__/Select/subcomponents/Option"
+import { Overlay } from "~components/__future__/Select/subcomponents/Overlay"
+import { SectionDivider } from "~components/__future__/Select/subcomponents/SectionDivider"
```
gyfchong, line 17: Could be good to have an `index.ts` for the sub components?

gyfchong:
Looking at this PR makes me feel like we also need an internal codemod to automate the transition of our future components

packages/components/src/__future__/Select/_docs/Select.stickersheet.stories.tsx
```diff
@@ -0,0 +1,465 @@
...
+              isReversed={isReversed}
+              data-sb-pseudo-styles="focus"
+            />
+            <Select
+              label="Label"
+              items={singleMockItems}
+              placeholder="Placeholder"
+              isDisabled
+              isReversed={isReversed}
+            />
```
gyfchong, line 96: Huh, this doesn't seem to trigger the contrast ratio error?

packages/components/src/__future__/Select/Select.tsx
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Use the `fullWidth` styles.
+   */
+  isFullWidth?: boolean
+  /**
+   * @deprecated: Either define `disabled` in your `Option` (in `items`), or use `disabledKeys`
+   */
+  disabledValues?: React.Key[]
+} & OverrideClassName<Omit<AriaSelectProps<Option>, OmittedAriaSelectProps>>
+
```
gyfchong, line 68: Needs the link comments
```
/**
 * {@link ... Guidance} |
 * {@link ... Storybook}
 */
```

packages/components/src/FilterSelect/FilterSelect.tsx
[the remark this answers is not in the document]
HeartSquared: Fair. Usually we don't, but given how many there are this might be a good exception.

packages/components/src/FilterSelect/types.ts
```diff
@@ -1,34 +1 @@
...
-  Omit<Node<SelectOptionGroup<Option>>, "childNodes"> & {
-    type: "section"
-    hasChildNodes: true
-    childNodes: Array<SelectOptionNode<Option>>
-  }
-
-export type SelectItemNode<Option extends SelectOption = SelectOption> =
-  | SelectOptionNode<Option>
-  | SelectOptionGroupNode<Option>
+export * from "~components/__future__/Select/types"
```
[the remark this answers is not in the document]
HeartSquared: For consumers. Otherwise if they want to use the types they'll have to dig into `future` and that would be pretty weird given `FilterSelect` isn't from there 😅 

I can leave a comment though. This is only temporary until they aren't split.

packages/components/src/__future__/Select/Select.tsx
```diff
@@ -0,0 +1,207 @@
...
+  items: Array<SelectItem<Option>>
+  id?: string
+  trigger?: (
+    selectToggleProps: SelectToggleProps & {
+      ref: UseFloatingReturn<HTMLButtonElement>["refs"]["setReference"]
+    },
+    // @deprecated: This arg is unnecessary now, but provided for legacy usages
+    ref: UseFloatingReturn<HTMLButtonElement>["refs"]["setReference"]
+  ) => JSX.Element
+  children?: (args: { items: Array<SelectItemNode<Option>> }) => React.ReactNode
```
HeartSquared, line 47: Should reference the prop from the popover contents

packages/components/src/FilterSelect/types.ts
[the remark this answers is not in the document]
gyfchong: Oh, I didn't see that these types are being made available to consumers, cool cool

gyfchong (approved):

PR #4111, opened 2023-09-29
KDS-1786-migrate-tag-from-kaizen-legacy-to-tag-legacy-in-kds

gyfchong:
Legacy Tag.

- Fix to the `LiveIcon` stroke colour
- Move the future Tag into a sub folder.


packages/components/src/Tag/Tag.tsx
```diff
@@ -0,0 +1,187 @@
...
+                fill="none"
+              />
+              <LiveIcon
+                role="presentation"
+                classNameOverride={styles.liveIcon_3}
+                width="16"
+                height="16"
+                viewBox="0 0 16 16"
+                fill="none"
+              />
```
HeartSquared, line 178: This should be consistent whenever we use the `LiveIcon` right? It should be baked into `LiveIcon` instead (maybe with an `isAnimated` prop if we want the animation to be optional).

packages/components/src/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+Tags help users quickly recognize important information about items that organize and categorize them. They visually label items with small amounts of information or the item's status, usually with keywords that help organize and categorize the items.
+
+<Canvas of={TagStories.Playground} />
+<Controls of={TagStories.Playground} />
+
+## API
+
+### Variants
+We have 3 different sets of combinations for various uses
```
HeartSquared, line 33:
```suggestion
We have 3 different sets of combinations for various uses.
```

packages/components/src/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+<Canvas of={TagStories.Status} />
+
+#### Validation
+Use validation tags for indicating the validation state such as the Cautionary style for cautioning users that the item has a validation warning that could be addressed.
+
+<Canvas of={TagStories.Validation} />
+
+#### Sentiments
+Use sentiment tags only for sentiment or other employee feedback data
```
HeartSquared, line 46:
```suggestion
Use sentiment tags only for sentiment or other employee feedback data.
```

packages/components/src/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,71 @@
...
+### Avatar
+**Note:** Avatar are not supported in `small` sizes.
+
+<Canvas of={TagStories.Profile} />
+
+### Dismissable
+<Canvas of={TagStories.Dismissable} />
+
+### TruncateWidth
+A pixel width which sets a limit for when the text starts to truncate
```
HeartSquared, line 64:
```suggestion
A pixel width which sets a limit for when the text starts to truncate.
```

packages/components/src/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+<Canvas of={TagStories.Truncate} />
+
+### Inline
+Removes the default right-hand margin
```
HeartSquared, line 69:
```suggestion
Removes the default right-hand margin.
```

packages/components/src/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </StickerSheet.Row>
+      </StickerSheet.Body>
+    </StickerSheet>
+  ),
+}
+
+export const StickerSheetDefault: StickerSheetStory = {
+  ...StickerSheetTemplate,
+  name: "Sticker Sheet (Default)",
+}
```
HeartSquared, line 85: Suggestion: We might actually want a Reversed stickersheet despite it not having different styles, simply because Tags are often used on the reversed background, so we can have SB a11y check the colour contrast.

packages/components/src/Tag/_docs/Tag.mdx
```diff
@@ -0,0 +1,71 @@
...
+Use sentiment tags only for sentiment or other employee feedback data
+
+<Canvas of={TagStories.Sentiments} />
+
+### Sizes
+**Note:** Icons and Avatars are not supported in `small` sizes.
+
+<Canvas of={TagStories.Sizes} />
+
+### Avatar
```
HeartSquared, line 55: Perhaps worth noting that there are two ways of adding an Avatar - JSX.Element and AvatarProps.

packages/components/src/Tag/Tag.tsx
gyfchong: I opted not to try to change this in the original icon cause its the only one that requires these extra props, and I'd rather consider a change in the SVG/design than to modify our template to accommodate.

packages/components/src/Tag/_docs/Tag.mdx
gyfchong: Oh, had no idea!

packages/components/svgoUtils.js
```diff
@@ -47,14 +47,20 @@ const replaceColor = child => {
   if (child.attributes.fill === "#000" || child.attributes.fill === "#000000") {
     child.attributes.fill = "currentColor"
   }
+  if (
+    child.attributes.stroke === "#000" ||
+    child.attributes.stroke === "#000000"
+  ) {
+    child.attributes.stroke = "currentColor"
+  }
```
gyfchong, line 55: Fixes `LiveIcon`'s case where the stroke is where the colour is applied.

packages/components/src/Tag/_docs/Tag.stickersheet.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { Avatar } from "~components/Avatar"
+import {
+  StickerSheet,
+  StickerSheetStory,
+} from "~storybook/components/StickerSheet"
+import { Tag } from "../index"
+import { TagVariants } from "../types"
+
+export default {
+  title: "KAIO-staging/Tag",
```
HeartSquared, line 12:
```suggestion
  title: "Staging/Tag",
```

packages/components/src/Tag/_docs/Tag.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { Avatar } from "~components/Avatar"
+import { Tag } from "../index"
+import {
+  SentimentTagVariants,
+  StatusTagVariants,
+  ValidationTagVariants,
+} from "../types"
+
+const meta = {
+  title: "KAIO-staging/Tag",
```
HeartSquared, line 12:
```suggestion
  title: "Staging/Tag",
```

HeartSquared (approved):

PR #4112, opened 2023-09-29
KDS-1750-migrate-input-to-kaio

gyfchong:
- Migrated `Input` to KAIO
- Added `InputSearch` docs
- Organised Inputs into their own dir


packages/components/src/Input/_docs/Input.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+    docs: {
+      source: { type: "dynamic" },
+    },
+  },
+}
+
+export const Reversed: Story = {
+  args: { reversed: true },
+  parameters: { backgrounds: { default: "Purple 700" } },
+}
```
gyfchong, line 94: Someone fix this please, no idea why it doesn't work...

packages/components/src/FilterMultiSelect/subcomponents/SearchInput/SearchInput.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useMemo } from "react"
 import { v4 } from "uuid"
```
mcwinter07, line 2: Should we replace this with React's useId?

packages/components/src/Input/Input/Input.module.scss
```diff
@@ -0,0 +1,359 @@
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "../../../styles/form-variables";
+@import "../../../styles/forms";
+
+$input-height: 48px;
+$input-base-padding-horizontal: $spacing-sm;
```
mcwinter07, line 8: Could we also replace the `sm` & `md` spacing tokens with the number spacing tokens

packages/components/src/Input/_docs/Input.stories.tsx
mcwinter07: I think you need to use a `DocStory` instead of a `Canvas` to get the reversed background

packages/components/src/Input/Input/Input.tsx
```diff
@@ -0,0 +1,72 @@
...
+        classNameOverride,
+        reversed ? styles.reversed : styles.default,
+        disabled && styles.disabled
+      )}
+      {...restProps}
+    />
+
+    {/* Inputs aren't able to have pseudo elements like ::after on them,
+          so we have to create an element ourselves for the focus ring */}
+    <div className={styles.focusRing} />
```
mcwinter07, line 64: Noticed a spacing issue on the focus ring for the input, could be that the offset hasn't properly calculated itself?

Screenshot 2023-10-02 at 11 17 19 am

Worth noting also that this empty `div`  feels like a bit of an older pattern since `focus-within` wasn't widely supported a couple of years back - If we move this to a pseudo after element on the `wrapper` this may clean up some of the sibling selectors in the stylesheet and make it cleaner the offset?

packages/components/src/Input/Input/_docs/Input.stickersheet.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        startIconAdornment={
+          <WritingIcon
+            role="presentation"
+            data-sb-a11y-color-contrast-disable={
+              props.disabled ? "true" : "false"
+            }
+          />
+        }
+        endIconAdornment={
+          <MeatballsIcon
```
mcwinter07, line 46: I'm wondering if we should change out the icon on the success stories to the tick svg? Otherwise there's no discernible difference from default and success states (though you could certainly argue that that would be be the point)

mcwinter07: We could potentially add a line and Canvas to provide more guidance in the MDX if we want to keep our chromatic tests with the same icons 🤔

packages/components/src/Input/InputSearch/_docs/InputSearch.stickersheet.stories.tsx
```diff
@@ -0,0 +1,177 @@
...
+      },
+    },
+  },
+} satisfies Meta
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed }) => (
+    <StickerSheet isReversed={isReversed}>
+      <StickerSheet.Header
+        headings={["Default", "Disabled", "Hover", "Focus"]}
```
mcwinter07, line 32: I feel like we could also have a "filled" or "active" column  with a `value` passed in so we can have a snapshot of the search input with a clear button

packages/components/src/Input/InputSearch/_docs/InputSearch.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const Playground: Story = {
+  parameters: {
+    docs: {
+      canvas: {
+        sourceState: "shown",
+      },
+    },
+  },
+}
+
```
mcwinter07, line 38: Did we want to give this local state so a user can type into this or save that for the `SearchField` component?

packages/components/src/FilterMultiSelect/subcomponents/SearchInput/SearchInput.tsx
gyfchong: Yes, but not in this PR 😄 
https://cultureamp.atlassian.net/browse/KDS-1673

packages/components/src/Input/Input/Input.module.scss
gyfchong: We could, but not the goal of these migrations.

packages/components/src/Input/Input/_docs/Input.stickersheet.stories.tsx
gyfchong: Not worth it for these stickersheets. Good callout that the styling isn't discernible enough, but we should consider whether or not to enforce a success styling to begin with, cause the design docs don't even mention "success" patterns. Which I'm assuming because forms tend to move on after being successful, and in other cases, we want errors to stand out. So I'd question the necessity of this variant anyway 😅

packages/components/src/Input/Input/Input.tsx
gyfchong: Let's log that as a ticket to solve separately. The issue here is resolved by not using `grid` or `flex` for alignment.

packages/components/src/Input/Input/_docs/Input.stickersheet.stories.tsx
mcwinter07: We seem to have a few of these across our form components that don't change styles. I would guess that the "success" variant changed the border to green in the past

gyfchong: Logged :D https://cultureamp.atlassian.net/browse/KDS-1821

packages/components/src/Input/Input/Input.tsx
gyfchong: https://cultureamp.atlassian.net/browse/KDS-1822

packages/components/src/FilterMultiSelect/subcomponents/SearchInput/SearchInput.tsx
mcwinter07: nice

mcwinter07 (dismissed):
Looks good to migrate 🚢 

I did notice that the `InputRange` has some text rendering that's marked as visually hidden, there is a comments in the stylesheet about it:

```
// If the .visually-hidden class is applied to natively focusable elements
// (such as a, button, input, etc) they must become visible when they receive
// keyboard focus. Otherwise, a sighted keyboard user would have to try and
// figure out where their visible focus indicator had gone to.
```

This doesn't feel like it part of migration - though I wanted to flag it in case there was any additional context to why its rendering / if its an issue that the text not our body typography

mcwinter07 (approved):
Nice ✅

PR #4124, opened 2023-10-02
Kds 1754 migrate guidance block scene to kaio

JakePitman:
## Why
Part of KAIO migrations

## What
- Migrates Scene


HeartSquared:
Already migrated

PR #4141, opened 2023-10-06
feat(icon): mirror directional icons in RTL

HeartSquared:
## Why

POC mirroring directional icons which should point in the opposite direction in RTL.

image

## What

- Add `RTLMirror` helper component
- Update `ArrowBackward` and `ArrowForward` to mirror in RTL

## Note

There is a slight conflict with icons which need this functionality, in that our `update-icons` script will rewrite and undo the implementation if there is a matching svg asset.


HeartSquared:
New Icon has RTL support now 😄

PR #4154, opened 2023-10-09
Fix support for tree-shaking

gyfchong:
## Why
KAIO's bundle does not export components as modules, causing un-treeshakeable code.

## What
Fix support for tree shaking by:
- Adding `babel-plugin-pure-static-props` plugin to Rollup config
- Adding `@babel/pugin-transform-react-pure-annotations` plugin to Rollup config
- Updating components to be tree-shakeable, more detail on [Confluence](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/3320647009/Tree+Shaking)
- Side affects: false

Addtionally:
- Replaced `esbuild` with `babel` to streamline tooling
- Added a ci check for tree shaking via [Agadoo](https://github.com/Rich-Harris/agadoo)
- Moved `react-intl` to peerDep
- Added `tslib` to dependencies
- ❗ Removed `ThemeManager` in favour of `theme` prop on `KaizenProvider` (simplified to remove unused code)


packages/components/package.json
```diff
@@ -79,37 +79,45 @@
...
     "query-string": "^8.1.0",
+    "react-intl": "^6.4.7",
     "rollup": "^3.28.1",
     "rollup-plugin-dts": "^6.0.2",
-    "rollup-plugin-esbuild": "^5.0.0",
+    "rollup-plugin-esbuild": "^6.1.0",
     "rollup-plugin-ignore": "^1.0.10",
     "rollup-plugin-node-externals": "^6.1.1",
     "rollup-plugin-postcss": "^4.0.2",
+    "rollup-plugin-pure": "^0.2.1",
```
gyfchong, line 108: Don't need this

turbo.json
```diff
@@ -22,6 +26,10 @@
       "inputs": ["src/**"],
       "outputs": ["dist/**"]
     },
+    "@kaizen/components#test:treeshake": {
+      "dependsOn": ["@kaizen/components#build"],
+      "inputs": ["dist/**"]
+    },
```
gyfchong, line 32: Oh, this is a weird requirement

HeartSquared: Yeah agreed. No idea why it didn't work otherwise though. It wouldn't stop caching even when I put `cache: false` 😅

HeartSquared (approved):

packages/components/package.json
```diff
@@ -70,46 +72,47 @@
...
     "rollup-plugin-dts": "^6.0.2",
-    "rollup-plugin-esbuild": "^5.0.0",
     "rollup-plugin-ignore": "^1.0.10",
     "rollup-plugin-node-externals": "^6.1.1",
     "rollup-plugin-postcss": "^4.0.2",
     "sass": "^1.66.1",
     "serialize-query-params": "^2.0.2",
     "svgo": "^3.0.2",
     "ts-jest": "^29.1.1",
+    "tslib": "^2.6.2",
```
HeartSquared, line 108: Move this to a `peerDep`

gyfchong: Decided to leave as dep, as this gets bundled into the final dist. Could be revisited if there's an issue.

packages/components/src/KaizenProvider/ThemeProvider/ThemeProvider.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+      color: {
+        ...heartTheme.color,
+        white: "#000",
+      },
+    }
+
+    render(<ThemeProvider theme={customTheme}>hello</ThemeProvider>)
+
+    await waitFor(() => {
+      assertThemeIsActive(customTheme, document.getElementById("theme-root"))
```
HeartSquared, line 43: Worth adding an explicit expect just for the new colour (maybe make it a silly key instead of overriding an existing one).

gyfchong: Couldn't be bothered having to create a new type for it 😂 Feel free though

HeartSquared (approved):

PR #4159, opened 2023-10-13
Add missing prop descriptions for Radio

sentience:
## Why

We just fixed an accessibility bug in performance-review-cycles-ui that was caused by an engineer who didn't understand the purpose of the `name` attribute on radio buttons (and therefore the `name` prop on `Radio` components):

- https://github.com/cultureamp/performance-review-cycles-ui/pull/788
- https://cultureamp.atlassian.net/browse/SR2-873

## What

To help avoid similar bugs in future, this adds documentation for the `name` prop (along with the `value` prop for good measure).

I couldn't get Storybook to infer these descriptions from my attempts to add JSDoc comments to the component prop types, so I've added them manually to the Story meta objects. Working JSDocs would be preferable, though, as it would provide helpful in-editor descriptions for these props too.


.changeset/tidy-turkeys-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Improve Radio/RadioField docs
```
gyfchong, line 5: Can be left empty since docs deploy independently of the versioning system

sentience: Can or must? Do you need me to remove this? Should I remove the changeset entirely? (In which case, how do I satisfy the changesets-exists build step?)

gyfchong: Sorry I said "Can" cause doing this doesn't technically break anything, but we don't want a version just for docs. You can use `yarn changeset --empty` to satisfy the ci step.

packages/components/src/Radio/Radio/_docs/Radio.stories.tsx
```diff
@@ -6,8 +6,20 @@ const meta = {
   component: Radio,
   args: {
     id: "radio",
-    name: "Radio",
-    value: "Radio",
+    name: "name",
+    value: "value",
+  },
+  argTypes: {
+    name: {
```
HeartSquared, line 13: I was able to use JSdocs by declaring the prop to be the same as the HTMLAttribute type signature, then putting the comment above. Storybook seemed to infer fine. Worth a try again, if that's what you aimed to achieve (sometimes SB caches, so that also might be why it didn't seem to work).

sentience: Thanks, I'll have another go! If it still doesn't work, can I hit you up for a quick pair?

HeartSquared: Yep no problem!

.changeset/tidy-turkeys-crash.md
sentience: @gyfchong Done, but I wonder if the fact that I have moved the docs to JSDoc changes this? The package now includes updated TypeScript code completion feedback; should that be a patch release?

gyfchong: Yep `patch` makes sense now

gyfchong (approved):

PR #4175, opened 2023-10-16
add hook to apply css vars to html element

gyfchong:
## Why
Previously thought applying CSS vars to a consistent node would work for all cases, however elements whch sit outside the root node (eg. portals) will not have access to the variables.

## What
Introduce a hook to load the variables onto the `html` element (as per original functionality)

### DIFF
--- a/.changeset/shiny-hotels-enjoy.md
+++ b/.changeset/shiny-hotels-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix asignment of CSS vars to load onto the `html` element.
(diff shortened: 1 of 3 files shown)


.changeset/shiny-hotels-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix asignment of CSS vars to load onto the `html` element.
```
HeartSquared, line 5:
```suggestion
Fix assignment of CSS vars to load onto the `html` element.
```

HeartSquared (approved):

gyfchong:
No needed, false alarm

PR #4176, opened 2023-10-16
feat: migrate DatePicker and DateRangePicker

HeartSquared:
## Why

Migration

## What

Migrate `DatePicker` and `DateRangePicker` and associated subcomponents.
- Remove duplication from FilterD(R)P


packages/components/src/Calendar/_docs/CalendarSingle.mdx
```diff
@@ -0,0 +1,28 @@
+import { Canvas, Controls, Meta } from "@storybook/blocks"
```
gyfchong, line 1: I've been co-locating the docs within their respective folders, so you'd have `src/Calendar/CalendarSingle/_docs/CalendarSingle.mdx`

packages/components/src/Calendar/Calendar.module.scss
```diff
@@ -0,0 +1,214 @@
+@import "~@kaizen/design-tokens/sass/typography";
```
gyfchong, line 1: What imports this `Calendar.module.scss` file?

packages/components/src/FilterDatePicker/subcomponents/FilterDatePickerField/FilterDatePickerField.tsx
```diff
@@ -78,7 +83,7 @@ export const FilterDatePickerField = ({
   classNameOverride,
   ...restProps
 }: FilterDatePickerFieldProps): JSX.Element => {
-  const [id] = useState<string>(propsId || v4())
+  const [id] = useState<string>(propsId || useId())
```
gyfchong, line 86: Should this be `const id = propsId ?? useId()`

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState } from "react"
 import { render, waitFor } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { FilterButton } from "~components/FilterButton"
-import { DateRange } from "~types/DatePicker"
+import { DateRange } from "~types/date-controls"
```
gyfchong, line 5: Should this type come from `DateRangePicker` or `CalendarRange`?

packages/components/src/Calendar/_docs/CalendarSingle.mdx
HeartSquared: Ah yeah fair. It's because I was tossing up between a single `Calendars` doc or splitting them. Will move.

packages/components/src/Calendar/Calendar.module.scss
HeartSquared: `calendarClassNames.ts` 🫠 Terrible isn't it haha. It's base styles that apply to both types of Calendars.

packages/components/src/FilterDatePicker/subcomponents/FilterDatePickerField/FilterDatePickerField.tsx
HeartSquared: Nope apparently it should be this (but yes thanks good catch)!
```
const reactId = useId()
const id = propsId ?? reactId
```

packages/components/src/FilterDateRangePicker/FilterDateRangePicker.spec.tsx
HeartSquared: Hmm... Yeah I really struggled with the location for the shared types. I wanted it in `DateRangePicker`, but that's currently a legacy version and I didn't want it to house non-legacy types (in case they deviate and it breaks `FilterDRP`).
But putting it in `FilterDRP` seemed wrong too, given it's meant to be somewhat of a wrapper only.

I didn't want to put it in `CalendarRange` because it doesn't actually use `DateRange` itself (well, it does, but it's in the spread props).

I put it in `date-controls/ReactDayPicker` along with `DateInterval` (exposed so consumers can use it to type against the disabled days abstraction; although now that I think about it, they can also just drill into the larger type so this probably isn't necessary), but open to changing it to wherever makes the most sense (I could move them all into Calendar maybe? Since DayPicker is the main reason for their existence).

packages/components/src/Calendar/Calendar.module.scss
HeartSquared: Should I rename this to `CalendarBase.module.scss` instead? Would that be clearer?

gyfchong: Probably align it to what imports it lol

packages/components/src/DateInput/DateInput/DateInput.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react"
+import classnames from "classnames"
+import { Input, InputProps } from "~components/Input"
+import { Label } from "~components/Label"
+import { isRefObject } from "~utils/isRefObject"
+import styles from "./DateInput.module.scss"
+
+type OmittedInputProps = "reversed" | "type" | "inputRef"
+
+export interface DateInputProps extends Omit<InputProps, OmittedInputProps> {
```
HeartSquared, line 10: `type`

packages/components/src/DateInput/DateInputDescription/DateInputDescription.tsx
```diff
@@ -19,7 +19,7 @@ export const DateInputDescription = ({
     React.isValidElement(description)
 
   const translatedInputFormatMessage = formatMessage({
-    id: "filterDateRangePicker.inputFormat",
+    id: "dateInputDescription.inputFormat",
```
HeartSquared, line 22: Renamed because moved out of FilterDRP. Also renamed within all the locale files.

packages/components/src/DatePicker/_docs/controls/weekStartsOnControls.ts
```diff
@@ -0,0 +1,28 @@
+import { DayOfWeek } from "~components/Calendar"
+
+export const weekStartsOnControls = {
```
HeartSquared, line 3: Move to Calendar and also implement in Calendar docs

packages/components/src/DateRangePicker/utils/formatDateRangeValue.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { DateRange } from "react-day-picker"
```
HeartSquared, line 1: Update import to Calendar

packages/components/src/DateRangePicker/utils/formatDateRangeValue.ts
```diff
@@ -0,0 +1,36 @@
+import { DateRange } from "react-day-picker"
```
HeartSquared, line 1: Update import to Calendar

packages/components/src/FilterDateRangePicker/subcomponents/DateRangeInputField/DateRangeInputField.module.scss
```diff
@@ -9,7 +9,7 @@ $input-disabled-opacity: 0.3;
   border: none;
 }
 
-.inputStartDate {
+.inputStartDate input {
```
HeartSquared, line 12: Moved `classNameOverride` in DateInput a level up to the wrapper to allow for stickersheet pseudo state tests

packages/components/src/Calendar/CalendarRange/_docs/CalendarRange.stickersheet.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+      hover: '[data-sb-pseudo-styles="hover"]',
+      focus: '[data-sb-pseudo-styles="focus"]',
+      focusVisible: '[data-sb-pseudo-styles="focus"]',
+    },
+  },
+  play: ({ canvasElement }): void => {
+    applyStickerSheetStyles(canvasElement)
+  },
+}
+const applyStickerSheetStyles = (canvasElement: HTMLElement): void => {
```
gyfchong, line 96: This is gnarly lol

packages/components/src/DateInput/DateInputWithIconButton/DateInputWithIconButton.module.scss
```diff
@@ -0,0 +1,52 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "~@kaizen/design-tokens/sass/border";
+@import "~@kaizen/component-library/styles/animation";
```
gyfchong, line 4: This is not available in KAIO

```suggestion
@import "../../../styles/animation";
```

packages/components/src/Calendar/CalendarRange/_docs/CalendarRange.stickersheet.stories.tsx
HeartSquared: Yeah it's pretty gross since we want to drill and target an element within elements.
Open to other suggestions if you can think of any!

packages/components/src/DateInput/DateInputWithIconButton/DateInputWithIconButton.module.scss
HeartSquared: Nice catch. Completely missed it.

packages/components/src/DatePicker/subcomponents/DateInputField/DateInputField.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+const DateInputFieldWrapper = (
+  props: Partial<DateInputFieldProps>
+): JSX.Element => <DateInputField {...defaultProps} {...props} />
+
+describe("<DateInputField />", () => {
+  describe("Input", () => {
+    it("associates the description with the input", () => {
+      render(<DateInputFieldWrapper />)
+      expect(
+        screen.getByRole("textbox", {
```
gyfchong, line 27: Why use `screen` here instead of getting it from the `render`? I've been grabbing from the `render`...

HeartSquared: Because I didn't write this 😄 I always grab from `render` too.
I did look into it, and it seems that `screen` is more of an alias which makes transitions from other testing libraries (eg. Cypress) a little easier.

Did a tiny bit of testing with an existing test:
- Using `screen` - Cached time `1.917 s`
- Using `render` - Cached time `1.8 s`

It seems `render` is consistently slightly more performant (which makes sense, given it can skip the step of searching for the container). Happy to change it.

gyfchong (approved):

gyfchong (approved):

PR #4174, opened 2023-10-16
KDS-1861-migrate-GenericModal

gyfchong:
Just GenericModal for now

Note: there are no stickersheets as this component isn't quite structured in a way where I could display all variations on a single story. Will keep this in mind for the rebuild

PR #4177, opened 2023-10-17
Kds 1757 migrate confirmationmodal

gyfchong:
Confirmation modal.

To be merged AFTER #4174


packages/components/src/Modal/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+  mood,
+  title,
+  onConfirm,
+  onAfterLeave,
+  confirmLabel = "Confirm",
+  dismissLabel = "Cancel",
+  confirmWorking,
+  automationId,
+  children,
+  ...props
```
HeartSquared, line 104: This appears to only contain `onDismiss`. Maybe just destructure it to reduce confusion (since the spread doesn't go anywhere - unless you want to add extra attributes, then it has a different purpose).

packages/components/src/Modal/ConfirmationModal/_docs/ConfirmationModal.mdx
```diff
@@ -0,0 +1,45 @@
...
+The Confirmation Modal confirms the end result of an interaction such as Positive, Informative, Cautionary or Negative.
+
+<Canvas of={ConfirmationModalStories.Playground} />
+<Controls of={ConfirmationModalStories.Playground} />
+
+## API
+
+### Prominant
+
+includes an animated spot illustration to help convey increased sentiment to the user.
```
HeartSquared, line 34:
```suggestion
Includes an animated spot illustration to help convey increased sentiment to the user.
```

packages/components/src/Modal/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+   */
+  mood: Mood
+  title: string
+  onConfirm?: () => void
+  onDismiss: () => void
+  onAfterLeave?: () => void
+  confirmLabel?: string
+  dismissLabel?: string
+  confirmWorking?: { label: string; labelHidden?: boolean }
+  automationId?: string
```
HeartSquared, line 46: 🔥

packages/components/src/Modal/ConfirmationModal/_docs/ConfirmationModal.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+      </button>
+      <ConfirmationModal
+        {...args}
+        isOpen={isOpen}
+        onConfirm={handleClose}
+        onDismiss={handleClose}
+      />
+    </>
+  )
+}
```
HeartSquared, line 30: If you make this into a template Story (have this as the `render`, then spread the whole Story into the other Stories), you won't need to type `args` nor the return type, and it will have intellisense (based on the meta) that tells you if you're missing args.

packages/components/src/Modal/ConfirmationModal/ConfirmationModal.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+    {...props}
+  >
+    {children}
+  </ConfirmationModal>
+)
+
+describe("<ConfirmationModal />", () => {
+  beforeEach(() => {
+    mockMatchMedia()
+  })
```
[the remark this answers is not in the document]
gyfchong: Hm, not enitrely sure how to clean this object from the window. `delete` seems like the obvious choice, by ts doesn't like it. This function sets matchers to false and if anything needs true, then they need to set it specifically, and then before the next one, it sets back to false. So I'm not sure it needs clean up.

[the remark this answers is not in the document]
HeartSquared: Hmm...but it declares multiple properties, and it doesn't undo it between tests. It's a little brittle if there's no clean up.

I found this which should be safer https://stackoverflow.com/a/56999581
```
let windowSpy;

beforeEach(() => {
  windowSpy = jest.spyOn(window, "window", "get");
});

afterEach(() => {
  windowSpy.mockRestore();
});
```

Although I'm not sure if it's as straightforward to replace as it looks. Maybe we can fix it later.

packages/components/src/Modal/ConfirmationModal/_docs/ConfirmationModal.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    chromatic: {
+      disable: false,
+      delay: 400, // match MODAL_TRANSITION_TIMEOUT in modals + 50ms
+      pauseAnimationAtEnd: true,
+    },
+  },
+  decorators: [HeightDecorator],
+}
+
+const ConfirmationModalTemplate: Story = {
```
gyfchong, line 36: @HeartSquared like this...?

packages/components/src/Modal/ConfirmationModal/ConfirmationModal.spec.tsx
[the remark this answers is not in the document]
gyfchong: Let's kick this down the road to the modal rebuild, I think matchMedia is a bit flawed in itself 😄

packages/components/src/Modal/ConfirmationModal/_docs/ConfirmationModal.stories.tsx
HeartSquared: Yep! This works!

Unsure if it makes the `Playground` a little unclear now, but I suppose if you put `decorators: undefined` (since it's currently still in there) it might be explicit that it's not.

Or maybe it's better to not have the Chromatic settings in the Template, so it's an explicit choice per story as opposed to a default 🤔 
Probably makes this change seem no different to before, but I've definitely had some where I put extra Story properties (eg. `play`) in so it's just a tad more scalable this way.

gyfchong: Probably leave it this way for now, I think when we revisit this would be during a rebuild that will hopefully be engineered for stickersheets, and this will all go away 😄

HeartSquared (approved):

PR #4180, opened 2023-10-18
KDS-1868 add styles.css to sideeffects

gyfchong:
Fixes this same issue detailed here: https://github.com/ng-select/ng-select/issues/1717

Kaizen CSS was being tree-shaken by Webpack as it is part of a package marked with no side effects and our styles are imported via JS yet not directly used. This fix adds our styles to the side effects parameter to prevent webpack from shaking it out.

### DIFF
--- a/.changeset/afraid-knives-approve.md
+++ b/.changeset/afraid-knives-approve.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fixes issue where styles.css was being tree-shaken by Webpack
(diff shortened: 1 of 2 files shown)


HeartSquared (approved):
PR description pls for records.

I wonder if it makes sense to be explicit about the file, or if we should glob `*.css` instead (assuming it works).

gyfchong:
> I wonder if it makes sense to be explicit about the file, or if we should glob *.css instead (assuming it works).

I don't want any other CSS files accidentally being included, despite there being no other CSS in the bundle, it's still a bit of a risk to me

PR #4184, opened 2023-10-19
Rename-modal-subcomponents

gyfchong:
## Why
Renamed components as part of migration but shouldn't have.

## What
Match naming scheme with kaizen-legacy


.changeset/strange-peas-confess.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
gyfchong, line 2: Empty cause nobody needs to know, and it'll go out with the next version anyway 😄

packages/components/src/Modal/GenericModal/subcomponents/ModalBody/ModalBody.tsx
```diff
@@ -1,17 +1,17 @@
...
+import styles from "./ModalBody.module.scss"
 
-export type ModalSectionProps = {
+export type ModalBodyProps = {
   inputEdit?: boolean
   children: React.ReactNode
 }
 
-export const ModalSection = ({
+export const ModalBody = ({
```
HeartSquared, line 10: This looks like a new rename 😛 I think `ModalSection` was fine (given the old name was `GenericModalSection`).

I'm not too fussed about it, except that it's kinda weird seeing `ModalHeader` and `ModalFooter` use `ModalBody` 😬

packages/components/src/Modal/GenericModal/index.ts
```diff
@@ -1,7 +1,7 @@
 export * from "./GenericModal"
 export * from "./context/ModalContext"
-export * from "./subcomponents/ModalDescription"
+export * from "./subcomponents/ModalAccessibleDescription"
 export * from "./subcomponents/ModalFooter"
 export * from "./subcomponents/ModalHeader"
-export * from "./subcomponents/ModalLabel"
-export * from "./subcomponents/ModalSection"
+export * from "./subcomponents/ModalAccessibleLabel"
+export * from "./subcomponents/ModalBody"
```
HeartSquared, line 7: Nit: Alphabet

HeartSquared (approved):

PR #4189, opened 2023-10-20
shifted styles back

gyfchong:
Just a cleanup of our style structure for Storybook. Doesn't help the `@layers` cause.


.stylelintignore
```diff
@@ -1,2 +1,3 @@
 packages/design-tokens/**/*.scss
 packages/components/locales/*
+storybook/styles/*
```
HeartSquared, line 3: A little iffy about ignoring all stylelint rules for the stylesheets just for probably one or two rules. Not too important though given it's local SB styles only, but might be worth revisiting later.

HeartSquared (approved):

PR #4206, opened 2023-10-24
feat: migrate TimeField

HeartSquared:
## What

Migrate `<TimeField/>` from `kaizen-legacy`.
- Installed deps:
	- "@internationalized/date"
	- "@react-aria/datepicker"
	- "@react-aria/i18n"
	- "@react-stately/datepicker"

## Why

Migration.


packages/components/src/TimeField/TimeField.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { useState } from "react"
+import { Time } from "@internationalized/date"
+import { useTimeField } from "@react-aria/datepicker"
+import { I18nProvider } from "@react-aria/i18n"
+import {
+  useTimeFieldState,
+  TimeFieldStateOptions,
+} from "@react-stately/datepicker"
+import classnames from "classnames"
+import { v4 } from "uuid"
```
HeartSquared, line 10: Update to `useId`

HeartSquared: Are you asking me not to do it now, or why not do it now?
If the former, then it can depend on if there's anything else to fix.
If the latter, I just missed it earlier.

gyfchong (approved):

PR #4208, opened 2023-10-25
Kds 1755/spot illustration migration

mcwinter07:
## Why
Part of the Great Migration

## What
-  Adds Spot Illustrations to KAIO
-  removes deprecated PositiveMale and PositiveFemale Spot illustations
  - The non-genered `Positive` has existed for a while now an is a suitable alternative for either
- Moves Base to subcomponents folder and Spot to a feature folder in preparation for Scene illustrations
  - I've opted for a similar folder structure to Notification
 
## Notes
- Since the illustration will be able to be destructured from the root @kaizen/component package, we may want to create a ticket update naming conventions to be more explicity for our users, ie:
```JSX
import { Cautionary} from "@kaizen/components"
// alt?
import { CautionarySpotIllustration} from "@kaizen/components"
```
- Currently no tests since the differences are largely visually and can be best assessed by chromatic. We'll probably need more tests later with Scene given it uses Lotti player


packages/components/src/Illustration/Spot/_docs/Spot.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+### Enabled Aspect Ratio
+
+When enabled this wraps the image in a `figure` tag and set the aspect ratio to 1/1.
+
+<Canvas of={SpotIllustrationStories.EnabledAspectRatio} />
+
+When not enabled, the `img` tag will attempt to fill a width of 100%.
+
+## All Spot illustrations
```
mcwinter07, line 41: I've included the stickersheet that has a list of the illustrations since our consumers would not be able to see the hidden stories on prod.

packages/components/src/Illustration/Spot/Spot.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { Base, BaseProps } from "../subcomponents/Base"
+
+export type SpotProps = Pick<BaseProps, "alt" | "classNameOverride"> & {
+  enableAspectRatio?: boolean
+}
+
+const SPOT_ILLUSTRATION_BASE_PATH = "illustrations/heart/spot/"
+const createSpotIllustration =
+  (fileName: string) =>
+  // eslint-disable-next-line react/display-name
```
mcwinter07, line 11: I feel like there might be a better way here to give these display names instead of just JSX elements, but I decided not to fall down the refactor rabbit hole - best this could be looked at when we revisit the names of these

packages/components/src/Illustration/Spot/_docs/Spot.stickersheet.stories.tsx
```diff
@@ -0,0 +1,781 @@
...
+  parameters: {
+    chromatic: { disable: false },
+    controls: { disable: true },
+  },
+} satisfies Meta
+
+const engagementSpots = [
+  {
+    Component: BenefitsSurvey,
+    name: "BenefitsSurvey",
```
mcwinter07, line 148: Originally these had more human readable names, but that doesn't help our consumer know what they will actually be exported as. I was thinking we wouldn't need this and could just use the displayName but that's not actually set for these components so I just when with the more verbose approach of remove the spaces and making sure they're in pascal case / align with the component name

packages/components/src/Illustration/Spot/_docs/Spot.mdx
gyfchong: I think just to reiterate, put an example of the import statement before the canvas

gyfchong (dismissed):
Just check your chromatic snapshots

packages/components/src/Illustration/Spot/_docs/Spot.mdx
mcwinter07: Good idea, I'll make a quick update to include that example

gyfchong (approved):

PR #4210, opened 2023-10-25
Migrate Tooltip component from kaizen-legacy

dougmacknz:
Notes:
* Got rid of the deprecated `inline` prop because I couldn't see any usages of it across our codebases.
* Added a bunch of docs in the MDX about accessibility
* Got rid of the 'without animation delay' story because I figure that can be achieved easily with the playground version.


packages/components/src/Tooltip/_docs/Tooltip.stickersheet.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+  ...StickerSheetTemplate,
+  name: "Sticker Sheet (Default)",
+}
+
+export const StickerSheetRTL: StickerSheetStory = {
+  ...StickerSheetTemplate,
+  name: "Sticker Sheet (RTL)",
+  parameters: {
+    /** @note: Only required if template has parameters, otherwise this spread can be removed */
+    ...StickerSheetTemplate.parameters,
```
HeartSquared, line 294: Clean up
```suggestion
```

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { Decorator, Meta, StoryFn, StoryObj } from "@storybook/react"
+import isChromatic from "chromatic/isChromatic"
+import { Tag } from "../../Tag"
+import { Button, InformationIcon, Text } from "../../index"
+import { Tooltip } from "../index"
+
+const openTooltipInChromatic: Decorator = (Story, { args }) => {
+  if (isChromatic()) args.isInitiallyVisible = true
+  return <Story />
+}
```
HeartSquared, line 11: This isn't necessary here since it's only for stories sent to Chromatic, which have now moved to stickersheet.

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+  },
+  render: args => (
+    <div
+      style={{ marginTop: "100px", display: "flex", justifyContent: "center" }}
+    >
+      <Tooltip {...args}>
+        <Button label="Hover or focus me" />
+      </Tooltip>
+    </div>
+  ),
```
HeartSquared, line 42: Optional clean up: Move the div to a decorator (the code snippet will be a bit cleaner for the consumers)

packages/components/src/Tooltip/utils/isSemanticElement.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react"
+import { Button, IconButton } from "@kaizen/button"
+import { Icon } from "@kaizen/component-library"
+import ArrowIcon from "@kaizen/component-library/icons/arrow-right.icon.svg"
```
HeartSquared, line 4: We don't want to bring `@kaizen/component-library` into KAIO, so best to refactor these usages.

HeartSquared (commented):
Can't comment on it, but there's a random file committed named `variables`

packages/components/src/Tooltip/_docs/Tooltip.mdx
```diff
@@ -0,0 +1,67 @@
...
+<!-- (in a portal) -->
+<div id="tooltip-id">Opens in a new tab</div>
+```
+
+Resulting in a screen reader announcement like:
+`Go somewhere, link, Opens in a new tab`
+
+Again, try to keep the contents of your children as simple as possible, ideally only a single natively focusable element like the example above.
+
+When your direct child is not semantic, you will get a console warning.
```
mcwinter07, line 67: I'd recommend just adding an example of the warning they might expect to see, in case they want to sift through the console, ie:
`<Tooltip /> is not directly wrapping a semantic element, screen reader users will not be able to access the tooltip info.`

packages/components/src/Tooltip/utils/isSemanticElement.spec.tsx
gyfchong: `<ArrowRightIcon />`

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+          justifyContent: "center",
+        }}
+      >
+        <Story />
+      </div>
+    ),
+  ],
+}
+
+export const OverflowScroll: StoryFn<typeof Tooltip> = props => (
```
HeartSquared, line 47:
```suggestion
export const OverflowScroll: Story = { render: props => (
```

packages/components/src/Tooltip/subcomponents/AppearanceAnim.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import { render } from "@testing-library/react"
+import useDebounce from "use-debounce"
+import { AnimationProvider } from "./AppearanceAnim"
+import "@testing-library/jest-dom"
```
HeartSquared, line 5: Nit: Pretty sure it's in the global config
```suggestion
```

packages/components/src/Tooltip/utils/isSemanticElement.spec.tsx
```diff
@@ -0,0 +1,66 @@
+import React from "react"
+import { ArrowRightIcon, Button, IconButton } from "../../index"
```
HeartSquared, line 2: Haha I don't think we've ever done this, but I suppose it's okay? Alt:
```suggestion
import { Button, IconButton } from "~components/Button"
import { ArrowRightIcon } from "~components/Icon"
```

dougmacknz: I like it this way but also don't feel strongly. @gyfchong any opinions?

gyfchong: Hmm I think stick with the aliases, incase we move the files around, I don't want to have to find and update individual paths

HeartSquared (approved):

PR #4212, opened 2023-10-26
feat: migrate Loading components

HeartSquared:
## Why

Migration.

## What

- Migrate `LoadingGraphic`, `LoadingHeading`, `LoadingInput`, `LoadingParagraph`
	- No stickersheet created for `LoadingInput` as it has no variants, so just sending the default and reversed stories
	- `LoadingParagraph` has an RTL stickersheet due to having RTL styles (unlike the others)
- Storybook:
	- Add global decorator for reversed background
	- Update plop template to include background for Reversed stories
	- Export `LinkTo` component from `~storybook/components`


.changeset/rotten-kings-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Migrate `LoadingGraphic`, `LoadingHeading`, `LoadingInput`, `LoadingParagraph` from `@kaizen/legacy`.
```
gyfchong, line 5:
```suggestion
Migrate `LoadingGraphic`, `LoadingHeading`, `LoadingInput`, `LoadingParagraph` from `kaizen-legacy`.
```

There's no `@kaizen/legacy` package

packages/components/src/Loading/LoadingParagraph/_docs/LoadingParagraph.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const MultiColumnExample: Story = {
+  render: () => (
+    <>
+      <div>
+        <LoadingParagraph isInline width={30} />
+        <LoadingParagraph isInline width={30} />
+        <LoadingParagraph isInline width={30} />
+      </div>
+      <div style={{ display: "grid", gridTemplateColumns: "1fr 1fr 1fr" }}>
```
gyfchong, line 61: No Tailwind? 😏

storybook/preview.tsx
```diff
@@ -31,15 +31,25 @@ const decorators = [
...
       if (document.body.getAttribute("dir") !== dir)
         document.body.setAttribute("dir", dir)
-    }, [props])
+    }, [context])
 
     return <Story />
   },
+  (Story, context) =>
+    context.args.isReversed &&
+    context.parameters.backgrounds.default !== "Purple 700" ? (
```
gyfchong, line 46: Seems a little strange to look out for specifically "Purple 700", what about an alias string? Otherwise it seems like you could do any colour.
```
parameters: { background: "reversed" }
```

.changeset/rotten-kings-eat.md
HeartSquared: Hahaha oops 😂 Thank goodness hey!

packages/components/src/Loading/LoadingParagraph/_docs/LoadingParagraph.stories.tsx
HeartSquared: Not from the copy/paste :P I can change it, but I also have no idea what the equivalent is hahaha.

storybook/preview.tsx
HeartSquared: Yeah... It's what's been set up tbh. We have a `backgrounds` array (see `storybook/backgrounds.ts`) and that's the key/name they've given it.
I also question whether anyone ever even uses any of the background colours offered (cos I don't). Didn't want to blow up the PR with a config refactor though, but happy to revisit later.

gyfchong: Oh it's existing, meh then lol

gyfchong (approved):
Assuming stickersheets are a-okay!

packages/components/src/Loading/LoadingParagraph/LoadingParagraph.tsx
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const LoadingParagraph = ({
+  isAnimated,
+  isCentred,
+  isReversed,
+  isInline,
+  noBottomMargin,
+  width,
+  isLink,
+  inheritBaseline,
```
mcwinter07, line 32: Do you think there's any value to have a story for this prop? As far as I can tell it just adds a `position: static` to the div, which I would assume it has by default? Just wander what this actually does 🤔

HeartSquared: You ask a good question. I also had no idea what the intent is. I temporarily made a story, then it looked absolutely no different. Given I had no explanation for its existence, I opted to not add a story so people hopefully don't ask about it 😬

packages/components/src/Loading/LoadingParagraph/_docs/LoadingParagraph.stories.tsx
mcwinter07: TW grid can get a little gnarly, but for that I think it might be:
```
grid grid-cols-[repeat(3,_1fr)]
```

packages/components/src/Loading/LoadingParagraph/LoadingParagraph.tsx
mcwinter07: 🤞 here's hoping! I would've assume in the name it would be to match to `vertical-align: baseline` but that's just a stab in the dark 🤷 It's probably not of any consequence in this migration but we may be able to remove it later if we find no one uses it

packages/components/src/Loading/LoadingParagraph/LoadingParagraph.module.scss
```diff
@@ -0,0 +1,50 @@
+@import "~@kaizen/design-tokens/sass/color";
+@import "~@kaizen/design-tokens/sass/spacing";
+@import "../mixins";
+
+.loadingParagraph {
+  @extend %loadingBase;
+
+  background-color: $color-gray-300;
+  height: calc(#{$spacing-md} * 0.625);
+  margin-bottom: calc(#{$spacing-md} * 0.5);
```
mcwinter07, line 10: optional knit:  we could change this to  `$spacing-12`. Unfortunately nothing to do with the gnarly one above 🥲 We could also swap out all the spacing-md's but not sure how much that would help the rando calcs

gyfchong (approved):

mcwinter07 (approved):

packages/components/src/Loading/LoadingParagraph/LoadingParagraph.module.scss
HeartSquared: Nice catch! I won't fix it now, but certainly worth updating the spacing tokens later. I'll put something in the Jira.

packages/components/src/Loading/LoadingParagraph/_docs/LoadingParagraph.stories.tsx
HeartSquared: That...honestly doesn't look any better 😂 More confusing if anything haha! Good to know though xD

PR #4213, opened 2023-10-26
KDS-1892-migrate-titleblock-to-kaio-1

gyfchong:
Extremely cut down version of the original docs. Not expecting anyone to actually need the docs given the latest designs have been phasing it out.


packages/components/src/TitleBlockZen/subcomponents/NavigationTabs.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react"
+import classnames from "classnames"
+import { TitleBlockVariant } from "../types"
+import { NON_REVERSED_VARIANTS } from "../utils"
+import styles from "./NavigationTabs.module.scss"
+
+export type CustomNavigationTabProps = Omit<NavigationTabProps, "render"> & {
+  className: string
```
HeartSquared, line 8: Oh no. Erm...should we flag it as deprecated? 😅 

Edit: Wait! It's used in `render` and is completely unusable by the consumers. So it's useless? Bah I guess just leave it. It doesn't make any sense anymore.

packages/components/src/TitleBlockZen/subcomponents/TitleBlockMenuItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export type TitleBlockMenuItemProps =
+  | ({
+      component: (props: CustomButtonProps) => JSX.Element
+    } & MenuItemProps)
+  | MenuItemProps
+
+export const TitleBlockMenuItem = (
+  props: TitleBlockMenuItemProps
+): JSX.Element | null => {
```
HeartSquared, line 16: Nit:
```suggestion
): JSX.Element => {
```

packages/components/src/TitleBlockZen/subcomponents/Toolbar.tsx
```diff
@@ -0,0 +1,46 @@
...
+          key={item.key}
+        >
+          {item.node}
+        </div>
+      ))}
+    </div>
+  )
+}
+
+export default Toolbar
```
HeartSquared, line 46: If it's internal, then nitpick, but if it's exported out, then we don't want default 😬

packages/components/src/TitleBlockZen/types.ts
```diff
@@ -0,0 +1,209 @@
...
+export type ButtonWithHrefNotOnClick = TitleBlockDistributiveOmit<
+  ButtonProps,
+  "onClick"
+>
+export type ButtonWithOnClickNotHref = TitleBlockDistributiveOmit<
+  TitleBlockButtonProps,
+  "href"
+>
+
+export type MenuGroup = {
```
HeartSquared, line 142: Fingers crossed this doesn't have any conflict with Menu types 🤞🏻 
Name spacing would be safest though 😉 (But I can see a lot of types without name spaces, so not sure if it would break too much, thus optional-ish for now)

packages/components/styles/_grid.scss
```diff
@@ -0,0 +1 @@
+$ca-grid: 1.5rem; // 24px @ default root font-size of 16px
```
HeartSquared, line 1: This is a legacy thing, so better to move it to `styles/legacy` o3o

packages/components/src/TitleBlockZen/_docs/TitleBlockZen.mdx
```diff
@@ -0,0 +1,23 @@
...
+<KaioNotification />
+
+<Installation
+  installCommand="yarn add @kaizen/components"
+  importStatement='import { TitleBlockZen } from "@kaizen/components"'
+/>
+
+## Overview
+
+<Canvas of={TitleBlockZenStories.Playground} />
```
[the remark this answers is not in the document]
gyfchong: Hell no haha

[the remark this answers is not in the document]
HeartSquared: 😂

PR #4214, opened 2023-10-26
Kds 1709 add remove button to selected options

JakePitman:
## Why
As per the designs, our MultiSelect tags need to be removable from the tags themselves, instead of just through the popover.

## What
Swaps `Tag` for `RemovableTag` in MultiSelect to represent selected options.


gyfchong:
Just a quick note, could you also write a test for the `onClick` functionality in `RemovableTag`? Cheers.

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/MultiSelectToggle.tsx
```diff
@@ -77,9 +79,14 @@ export const MultiSelectToggle = forwardRef<
                   // This stops the underlying toggle collapsing the popover when interactive with Tags
                   // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/click-events-have-key-events
                   <li key={value} onClick={e => e.stopPropagation()}>
-                    <Tag classNameOverride={styles.selectedItemTag}>
+                    <RemovableTag
+                      removeButtonProps={{
+                        ariaLabel: "Remove option",
+                        onClick: () => deselectOption(value),
```
HeartSquared, line 85: As discussed in [Slack](https://cultureamp.slack.com/archives/C05D8V4GALX/p1698373147766779), we're going to stick with the convention where props will follow the `onEvent` pattern, as it is more scalable and gives control to the consumer without restricting api (any refactors would not result in a huge change either).

So something on the line of `onRemoveOption` would suffice, and the deselecting functionality handled by the consuming component (which you've done).

HeartSquared (commented):
We'll want some tests too. There were some tests relating to keyboard navigation prepared which may also be relevant to be uncommented out now as well.

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -57,6 +57,12 @@ export const MultiSelect = ({
     {} as Record<MultiSelectOption["value"], MultiSelectOption>
   )
 
+  const deselectOption = (optionValue: MultiSelectOption["value"]): void => {
```
gyfchong, line 60:
```suggestion
  const deselectOption = (optionValue: MultiSelectOption["value"]): void => {
```

gyfchong (approved):
Just check off your Chromatic!

PR #4216, opened 2023-10-26
feat: migrate SplitButton

HeartSquared:
## Why

Migration.

## What

- Migrate `SplitButton` from `kaizen-legacy`.
- Replace usages of `@kaizen/draft-menu` with `~components/Menu`
	- Export Menu subcomponents which were missed


packages/components/src/SplitButton/_docs/SplitButton.stickersheet.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+    <MenuItem
+      // @todo: Add icon when Menu migrated
+      // icon={duplicateIcon}
+      label="Menu Item 2"
+    />
+  </MenuList>
+)
+
+export default {
+  title: "Components/SplitButton",
```
gyfchong, line 53: Mayyybe `Components/Button/SplitButton` ?

gyfchong (dismissed):

gyfchong (approved):

gyfchong:
Optional: if they didn't exist before than all good, but there doesn't seem to be any tests for when its open?

HeartSquared:
> Optional: if they didn't exist before than all good, but there doesn't seem to be any tests for when its open?

Yup didn't exist before. Also, the open state is actually controlled by Menu 😬

PR #4218, opened 2023-10-27
docs(Button): update format and fix icon docs

HeartSquared:
## Why

Button's `icon` docs were incorrect and confused consumers attempting to use it.

## What

- Update Button docs (moved to MDX to be able to add links to other stories)
- Split Icon story into two and update descriptions
- Update Storybook decorator to also check for the `reversed` prop (in addition to `isReversed`) for adding a purple background


gyfchong (approved):
Optional: Maybe also update `IconButton` with a similar description

gyfchong (approved):

PR #4246, opened 2023-10-30
Migrate Toast Notifications

mcwinter07:
## Why
Part of the Migration to KAIO

## What
- Replace `ReactDOM.render` with `createRoot`
- Remove deprecated `autohide` functionality
  - Add notes in documentation on a11y risk and reason for removal
- Update test description and cover a few more of scenarios
- Update documentation of API with more descriptions exported methods
- Split `ToastNotificationsListContainer` and `ToastNotificationsList` into two separate sub-components
  - Originally these were both exported from the same file - this is just a bit of house keeping to make it easier to read


packages/components/src/Notification/ToastNotification/ToastNotification.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type ToastNotificationProps = Omit<
+  ToastNotificationWithOptionals,
+  "message" | "persistent"
+> & {
+  /**
+   * Removes the dismiss trigger. functions the same as `persistent` in `addToastNotification`. If this is true you will need to manage the removal of notifications manually.
+   * @default false
+   */
+  hideCloseIcon?: boolean
```
mcwinter07, line 13: Functionally this does the same thing as persistent, so I've updated the type for the omit that from the ToastNotificationProps so there's less likelihood someone will pass in both into the component

packages/components/src/Notification/ToastNotification/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  title: string
+  message: React.ReactNode
+  onHide?: () => void
+  /**
+   * Removes the dismiss trigger. If this is true you will need to manage the removal of notifications manually.
+   * @default false
+   */
+  persistent?: boolean
+} & {
+  [key: `data-${string}`]: unknown
```
mcwinter07, line 17: The original component took automationId as a prop, which was being used to pass it in as a test id, which felt very out of date with modern practice. I figured pulling in data attribution in a `restProps` was a better way tp pass test ids into the component

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.spec.tsx
```diff
@@ -0,0 +1,145 @@
...
+describe("ToastNotificationsManager", () => {
+  beforeEach(() => {
+    act(() => {
+      clearToastNotifications()
+    })
+  })
+
+  describe("addToastNotification", () => {
+    it('creates a single "status" container for notifications', async () => {
+      await act(async () => {
```
mcwinter07, line 18: Noting here that since the change to `createRoot` this now throw a test warning about the use of act:

Screenshot 2023-10-30 at 12 16 52 pm

I haven't figured a way to fix this, though I suspect that it may require a bit more of a refactor. If there any thoughts, would welcome input :)

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+
+### clearToastNotifications
+
+Clears all notifications currently held in state.
+
+<Canvas of={ToastNotificationStories.ClearNotifications} />
+
+
+## Accessibility and autohide
```
mcwinter07, line 60: @dougmacknz wouldn't mind getting you're eye across the comments below to make sure I'm not off base with the reasoning and justification for removing autohide

packages/components/src/Notification/ToastNotification/types.ts
HeartSquared: We have a `DataAttributes` type which helps with this 😉 That being said, if we aren't extending HTML attributes and spreading the props, it might be safer in the meantime to just do what G has been doing and only add `data-testid?: string` as a prop to replace `automationId`

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.spec.tsx
HeartSquared: I tried to look into it, and tried multiple things and failed. 
- Wrap `expect`s in `await waitFor`
- Use `createPortal` within the util (fails because it only works inside a JSX.Element)
- Wrap `render` in `act` (sync and async)
- Wrap the util call within `ToastNotification` in a `useEffect`

I believe it is likely caused by `createRoot.render` being called conditionally (it waits for a portal to first be created) which has caused confusion in testing. We'll just need to live with this for now and hope it doesn't cause weird problems.

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+As `autohide` presents an [accessibility issue](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits.html), it is no longer a default feature of this component.
```
dougmacknz, line 62: > it is no longer a default feature

It's no longer a feature at all anymore right?

packages/components/src/Notification/ToastNotification/types.ts
```diff
@@ -0,0 +1,33 @@
...
+type Modify<T, R> = Omit<T, keyof R> & R
+
+export type ToastNotification = {
+  id: string
+  type: NotificationType
+  title: string
+  message: React.ReactNode
+  onHide?: () => void
+  /**
+   * Removes the dismiss trigger. If this is true you will need to manage the removal of notifications manually.
```
dougmacknz, line 12: I don't think this is what was envisioned for this prop (for people implementing to find their own way to hide it). I reckon it was probably just implemented because the other notifications had it as a way of saying this notification will never disappear. Which probably should have never been an option for toasts. I reckon we should just get rid of this

packages/components/src/Notification/ToastNotification/types.ts
```diff
@@ -0,0 +1,33 @@
+import { NotificationType } from "../types"
+
+type Modify<T, R> = Omit<T, keyof R> & R
+
+export type ToastNotification = {
```
HeartSquared, line 5: I'm wary this may cause a conflict with the type generated for the `ToastNotification` component, but it looks like it was already like that so I suppose we leave it lest we break it. Something to keep in mind though.

HeartSquared:
Is it just me, or are the Chromatic snapshots empty?

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
mcwinter07: @dougmacknz This is correct, the wording could be changed to clarify that

packages/components/src/Notification/ToastNotification/types.ts
mcwinter07: That would explain why the JSX version of the component has a `hideCloseIcon` prop as well 🤔

packages/components/src/Notification/ToastNotification/types.ts
mcwinter07: Felt a little ick, but hesitant to change the name since it was part of the original types

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.spec.tsx
mcwinter07: Thanks for validating, I tried a few of those so good to know it wasn't just me 😆 It does feel like this is linked to that. I figure we can keep it as is and come back to this as part of a much needed refactor when we have capacity / if it causes an issue 👍

packages/components/src/Notification/ToastNotification/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./ToastNotification"
+export * from "./subcomponents/ToastNotificationManager"
+export * from "./subcomponents/ToastNotificationsListContainer"
+export * from "./subcomponents/ToastNotificationsList"
```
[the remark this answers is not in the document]
mcwinter07: There was a few hits for the `ToastNotificationsList`, though could probably remove the list container

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
gyfchong: I would remove it from the docs completely. No need to explain something that doesn't exist!

gyfchong:
Chromatic is empty

packages/components/src/Notification/ToastNotification/types.ts
mcwinter07: So do think both `hideCloseIcon` and `persistent` should be removed entirely?

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
mcwinter07: I've reworded this to remove the old prop name and provide guidance if they need this kind of functionality. I think there are valid use cases so worth having here just to ensure that they're aware

packages/components/src/Notification/ToastNotification/types.ts
gyfchong: Let's clarify the purpose of Toast Notifications with the designers to make sure it aligns with their understanding of the feature set in relation to the other "toasts" we have. But doing anything about it should be out of scope for the migration.

mcwinter07: Sounds good, will leave as is for now

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+      />
+      <Button
+        label="Clear notifications"
+        onClick={() => clearToastNotifications()}
+      />
+    </>
+  ),
+}
+
+export const MultipleNotifications: Story = {
```
HeartSquared, line 144: This doesn't appear used (we don't expose individual stories on prod, only the Docs page). Was this meant to be the stickersheet?

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stickersheet.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+  parameters: {
+    chromatic: { disable: false, delay: 600 },
+    controls: { disable: true },
+  },
+} satisfies Meta
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: () => {
+    React.useEffect(() => {
+      addToastNotification({
```
HeartSquared, line 17: Hmm... this seems a little iffy. I was expecting a bunch of `<ToastNotification />`.
But just checked and it's how it was in legacy, so I suppose it's probably okay for now :s

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stickersheet.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+            semper odio vitae sem gravida rutrum. Praesent vel sapien eget eros
+            dictum luctus scelerisque eu nibh. Etiam ullamcorper lobortis
+            gravida. Suspendisse massa tortor, ultricies et ipsum at, iaculis
+            bibendum est.
+          </>
+        ),
+      })
+    }, [])
+
+    return <></>
```
HeartSquared, line 82: Unless it is important for it to return a fragment? Usually `null` is used for not rendering.

```suggestion
    return null
```

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
```diff
@@ -0,0 +1,67 @@
...
+### addToastNotification
+Creates an instance of the the Toast Notification manager and appends new notifications to the list.
+
+<Canvas of={ToastNotificationStories.CreateNotification} />
+
+If `addToastNotification` is called with an `id` that matches an existing notification, then it will update the content of that component.
+
+This will also be invoked if you use the `ToastNotification` component.
+
+<Canvas of={ToastNotificationStories.UpdateNotification} />
```
HeartSquared, line 43: Do we actually want to showcase this? It sounds a bit like an anti-pattern for someone to just replace the content of a Toast 😅

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
mcwinter07: Actually I should remove this, it's an old story but I don't think there's any valid use case for it that hasn't been tested in other stories

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stickersheet.stories.tsx
mcwinter07: Yeah I'll leave it as is this round. I reckon some of the jank can be fixed refactoring the ToastNotificationManager

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stickersheet.stories.tsx
mcwinter07: Null should be fine 👍

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.tsx
```diff
@@ -0,0 +1,139 @@
...
+ */
+const createToastNotificationManager = (): ToastNotificationApi => {
+  let setNotifications:
+    | React.Dispatch<React.SetStateAction<ToastNotification[]>>
+    | undefined
+  if (portal === undefined && typeof window !== "undefined") {
+    // automation and testid appear to be consumed in other repos, we'll need to refactor before we remove so they can be added in
+    portal = document.createElement("div")
+    portal.setAttribute(
+      "data-automation-id",
```
gyfchong, line 51: Are you replacing these with `data-testid`?

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.tsx
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Set default values for optional params
+ */
+const setDefaults = (
+  notificationMaybe: ToastNotificationWithOptionals
+): ToastNotification => {
+  const copy = { ...notificationMaybe }
+  if (typeof copy.id === "undefined") {
+    copy.id = v4()
+  }
```
gyfchong, line 34: Should swap this out for `useId()`

```
const reactId = useId()

if(...) { copy.id = reactId }
```

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationsList/ToastNotificationsList.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { HTMLAttributes } from "react"
+import { OverrideClassName } from "~types/OverrideClassName"
+import { GenericNotification } from "../../../subcomponents/GenericNotification"
```
gyfchong, line 3:
```suggestion
import { GenericNotification } from "~components/Notification/subcomponents/GenericNotification"
```

gyfchong: Feels like not quite a subcomponent given it gets used like this, can reorg later

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationsListContainer/ToastNotificationsListContainer.spec.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react"
+import { fireEvent } from "@testing-library/dom"
+import { render, act, waitFor, screen } from "@testing-library/react"
+import { ToastNotification } from "../../types"
+import { ToastNotificationsListContainer } from "."
```
gyfchong, line 5:
```suggestion
import { ToastNotificationsListContainer } from "./ToastNotificationsListContainer"
```
Better to be explicit I think

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.tsx
mcwinter07: my one concern is [this file](https://github.com/cultureamp/recognition-front-end/blob/d85ee88a09ad7c9f82d2e2f48d0674d6330ec504/apps/recognition-feed/src/templates/Admin/styles.scss#L42) that seems to use it as a selector. That said, doesn't feel like the best practice anyway and they could just change it to data-testid

gyfchong: A good catch, but I'll add the guidance for the removal of `data-automation-id` in the migration docs so that everyone is aware. So we should remove this.

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.tsx
HeartSquared: Note: Sounds good, might not work. `useId` is a hook, which must be called within a component. This file is not comprised of components, it's just a bunch of functions (this one in particular is just a util for setting defaults) :( Worth trying though.

gyfchong: Ah right, seems like you can't even render a Toast on the server side anyway? Ignore the comment lol

mcwinter07: Interesting 🤔 If I try to lift and shift this with `useId` its seems to break the component or at least the functions that are exported (ie: `addToastNotification`). While the component still renders when using the component version of `ToastNotification` all the function `addToastNotification` etc fail:

`Invalid hook call. Hooks can only be called inside of the body of a function component`

Starting to feel like we need to refactor the ToastNotifcationManager for this to work 🤔 Happy to hold on this to discuss at work review

mcwinter07: wüps, didn't see these comments before my schpeel 😆 I'll move these thoughts over to a ticket 😉

packages/components/src/Notification/ToastNotification/subcomponents/ToastNotificationManager/ToastNotificationManager.tsx
mcwinter07: Brill, will put the 🪓  to it!

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
mcwinter07: Tis' gone with the wind 🎐

HeartSquared (approved):

PR #4247, opened 2023-10-30
Migrate Tabs component from kaizen-legacy

dougmacknz:
Notes:
* Added classNameOverride and native props to each of the components so that I could replace Box usages out in the stories.
* Got rid of the `keyboardActivation` prop because I couldn't think of a good reason to need it, and no one has used it.


packages/components/src/Tabs/_docs/Tabs.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { Meta, StoryObj } from "@storybook/react"
+import { Button } from "~components/Button"
+import { Card } from "~components/Card"
+import { Text } from "~components/Text"
+import { Tabs, TabList, Tab, TabPanels, TabPanel } from "../index"
+
+const meta = {
+  title: "Components/Tabs",
+  component: Tabs,
+  args: {},
```
gyfchong, line 11: You can add the `children` here which should apply them to all stories

gyfchong (dismissed):
Just a nit to make your stories neater

packages/components/src/Tabs/_docs/Tabs.stories.tsx
dougmacknz: Nice good call

packages/components/src/Tabs/_docs/Tabs.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+    ),
+  },
+} satisfies Meta<typeof Tabs>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Playground: Story = {
+  parameters: {
```
gyfchong, line 41: Add `chromatic: { disable: false }` here

gyfchong (approved):

PR #4252, opened 2023-10-31
fix: update CloseIcon to be able to inherit color

HeartSquared:
## Why

`CloseIcon` does not inherit colour.

## What

- Update stickersheet for Icon to show colour inheritance
- Update icon script to also replace Purple 800 hex (as UI Kit stores them as such - easier moving forward)
- Regenerate `CloseIcon`
	- Note: There are some other icons which are broken, but will fix them separately as they have other issues (eg. doesn't exist in UI Kit or has a complex svg)

### DIFF
--- a/packages/components/svgoUtils.js
+++ b/packages/components/svgoUtils.js
@@ -44,13 +44,17 @@ const replaceAttrKeys = child => {
 // Figma/ Sketch don't allow us to set 'currentColor' when exporting SVGs.
 // We have a convention to export the color as #000 or #0000000, then change it here.
 const replaceColor = child => {
-  if (child.attributes.fill === "#000" || child.attributes.fill === "#000000") {
+  const hexCodesToReplace = [
+    "#000",
+    "#000000",
+    "#2F2438", // Purple-800 from Figma export
+  ]
+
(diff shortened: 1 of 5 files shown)


packages/components/svgoUtils.js
```diff
@@ -44,13 +44,17 @@ const replaceAttrKeys = child => {
...
 // We have a convention to export the color as #000 or #0000000, then change it here.
 const replaceColor = child => {
-  if (child.attributes.fill === "#000" || child.attributes.fill === "#000000") {
+  const hexCodesToReplace = [
+    "#000",
+    "#000000",
+    "#2F2438", // Purple-800 from Figma export
+  ]
+
+  if (hexCodesToReplace.includes(child.attributes.fill)) {
```
gyfchong, line 53: Good stuff, just wish JS read better in an english sense lol ie. "fill includes a hex code to replace"

gyfchong (approved):

gyfchong:
We use the close icon in a lost of places haha

packages/components/svgoUtils.js
HeartSquared: Yeah I actually had to double take to ensure this wasn't changing the condition 😅

PR #4253, opened 2023-10-31
add LikertScaleLegacy

gyfchong:
## Why
Migration!

## What
`LikertScaleLegacy`

https://cultureamp.design/?path=/docs/legacy_components-likert-scale--docs


packages/components/src/LikertScaleLegacy/_docs/LikertScaleLegacy.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  },
+} satisfies Meta<typeof LikertScaleLegacy>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Playground: Story = {
+  render: args => {
+    const [selectedItem, setSelectedItem] = useState<ScaleItem | null>(null)
```
gyfchong, line 50: Debating whether or not I should make the Playground work...

packages/components/src/LikertScaleLegacy/LikertScaleLegacy.module.scss
```diff
@@ -0,0 +1,369 @@
...
+  .likertItem5 & {
+    margin-left: -12px;
+
+    [dir="rtl"] & {
+      margin-left: -9px;
+    }
+  }
+}
+
+:where(.likertItem4, .likertItem5) .check {
```
mcwinter07, line 306: Is there any potential issues with `:where` or `.is` not being scoped to a class like `.container :where`? I think the specificity of the `.likertItem` within the `where` selector should be ok, but it still feels a little iffy

gyfchong: Good spot, we should change this to be safer

packages/components/src/LikertScaleLegacy/_docs/LikertScaleLegacy.stories.tsx
mcwinter07: The story seems to work on the mdx page 😄 or do you mean show we not have it work because its a legacy likert scale?

gyfchong: I made it work, but I'm debating removing the functionality in the example cause it might imply no effort on the consumer side to make it work in their code base...

mcwinter07: Ah gotcha. Couldn't we update the canvas src to showcase the mimimum code required to get that story up and running?

gyfchong: Hm I guess there's not much to toggle in controls, so may as well hardcode it

mcwinter07: Nice 👍 I can't imagine it changing so I think that should be good

mcwinter07 (approved):

PR #4254, opened 2023-11-01
Migrate ProgressBar from kaizen-legacy

mcwinter07:
## Why
Migration task from kaizen legacy

## What
- Update folder architecture
- Move Label into subcomponents
- Add basic test coverage for utils and a11y coverage
- Move calculatePercentage into utils
- Update stories and documentation to modern pattern

## Notes for potential improvements / things not tackled in this PR
- Potential to revisit the a11y of the component
  - specifically connect the label and subtext to an aria-labelledby and aria-describedby
  - potential to connect the isAnimating prop to some kind of descriptive message for assistive technologies (may also need and aria-live region)
- Add RTL support
  - currently this does not render as is describe in design guidance on confluence. This also need to be updated in Figma
- Better guidance of the difference between the purpose of label vs subtext
- Sensible default props
  - props like `isReversed` `isAnimated` could potentially have a default value of `false`


gyfchong:
Double check `yarn test:storybook` locally and disable the a11y rules which are expected to fail eg. `aria-progressbar-name`

mcwinter07:
> Double check `yarn test:storybook` locally and disable the a11y rules which are expected to fail eg. `aria-progressbar-name`

@gyfchong I was thinking about this and while we could probably solve some use by attaching aria-labelledby to the Label, given its an optional prop it doesn't feel like that was the initial intention.

Makes me wonder if its because folks have their own labels and use `aria-labelledby` to ignore that prop or are just passing in `aria-label`, which resolves that issue 🤔

In any case I'll disable it for now given there are alternatives and I'll add some notes in the [update ticket](https://cultureamp.atlassian.net/browse/KDS-1906).

gyfchong:
>  solve some use by attaching aria-labelledby to the Label, given its an optional prop it doesn't feel like that was the initial intention.

IMO we shouldn't try to fix these errors in this case, cause we made this component with the intention of attaching a label manually, similar to the `input` components that don't have labels either. Although the difference here is that we don't have a `ProgressField` that composes it properly for our consumers.

I agree that chances are consumers are misusing this, but nevermind them for our testing. Just add a comment to the a11y disable to indicate that we don't have a label on purpose, and we can remove that if we ever add a label.

JakePitman (approved):
Looks good dude! Couldn't spot any issues

PR #4255, opened 2023-11-01
KDS-1416 Migrate RTE & Co

gyfchong:
## Why
Migrate RichTextEditor, RichTextContent, EditableRichTextContent from `kaizen-legacy`

## Notes
- `@cultureamp/rich-text-toolkit` added as an optional Peer Dep, meaning consumers can choose not to install the package if they do not use RTE. This does break the circular dependency as we are no longer trying to bundle RTT, however Ideally we would want to move towards removing this peer dep altogether.


packages/components/src/RichTextEditor/RichTextEditor/_docs/RichTextEditor.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    defaultValue: [],
+    onChange: () => undefined,
+  },
+} satisfies Meta<typeof RichTextEditor>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Playground: Story = {
```
mcwinter07, line 23: Potential improvement later is to add a drop down for `controls` to let our consumer add items to the toolbar in the playground story

packages/components/src/RichTextEditor/RichTextEditor/_docs/RichTextEditor.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+        {...args}
+        defaultValue={rteData}
+        onChange={handleOnChange}
+      />
+    )
+  },
+  parameters: {
+    docs: {
+      canvas: {
+        sourceState: "shown",
```
mcwinter07, line 41: One call out that Dale mentioned a while back was that when he first used the RTE, it was that it wasn't clear to him that `handleOnChange` would need to setter with `editorState.toJSON().doc.content`.

Something we could either document later or expose in the source code of the canvas

packages/components/src/RichTextEditor/RichTextEditor/_docs/RichTextEditor.mdx
```diff
@@ -0,0 +1,51 @@
...
+## Overview
+
+A text area with additional capabilities for users to format the input text.
+
+<Canvas of={RichTextEditorStories.Playground} />
+<Controls of={RichTextEditorStories.Playground} />
+
+## API
+
+<DocsStory of={RichTextEditorStories.Controls} />
```
mcwinter07, line 32: Would be good to circle back later and add more to this in terms of current controls available and how to handle grouping

mcwinter07 (dismissed):
I've just left a few comments some stories, more around uplift and things we could improve the docs later but the migration looks hunky dory to me so giving this a ✅

packages/components/src/RichTextEditor/RichTextEditor/_docs/RichTextEditor.stories.tsx
gyfchong: I can do a hardcode here again 👍

gyfchong:
Might pause this and wait for the RTT to be migrated first, then we won't be shipping a massive RTE and the peer dep wouldn't be necessary

HeartSquared (approved):

PR #4257, opened 2023-11-02
Fix export of JS tokens

gyfchong:
JS tokens weren't being compiled...


packages/components/jest.config.ts
```diff
@@ -12,6 +12,7 @@ const jestConfig: JestConfigWithTsJest = {
     "~types/(.*)$": "<rootDir>/src/types/$1",
     "~utils/(.*)$": "<rootDir>/src/utils/$1",
     "~components/(.*)$": "<rootDir>/src/$1",
+    "~tokens/(.*)$": "packages/design-tokens/src/$1",
```
HeartSquared, line 15: Wouldn't `~design-tokens` be better?

gyfchong: Too long haha

HeartSquared (approved):

PR #4260, opened 2023-11-02
fix(ErrorPage): Add 413 error page

kaydwithers:
## Why
This PR adds a new 413 Error Page.

## What
A 413 Error Page is rare, a client has encountered it in Production when a URL string is too large for the server to handle. 

When this occurred with the client, it showed an empty page. An Ops ticket was raised and this will partially cover some of that ticket when we add our NextJS error pages.

https://cultureamp.atlassian.net/browse/CABB-2775

### DIFF
--- a/.changeset/strong-donkeys-tell.md
+++ b/.changeset/strong-donkeys-tell.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+fix: add 413 error page
--- a/.changeset/strong-donkeys-tell.md
+++ b/.changeset/strong-donkeys-tell.md
@@ -2,4 +2,4 @@
 "@kaizen/components": minor
 ---
 
-fix: add 413 error page
+Add 413 error page.
--- a/packages/components/locales/mi.json
+++ b/packages/components/locales/mi.json
@@ -49,10 +49,6 @@
     "description" : "Main title for page",
     "message" : "You can't view this page"
   },
-  "kzErrorPage.404.message" : {
-    "description" : "Call to action instructions for the user",
-    "message" : "Sorry but we can't fing the page you're looking for. Go back and try again, or head to Home"
-  },
   "kzErrorPage.404.title" : {
     "description" : "Main title of page",
     "message" : "Missing pages are one of life's mysteries"
--- a/packages/components/locales/mi.json
+++ b/packages/components/locales/mi.json
@@ -49,6 +49,10 @@
     "description" : "Main title for page",
     "message" : "You can't view this page"
   },
+  "kzErrorPage.404.message" : {
+    "description" : "Call to action instructions for the user",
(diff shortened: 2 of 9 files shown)


christian314159 (approved):

.changeset/strong-donkeys-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+fix: add 413 error page
```
HeartSquared, line 5: This would be enough :) Or if you'd like to keep to the conventional commit, it should be `feat`
```suggestion
Add 413 error page.
```

packages/components/locales/mi.json
```diff
@@ -49,10 +49,6 @@
     "description" : "Main title for page",
     "message" : "You can't view this page"
   },
-  "kzErrorPage.404.message" : {
```
HeartSquared, line 52: Hmm... I think this would've been related to the other PR. It's probably safe to actually update the message with `find` instead of having it deleted (I know it's auto generated, but maybe this can save us some cost from getting new translations? 🤞🏻)

packages/components/locales/si-LK.json
```diff
@@ -51,7 +51,7 @@
   },
   "kzErrorPage.404.message" : {
     "description" : "Call to action instructions for the user",
-    "message" : "කණගාටුයි, නමුත් අපි ඔබ සොයන පිටුව fing නොහැක.  ආපසු ගොස් නැවත උත්සාහ කරන්න, නැතහොත් නිවසට යන්න"
+    "message" : "කණගාටුයි, නමුත් අපි ඔබ සොයන පිටුව find නොහැක.  ආපසු ගොස් නැවත උත්සාහ කරන්න, නැතහොත් නිවසට යන්න"
```
HeartSquared, line 54: Omg lmao 🫠 Okay maybe this one needs to actually be re-translated.

packages/components/locales/mi.json
kaydwithers: @HeartSquared yes, the previous PR got merged without a translation extraction. 

I've added it back and just replaced fing with find. For our translations we usually delete our old text so that the translator knows that it needs to be re-translated.

.changeset/strong-donkeys-tell.md
kaydwithers: @HeartSquared this has been updated.

packages/components/locales/mi.json
HeartSquared: Ah I see! That's probably a good call then. Didn't really think about it since we don't do i18n that often 😬 You've convinced me!

packages/components/locales/si-LK.json
kaydwithers: haha yeah okay, it's been removed for re-translation.

HeartSquared (approved):

PR #4270, opened 2023-11-06
add Scene\'s

gyfchong:
Migrate Scene from kaizen-legacy


packages/components/src/Illustration/Scene/_docs/Scene.mdx
```diff
@@ -0,0 +1,26 @@
...
+<Installation
+  installCommand="yarn add @kaizen/components"
+/>
+
+## Overview
+
+<Canvas of={SceneStories.Playground} />
+<Controls of={SceneStories.Playground} />
+
+<Canvas of={Stickersheets.StickerSheetDefault} />
```
gyfchong, line 26: Opted to just use the stickersheet, following our docs patterns for this type of component would just take too long for not much benefit

packages/components/src/Illustration/Scene/_docs/Scene.stickersheet.stories.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed }) => (
+    <StickerSheet isReversed={isReversed}>
+      <StickerSheet.Body>
+        <StickerSheet.Row>
+          EmptyStatesAction
+          <StickerSheet.Cell width={400}>
+            <EmptyStatesAction />
+          </StickerSheet.Cell>
```
gyfchong, line 72: I'm sure there's some smart way to iterate, but the way the original did it in `kaizen-legacy` was way too involved for the value we gain

dougmacknz (approved):

PR #4272, opened 2023-11-06
KDS-1914-migrate-guidance-block-to-kaio

gyfchong:
Migrate Guidance Block.


mcwinter07 (commented):
The scene illustration Story doesn't seem to be rendering as expected. I threw a width on the figure and it looks like the SVG is there, it's doesn't seem to be inheriting the width?
Screenshot 2023-11-08 at 3 33 59 pm

packages/components/src/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -0,0 +1,277 @@
...
+  smallScreenTextAlignment?: TextAlignment
+  actions?: GuidanceBlockActions
+  /*
+   * This will still require the secondary object to be passed into the actions ie: {secondary: { label: "Dismiss action" }}`
+   */
+  secondaryDismiss?: boolean
+  /*
+   * `persistent` should always return true and will soon be deprecated.
+   * The X close icon has been superseded with the pattern "dismiss" or "cancel" using the secondary action.
+   */
```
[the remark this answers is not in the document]
gyfchong: Just got rid of it altogether :P

gyfchong:
Interesting @mcwinter07 my Scene renders fine

packages/components/src/GuidanceBlock/_docs/GuidanceBlock.stickersheet.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+  content: (
+    <>
+      <Tag>Early Access</Tag>
+      <Heading variant="heading-3">{TEXT_PROPS.text.title}</Heading>
+      <Text variant="body">{TEXT_PROPS.text.description}</Text>
+    </>
+  ),
+}
+
+const StickerSheetTemplate: StickerSheetStory = {
```
mcwinter07, line 62: Should we also have the stacked variants sent to chromatic for safety since there a visual difference in the layout? either that or send the individual stories

mcwinter07:
> Interesting @mcwinter07 my Scene renders fine

Strange, let me try in another browser

gyfchong:
Is this an issue in legacy @mcwinter07?

mcwinter07:
@gyfchong  Found the issue (ish), seems like its a problem with that specific SVG, if you replace it with another scene it renders ok in the sticker sheet. I'll create a separate ticket to investigate that Scene illustration since it's not the Guidance block (though I would recommend replacing it in the story for now)

packages/components/src/Filter/FilterDatePicker/FilterDatePicker.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 import { formatDateAsText } from "~components/Calendar"
 import { getLocale } from "~components/DatePicker/utils/getLocale"
-import { Filter, FilterContents, FilterProps } from "../Filter"
+import { Filter, FilterContents, FilterProps } from "~components/Filter"
```
mcwinter07, line 4: Just checking if its intended to have these changes for the Filter in? It look like its just a bit of house keeping with the import path but wanted to confirm 😉

gyfchong: Good catch

mcwinter07 (approved):
Other than the non-blocking comment, this looks good to me 👍

packages/components/src/GuidanceBlock/GuidanceBlock.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+  onchange: null,
+  addListener: jest.fn(),
+  removeListener: jest.fn(),
+}))
+
+describe("GuidanceBlock", () => {
+  afterEach(cleanup)
+
+  it("is initially visible", () => {
+    const { container } = render(
```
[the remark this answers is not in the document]
gyfchong: Actually, this is a useless test, so goodbye!

mcwinter07 (approved):

PR #4274, opened 2023-11-07
add Icon reference docs

gyfchong:
Add an Icon docs page with a reference to all the available icons + search and quick copy


mcwinter07 (commented):
Awesome work, I really like the updated docs with the search / copy util 👨‍🍳 One thought for an additional story is how to apply color. It could be helpful for new comers to understand how the SVG receives the fill color using currentColor

mcwinter07 (approved):

PR #4280, opened 2023-11-08
Tailwind-migration-docs

gyfchong:
Teaching people how to migrate to V2 + install the reset all on one page


docs/Systems/Tailwind/migration-to-v2.mdx
```diff
@@ -0,0 +1,25 @@
+import { Meta } from "@storybook/blocks"
+
+<Meta title="Systems/Tailwind/Migration to v2" />
+
+# Migration to v2
```
sentience, line 5: Sorry for the basic question, but v2 of what? Not of Tailwind, surely? Of KAIO? It might be good to be explicit to avoid confusion.

docs/app-starter.stories.mdx
```diff
@@ -53,17 +53,19 @@ The order of these CSS stylesheets is **very important**:
 
 No matter how you choose to import the following, this order must be upheld.
 
+**We recommend you use your `tailwind.css` file to ensure this order (JS files will order them differently)**
+
 ### Reset
 
 Kaizen's CSS reset aims to add any global default styles such as typography or base styles which enable Tailwind utilities (you can think of this as Kaizen's own Preflight).
```
sentience, line 60: I suggest linking this mention of Preflight to the Tailwind docs on the subject, since many readers will be unfamiliar with this: https://tailwindcss.com/docs/preflight

docs/app-starter.stories.mdx
```diff
@@ -53,17 +53,19 @@ The order of these CSS stylesheets is **very important**:
...
 
 Note: this is different from a Normalize which aims to create uniformity in basic elements across the multiple browsers. (eg. Corrects bugs and common browser inconsistencies).
```
sentience, line 62: Can we also explain in a sentence or two:
1. Why we aren't able/choosing to use Tailwind's Preflight.
2. What kinds of things our reset does and why. (Perhaps, as Tailwind's docs do, we could link to the source code and ensure it contains clear comments to document the purpose of each style rule.)

docs/app-starter.stories.mdx
```diff
@@ -53,17 +53,19 @@ The order of these CSS stylesheets is **very important**:
 
 No matter how you choose to import the following, this order must be upheld.
 
+**We recommend you use your `tailwind.css` file to ensure this order (JS files will order them differently)**
```
sentience, line 56: Below this same document says:
> You can either import this via JS or CSS,

Do we no longer recommend that? If so, recommend updating that other wording.

docs/app-starter.stories.mdx
gyfchong: I don't think this is the appropriate place to explain these things, and doesn't help the reader achieve the intended goal of the page, which is to let someone know how to manually retro-fit an existing code base, which should become rarer as more repos are spun up with `next-template`. 

I'll remove the reference to Preflight to reduce any potential questions, and I'd be happy to write a decision log with these details, in case someone would like to know what led us down this path.

docs/Systems/Tailwind/migration-to-v2.mdx
gyfchong: To try to avoid being verbose in the title, I've renamed the entire section to "Tailwind Preset" on our storybook to be more explicit. Hope that helps!

docs/app-starter.stories.mdx
sentience: Sounds good. Having the reasoning for a custom preflight/reset recorded somewhere easy to find (IMO a comment in the file itself that compiles away would be ideal for this) will be valuable.

PR #4281, opened 2023-11-09
Exclude-renovate-chromatic

gyfchong:
Context:
https://cultureamp.slack.com/archives/C2ZNME08P/p1699512153542499


.github/workflows/test-stories.yaml
```diff
@@ -9,6 +9,8 @@ on:
       - opened
       - synchronize
       - ready_for_review
+    branches-ignore: # on following branches we don't want to run chromatic automatically, but only on manual check
```
HeartSquared, line 12: We'll need to add a way to manual check then

HeartSquared (approved):

PR #4282, opened 2023-11-10
Migrate BrandMomentCaptureIntro from kaizen-legacy

mcwinter07:
## Why
Last of the components to migrate

## What
- Migrate BrandMomentCaptureIntro as nested Scene component
  - Replace setting of inline styles and display non with conditional render
- Adds stories and documentation to illustrate specific use case / animation

## Notes and context
Based on code comment and a quick search of where this is used (almost exclusively Murmur) this appears to be a mostly a snowflake. 

Given the minimal differences between this and Scene, I think with a refactor of Scene we could even remove this from KAIO entirely and let this built from its smaller parts, but for now I've avoided touching the Video and Base subcomponents that could likely achieve.


packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.stickersheet.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <StickerSheet.Body>
+        <StickerSheet.Row>
+          <BrandMomentCaptureIntro />
+        </StickerSheet.Row>
+      </StickerSheet.Body>
+    </StickerSheet>
+  ),
+  parameters: {
+    pseudo: {
+      hover: true,
```
mcwinter07, line 30: I've blanketed this with a hover true for chromatic to snapshot the autoplay pause button on the video render of the illustration. 

I started going down the route of updating the props to allow for data attributes so the selector could be use for a hover state but we would need to update several types and how props are spread in its various shared subcomponents so I opted to leave this out (especially due to the snowflake nature of the component)

gyfchong: Oh good spot, thanks

gyfchong (dismissed):

packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.mdx
```diff
@@ -0,0 +1,45 @@
...
+- An ambient animation that can be looped.
+- A static image that is not animated.
+
+<Canvas of={BrandMomentCaptureIntroStories.Playground} />
+<Controls of={BrandMomentCaptureIntroStories.Playground} />
+
+## API
+
+### Single animation
+Will trigger the animate once.
```
HeartSquared, line 36:
```suggestion
Will trigger the animation once.
```

packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.stickersheet.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  parameters: {
+    chromatic: { disable: false },
+    controls: { disable: true },
+  },
+} satisfies Meta
+
+const StickerSheetTemplate: StickerSheetStory = {
+  render: ({ isReversed }) => (
+    <StickerSheet isReversed={isReversed}>
+      <StickerSheet.Header headings={["Default", "Autoplay (hover)"]} />
```
HeartSquared, line 20: Looks like there's only one example, so no need for headings right?

packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import { Meta, StoryObj } from "@storybook/react"
+import { BrandMomentCaptureIntro } from "../index"
+
+const meta = {
+  title: "Components/Illustrations/BrandMomentCaptureIntro",
```
HeartSquared, line 5: Should it be `Illustrations/Scene/Brand...`?

packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.stickersheet.stories.tsx
mcwinter07: Ah fudge, must of unstaged one, will fix!

packages/components/src/Illustration/Scene/BrandMomentCaptureIntro/_docs/BrandMomentCaptureIntro.stories.tsx
mcwinter07: hmm yeah, its a strange one. Technically it is its own thing, but logically it would make better sense to group it in there since it's only ever used as a stand-in for scene

packages/components/src/Illustration/Scene/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./Scene"
+export * from "./BrandMomentCaptureIntro"
```
gyfchong, line 2: Alphabetical order please 😄

HeartSquared (approved):

gyfchong (approved):

PR #4283, opened 2023-11-10
Add description to MultiSelect

dougmacknz:
https://cultureamp.atlassian.net/browse/KDS-1714

Notes:
* Strange that `MultiSelectToggle` takes a bunch of props and then splits them between a containing div and the button itself.

### DIFF
--- a/.changeset/wet-emus-invite.md
+++ b/.changeset/wet-emus-invite.md
@@ -1,5 +1,2 @@
 ---
-"@kaizen/components": minor
 ---
-
-Add ability to add a description to MultiSelect
--- a/.changeset/wet-emus-invite.md
+++ b/.changeset/wet-emus-invite.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add ability to add a description to MultiSelect
--- a/packages/components/src/MultiSelect/MultiSelect.spec.tsx
+++ b/packages/components/src/MultiSelect/MultiSelect.spec.tsx
@@ -30,7 +30,6 @@ const MultiSelectWrapper = ({
     <MultiSelect
       items={defaultOptions}
       label="Jalapeno"
-      description="A short description"
       {...otherProps}
       isOpen={isOpen}
       onOpenChange={setIsOpen}
--- a/packages/components/src/MultiSelect/MultiSelect.spec.tsx
+++ b/packages/components/src/MultiSelect/MultiSelect.spec.tsx
@@ -30,6 +30,7 @@ const MultiSelectWrapper = ({
     <MultiSelect
       items={defaultOptions}
       label="Jalapeno"
+      description="A short description"
       {...otherProps}
       isOpen={isOpen}
       onOpenChange={setIsOpen}
@@ -42,6 +43,20 @@ const MultiSelectWrapper = ({
 const user = userEvent.setup()
 
 describe("<MultiSelect />", () => {
+  describe("accessible name and description", () => {
+    it("has an accessible name and description when provided a description", async () => {
+      const { getByRole } = render(
+        <MultiSelectWrapper id="jalapeno" description="A short description" />
+      )
+      const toggleButton = getByRole("button", {
+        name: "Jalapeno",
+        description: "A short description",
+      })
+
+      expect(toggleButton).toBeInTheDocument()
+    })
+  })
+
   describe("id", () => {
     it("uses the consumer-provided id", async () => {
       const { getByTestId, getByRole } = render(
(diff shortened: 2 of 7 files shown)


mcwinter07:
> Strange that MultiSelectToggle takes a bunch of props and then splits them between a containing div and the button itself.

@dougmacknz Do you mean why did we not spread the restProps onto the button as appose to the surrounding div?

mcwinter07 (approved):
Looks hunky dory to me (appreciate keeping with the breaky theme 😉). Left a response to the note for follow up. Also happy to continue discussion in slack if its non blocking to the PR

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -16,6 +17,10 @@ export type MultiSelectProps = {
   label: string
   items: MultiSelectOptionsProps["options"]
   selectedValues: Set<MultiSelectOption["value"]>
+  /**
+   * A description that provides context for the field
+   */
+  description?: ReactNode
```
gyfchong, line 23: Could we use the type from `FieldMessage` itself? Just to make sure it's always passing the correct type

dougmacknz:
> Do you mean why did we not spread the restProps onto the button as appose to the surrounding div?

@mcwinter07 Yeah so just when I pass props to `MultiSelectToggle` some go to the button (aria-labelledby, etc), and then some go to the div (any native stuff not defined). Created a bit of confusion for me and feels like it could pose some issues down the line.

.changeset/wet-emus-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
gyfchong, line 2: Let's empty the changeset, we haven't officially released MultiSelect

mcwinter07:
> > Do you mean why did we not spread the restProps onto the button as appose to the surrounding div?
> 
> @mcwinter07 Yeah so just when I pass props to `MultiSelectToggle` some go to the button (aria-labelledby, etc), and then some go to the div (any native stuff not defined). Created a bit of confusion for me and feels like it could pose some issues down the line.

I think the intention was for the `div` to have the restProps so it can have data attributes, specially `data-testid`, whereas button element only receive whgat controls the open and close state

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -26,13 +31,15 @@ export const MultiSelect = ({
...
   selectedValues,
+  description,
   onSelectedValuesChange,
   isOpen,
   onOpenChange,
   classNameOverride,
   ...restProps
 }: MultiSelectProps): JSX.Element => {
   const id = propsId ?? useId()
+  const descriptionId = useId()
```
HeartSquared, line 42: Might not be an issue, but do you think we should do the thing where if the consumer passes in an `id`, the description id is based on it? Eg. `{id}--description`?

packages/components/src/MultiSelect/_docs/MultiSelect.stories.tsx
```diff
@@ -25,6 +25,8 @@ const meta = {
   },
   args: {
     label: "Breakfast order",
+    description:
```
HeartSquared, line 28: Given this is an optional prop, this should have its own story and docs description instead of being part of the default.

packages/components/src/MultiSelect/MultiSelect.tsx
dougmacknz: I've switched it for consistency

packages/components/src/MultiSelect/MultiSelect.spec.tsx
```diff
@@ -30,6 +30,7 @@ const MultiSelectWrapper = ({
     <MultiSelect
       items={defaultOptions}
       label="Jalapeno"
+      description="A short description"
```
HeartSquared, line 33: Oops missed this. Once again since it's optional (same logic for tests as stories), we don't want it there always 😬

packages/components/src/MultiSelect/_docs/MultiSelect.stories.tsx
```diff
@@ -143,3 +143,10 @@ export const OpenState: Story = {
   ...MultiSelectTemplate,
   parameters: { docs: { source: { code: sourceCodeOpenState } } },
 }
+
+export const Description: Story = {
```
HeartSquared, line 147: You'll need to add this to the MDX :)

packages/components/src/MultiSelect/MultiSelect.spec.tsx
```diff
@@ -42,6 +42,20 @@ const MultiSelectWrapper = ({
 const user = userEvent.setup()
 
 describe("<MultiSelect />", () => {
+  describe("accessible name and description", () => {
+    it("has an accessible name and description when provided a description", async () => {
```
HeartSquared, line 46: Nit (kinda annoying that I don't see everything at once :s)
```suggestion
    it("has an accessible name and description when provided a description", () => {
```

HeartSquared (dismissed):

HeartSquared (approved):

PR #4306, opened 2023-11-13
MultiSelect: Add status and validationMessage props

mcwinter07:
## Why
Adding missing features

## What
-  Add `status` variants styles and sitckersheet variants to MultiSelectToggle
- Add `validationMessage` and `status` to MultiSelect
  - Connect `validationMessage` to `aria-describedby`
  - Add test coverage for validation states
  - Add sitckersheet variants


packages/components/src/MultiSelect/MultiSelect.spec.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from "react"
-import { render, waitFor } from "@testing-library/react"
+import { render, waitFor, screen } from "@testing-library/react"
```
mcwinter07, line 2: Using screen here since this is the decided way going forward but will hold off on refactoring the others to use this to a later time

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -19,6 +20,10 @@ export type MultiSelectProps = {
   onSelectedValuesChange: MultiSelectOptionsProps["onChange"]
   isOpen: boolean
   onOpenChange: (isOpen: boolean) => void
+  /** A status to convey `error` or `caution` states for the input  */
+  status?: FieldMessageProps["status"]
+  /** A message to provide context for the change in `status`, most commonly validation.  */
+  validationMessage?: FieldMessageProps["message"]
```
mcwinter07, line 26: Noting that we are only really using `error` and `caution` from this. Is it worth just typing those two or leaving as is? Neither our `default` or `success` state impact the UI?

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -76,18 +84,27 @@ export const MultiSelect = ({
           ref={toggleButtonRef}
           id={`${id}--toggle`}
           aria-labelledby={`${id}--label`}
+          aria-describedby={validationId}
```
mcwinter07, line 87: This will probably conflict with Dug's PR so may need to loop back here if that gets merged first

packages/components/src/MultiSelect/MultiSelect.tsx
HeartSquared: You could use `Pick`? Probably better to only include the ones necessary, otherwise it will look like a bug 😬

packages/components/src/MultiSelect/MultiSelect.spec.tsx
```diff
@@ -281,3 +281,32 @@ describe("Removing an option", () => {
...
+    render(
+      <MultiSelectWrapper
+        selectedValues={new Set(["waffle"])}
+        status="error"
+        validationMessage="No waffles are available"
+      />
+    )
+    expect(screen.getByText("No waffles are available")).toBeInTheDocument()
+  })
+  it("describes the toggle", () => {
```
HeartSquared, line 296: Definitely worth adding a test (or amending this one) to also showcase the order relative to the field description when Doug's PR is merged in.

packages/components/src/MultiSelect/MultiSelect.tsx
```diff
@@ -30,9 +35,12 @@ export const MultiSelect = ({
   isOpen,
   onOpenChange,
   classNameOverride,
+  status,
+  validationMessage,
   ...restProps
 }: MultiSelectProps): JSX.Element => {
   const id = propsId ?? useId()
+  const validationId = `${id}-description`
```
HeartSquared, line 43: You'll be needing to combine with the description, so they need different ids :)
```suggestion
  const validationId = `${id}--validation-message`
```

packages/components/src/MultiSelect/_docs/MultiSelect.stickersheet.stories.tsx
```diff
@@ -51,48 +51,90 @@ const StickerSheetTemplate: StickerSheetStory = {
...
     )
   },
+  parameters: {
+    pseudo: {
+      hover: '[data-sb-pseudo-styles="hover"]',
+      active: '[data-sb-pseudo-styles="active"]',
+      focus: '[data-sb-pseudo-styles="focus"]',
+      focusVisible: '[data-sb-pseudo-styles="focus"]',
+    },
+  },
```
HeartSquared, line 137: This shouldn't be needed?

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/MultiSelectToggle.tsx
```diff
@@ -40,14 +44,17 @@ export const MultiSelectToggle = forwardRef<
        */}
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
       <div
-        className={classnames(styles.multiSelectToggle, classNameOverride)}
+        className={classnames(styles.multiSelectToggle, classNameOverride, [
+          status && styles[status],
+        ])}
```
HeartSquared, line 49: The array isn't needed :)
```suggestion
        className={classnames(styles.multiSelectToggle, classNameOverride,
          status && styles[status],
        )}
```

packages/components/src/MultiSelect/MultiSelect.spec.tsx
```diff
@@ -281,3 +281,32 @@ describe("Removing an option", () => {
...
+describe("Has validation status", () => {
+  it("has a validation message with an accessible name", () => {
+    render(
+      <MultiSelectWrapper
+        selectedValues={new Set(["waffle"])}
+        status="error"
+        validationMessage="No waffles are available"
+      />
+    )
+    expect(screen.getByText("No waffles are available")).toBeInTheDocument()
```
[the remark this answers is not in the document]
mcwinter07: True, I think I'll change this to "has a validation message" since I broke out the accessibility testing into the other test

packages/components/src/MultiSelect/MultiSelect.tsx
mcwinter07: Ah I think this has to be `Extract` since `FieldMessageStatus` is a union of string literals

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/_docs/MultiSelectToggle.stickersheet.stories.tsx
```diff
@@ -91,6 +91,46 @@ const StickerSheetTemplate: StickerSheetStory = {
...
+              <MultiSelectToggle status="caution" {...defaultProps} />
+              <MultiSelectToggle
+                status="caution"
+                {...defaultProps}
+                data-sb-pseudo-styles="hover--button"
+              />
+              <MultiSelectToggle
+                status="caution"
+                {...defaultProps}
+                data-sb-pseudo-styles="focus--button"
```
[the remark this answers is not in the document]
mcwinter07: You could argue that the hover and focus change is tested in the previous story since our border don't change for validation states. Though I do wonder if there's value in it to ensure that there is a11y contrast in the story.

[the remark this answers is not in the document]
mcwinter07: Maybe I'll remove them and we can add the hover and focus states back into the Toggle if ever add any additional styles (like a darkened border)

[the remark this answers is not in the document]
HeartSquared: Yeah I think if they don't have anything specific, it should be safe to not have them :)

packages/components/src/MultiSelect/MultiSelect.spec.tsx
mcwinter07: Added in a new test to testing the concatenation specifically in the accessible description 👍

packages/components/src/MultiSelect/MultiSelect.tsx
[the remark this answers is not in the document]
gyfchong: Yes please!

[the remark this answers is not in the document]
gyfchong: (Apologies @dougmacknz this wasn't a very well documented pattern)

packages/components/src/MultiSelect/types.ts
```diff
@@ -1,5 +1,12 @@
...
 export type MultiSelectOption = {
   label: string
   // Used for "value" attribute in option - must be unique
   value: string | number
 }
+
+export type ValidationMessage = {
+  status?: Extract<FieldMessageStatus, "error" | "caution">
+  message: string | React.ReactElement
+}
```
gyfchong, line 12: Seems like something we need to make more widely accessible to the other fields

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/MultiSelectToggle.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { HTMLAttributes, forwardRef } from "react"
 import classnames from "classnames"
 import { ClearButton } from "~components/ClearButton"
+// import { ClearButton } from "~components/ClearButton"
```
gyfchong, line 4: Should this be here?

packages/components/src/MultiSelect/types.ts
mcwinter07: You mean like a shared types file in the root of the src similar to the style utils?

gyfchong: Could it belong to a component?

mcwinter07: We could export the type from FieldMessage since these props are consumed in that component - assuming that we only want validation messages to have `error` and `caution` states.

gyfchong: We can always extend later, but its your call whether it makes sense to do it now 😄

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/MultiSelectToggle.tsx
mcwinter07: Not anymore, just a hangover of a rebase, will 🪓

packages/components/src/MultiSelect/types.ts
mcwinter07: My main concern was the existing one in the Filter Date Picker, since it has all states accounted for, but looking at the story its only ever used for a error state. As you say we can always extend on it and it doesn't hurt to be more specific in what we're expecting

mcwinter07: Hmmm gave it a shot and realized I'd have to make a fair few changes to the imports in the Filter DP. I think I'll revert that back and create a ticket following this one to move that type to the FieldMessage component once this is merged. I'd rather untangle this release and get a specific optimization out as its own thing :)

gyfchong (approved):

PR #4307, opened 2023-11-14
Implement clear button

JakePitman:
## Why
So that user's can remove all selected options without having to clear each one individually.

## What
Adds a 'clear all' button to the MultiSelectToggle when an option has been selected.


.changeset/large-falcons-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add Clear all button to Multi Select
```
gyfchong, line 5: This should be empty because we aren't releasing the component yet

gyfchong (approved):

PR #4309, opened 2023-11-16
refactor(ToastNotification): update to be tree-shakeable

HeartSquared:
## Why

[KDS-1957](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-1957)

If one part is not tree-shakeable, the whole package can fail to tree-shake.

## What

Make `ToastNotification` tree-shakeable.

- Refactored to use context instead of ToastNotificationManager (this was the main cause)
- Renamed the object `ToastNotification` type as it clashed with the component
- Separated the util for updating a notification from adding
- Add a hook to remove the need to expose the context to consumers
	- The context is consumed by KaizenProvider, therefore must also be used by consumers

[KDS-1957]: https://cultureamp.atlassian.net/browse/KDS-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gyfchong:
Oh, cause of the portals, does that mean Toasts won't appear on top of Workflow or Modals/Inspector?

packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
```diff
@@ -0,0 +1,40 @@
...
+  if (containers) {
+    // Remove any duplicate instances
+    // (eg. Storybook docs page has multiple stories each with their own context)
+    containers.forEach((c, i) => {
+      if (i === 0) return
+      c.remove()
+    })
+  }
+
+  return createPortal(
```
gyfchong, line 24: Does this mean toasts won't appear on top of modals/inspector/workflow?

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.mdx
```diff
@@ -17,51 +17,56 @@ import * as ToastNotificationStories from "./ToastNotification.stories"
 
 <Installation
   installCommand="yarn add @kaizen/components"
-  importStatement='import { addToastNotification, removeToastNotification, clearToastNotifications, ToastNotification } from "@kaizen/components"'
+  importStatement='import { addToastNotification, updateToastNotification, removeToastNotification, clearToastNotifications, ToastNotification } from "@kaizen/components"'
```
gyfchong, line 20: This is a bit long, maybe we focus on the main function we want to champion? Was it `addToastNotification`? and move the rest of the imports to their respective sections.

Maybe even leave out the `<ToastNotification />` component to deter further use.

packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
HeartSquared: Hmm...did they ever though? It was injected into `document.body` from the start 🤔

gyfchong: Seems the styles include a `z-index`, not high enough for workflow, but seems like toasts should be considered the highest priority over any other floating element. Perhaps this just means we need a token system for `z-index` so we rely on a consistent environment to guarantee that important things are surfaced across the platform.

OOS of course.

gyfchong (approved):

PR #4327, opened 2023-11-20
Kds 1953 port button fix from kaizen legacy

JakePitman:
## Why
There was an issue where screen readers would announce the button's contents on page load.

## What
Sets `aria-live` attribute to `polite` only when a `workingLabel` is present.
Adds _tests_ ✨
Adds a story to demonstrate button transitioning from a 'ready' state, to 'working', to 'completed'. This story also demonstrates how the screen reader behaves when the button transitions in and out of the 'working' state.


.changeset/popular-gifts-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
gyfchong, line 2: Should this be a patch as it seems to fix an issue?
```suggestion
"@kaizen/components": patch
```

JakePitman: Yep, thanks

mcwinter07 (commented):
Nice stuff. The stories a good example for our test cases, but I'm wondering if we add it to our docs page to show a basic pattern in action? Potentially we could add it as a canvas below the story that shows its multiple states. Either than or just add a button to the original story to end the working status

mcwinter07 (approved):

PR #4328, opened 2023-11-20
Add the portalContainerId in Select Future release

mcwinter07:
## Why
This is blocking adoption of the Select component in KAIO.  Consumers were having issues with the select not having an expose portal prop.

## What
- Introduce the `portalContainerId` prop in future `Select`
  - This is then used to pass a new portal element down to the selects `Popover` primitive
- adds test cases to ensure portal appends to correct DOM nodes
- adds a new story to showcase and document how to use


packages/components/src/__future__/Select/Select.tsx
```diff
@@ -67,6 +67,7 @@ export type SelectProps<Option extends SelectOption = SelectOption> = {
    * @deprecated: Either define `disabled` in your `Option` (in `items`), or use `disabledKeys`
    */
   disabledValues?: Key[]
+  portalContainer?: Element | DocumentFragment
```
gyfchong, line 70: Could we instead let the consumer pass in the id of the component they would like to portal to? 

Imagine they have a modal that they're nesting the select into, they could add an `id` to their modal and then the same to their select to create the affect.

mcwinter07: Probably create an easier API to interact with if they need a portal 🤔 Will give it a crack

mcwinter07: Updated the component to take an id as recommended. It should offload some of the work folks may have had to do to pass in the element 🕺

packages/components/src/__future__/Select/_docs/Select.mdx
```diff
@@ -98,3 +98,23 @@ Add validation messages using `status` and `validationMessage`.
...
+
+### Portals
+
+By default, the Select's popover will attach itself to the `body` of the document using React's `createPortal`.
+
+You can change the default behaviour by providing a `portalContainerId` to attach this to different element in the DOM. This can help to resolve issues that may arise with `z-index` or having a Select in a modal.
+
+<Canvas of={SelectStories.PortalContainer} />
+
+It is important to note that a `relative` position should be applied to container element that contains the portal, ie:
```
gyfchong, line 110: I'd say "not absolute" rather than "is relative", as it will explain it clearer without the need for an example since everything is relative by default.

mcwinter07: But the container needs a `position: relative` for the popover to determine its position document since it uses `absolute`. Otherwise the static positioning on most elements could risk the Popover spilling out of the confines of it container.

Relative on container:
Screenshot 2023-11-22 at 9 24 59 am

Classless container:
Screenshot 2023-11-22 at 9 24 53 am

gyfchong: Huh very curious, but do we want it to cut off..?

HeartSquared: I wouldn't cut it off as a default, but if someone wants to cut it off, I'm sure they can figure it out :) I think we could probably leave the advice off 😄

packages/components/src/__future__/Select/Select.spec.tsx
```diff
@@ -356,4 +356,83 @@ describe("<Select />", () => {
       })
     })
   })
+
+  describe("Popover portal", () => {
+    it("has accessible trigger controls", async () => {
+      render(<SelectWrapper id="id--select" isOpen />)
```
HeartSquared, line 362: No need to declare the id in these tests

packages/components/src/__future__/Select/Select.tsx
```diff
@@ -67,6 +67,10 @@ export type SelectProps<Option extends SelectOption = SelectOption> = {
    * @deprecated: Either define `disabled` in your `Option` (in `items`), or use `disabledKeys`
    */
   disabledValues?: Key[]
+  /**
+   * Creates a portal for the Popover to the matching element id
+   */
+  portalContainerId?: string
```
HeartSquared, line 73: I think instead of confining it to an id, we should expose the same api that we've given the Popover, which accepts `portalContainer` as an element. This way, the consumer can use a ref if they wanted and do not require an id :)

gyfchong: This was my suggestion to simplify the API for the most common use case

HeartSquared: I'm in two minds about this, but having just tested variations of consumer code, I suppose this will be easiest for the consumer since they don't need a `useEffect`. Happy to go this way then :)

packages/components/src/__future__/Select/_docs/Select.mdx
mcwinter07: We can remove it, but my thinking here is that this is a visual shorthand of how the portal container can effect the popover, instead of it just being on the body. 

I'm less beholden to having the advice around relative at the bottom since the example code can be seen in the canvas anyway.

packages/components/src/__future__/Select/Select.spec.tsx
mcwinter07: And they're outa here ⚾ 🌬️ Nice catch, some left overs of when I had some janky selectors in there

.changeset/happy-tips-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
```
gyfchong, line 2:
```suggestion
"@kaizen/components": minor
```
Feels like a new feature 😄

mcwinter07: True!

gyfchong (approved):

PR #4329, opened 2023-11-20
Remove value prefix on Select combobox accessible name

dougmacknz:
From the code comment I wrote:
react-aria/useSelect wants to prefix the combobox's accessible name with the value of the select.
We use role=combobox, meaning screen readers will read the value.
So we're modifying the `aria-labelledby` property to remove the value element id.
Issue: https://github.com/adobe/react-spectrum/issues/4091

### DIFF
--- a/packages/components/src/__future__/Select/Select.tsx
+++ b/packages/components/src/__future__/Select/Select.tsx
@@ -116,12 +116,22 @@ export const Select = <Option extends SelectOption = SelectOption>({
 
   const {
     labelProps,
-    triggerProps,
+    triggerProps: reactAriaTriggerProps,
     valueProps,
     menuProps,
     errorMessageProps,
     descriptionProps,
   } = useSelect(ariaSelectProps, state, triggerRef)
+
+  // Hack incoming:
+  // react-aria/useSelect wants to prefix the combobox's accessible name with the value of the select.
+  // We use role=combobox, meaning screen readers will read the value.
+  // So we're modifying the `aria-labelledby` property to remove the value element id.
+  // Issue: https://github.com/adobe/react-spectrum/issues/4091
+  const triggerProps = {
+    ...reactAriaTriggerProps,
+    "aria-labelledby": reactAriaTriggerProps["aria-labelledby"]?.split(" ")[1],
+  }
 
   const { buttonProps } = useButton(triggerProps, triggerRef)
   const selectToggleProps = {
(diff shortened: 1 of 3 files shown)


mcwinter07 (approved):
Nice fix, this really cleans up those tests. Hearing that they're on a pattern from WAI-ARIA 1.2 makes a lot more sense why the value was prepended into the accessible label label. 

This may break the tests for anyone consuming the future Select package so worth calling out in the updates channel 👍

gyfchong (dismissed):
Beautiful hack

packages/components/src/__future__/Select/Select.tsx
```diff
@@ -116,13 +116,23 @@ export const Select = <Option extends SelectOption = SelectOption>({
...
   } = useSelect(ariaSelectProps, state, triggerRef)
 
+  // Hack incoming:
+  // react-aria/useSelect wants to prefix the combobox's accessible name with the value of the select.
+  // We use role=combobox, meaning screen readers will read the value.
+  // So we're modifying the `aria-labelledby` property to remove the value element id.
+  // Issue: https://github.com/adobe/react-spectrum/issues/4091
+  const triggerProps = {
+    ...reactAriaTriggerProps,
+    "aria-labelledby": reactAriaTriggerProps["aria-labelledby"]?.split(" ")[1],
```
HeartSquared, line 133: If there's no item selected, then would you end up with no label?

dougmacknz: They still add both ids even when there's no value or placeholder.

Screenshot 2023-11-21 at 10 17 22 am

mcwinter07: I think there was some weird behavior around it add the default value in if no item was selected. That said, if you have neither that's still a very good question

dougmacknz: I'll add a test for the case where it doesn't have a selected value if it doesn't already exist

HeartSquared: Do we pass in the label as a prop? What if we have `aria-label={label}` and `aria-labelledby={undefined}` (to remove what react-aria puts in)?

dougmacknz: I don't think it's possible to do that. If you pass in `aria-labelledby={undefined|null|" "}` react-aria just ignores it and does its usual thing.

Passing in an extra `aria-labelledby` was possible though, so I've adjusted the string manipulation now to take everything beyond the first space character and added a test for that case.

HeartSquared: It should only ignore it if you've declared your prop before the spread. If you put it after, it shouldn't be ignored 🤔 Nonetheless, whatever works! No idea what strange magic they may have built in haha.

mcwinter07 (approved):

PR #4330, opened 2023-11-20
Pass aria-describedby into RTE when sent via props

dougmacknz:
## Why
RTE currently allows you to pass in `aria-describedby` but it never gets used.

## What
Pass `aria-describedby` from props through to the RTE


packages/components/src/RichTextEditor/RichTextEditor/RichTextEditor.tsx
```diff
@@ -144,7 +145,11 @@ export const RichTextEditor = ({
     : ""
   const descriptionAria = description ? `${editorId}-rte-description` : ""
 
-  const ariaDescribedBy = `${validationMessageAria} ${descriptionAria}`
+  const ariaDescribedBy = classnames(
```
gyfchong, line 148: Why the use of `classnames` here?

dougmacknz: Clean way to concat these strings with a space in between

dougmacknz: Compared to string interpolation, if one of them is falsy, it won't add the extra space or the word `undefined` which happens sometimes.

dougmacknz: image

gyfchong: Oh, that's a creative use, but I don't think its very clear. Maybe add a comment or find a utility that is a bit more closer to the use?

dougmacknz: Respectfully disagree because this is the one thing that the function does. Wanna put in a review suggestion for what you think the comment should be?

gyfchong: Hm fair enough

gyfchong (approved):

PR #4332, opened 2023-11-21
Remove redundant buildkite config

mcwinter07:
## Why
We no long use buildkite as our primary CI pipeline so this is cleaning up the now redundant expense.

## What
- removes `.buildkite` folder and config


gyfchong (approved):

mcwinter07:
Coverting this back to draft until we land on the BK approach

PR #4335, opened 2023-11-22
Add conditional check for release tag in Buildkite pipeline

mcwinter07:
## Why
This aims to stop all buildkite builds and publishing if a merge to the main brain does not have a release tag

## What
- adds a conditional in the Buildkite steps to check for the existance of a tag and that it is in release format, ie: `blah@12.3.4`
  - Theoretically this means that it should fail on the initial step, which should early exit of the pipeline
  - This also mean we don't have to supply any more information since all releases should have tags (hopefully!)

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -15,7 +15,7 @@ steps:
     key: "is_released"
     branches: "main"
     # we assume that all releases with have a tag and that tag with have a @ version format
-    if: build.tag =~ /@(\d+\.\d+\.\d+)/
+    if: build.tag =~ /@(\d+\.\d+\.\d+)/ && build.tag =~ /^(?:(?!canary).)*$/
 
   - wait
 
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -11,8 +11,17 @@ x-defaults: &defaults
         run: build
 
 steps:
+  - name: ":sleuth_or_spy: Checking for release tag"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yaml
```diff
@@ -11,8 +11,17 @@ x-defaults: &defaults
         run: build
 
 steps:
+  - name: ":sleuth_or_spy: Checking for release tag"
+    key: "is_released"
+    branches: "main"
+    # we assume that all releases with have a tag and that tag with have a @ version format
+    if: build.tag =~ /@(\d+\.\d+\.\d+)/
```
gyfchong, line 18: Would canary's be caught up in this?

mcwinter07: That's a good point, could potentially have a negative assertion for the string "canary" and check like this:
```
build.tag =~ /@(\d+\.\d+\.\d+)/ && build.tag =~ /^(?:(?!canary).)*$/
```

mcwinter07: Which should work:
Screenshot 2023-11-23 at 9 56 48 am
Screenshot 2023-11-23 at 9 58 28 am

mcwinter07: I'd combine them together but the regex looks pretty gnarly 😆 
`(?:(?!canary)|@(\d+\.\d+\.\d+)).*$`

gyfchong (approved):

gyfchong (approved):

PR #4339, opened 2023-11-23
chore: run jest tests using turbo

HeartSquared:
## Why

Speed up running tests by using turbo.

## What

- Consolidate base jest config between packages
- Consolidate base tsconfig between packages
- Update jest setup to throw error on console error/warn
	- https://www.benmvp.com/blog/catch-warnings-jest-tests/
- Fix tests
	- ContextModal was spreading some props when they should've been destructured, so a small patch to fix these


package.json
```diff
@@ -10,9 +10,8 @@
   "scripts": {
     "build": "yarn turbo build",
     "clean": "yarn turbo clean && yarn storybook:clean && rimraf node_modules",
-    "test": "jest && yarn test:pkg:aio",
-    "test:pkg:aio": "yarn pkg:aio test && yarn pkg:aio build",
```
HeartSquared, line 14: Oops accidentally deleted this. Does anyone use it (nothing automated calls it)?

gyfchong: We won't need it if turbo works

gyfchong (approved):

PR #4350, opened 2023-11-23
fix(VisuallyHidden): export in components barrel file

BrendanTimmons:
## Why
Fix `<VisiallyHidden/>` component not being exported. We want to use it for [this ticket](https://cultureamp.atlassian.net/jira/software/c/projects/TCP/boards/577?selectedIssue=TCP-812) on our anytime feedback project.

## What
- Add export to the components barrel file.
- Additionally I saw the list was organised alphabetically so i just ran a vim `:sort` on the list to make sure my export was alphabetical, but it also sorted a few others that weren't correctly sorted.


dougmacknz (approved):
ty

PR #4354, opened 2023-11-24
fix(KaizenProvider): await document before rendering ToastNotificationsList

HeartSquared:
## What

Add a `useEffect` to await document to exist before rendering the `ToastNotificationsList`.

## Why

`KaizenProvider` contains the `ToastNotificationsList` render, which must be awaited in Next since it is SSR where `document` does not exist on first paint.

https://stackoverflow.com/questions/24647839/referenceerror-document-is-not-defined-in-plain-javascript

### Notes

Considered adjusting our Next template by using `dynamic`, but given it only consumes `KaizenProvider` and not `ToastNotificationsList`, it would prevent the whole app from SSR, which would defeat the point. The alternative would be to pull `ToastNotificationsList` out of `KaizenProvider` and have the template pull it in as `children`, but that's not a great abstraction, and would require repeating in every template, so decided to confine the fix within Kaizen.


packages/components/src/KaizenProvider/KaizenProvider.tsx
```diff
@@ -12,16 +12,30 @@ export type KaizenProviderProps = {
...
+  const [notificationsList, setNotificationsList] = useState<JSX.Element>()
+
+  useEffect(() => {
+    // SSR does not have a document, which is required for ToastNotificationsList.
+    // Await document render before rendering the component.
+    if (document !== undefined) {
+      setNotificationsList(<ToastNotificationsList />)
+      setDocumentIsAvailable(true)
+    }
+  }, [documentIsAvailable])
```
HeartSquared, line 26: Notes:
- If we do not define a dependency array, then we will have an endless amount of renders, which we do not want
- If we define it as `[]`, then the first render will be a step behind and we will not see any notifications which are meant to show on page load

mcwinter07 (approved):
Nice fix 👍

PR #4363, opened 2023-11-27
Test buildkite pipeline

mcwinter07:
## Why
testing if the publish flow works with a correct label

## What
- small arbitrary update in shell script


mcwinter07:
Accidental PR, closing

PR #4367, opened 2023-11-27
Release Multi Select

dougmacknz:
Swooping in to take all the glory from the rest of the team's work.

* Exports the Multi Select from components/index
* Moves the Multi Select stories out of staging and into the components category.
* Fixes an issue with the accessible name of the clear all button. It was hardcoded to `Clear all waffles`. It now uses a reference to itself with the label of `Remove all options from ` and then suffixes the label element for the multi select.
* Uncomments some tests that were waiting for the clear all functionality
* Moves focus to the dialog element when the popover is opened
* Fixes an issue on our future Select: it had focus lock after it was ported to use MultiSelect's Popover component. That's now removed and some tests needed to be updated.


packages/components/src/MultiSelect/_docs/MultiSelect.stories.tsx
```diff
@@ -3,7 +3,7 @@ import { Meta, StoryObj } from "@storybook/react"
 import { MultiSelect, MultiSelectProps } from "../index"
 
 const meta = {
-  title: "Staging/Multi Select",
+  title: "Components/Multi Select",
```
HeartSquared, line 6: This will be the new convention (not sure where it was documented) - we can fix it separately if there's nothing else to fix though.

```suggestion
  title: "Components/MultiSelect",
```

packages/components/src/MultiSelect/_docs/MultiSelect.stickersheet.stories.tsx
```diff
@@ -11,7 +11,7 @@ import { MultiSelect, MultiSelectProps } from "../index"
 const IS_CHROMATIC = isChromatic()
 
 export default {
-  title: "Staging/Multi Select",
+  title: "Components/Multi Select",
```
HeartSquared, line 14:
```suggestion
  title: "Components/MultiSelect",
```

packages/components/src/MultiSelect/subcomponents/MultiSelectToggle/MultiSelectToggle.tsx
```diff
@@ -38,84 +38,89 @@ export const MultiSelectToggle = forwardRef<
...
+                        {label}
+                      </RemovableTag>
+                    </li>
+                  ))}
+                </ul>
+
+                <ClearButton
+                  id={clearAllId}
+                  aria-label="Remove all options from "
+                  aria-labelledby={`${clearAllId} ${ariaLabelledBy}`}
```
HeartSquared, line 110: :O I didn't know this worked. I thought only one of them would be chosen as the label so we had to construct it?

dougmacknz: Yeah interesting one ay! I noticed react-aria was doing it and it seems to work, though maybe I'll go double check on all of our supported browsers.

The reason why I want to do this instead of creating another visually hidden element is that creating another element would add a nonsensical element to the document for screen reader users.

dougmacknz: `aria-labelledby` gets priority over `aria-label`, and I guess browsers are smart enough to avoid the recursive issue and reference the accessible name of itself with the `aria-labelledby` ignored.

image

dougmacknz: Tested in Chrome, Firefox, Edge on Windows and Chrome, Firefox and Safari on Mac and all working as expected

HeartSquared (approved):

packages/components/src/__future__/Select/Select.spec.tsx
```diff
@@ -315,13 +315,14 @@ describe("<Select />", () => {
...
-        const { getByRole } = render(<SelectWrapper defaultOpen />)
+        const { getByRole } = render(<SelectWrapper />)
 
         await user.tab()
+        await user.keyboard("{ArrowDown}")
         await waitFor(() => {
           expect(
             getByRole("option", { name: "Short black", selected: false })
-          ).toHaveFocus()
+          ).toBeVisible()
```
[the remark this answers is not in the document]
dougmacknz: Options don't get real focus in this component. It was only getting focus here because of the focus lock that was added in the migration PR.

No idea why, but the `defaultOpen` was causing these tests to no longer work. It should have been as simple as reverted back to what kaizen-legacy is doing with these tests but something else must have changed in the migration.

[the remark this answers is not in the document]
dougmacknz: The final selection verifies that the correct thing was "focused" no?

I assumed this was just in here so that the test runner didn't press enter before the listbox is visible.

HeartSquared (approved):

PR #4369, opened 2023-11-27
Increase opacity on placeholder in Select

harrisrachel:
## Why
Team Athena is using the Select component in Anytime Feedback. We are running the storybook tests and are addressing all accessibility issues that are raised. The Select had a placeholder opacity of 0.5 (whereas other placeholders were 0.7 in Kaizen) and this was failing the minimum contrast rule.

We've not been able to override this using react-select properties as it's overridden by these styles from what I can see.

## What
1. Change placeholder opacity from 0.5 to 0.7
2. Remove override of a11y test as it should now pass the rule

### DIFF
--- a/.changeset/honest-weeks-sniff.md
+++ b/.changeset/honest-weeks-sniff.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Change placeholder opacity on Select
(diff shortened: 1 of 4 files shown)


HeartSquared:
I think you mean "increase" 😉

harrisrachel:
@HeartSquared Thank you! I was thinking "decrease transparency" when I was writing it I think! But yes, it should be increase opacity.

HeartSquared (approved):

PR #4383, opened 2023-12-01
feat(LinkStyle): add LinkStyle component

HeartSquared:
## What

Add `LinkStyle` component, which follows this proposed pattern https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/2993324684/Improve+Kaizen+Button+and+Link+components+-+Solution+Preview

**Note:** Design review required, as there aren't proper designs in our UI Kit, and after briefly showing Winter and Doug, we made up some accessible styles (accessibility requires that the underline always be shown).

|Default|Reversed|
|---|---|
|image|image|

## Why

P &L fun time! [KDS-2001](https://cultureamp.atlassian.net/jira/software/projects/KDS/boards/289?selectedIssue=KDS-2001)
But also because we don't have one, and every team makes their own Link component, which isn't great.

[KDS-2001]: https://cultureamp.atlassian.net/browse/KDS-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HeartSquared:
Leadership going in a different direction

PR #4394, opened 2023-12-05
docs(MultiSelect): add story variant of inline notification to docs

mcwinter07:
## Why
We released the component prior to the design docs being ready  / announced

## What
- adds an update notification
- adds a story variant of inline notifications to allow for JSX to be passed in as an arg (avoiding the MDX parser styles for p tags)

PR #4395, opened 2023-12-05
docs(components): Remove KAIO inline notification

mcwinter07:
## Why
With all components migrated, KAIO config should be the norm so we no longer need this caveat

## What
- remove KaioNotification


packages/components/src/Avatar/_docs/Avatar.mdx
```diff
@@ -13,7 +13,6 @@ import * as AvatarStories from "./Avatar.stories"
   className="!mb-8"
 />
 
-<KaioNotification />
 
 <Installation
```
HeartSquared, line 17: OOS, but I reckon we could probably make this one simpler too

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
```diff
@@ -12,7 +12,11 @@ import * as MultiSelectStories from "./MultiSelect.stories"
   className="!mb-8"
 />
 
-<KaioNotification />
+<StoryInlineNotification
+  type="negative"
+  persistent
+  content={<span>This component  has not been officially released. Please wait for the an announcement in <code>#update_design_system</code> and for the design guidance to be published to Figma and Confluence before consuming.</span>}
```
HeartSquared, line 18:
```suggestion
  content={<span>This component has not been officially released. Please wait for the an announcement in <code>#update_design_system</code> and for the design guidance to be published to Figma and Confluence before consuming.</span>}
```

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
```diff
@@ -12,7 +12,11 @@ import * as MultiSelectStories from "./MultiSelect.stories"
   className="!mb-8"
 />
 
-<KaioNotification />
+<StoryInlineNotification
+  type="negative"
+  persistent
```
HeartSquared, line 17: I reckon bake this into the component

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
```diff
@@ -12,7 +12,11 @@ import * as MultiSelectStories from "./MultiSelect.stories"
   className="!mb-8"
 />
 
-<KaioNotification />
+<StoryInlineNotification
```
HeartSquared, line 15: Maybe instead of being generic, we should the content part of the component? `<SbNotificationUpcomingRelease>` or something?

storybook/components/DocsContainer/ComponentDocsTemplate/ComponentDocsTemplate.tsx
```diff
@@ -41,7 +40,6 @@ export const ComponentDocsTemplate = ({
       <Content>
         <Title context={context} />
         <Links context={context} />
-        <KaioInstallation context={context} />
```
HeartSquared, line 44: OOS - We should replace usage of this template with MDX and then remove this instead

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
mcwinter07: We could, but we may other use cases outside upcoming release warnings / statements, maybe just `SbNotification`?

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
mcwinter07: I think type should probs be `cautionary`, since `negative` sentiment by default seems a little off, but defo with the persistent!

packages/components/src/MultiSelect/_docs/MultiSelect.mdx
HeartSquared: In that situation, we can make another component. This particular notification has a very reusable nature to it for any upcoming components. If the message was very specific to the component, it would be a little more justified, but it may not even be the same problem.

mcwinter07: I suppose my concern is more around recreating similar components that are, in all but name, an Inline Notification for storybook. That said if there is a high likelihood that we will need more of these in the future and they these would only be used with upcoming releases I don't have an opposition scoping the name

mcwinter07: Also lets move any more comments over to this over to the [PR](https://github.com/cultureamp/kaizen-design-system/pull/4394#discussion_r1414843236), since this is unrelated to removal work done in this PR :)

storybook/components/DocsContainer/ComponentDocsTemplate/ComponentDocsTemplate.tsx
mcwinter07: Will create a ticket after this work 👍

HeartSquared (approved):

HeartSquared (approved):

PR #4396, opened 2023-12-06
fix(MultiSelect Popover): Use :focus-visible instead of :focus for focus ring

dougmacknz:
Having this styling on `:focus` means we get this happening:

https://github.com/cultureamp/kaizen-design-system/assets/1811583/cb748bbb-1f07-4547-9808-b82c88dc7f75

Switching to `:focus-visible`, we still lose focus on the checkbox on mouse down for some reason (happens on our regular CheckboxField as well) but at least it's no longer as obviously janky.

Focus rings are ideally only set on `:focus-visible` as well anyway.

### DIFF
--- a/packages/components/src/MultiSelect/subcomponents/Popover/Popover.module.scss
+++ b/packages/components/src/MultiSelect/subcomponents/Popover/Popover.module.scss
@@ -11,7 +11,7 @@
   background: $color-white;
   overflow: auto;
 
-  &:focus {
+  &:focus-visible {
     outline: none;
     border-color: $color-blue-500;
   }
--- a/packages/components/src/MultiSelect/subcomponents/Popover/Popover.module.scss
+++ b/packages/components/src/MultiSelect/subcomponents/Popover/Popover.module.scss
@@ -11,8 +11,11 @@
   background: $color-white;
   overflow: auto;
 
-  &:focus-visible {
+  &:focus {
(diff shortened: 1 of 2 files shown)


packages/components/src/MultiSelect/subcomponents/Popover/Popover.module.scss
```diff
@@ -11,7 +11,7 @@
   background: $color-white;
   overflow: auto;
 
-  &:focus {
+  &:focus-visible {
     outline: none;
```
HeartSquared, line 15: Should be like this instead right?

```suggestion
  &:focus {
    outline: none;
  }
    
  &:focus-visible {
```

dougmacknz: It didn't seem to be adding an outline on Chrome or Firefox but I've committed your suggestion anyway just in case.

dougmacknz: My guess is that browsers don't do an outline on a div:focus so this line wasn't actually needed anyway, but not bad to have just in case.

HeartSquared (approved):

dougmacknz:
@HeartSquared Had to amend commit because there was a formatting issue

HeartSquared (approved):

PR #4398, opened 2023-12-07
Unpin headlessui/react

dougmacknz:
Looking good to go from the UI perspective, but failing a test for some reason.


gyfchong:
🤷

dougmacknz:
(╯°□°)╯︵ ┻━┻

gyfchong:
Hmm, I narrowed the problem down to possibly Turbo, cause if you run the tests individually, they pass. But if it's through turbo, then they fail...

PR #4405, opened 2023-12-12
chore: replace yarn with pnpm

HeartSquared:
## What

- Replace `yarn` with `pnpm`
- Removed KAIO as peerDep from kaizen/tailwind as it was causing a circular symlink
	- Given KAIO has a devDep on K/TW, perhaps the peer dep should go there instead?
- Clean up dependencies
- Install `playwright` as a devDep

## Why

CA is moving towards using `pnpm`, so let's try and keep up with the times :)

## TO DO
- [x] Add changeset
- [x] Test canary
- [ ] Remove yarn-integrity from required CI

## Notes
- https://github.com/pnpm/pnpm/issues/5522
- There doesn't appear to be an equivalent for `yarn check --integrity`, but it seems that pnpm has a more stable lock file than yarn in general, so maybe we can get away without it?
	- https://github.com/orgs/pnpm/discussions/3202
- https://storybook.js.org/docs/writing-tests/test-runner#test-hook-api
	- `preRender` renamed to `preVisit`
	- `postRender` renamed to `postVisit`


packages/components/src/Avatar/_docs/Avatar.mdx
```diff
@@ -15,7 +15,7 @@ import * as AvatarStories from "./Avatar.stories"
 
 
 <Installation
-  installCommand="yarn add @kaizen/components"
+  installCommand="pnpm add @kaizen/components"
```
gyfchong, line 18: I wonder about this, I think technically the majority of consuming apps are still `yarn`

gyfchong: Also we could probably bake the install command into the component since its the same for them all? Maybe even the import statement could be partially templated

HeartSquared: Yeah I was actually thinking that haha. Maybe it can just list both 🤷🏻‍♀️

packages/tailwind/package.json
```diff
@@ -25,10 +25,9 @@
   "private": false,
   "license": "MIT",
   "dependencies": {
-    "@kaizen/design-tokens": "^10.3.16"
+    "@kaizen/design-tokens": "workspace:*"
```
gyfchong, line 28: I think we can only do this with peers or devDeps, cause wouldn't the dep be published like this and consumers wouldn't get the right version?

HeartSquared: >This feature allows you to depend on your local workspace packages while still being able to publish the resulting packages to the remote registry without needing intermediary publish steps - your consumers will be able to use your published workspaces as any other package, still benefitting from the guarantees semver offers.

https://pnpm.io/workspaces#publishing-workspace-packages

:)

gyfchong: So fancy

packages/components/src/Avatar/_docs/Avatar.mdx
gyfchong: Might be too thick...

HeartSquared: `(yarn|pnpm) add @kaizen/components`? 😂 Will see

gyfchong (approved):

PR #4415, opened 2023-12-18
Fix SSR hydration warning when using Illustration

lisatassone:
## Why
When using the Illustration component either directly or through another component (BrandMoment for instance), you will receive a hydration warning in the console:
image

Without this fix, every consumer that uses a component with VideoPlayer in it needs to dynamically import it into their Next.js app and disable prerendering for it.

## What
This PR moves the source element logic into a `useEffect` so we can ensure that it only runs once on the client - thus keeping the server HTML and initial client render consistent.

I encountered an issue with updating the inline snapshots as that feature is no longer compatible with prettier 3.0. I have converted these to external snapshots after [confirming with the DS team.](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1702857404345209)

Tested it fixed the issue via canary release

### DIFF
--- a/packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.spec.tsx
+++ b/packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.spec.tsx
@@ -65,24 +65,7 @@ describe("<VideoPlayer />", () => {
         />
       )
       const videoPlayer = screen.getByTestId("kz-video-player")
-      expect(videoPlayer).toMatchSnapshot(`
-        <video
-          aria-hidden="true"
-          class="wrapper"
-          data-testid="kz-video-player"
-          muted=""
-          playsinline=""
-          poster="https://d1e7r7b0lb8p4d.cloudfront.net/illustrations/heart/spot/moods-cautionary.svg.png"
-          preload="metadata"
-          tabindex="-1"
-          width="100%"
-        >
-          <source
-            src="https://d1e7r7b0lb8p4d.cloudfront.net/illustrations/heart/spot/moods-cautionary.webm.mp4"
-            type="video/mp4"
-          />
-        </video>
-      `)
+      expect(videoPlayer).toMatchSnapshot()
       expect(mockPause).toBeCalled()
     })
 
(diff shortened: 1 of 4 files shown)


packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.tsx
```diff
@@ -149,6 +150,24 @@ export const VideoPlayer = ({
...
+    /**
+     * This seems counter-intuitive, but webm support is codec specific.
+     * Only offer webm if we are positive the browser supports it.
+     * Reference: https://bugs.webkit.org/show_bug.cgi?id=216652#c1
+     */
+    canPlayWebm() ? (
+      setSourceEl(
+        <>
+          <source src={assetUrl(`${source}.webm`)} type="video/webm" />
+          <source src={assetUrl(`${source}.mp4`)} type="video/mp4" />
```
dougmacknz, line 163: This is meant to have both the webm and mp4 versions?

packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe("<VideoPlayer />", () => {
         />
       )
       const videoPlayer = screen.getByTestId("kz-video-player")
-      expect(videoPlayer).toMatchInlineSnapshot(`
+      expect(videoPlayer).toMatchSnapshot(`
         <video
```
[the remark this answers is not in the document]
lisatassone: Oh of course. WIll update

packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.tsx
lisatassone: This is what the original would have had from the original code and you can have [multiple sources](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#usage_notes) in the video element.

As such I've replicated the functionality of the original implementation - but since we are checking for webm support, it does seem redundant (but maybe we want a fallback anyway?)

dougmacknz: Oh right, my bad I read the old code as an if else. All good 👍

dougmacknz (approved):

PR #4417, opened 2023-12-18
Add readonly input state

haustraliaer:
## Why
Noticed that the email input on the Login screen became `disabled` when stepping through to the choice between Google and SSO. Should be read only instead.

image

Also seems like useful functionality for elsewhere in the platform.

## What
Adds a `readOnly` prop to the Input and TextArea components – flowing up to TextField and TextAreaField as well.

readonly

### DIFF
--- a/.changeset/two-cooks-fetch.md
+++ b/.changeset/two-cooks-fetch.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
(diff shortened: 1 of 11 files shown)


gyfchong:
The guidance from the team has been that this isn't the right way to solve the initial problem, and to build a `dl` in the repo that needs this solution.

PR #4418, opened 2023-12-18
Clean up remnants of TitleBlock's action prop on menu items

dougmacknz:
Bit of a confusing mess here in typical TitleBlockZen fashion:
We had 2x `TitleBlockMenuItemProps` type definitions being exported, one which had an old and required `action` prop.

I've consolidated those into one (living in types.ts) and cleaned up any remnants of the old `action` prop in stories.

### DIFF
--- a/packages/components/src/TitleBlockZen/utils.ts
+++ b/packages/components/src/TitleBlockZen/utils.ts
@@ -1,4 +1,3 @@
-import { MenuItemProps } from "~components/Menu"
 import {
   TitleBlockButtonProps,
   TitleBlockCustomButtonProps,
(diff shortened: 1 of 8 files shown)


packages/components/src/TitleBlockZen/utils.ts
```diff
@@ -8,10 +7,6 @@ import {
   TitleBlockMenuItemProps,
 } from "./types"
 
-export const isMenuItemNotButton = (
```
dougmacknz, line 11: This wasn't being used anywhere

mcwinter07 (approved):
Nice clean up 👍

PR #4422, opened 2023-12-21
Move MultiSelect to __future__

dougmacknz:
https://cultureamp.atlassian.net/browse/KZN-1983


mcwinter07 (commented):
Just with the docs for the MultiSelect, we'll need to update the import statement similar to select.


Screenshot 2023-12-21 at 1 56 11 pm

I also noticed that's not in a future docs like the Select. Not sure if MultiSelect needs a non-future MDX to act as the root or if this is fine for now?

dougmacknz:
Changed our mind, not going to do this in the end.

PR #4469, opened 2024-01-10
feat: add ids to title block zen

louiseliang:
## Why

To manage the focus when close a modal/drawer, ids are required to be exposed in TitleblockZen's subcomponents.

## What

`<MenuItem>`: accept id in props and pass it to child button/link
`<TitleBlockZen>`
  - Add constant ids for TitleBlockZen's secondary menu and other actions menu in mobile view
  - accept id in props to be assigned to the root div


dougmacknz (approved):
Nice

PR #4471, opened 2024-01-10
feat(TitleBlockZen): allow mobile actions menu auto hide in title block

louiseliang:

## Why

When menu item in mobile menu open a drawer when clicked, the focus will still stay within the mobile menu as the focus lock is still engaged. 

Expected behaviour: focus can be released so that it can be moved into opened drawer.

## What

- Pass an optional props to allow the auto hide behaviour for mobile menu:
  `<TitleBlockZen />` - autoHideMobileActionsMenu
  `<MobileActions />` - autoHide
- Follow the approach in `<StatelessMenu />` : append close menu to children elements' on click event.


dougmacknz (approved):
I crown you the new TitleBlock expert 👑

.changeset/mean-sheep-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
```
dougmacknz, line 2: This probably should have been a minor in hindsight, doesn't matter too much.

PR #4507, opened 2024-01-30
Add-volta-settings

gyfchong:
## Why
Volta good, NVM bad
https://docs.volta.sh/guide/

## What
Added a volta config to package json to let volta know what to use


package.json
```diff
@@ -141,5 +141,9 @@
     "turbo": "^1.11.3",
     "typescript": "^5.3.3",
     "webpack": "^5.89.0"
+  },
+  "volta": {
```
HeartSquared, line 145: Do you need to remove `.nvmrc` then?

gyfchong: Nah, they can work independently in the same repo. Also I think NVM is more popular at CA.

HeartSquared: Should we even be committing this change then if no one else is meant to use it 😛

gyfchong: I don't think there's any explicit standard that says we must use NVM... Just so happens to be the most common one found.

HeartSquared (approved):
Can't hurt, right?

PR #4513, opened 2024-02-01
ci: split storybook build scripts for tests vs site

HeartSquared:
## Why

The `--test` flag makes Storybook only have stories without docs. The site needs the reverse.

## What

Split `storybook:build` into two scripts - one for tests, one for the site.


package.json
```diff
@@ -14,7 +14,9 @@
     "storybook": "storybook dev -p 6006 -c storybook",
     "storybook:preview": "NODE_ENV=production pnpm storybook --docs",
     "storybook:clean": "rm -rf storybook/public",
-    "storybook:build": "storybook build --test -c storybook -o storybook/public --docs",
+    "storybook:build": "storybook build -c storybook -o storybook/public",
+    "storybook:build:test": "pnpm storybook:build --test",
+    "storybook:build:prod": "pnpm storybook:build --docs",
```
HeartSquared, line 19: Was contemplating whether to use `:site` instead of `:prod`. Or maybe `:docs`. Can change to whatever makes the most sense.

gyfchong: I think this is fine, its just an internal script being terse is probably better

gyfchong (approved):

PR #4522, opened 2024-02-14
chore(deps): manual deps updates

mcwinter07:
## Why
Creating a PR to manually bump deps while renovate is down

## What
- update deps across all our packages


packages/design-tokens/package.json
```diff
@@ -38,8 +38,8 @@
     "json-to-flat-sass": "^1.0.0",
     "lodash.flatmap": "^4.5.0",
     "object-to-css-variables": "^0.2.1",
-    "postcss": "^8.4.34",
-    "prettier": "*",
+    "postcss": "^8.4.35",
+    "prettier": "^3.2.5",
```
HeartSquared, line 42: Hmm... I don't exactly recall about pnpm workspaces, but should we be updating this from `*`?

mcwinter07: I mean good question, I will say the default behavior an pnpm `up` will do is replace the * with the caret version. Not apposed to it, but do we have a reason for using the wildcard?

mcwinter07: Also noting two other things it does:
-     `"@headlessui/react": "<=1.5.0"` - changes to `"@headlessui/react": "^1.5.0",`
  - that seems iffy
-  ` "prosemirror-utils": "^1.2.1-0",` becomes `"prosemirror-utils": "1.2.1-0",` 
  - odd it removes the caret 🤔

HeartSquared: `<=` is definitely not equivalent to `^` 😱 

I think the wildcard usage is so that it uses the version from root? Unsure though, I could be wrong, but I think that would be the intent.

Removing the carat seems wrong too :(

mcwinter07: > I think the wildcard usage is so that it uses the version from root? Unsure though, I could be wrong, but I think that would be the intent.

Ah ok, I mean that kind of make sense. So for us, all we need to do is update our prettier at the root, which would be the same for consuming repos so there's no double up of different versions

mcwinter07: > Removing the carat seems wrong too :(

Yeah, I think I'll leave those as is 🤔 Wonder if there's config for pnpm so it doesn't force these changes with things like `^` or `<=`

HeartSquared: Yeah something to look into some time I reckon :) But not now 😉

packages/components/package.json
```diff
@@ -49,15 +49,15 @@
...
-    "@react-aria/button": "^3.9.1",
-    "@react-aria/datepicker": "^3.9.1",
-    "@react-aria/focus": "^3.16.0",
-    "@react-aria/i18n": "^3.10.0",
-    "@react-aria/listbox": "^3.11.3",
-    "@react-aria/menu": "^3.12.0",
-    "@react-aria/overlays": "^3.20.0",
-    "@react-aria/select": "^3.14.1",
-    "@react-aria/utils": "^3.23.0",
+    "@react-aria/button": "^3.9.2",
```
HeartSquared, line 52: These ones will need a changeset o3o

HeartSquared (approved):

PR #4526, opened 2024-02-21
refactor(major!): add a prefix for Tailwind classes

kaydwithers:
## Why
This PR adds a `kz-` prefix to Tailwind classes.

This prefix prevents clashes with Tailwind classes from other apps.

## What
Here is a document explaining the issue in more depth.
https://cultureamp.atlassian.net/wiki/spaces/TV/pages/3450274547/Solving+TailwindCSS+namespace+issues+with+shared+components


gyfchong:
Spoke to @sentience and we're looking at a different solution to the problem. Apologies we have to close this.

sentience:
For @kaydwithers and future readers, here's what we discussed:

All of the Tailwind styles used in the Kaizen codebase (and prefixed in this PR) are actually in the Storybook docs for Kaizen. The fact that these styles are distributed in the KAIO package is actually a bug! The KAIO distribution CSS bundle should only contain styles required by Kaizen components (which currently contains zero Tailwind styles).

So rather than add prefixes to the styles that should not be in the KAIO CSS output anyway, the Kaizen team will instead fix this issue as a short-term priority. To do this, the Tailwind configuration used to generate the CSS bundle for the KAIO package will exclude the Kaizen Storybook/docs directories, so it effectively will not contain any Tailwind CSS output.

Eventually, we do foresee the possibility that some Kaizen components will make use of Tailwind styles. If and when this happens we expect to need to introduce a Tailwind CSS prefix, at which time the changes proposed in this PR would make sense to bring in. But in he meantime, the Kaizen team prefers to retain the ability maintain its docs without having to type Tailwind prefixes.

HeartSquared:
> Ping @HeartSquared in case I got any of the above wrong.

Pretty much on the money Kevin :) We'll look at releasing a short-term fix where we remove the TW styles from our dist, and then later on (pending roadmap priorities) we'll investigate a proper solution for distributing KAIO TW styles with a prefix.

PR #4527, opened 2024-02-22
chore: update vscode settings

HeartSquared:
## What and Why

VS Code keeps making this change itself whenever I update. It probably knows better what it wants.


gyfchong (approved):
Not all heroes wear capes

PR #4528, opened 2024-02-23
chore: dependency updates

HeartSquared:
Dep updates in `package.json`


HeartSquared:
Renovate has revived and created a rollup with the same changes

PR #4557, opened 2024-03-04
remove compiled tailwind from kaio

mcwinter07:
## Why

Distributing Tailwind CSS in our bundle is causing several issues in consuming repos

## What

We need to update our bundle config to exclude tailwind from our distributed styles, i.e:
```
// currently has compiled tailwind from @kaizen/components
@import "@kaizen/components/dist/styles.css";

@tailwind base;
@tailwind components;
@tailwind utilities;
```

The styles in /dist currently will contain compile tailwind. We need it to only contain the compiled CSS of our components.

This task is to remove tailwind from our bundle, NOT from the entire repo as it still has utility for our storybook.

## Changes

- The tailwind documentation template (formerly: `UtilityClassTemplate`) has been moved out of the `System/tailwind` folder and into `storybook/components`
  - There are a number doc components and templates that use tailwind at the moment so I've allowed `storybook/components` as a content source tailwind, i.e: `content: ["./**/*.{stories.tsx,mdx}", "./storybook/components/**/*.tsx"]`
- Components with tailwind in their class names have been replaced with CSS (or a component where possible)
  - This includes:  Slider, Workflow's subcomponents, GenericTile, MultiActionTile
- turbo config for `@kaizen/components#build` no longer depends on `@kaizen/components#build`
- removed tailwind directives from global.css

## Checklist:

- [x]     Tailwind must not be in our final CSS bundle
- [x]     Tailwind must be able to be consumed in our storybook (stories should remain unchanged)
- [x]     Tailwind must be able to be passed by our components in repos
- [x]     Update documentation to outline that tailwind is not support within components


packages/components/tailwind.config.js
```diff
@@ -3,7 +3,7 @@
 const { Preset } = require("@kaizen/tailwind")
 
 module.exports = {
-  content: ["./src/**/*.tsx"],
+  content: ["./src/**/*.{stories.tsx,mdx}"],
```
HeartSquared, line 6: Perhaps we don't even need this file at all and just rely on the global one to cater for all the stories (haven't checked, unsure if true)?

HeartSquared:
Probably worth adding to the checklist a little documentation somewhere that we can't use TW in any distributed code in the meantime (we'd need to solve the prefix thing if we were to do it).

packages/components/tailwind.config.js
mcwinter07: I axed it with no issues :) looks like we're in the clear for `@kaizen/components`

mcwinter07:
I've checked  in a consuming repo with a canary release and Tailwind still works as expected, so that's a green light to review this PR 👍

mcwinter07:
Oh, calling out one thing I didn't solve - inhibiting Tailwind intellesense on our components, i.e: in Button. I thought updating the content source would do this automagically but sadly not - Might be worth making a ticket to tract that as a potential task later, but didn't want to block this update going out

gyfchong:
> inhibiting Tailwind intellesense on our components

Does this mean intellisense will still suggest Tailwind class names?

HeartSquared (approved):

PR #4561, opened 2024-03-08
Add new ThumbUp and ThumbsDown icon set

mcwinter07:
## Why
- Request for a new set of thumbs up and down with on and off states

## What
- Adds new thumbs up and down SVGs with on and off state to assets.
  - These have been manually modified to remove the mask that was causing clipping issues.
    - The investigation as to why the Figma export included this, which is seemingly unnecessary is not in scope of this ticket.
- Removes old thumbs up and down SVGs from assets.
  - This will not remove the component but does mean it won't be recreated when we run the update script.
  - The `.tsx` component from these now have a `@deprecated` doc block pointing to the new ones.
- Adds ts-node as a dev dep for Kaizen/components so we can run the update-icons script without a global dependency.
- Some clean up of the Icon reference story.
  - Add a basic filter to remove deprecated components from the story
  - Removes broken search that doesn't bring a lot of functionality.
  - Updates icon cards to make them more readable in their alignment.

## Out of scope
- Investigating out why the Figma export for these icons needed to be manually fixed
  - See this [comment](https://cultureamp.slack.com/archives/C05D8V4GALX/p1709787885978789?thread_ts=1709786164.612159&cid=C05D8V4GALX) in this [thread](https://cultureamp.slack.com/archives/C05D8V4GALX/p1709786164612159) for more
- Refactoring the reference page to look more like the grouping in Figma
  - This is worth revisiting when / if we take a deeper dive into icons and can collab more closely with design
- Updating icon internals outside of the ThumbsUp and Down variants
  - We should run the update icons as a separate script to let SVGo update the internals for all icons to reduce the risk against missed changes in chromatic


mcwinter07:
Noting here that there's many minor changes outline in the diff for all other icon components. This is likely a result of SVGO being updated since it was last run, which has removed some more SVG bloat. Based on the chromatic snapshot, this does  visually impact or change the behavior of our icons

package.json
```diff
@@ -134,6 +134,7 @@
     "stylelint-config-standard-scss": "^13.0.0",
     "tailwindcss": "^3.4.1",
     "ts-jest": "^29.1.2",
+    "ts-node": "^10.9.2",
```
HeartSquared, line 137: Given the script lives in `package/components`, this dev dep should actually belong there.

packages/components/src/Icon/AddLinkIcon.tsx
```diff
@@ -10,7 +10,7 @@ export const AddLinkIcon = (props: IconProps): JSX.Element => {
     <>
       <path
         fill="currentColor"
-        d="M9.125 14.167H5.833c-1.153 0-2.135-.406-2.948-1.219-.812-.813-1.218-1.795-1.218-2.948s.406-2.135 1.218-2.948c.813-.813 1.795-1.219 2.948-1.219h3.292v1.75H5.833c-.694 0-1.27.23-1.729.688-.458.458-.687 1.034-.687 1.729 0 .695.229 1.271.687 1.729.459.459 1.035.688 1.729.688h3.292zm-2.458-3.292v-1.75h6.666v1.75zm4.208 3.292v-1.75h3.292c.694 0 1.27-.23 1.729-.688.458-.458.687-1.034.687-1.729 0-.695-.229-1.271-.687-1.729-.459-.459-1.035-.688-1.729-.688h-3.292v-1.75h3.292c1.153 0 2.135.406 2.948 1.219.812.813 1.218 1.795 1.218 2.948s-.406 2.135-1.218 2.948c-.813.813-1.795 1.219-2.948 1.219z"
+        d="M9.125 14.167H5.833q-1.729 0-2.948-1.219T1.667 10t1.218-2.948q1.219-1.22 2.948-1.219h3.292v1.75H5.833q-1.04 0-1.729.688-.687.687-.687 1.729t.687 1.729q.688.688 1.729.688h3.292zm-2.458-3.292v-1.75h6.666v1.75zm4.208 3.292v-1.75h3.292q1.04 0 1.729-.688.687-.687.687-1.729t-.687-1.729q-.688-.688-1.729-.688h-3.292v-1.75h3.292q1.729 0 2.948 1.219T18.333 10t-1.218 2.948q-1.219 1.22-2.948 1.219z"
```
HeartSquared, line 13: I'd hope these mass changes are only related to you installing `ts-node`, otherwise if svgo decides it wants to change the code every so often for no good reason, it opens us up to missing things changing. Even if we have Chromatic, as shown in the latest snapshot, it's flagged for changes anyway due to you adding new icons to the list, so if any of the icons after it did change, it can easily be missed thinking it was just a layout shift.

Personally I wouldn't have committed these changes in this PR, but instead have it separately so we can ensure it's consistent (or if it's just some once off change that reverts the next time we run the script) and safe.

packages/components/src/Icon/_docs/Icon.stories.tsx
```diff
@@ -63,52 +63,49 @@ const ReferenceButton = ({
...
         )}
       </button>
     </li>
   )
 }
 
-export const Reference: Story = {
-  render: () => {
-    const [searchTerm, setSearchTerm] = useState<string>("")
+const deprecatedList: string[] = ["ThumbsUpIcon"]
```
HeartSquared, line 84: I assume there should be a ThumbsDown one too?

packages/components/src/Icon/_docs/icon.stories.scss
```diff
@@ -16,9 +16,9 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  justify-content: center;
-  gap: var(--spacing-16);
-  padding: var(--spacing-8);
+  justify-content: flex-start;
+  gap: var(--spacing-8);
+  padding: var(--spacing-16);
```
HeartSquared, line 21: Perhaps debatable, but I feel like the balance seems off when I look at the new styles. I see they're more aligned though, but the spacing gaps look a bit wrong (likely related to the fixed width/height).

Maybe if you're playing with visuals, using `display: grid` might be better than flex :)

packages/components/src/Icon/AddLinkIcon.tsx
mcwinter07: I feel like its more likely its SVGo trying to optimize the SVG further rather than ts-node. But I take you're point about the risk 🤔 Might exclude it and do a follow up PR to update

packages/components/src/Icon/_docs/Icon.stories.tsx
mcwinter07: yes... must do that one too 😆

package.json
mcwinter07: Fair point 👍

packages/components/src/Icon/_docs/icon.stories.scss
mcwinter07: I've updated it a bit using grid and removed the set height. It still looks a little off when you have a row where some break over multiple lines and some don't but I'd rather they all start at the same reading position than jump around like they use to.

## before
Screenshot 2024-03-12 at 10 49 48 am

## after
Screenshot 2024-03-12 at 10 50 28 am

HeartSquared: Much better 😄 I think it was just the uneven spacing caused by the set height which was throwing me off :)

packages/components/src/Icon/_docs/icon.stories.scss
```diff
@@ -12,15 +12,12 @@
...
   @include button-reset;
-
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
-  gap: var(--spacing-16);
-  padding: var(--spacing-8);
+  display: grid;
+  grid-template-rows: min-content 1fr;
```
HeartSquared, line 16: Oops my bad! Didn't see that there was actually a `.grid` controlling the grid - thought that `.button` was the grid.
So technically it does make sense for this to be display flex, but the height not being a set value is still valid.

Nonetheless, up to you if you want to change it or not - visually it looks much better than before already (and it's internal docs, so I'm not so fussed) :)

mcwinter07: All breezy, I've revert part of this back and leveraged the upper grid to control the width of the columns

packages/components/src/Icon/AddLinkIcon.tsx
mcwinter07: I dropped the old commit and re-added these :) should be a much small change

packages/components/src/Icon/_docs/Icon.stories.tsx
```diff
@@ -63,52 +63,53 @@ const ReferenceButton = ({
...
+      <ul className={styles.grid}>
+        {Object.keys(ICONS)
+          .filter(iconName => {
+            if (
+              deprecatedList.find(deprecatedIcon => deprecatedIcon === iconName)
+            ) {
+              return
+            }
+            return iconName
+          })
```
HeartSquared, line 102: This should be enough since `filter` is based on booleans :)
```suggestion
          .filter(iconName => !deprecatedList.includes(iconName))
```

packages/components/src/Icon/_docs/icon.stories.scss
```diff
@@ -13,14 +13,12 @@
 .button {
   @include button-reset;
 
-  display: flex;
+  display: inline-flex;
```
HeartSquared, line 16: Might need to be `flex` instead of `inline-flex`? The layout in the docs have changed

packages/components/src/Icon/_docs/Icon.stories.tsx
mcwinter07: Cleaner, I like it 👍

packages/components/src/Icon/_docs/icon.stories.scss
mcwinter07: That won't make any change. The main diff is the gaps between the cards are even because the columns are now controlled with a set value as apposed to 1fr - to me its neater

HeartSquared (approved):

PR #4604, opened 2024-04-05
feature: Allow label and trigger props in FilterBarMultiSelect

maiconmborges:

## Why
`FilterBarMultiSelect` types were not allowing a `label` or a `trigger` to be passed, even though they were being spread. Also, the label prop was not being used at all.

## What
We are now allowing the `label` to be customised, but also making sure we are still using the filter's name in the remove button's tooltip. This way, we can achieve something like this:

https://github.com/cultureamp/kaizen-design-system/assets/4048660/3c25bcb6-7914-4bf4-b8b7-6e19dfe78fe5

As part of this PR, we are now also localising the `Remove filter` tooltip text.

### DIFF
--- a/packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
+++ b/packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
@@ -4,6 +4,7 @@ import {
   FilterMultiSelect,
   FilterMultiSelectProps,
   ItemType,
+  MenuTriggerProviderContextType,
   getSelectedOptionLabels,
 } from "~components/Filter/FilterMultiSelect"
 import { useFilterBarContext } from "../../context/FilterBarContext"
@@ -11,14 +12,11 @@ import { checkArraysMatch } from "../../utils/checkArraysMatch"
 
 export type FilterBarMultiSelectProps = Omit<
   FilterMultiSelectProps,
-  | "isOpen"
-  | "setIsOpen"
-  | "renderTrigger"
-  | "label"
-  | "selectedKeys"
-  | "trigger"
+  "isOpen" | "selectedKeys" | "label" | "trigger"
 > & {
   id?: string
+  label?: string
+  trigger?: (value?: MenuTriggerProviderContextType) => React.ReactNode
 }
 
 // This should technically be handled within the FilterMultiSelect
@@ -44,6 +42,7 @@ export const FilterBarMultiSelect = ({
   items: propsItems,
   children,
   onSelectionChange,
+  label,
   ...props
 }: FilterBarMultiSelectProps): JSX.Element | null => {
   const { getFilterState, setFilterOpenState, updateValue, hideFilter } =
@@ -94,12 +93,13 @@ export const FilterBarMultiSelect = ({
                 items
               )
             : [],
-          label: filterState.name,
+          label: label || filterState.name,
         }
 
         return filterState.isRemovable ? (
           <FilterMultiSelect.RemovableTrigger
             {...triggerProps}
+            removeButtonTooltipText={filterState.name}
             onRemove={() => hideFilter(id)}
           />
         ) : (
--- a/packages/components/src/Filter/FilterButton/FilterButtonRemovable/FilterButtonRemovable.tsx
+++ b/packages/components/src/Filter/FilterButton/FilterButtonRemovable/FilterButtonRemovable.tsx
@@ -1,4 +1,5 @@
 import React, { forwardRef } from "react"
+import { useIntl } from "@cultureamp/i18n-react-intl"
 import { ButtonGroup, ButtonGroupProps } from "~components/ButtonGroup"
 import { FilterTriggerRef } from "~components/Filter/Filter"
 import { ClearIcon } from "~components/Icon"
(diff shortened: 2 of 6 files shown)


packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -4,21 +4,19 @@ import {
...
-  | "setIsOpen"
-  | "renderTrigger"
-  | "label"
-  | "selectedKeys"
-  | "trigger"
+  "isOpen" | "selectedKeys" | "label" | "trigger"
 > & {
   id?: string
+  label?: string
+  trigger?: (value?: MenuTriggerProviderContextType) => React.ReactNode
```
HeartSquared, line 19: You should be able to do this :)

```suggestion
  label?: FilterMultiSelectProps["label"]
  trigger?: FilterMultiSelectProps["trigger"]
```

packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -94,12 +93,13 @@ export const FilterBarMultiSelect = ({
                 items
               )
             : [],
-          label: filterState.name,
+          label: label || filterState.name,
```
HeartSquared, line 96: If we are to allow overriding, we'd want it to flow onto all the instances of `filterState.name`, so you'll want to ensure L77 ~~and L102~~ also follow the pattern (you can make a variable).

Any usages which want the labels to differ (such as your use case) will need to be a customisation of the `trigger` prop by the consumer instead of being the default :)

packages/components/src/Filter/FilterButton/FilterButtonRemovable/FilterButtonRemovable.tsx
```diff
@@ -25,12 +26,20 @@ export const FilterButtonRemovable = forwardRef<
...
+  const removeButtonLabel = formatMessage(
+    {
+      id: "filterButtonRemovable.tooltipText",
+      defaultMessage: "Remove filter - {filter}",
+      description: "Label for the tooltip used in a filter remove button",
+    },
+    {
+      filter: removeButtonProps?.tooltipText || triggerButtonProps?.label,
+    }
+  )
```
HeartSquared, line 42: I can't say I'm too familiar with `formatMessage`, but is this the equivalent of writing this?

```
const removeButtonLabel =
    removeButtonProps?.tooltipText ??
    formatMessage({ id: "filterButtonRemovable.tooltipText" ... })
```

If so, what's the benefit of the format? Just wondering since it makes `formatMessage` always get called as opposed to only when `tooltipText` doesn't exist.

Also noticed that the optional chaining on `triggerButtonProps.label` isn't necessary since it's actually a required field.

packages/components/src/Filter/FilterMultiSelect/subcomponents/Trigger/RemovableFilterTrigger/RemovableFilterTrigger.tsx
```diff
@@ -7,13 +7,15 @@ import { FilterTriggerButtonProps } from "../FilterTriggerButton"
 
 export type RemovableFilterTriggerProps = FilterTriggerButtonProps & {
   onRemove: () => void
+  removeButtonTooltipText: string
```
HeartSquared, line 10: This should be optional.

packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
```diff
@@ -94,12 +93,13 @@ export const FilterBarMultiSelect = ({
...
         }
 
         return filterState.isRemovable ? (
           <FilterMultiSelect.RemovableTrigger
             {...triggerProps}
+            removeButtonTooltipText={filterState.name}
```
HeartSquared, line 102: We don't want to be overriding the default value from this component - it should be a consumer customisation if need be.

```suggestion
```

HeartSquared:
Worth adding some tests too to ensure it works as expected :)

packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
maiconmborges: We need this to ensure the name of the filter is the one actually being used in the tooltip of the remove button. If a user selects `Filter 1` and we pass a label `These are your choices: `, we should, by default, have the tooltip saying `Remove filter - Filter 1` and not `Remove filter - These are your choices`.

packages/components/src/Filter/FilterButton/FilterButtonRemovable/FilterButtonRemovable.tsx
maiconmborges: The way I made it is we would always have `Remove filter` by default in the message, even when passing a tooltip text, but I just realised it might be to strict.

maiconmborges: Actually, this is necessary for the change I intended. If we pass a `tooltipText`, which is the case of the `FilterBarMultiSelect` passing down the filter name instead of the custom label (explained [here](https://github.com/cultureamp/kaizen-design-system/pull/4604#discussion_r1552733743)).

packages/components/src/Filter/FilterBar/subcomponents/FilterBarMultiSelect/FilterBarMultiSelect.tsx
maiconmborges: https://github.com/cultureamp/kaizen-design-system/pull/4604#discussion_r1552733743

PR #4615, opened 2024-04-09
Update VideoPlayer component to be SSR friendly

mcwinter07:
## Why
- The helper to check for WebM relied on `window`, which was not available during the Next SRR cycle

## What
- wraps the helper in a useEffect

### DIFF
--- a/packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.tsx
+++ b/packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.tsx
@@ -152,16 +152,16 @@ export const VideoPlayer = ({
     }
   }, [videoRef])
 
-  const pausePlay = usePausePlay(videoRef)
+  useEffect(() => {
+    // SSR does not have a window, which is required for canPlayWebm.
+    if (window !== undefined) setWindowIsAvailable(true)
+  }, [])
 
   useEffect(() => {
-    // SSR does not have a window, which is required for for the canPlayWebm helper.
-    // Await window render before rendering the component.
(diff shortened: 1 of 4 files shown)


HeartSquared (dismissed):

packages/components/src/Illustration/subcomponents/VideoPlayer/VideoPlayer.tsx
```diff
@@ -151,6 +154,15 @@ export const VideoPlayer = ({
...
   const pausePlay = usePausePlay(videoRef)
 
+  useEffect(() => {
+    // SSR does not have a window, which is required for for the canPlayWebm helper.
+    // Await window render before rendering the component.
+    if (windowIsAvailable !== undefined) {
+      setIsWebmCompatible(canPlayWebm())
+      setWindowIsAvailable(true)
+    }
+  }, [windowIsAvailable])
```
gyfchong, line 164: I don't think its performant to have a useEffect rely on a state that it itself is updating, so probably best to use separate Effects for individual checks, one to detect window and one to check for webm.

This way one effect can change the window state and the other is waiting for that window state before checking webm.

Another way could be to have one effect for the window and just do an inline check within the render:
`isWindowAvailable && canPlayWebm() && ( ... <source>)`

HeartSquared: Oops, yeah it actually should be `if (window !== undefined)`, not `if (windowIsAvailable)` 😅

gyfchong (approved):

PR #4616, opened 2024-04-09
Add focus lock to Menu

dougmacknz:
## Why
Addresses two accessibility issues:
- Adds 'Esc' to close functionality.
- Restores focus back to the menu button after an interaction from the menu. This even works if the menu action opens a modal first. Focus will go back to the menu button after the modal closes, which has been a pain point in the past, with some people manually implementing fixes.

https://github.com/cultureamp/kaizen-design-system/assets/1811583/96ae967d-a6e1-4a2d-9a54-7edc99364ce9

## What
Adds focus lock to the `MenuDropdown` component in a very similar way to what we did on `MultiSelect`: the button is included as part of the items able to be focused inside the lock.


gyfchong:
To simplify ongoing debugging, we're going to wait for 2 fixes to go out in separate versions before merging this

gyfchong (approved):

PR #4642, opened 2024-04-19
Add migration docs for future Tag

mcwinter07:
## Why
Given the change in the API between legacy and `future` Tag, migration docs are required to assist team migrate. 

## What
- Adds migration documentation for the future Tag with examples


tailwind.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     "./storybook/components/**/*.tsx",
   ],
   presets: [Preset],
+  important: "body",
```
mcwinter07, line 14: I realized that classNameOerride with Tailwind utility classes did not have specificity and was no applying in the story without `!`'s if the style was previously declared. I think this is just in the way we consume tailwind in our repo not a reflection of specificity problems in how we ship our components, which this config won't impact

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+While the `LiveIcon` is exported from `@kaizen/components` the legacy `Tag` used an animated implementation, which is not currently available. As an interim solution you can use the following story's code snippet to create a `LiveIconComponent`. This can then be passed into as `Children` of the `Tag`.
+
+<Canvas of={TagMigrationStories.LiveIconComponentStory} sourceState="hidden"/>
+
+### Validation
+
+Following the same pattern as status, `validationPositive`, `validationInformative`, `validationNegative`, `validationCautionary` can all be replace with the `color` and `icon` prop.
+
+Note: because the icons convey meaning, they should have an accessible label that provides context to the tag. A contextual `aria-label`and role of `img` addresses one of the [accessibility issues](https://github.com/cultureamp/kaizen-discourse/issues/102) with the legacy Tag.
```
HeartSquared, line 58:
```suggestion
Note: Because the icons convey meaning, they should have an accessible label that provides context to the tag. A contextual `aria-label` and role of `img` addresses one of the [accessibility issues](https://github.com/cultureamp/kaizen-discourse/issues/102) with the legacy Tag.
```

Rather than link to the a11y issue, I'd probably just provide an example snippet with the correct expected code.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+<Canvas of={LegacyTagStories.Validation} />
+
+#### After
+
+<Canvas of={TagMigrationStories.ValidationMigration} sourceState="shown"/>
+
+
+Note: When using the `icon` prop, the icon's color will be determined by `color` value that has been set. This is to ensure that and accessible contrast ratio is achieved.
+
+Any Icons added as a child of the `Tag` will need to be manually set to ensure the right contrast ratio.
```
HeartSquared, line 71:
```suggestion
Any Icons added as `children` of the `Tag` will need to be manually set to ensure the right contrast ratio.
```

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+<Canvas of={LegacyTagStories.Sentiments} />
+
+#### After
+
+<Canvas of={TagMigrationStories.SentimentsMigration} sourceState="shown"/>
+
+If required, you could use `classNameOverride` to implement the same UI. It is important that any override of the default styles still meets minimum contrast ratios to ensure the component is accessible.
+
+<Canvas of={TagMigrationStories.SentimentNone} />
```
HeartSquared, line 87: Perhaps the correct suggestion is to actually use one of the provided colours. Check with the designers though :)

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+#### Before
+
+<Canvas of={LegacyTagStories.Dismissible} />
+
+#### After
+
+<Canvas of={TagMigrationStories.DismissibleMigration} sourceState="shown"/>
+
+Remember that when handling dismiss (or any destructive action), a users focus should not be lost on the page.
```
HeartSquared, line 104:
```suggestion
Remember that when handling dismiss (or any destructive action), a user's focus should not be lost on the page.
```

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+<Canvas of={LegacyTagStories.Inline} />
+
+#### After
+
+<Canvas of={TagMigrationStories.InlineMigration} sourceState="shown"/>
+
+
+### Truncation
+
+This was original controlled by the component and set the `max-width` of a component via inline styles. This can now be easily controlled with Tailwind and was not needed in the new API.
```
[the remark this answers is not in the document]
mcwinter07: I can probably elaborate on some guidance here - ie: make sure it is screen readable and that you are not hiding important information. I don't know how realistic it is to say out and out don't do it, but could definitely elaborate on this

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+#### After
+
+<Canvas of={TagMigrationStories.DismissibleMigration} sourceState="shown"/>
+
+Remember that when handling dismiss (or any destructive action), a users focus should not be lost on the page.
+
+
+### Avatars
+
+Avatars are no longer provided directly to the `Tag` as a prop. Instead they will now me passed in as children.
```
HeartSquared, line 109:
```suggestion
Avatars are no longer provided directly to the `Tag` as a prop. Instead pass them through `children`.
```

Worth writing a note on the a11y requirements? I believe Avatars in Tag are usually presentational only, so we should ensure the correct attributes are added.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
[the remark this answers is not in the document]
mcwinter07: Would be better to push guidance to a central doc on ellipsis and a11y guidance if we know of one. The reason I didn't go hard on this is because it wasn't raised as an a11y issue in the intopia report

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+#### Before
+
+<Canvas of={LegacyTagStories.ProfileWithAvatarProps} />
+
+#### After
+
+<Canvas of={TagMigrationStories.AvatarMigration} sourceState="shown"/>
+
+
+Note: The use of `ps-4` (or `padding-inline-start`) in the `classNameOverride` is to override the default spacing at start of the `Tag`. This will ensure the Avatar can contour the border radius of the `Tag`.
```
HeartSquared, line 120: This is definitely an indication that we should be including the Avatar within our component. But it's fine in the interim.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+#### Dismissible Avatars
+
+Following a similar pattern to the above, you can use the `RemovableTag` with an `Avatar` and plain text as its children to achieve a dismissible version.
```
HeartSquared, line 124:
```suggestion
Following a similar pattern to the above, you can use the `RemovableTag` with an `Avatar` and plain text as its `children` to achieve a dismissible version.
```

I don't know whether `plain text` is required to be mentioned. Technically it can accept a ReactNode right?

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+<Canvas of={TagMigrationStories.AvatarRemovableMigration} />
+
+### Inline
+
+Originally the `Tag` was built with the assumption that it would always required spacing at the end. This necessitated the need for the `inline` prop to remove the margin from end when it was not required.
+
+In favour of simplifying the component, `inline` and the default `margin-inline-end` has been removed. It is recommended now to control spacing between tags to use a `gap` of 12 pixels on the container of the tags, ie: `gap-12`. In instances where more specific margins are require, opt for `margin-inline-end` and `classNameOverride` on the tag to ensure RTL support is baked in.
```
HeartSquared, line 132: I think there's too much technicality mentioned here. I'm seeing so many attributes and classes that I can't figure out what is relevant.

Perhaps just outline that we no longer include a margin in the component, so spacing should be resolved by the consumer in a container style.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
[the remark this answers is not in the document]
HeartSquared: I agree. Truncation isn't an accessible pattern, so it shouldn't be recommended to be done.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+<Canvas of={LegacyTagStories.Truncate} />
+
+#### After
+
+<Canvas of={TagMigrationStories.TruncateMigration} sourceState="shown"/>
+
+### Size
+
+The two sizes for tag have been remove in favour of a default size in between `medium` and `small`. The `small` specifically was removed due to the reduced hit radius for the dismissible button and poor compatibility with avatars and icon.
```
HeartSquared, line 157:
```suggestion
The two sizes for tag have been removed in favour of a default size in between `medium` and `small`. The `small` specifically was removed due to the reduced hit radius for the dismissible button and poor compatibility with avatars and icon.
```

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+`classNameOverride` can be leveraged to add padding and modify properties but should be done with caution to reduce inconsistencies in the platform and reduce the risk of accessibility violations/
```
HeartSquared, line 159:
```suggestion
If you must use an old size, `classNameOverride` can be leveraged to add padding and modify properties.
```

I'd probably mention a11y in the previous sentence instead.

packages/components/src/__future__/Tag/Tag/_docs/Tag.mdx
```diff
@@ -52,3 +52,11 @@ in LTR, and on the right of `children` in RTL. For a list of icons, see [Icons](
 Allows classnames to be attached to the outermost element of Tag.
 
 <Canvas of={TagStories.ClassNameOverride} />
+
+### Migrating from legacy tag
```
HeartSquared, line 56:
```suggestion
### Migrating from legacy Tag
```

HeartSquared: I'd just link to the other doc instead of listing things in here.

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+  ExclamationIcon,
+  InformationIcon,
+  SuccessIcon,
+  LiveIcon,
+} from "~components/Icon"
+import { Tag, RemovableTag } from "../../"
+import LegacyStyles from "./TagLegacy.module.scss"
+
+const meta = {
+  title: "Components/Tag/Future/Migration Guide",
```
HeartSquared, line 16:
```suggestion
  title: "Components/Tag/Migration Guide",
```

I think? Feels like the guide should be on the top level, but not too fussed.

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+import { Avatar } from "~components/Avatar"
+import {
+  CautionIcon,
+  ExclamationIcon,
+  InformationIcon,
+  SuccessIcon,
+  LiveIcon,
+} from "~components/Icon"
+import { Tag, RemovableTag } from "../../"
+import LegacyStyles from "./TagLegacy.module.scss"
```
HeartSquared, line 13:
```suggestion
import legacyStyles from "./TagLegacy.module.scss"
```

Shouldn't the styles actually exist in `Tag.module.scss` already though?

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+  args: {
+    children: "My tag",
+  },
+  parameters: {
+    docs: {
+      source: {
+        type: "dynamic",
+      },
+    },
+  },
```
HeartSquared, line 27: Shouldn't be needed here unless you do want all the sources to be dynamic

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+  </style>
+        `,
+      },
+    },
+  },
+}
+
+export const StatusMigration: StoryObj = {
+  render: () => (
+    <div className="flex gap-12">
```
HeartSquared, line 127: Should use a decorator for this container 😉

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+    </div>
+  ),
+}
+
+export const ValidationMigration: StoryObj = {
+  render: () => (
+    <div className="flex gap-12">
+      <Tag
+        color="green"
+        icon={<SuccessIcon role="img" aria-label="Success," />}
```
HeartSquared, line 144: The commas look kinda strange? Is it necessary?

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+            avatarSrc="https://www.cultureampcom-preview-1.usw2.wp-dev-us.cultureamp-cdn.com/assets/slices/main/assets/public/media/chapters-card-1@2x.05e547444387f29f14df0b82634bf2b6.png"
+          />
+          Tag
+        </span>
+      </RemovableTag>
+    </div>
+  ),
+}
+
+export const InlineMigration: StoryObj = {
```
HeartSquared, line 283: I don't think any of these look like optimal implementations? Just one clean example should be enough.

Something like this (assuming gap-12 is the correct size, otherwise whatever it should be according to guidelines).
```tsx
<div className="flex gap-12">
  <Tag>Tag</Tag>
  <Tag>Tag</Tag>
  <Tag>Tag</Tag>
</div>
```

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+          </span>
+        </Tag>
+      </div>
+    </>
+  ),
+}
+
+export const SizesMigration: StoryObj = {
+  render: () => <Tag>Tag</Tag>,
+}
```
HeartSquared, line 321: Don't forget that StoryObj is usually used like this :)

```suggestion
export const SizesMigration: StoryObj = {
  args: {
    children: "Tag"
  }
}
```

Or if `children` is pre-defined in the Meta, it can just be this:
```suggestion
export const SizesMigration: StoryObj = {}
```

tailwind.config.js
HeartSquared: Yeah our Storybook TW is flawed 😬 That's a different task to tackle though haha.

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: Basing it on the intopia report, from what I gather it reads as "Success, Tag" to some screen readers

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: When you say optimal implementation, do you mean not in a`ul` with `li`'s? Granted they probably should be since it is a list of tags, or do you mean more use gap as a spacing mechanism?

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: Yeah, always forget about that one ^ will update 👍

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: good call out, there's a few I can add that in for 👍

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: How about something more like:
```
<>
  <ul className="flex m-0 p-0 gap-12 list-none">
    <li>
      <Tag>
        Tag
      </Tag>
    </li>
    <li>
      <Tag>
        Tag
      </Tag>
    </li>
    <li>
      <Tag>
        Tag
      </Tag>
    </li>
  </ul>
</>
```

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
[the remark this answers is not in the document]
mcwinter07: I think there room for pause on saying truncation is not accessible - there is an accessible way to do it and probably valid reasons, i.e: if someone has a long name and that's in a tag. It does need to be handled properly to pass WCAG criteria (reveal on focus or reveal on activation like click) but its not a flat no from my understanding of the standard.

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
HeartSquared: Just the use of `gap-12`. I'd remove the `m-0 p-0` from the snippet since that's related to consumer layout placement.

It's a bit long, so I'd try to reduce the number of lines :) No issues with you using a list though 👍🏻

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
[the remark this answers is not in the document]
mcwinter07: I suppose the thing I'm really latching onto is this line ([in this doc](https://www.w3.org/WAI/WCAG21/Understanding/resize-text)):

> truncation is acceptable if the component's full content is available on focus or after user activation and an indication that this information can be accessed

Granted this is above resizing containers with zoom, but I wonder if that still stand 🤔 @dougmacknz may know more / have some handy dandy docs I can refer to, but will pester him next week to confirm :)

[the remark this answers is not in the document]
HeartSquared: I think if you're going to give an example migration path, it might as well be something actually usable that meets a11y standards. Just the closest solution you can think of as a suggestion.

Otherwise I'd be looking at the docs thinking you've told me the old pattern is bad, but that the new one is just as bad.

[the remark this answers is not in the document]
mcwinter07: Yeah, the Tailwind implementation of a potentially problematic truncation is still an issue, just wasn't sure if I should be trying to solve that here or not, or just provide an example of where to start before they make it properly accessible - Maybe having the examples is a bad idea and should just leave it as why its not in the new API?

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: I've updated the examples to just use the gap 12 in the after but had to leave the other tailwind in for the UL. I didn't think we'd need them but looks like our reset doesn't apply to our Stories

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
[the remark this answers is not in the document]
mcwinter07: I'm going to axe the "after"  Story and just provide some guidance on why it shouldn't be done 👍

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
gyfchong: As much as I like to create semantically correct examples as a sort of guide, I don't think we need to demonstrate it in this case. The focus should be on the `gap` TW class affording the same functionality and we should keep it at that. Especially if we need to remove the focus for other things to be true.

packages/components/src/__future__/Tag/Tag/_docs/MigrationGuide.mdx
gyfchong: I agree, I don't think we need to expound this much. I think we leave the recommended method to the example

packages/components/src/__future__/Tag/Tag/_docs/Tag.mdx
gyfchong:
```suggestion
### Migrating to Future Tag
```
Probably better to rephrase it like this, so we remove the Legacy

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
gyfchong:
```suggestion
  title: "Components/Tag/Migration to Future",
```

packages/components/src/__future__/Tag/Tag/_docs/Tag-migration-guide.mdx
```diff
@@ -0,0 +1,170 @@
...
+#### Before
+
+<Canvas of={LegacyTagStories.Inline} />
+
+#### After
+
+<Canvas of={TagMigrationStories.InlineMigration} sourceState="shown"/>
+
+
+### Truncation
```
gyfchong, line 145:
```suggestion
### truncationWidth
```
Probably want to keep the headings to the properties

packages/components/src/__future__/Tag/Tag/_docs/TagMigration.stories.tsx
mcwinter07: I'll strip it back to the basics with Gap and just add a small comment on lists to leave it in our consumers hands

gyfchong (approved):

PR #4647, opened 2024-04-22
Add focus ring styles to TiltleBlockZen NavifationTabs

mcwinter07:
## Why
This fix is to resolve the standing accessibility issue with the current implementation reliance on outline and default browser focus rings (see issue [here](https://github.com/cultureamp/kaizen-discourse/issues/99))

Screenshot 2024-04-22 at 4 31 00 pm

Screenshot 2024-04-22 at 4 30 55 pm

## Context
The `TitleBlockZen` appeared to have no focus styles declared so a browser default was applied to both the breadcrumbs and the tabs. On safari, this failed minimum contrast ratios for accessibility. The breadcrumbs focus work will be split into a separate PR.

## Proposed changes
This is the proposed change (to be clarified with design)
Screenshot 2024-04-22 at 4 30 46 pm

## Other considerations
Should we be aligning this (if possible) to our other [Tabs](https://cultureamp.design/?path=/docs/components-tabs--docs) focus strategy?

Screenshot 2024-04-22 at 4 41 46 pm

## Update on focus state - 23/04

After catching up with Jess we decided that it would be best to replicate the focus ring (as close as possible) to the modern Tab component. This should meet contrast ratio requirements and create a more cohesive experience for users.

focus-desktop

focus-small-med

Noting here that the TitleBlockZen navigation tabs have less padding than our other Tabs so the focus ring will be a little narrower. I think this is fine as a it will only show on focus - update the padding may cause a issue is blowing out content in consuming repos if we try to aim for an exact replica.

## What
- removed the default `outline` for the focus state and adds in a 'similarly' focus ring similar to [Tabs](https://cultureamp.design/?path=/docs/components-tabs--docs) component


packages/components/src/TitleBlockZen/_docs/TitleBlockZen.stories.tsx
```diff
@@ -97,76 +98,30 @@ export const Playground: Story = {
   },
 }
 
-// export const WithBadge: Story = {
```
mcwinter07, line 100: These were a bunch of commented out stories, I figure we don't need them any more

packages/components/src/TitleBlockZen/_docs/TitleBlockZen.stories.tsx
```diff
@@ -97,76 +98,30 @@ export const Playground: Story = {
...
+      canvas: {
+        sourceState: "shown",
+      },
+    },
+    pseudo: {
+      hover: [
+        '#tab-hover-example [class^="TitleBlockZen-TitleBlockZen-module__navigationTabsList"] li:nth-child(2) a',
+      ],
+      focus: [
+        '#tab-focus-example [class^="TitleBlockZen-TitleBlockZen-module__navigationTabsList"] li:nth-child(2) a',
```
mcwinter07, line 113: Our usual strategy is to use data attributes but given TitleBlock doesn't accept them, I've used ID's instead for the the examples and selector

packages/components/src/TitleBlockZen/_docs/TitleBlockZen.stories.tsx
```diff
@@ -97,76 +98,30 @@ export const Playground: Story = {
...
+      ],
+    },
+  },
+  render: args => (
+    <StickerSheet>
+      <StickerSheet.Row rowTitle="Tab hover">
```
mcwinter07, line 119: I've added basic coverage for the hover and focus state. I'll need to to the same thing with breadcrumbs so I figured I'd chuck this into a stickersheet for chromatic

gyfchong (approved):

PR #4656, opened 2024-04-29
Kzn 2294 bundler package + style injection

gyfchong:
## Introduction
This PR introduces the new package `@kaizen/package-bundler ` which is a tool that bundles packages to ESM/CJS as well as optionally CSS via the CLI.

## Why
To enable teams to consistently distribute Tier 3 UI Libraries which are compatible with the rest of Culture Amp engineering stack.

## What
A package which compiles Typescript, React and CSS Modules (SCSS/CSS/Tailwind) into ESM/CJS bundle.


packages/design-tokens/package.json
```diff
@@ -21,11 +21,12 @@
   "private": false,
   "license": "MIT",
   "sideEffects": false,
+  "type": "module",
   "scripts": {
     "test": "FORCE_COLOR=1 jest",
     "test:ci": "pnpm test -- --ci",
     "build": "pnpm clean && pnpm prepublishOnly",
-    "build:json": "pnpm ts-node --files ./bin/buildCssVarTokens.ts",
+    "build:json": "tsx ./bin/buildCssVarTokens.ts",
```
gyfchong, line 29: `ts-node` decided to stop working for some reason, switched to the better `tsx`

packages/components/package.json
```diff
@@ -26,20 +26,11 @@
     "styles.css"
   ],
   "scripts": {
-    "prepublishOnly": "pnpm build:components && pnpm build:styles && pnpm postBuild && pnpm dist:clean",
-    "build": "pnpm clean && pnpm prepublishOnly",
+    "build": "pnpm package-bundler build-shared-ui && pnpm styles:global",
+    "styles:global": "postcss ./styles/global.css --output dist/styles.css",
```
jakubriedl, line 30: generally across tools we use a pattern for commands [like here](https://github.com/cultureamp/next-template/blob/main/apps/app-template/package.json#L10-L11) which in this case could be `build:styles` as previously or if you want to highlight it's just the global styles `build:styles:global` or `build:global-styles`

packages/design-tokens/package.json
jakubriedl: good choice, we should switch to `tsx` everywhere. `ts-node` is full of problems

packages/package-bundler/README.md
```diff
@@ -0,0 +1,196 @@
...
+You will also need to add the compiled Tailwind stylesheet to `sideEffects` in your `package.json` to ensure it is not tree-shaken:
+```json
+"sideEffects": [
+  "tailwind.css.*"
+]
+```
+
+### Alias
+
+If you are using aliases, ensure you have them listed in your `tsconfig.json` (the `tsconfig.dist` and `tsconfig.types` should extend this) and in `rollup.config`.
```
jakubriedl, line 168: 🤔 for libs like `i18n-react-intl` and `frontend-apis` use aliases. Do we need to list them in here? Should we export from them a `tsconfig.json` that people can just extend so it configures it?

packages/package-bundler/src/presets/shared-ui/bin/addTailwindStylesImport.ts
```diff
@@ -0,0 +1,87 @@
...
+  DIST_DIRS.forEach(dir => {
+    ;(async () => {
+      try {
+        addGlobalStylesImport(`${args.packagePath}/${dir}`)
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error("Directory not found", e)
+      }
+    })()
+  })
```
jakubriedl, line 84: you can cleanup this code a fair bit
```suggestion
  DIST_DIRS.map(dir => 
    addGlobalStylesImport(`${args.packagePath}/${dir}`)
      .catch(e => console.error("Directory not found", e))
  )
```

jakubriedl: also we shouldn't have a no-console lint for `error` and `warn` levels. These are legit

jakubriedl (commented):
great work, looks good. Just few nit comments

packages/package-bundler/README.md
HeartSquared: Locales is catered for within the rollup settings provided. It's one of the exceptions I suppose because the i18n set up is opinionated in its own way.

That being said, we currently have it mapped to `locales` with the assumption that the KAIO implementation had followed the rules. We can adjust it otherwise if that's not the case 😬 

The aliases we mention are for local development, so it depends with the frontend-apis one. Might be better to just have it in the template? 🤔

packages/package-bundler/src/presets/shared-ui/bin/addTailwindStylesImport.ts
HeartSquared: Not sure if it still needs to be async or not. I'll check.

Actually we didn't want people putting console anything into components because it was creating noise that we received complaints for. But given this is a script, it can probably be listed as an exception instead. I'll look into it.

turbo.json
```diff
@@ -24,9 +24,13 @@
...
     "@kaizen/components#build": {
-      "dependsOn": ["@kaizen/design-tokens#build"],
+      "dependsOn": [
+        "@kaizen/design-tokens#build",
+        "@kaizen/package-bundler#build"
+      ],
       "inputs": ["src/**", "rollup.config.mjs", "tsconfig*.json"],
-      "outputs": ["dist/**"]
+      "outputs": ["dist/**"],
+      "cache": false
```
HeartSquared, line 33: Reminder: Remove this

turbo.json
```diff
@@ -40,6 +44,11 @@
     "@kaizen/design-tokens#build": {
       "inputs": ["css/**", "less/**", "sass/**", "src/**", "tokens/**"],
       "outputs": ["dist/**"]
+    },
+    "@kaizen/package-bundler#build": {
+      "inputs": ["bin/*", "src/**", "tsconfig*.json"],
+      "outputs": ["dist/**"],
+      "cache": false
```
HeartSquared, line 51: Reminder: Remove this

HeartSquared (approved):
🔥

jakubriedl (approved):
great work team

PR #4660, opened 2024-04-30
Add onAfterEnter callback to Modals 

mcwinter07:
## Why
This update to the modals is a more holistic way of addressing the issue raised by Will regarding the RTE's Link modal (see thread [here](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1713422268454219))

## Original issue overview
https://github.com/cultureamp/kaizen-design-system/assets/36558508/ebd5d18d-559f-4935-96bb-60bf5669b9c1

## Background
Initially, to resolve this we were going to intercept the onKeyDown event for a `paste` and shift focus to the input. This was more a monkey patch that a long term solution. Instead we opted to provide an `onAfterEnter` callback that is triggered when the Modal is opened.

## Maintaining a11y
The default behaviour that we have baked into our modals is to focus on the accessible title. This is to ensure that screen reader users have context to where there focus has been shifted. This guard rail still exists if the onAfterEnter callback does not shift focus to a child of the modal. 

Additional documentation has been added to the InputEditModal docs since this is where it would be likely used most.

## What
- Adds `onAfterEnter` callback to GenericModal
  - `onAfterEnter` to other modals to allow for callbacks to be consumed
- Update RTE's LinkModal to shift focus after open 
- Adds storybook play tests to GenericModal to simulate the focus management

### DIFF
--- a/packages/components/src/Modal/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/Modal/ConfirmationModal/ConfirmationModal.tsx
@@ -39,6 +39,9 @@ export type ConfirmationModalProps = {
   title: string
   onConfirm?: () => void
   onDismiss: () => void
+  /** A callback that is triggered after the modal is opened. */
+  onAfterEnter?: () => void
+  /** A callback that is triggered after the modal is closed. */
   onAfterLeave?: () => void
   confirmLabel?: string
   dismissLabel?: string
@@ -99,6 +102,7 @@ export const ConfirmationModal = ({
   title,
   onConfirm,
   onAfterLeave,
+  onAfterEnter,
   confirmLabel = "Confirm",
   dismissLabel = "Cancel",
   confirmWorking,
@@ -134,6 +138,7 @@ export const ConfirmationModal = ({
       onEscapeKeyup={onDismiss}
       onOutsideModalClick={onDismiss}
       onAfterLeave={onAfterLeave}
+      onAfterEnter={onAfterEnter}
     >
       <div className={styles.modal} data-modal {...props}>
         <ModalHeader onDismiss={onDismiss}>
--- a/packages/components/src/Modal/GenericModal/GenericModal.tsx
+++ b/packages/components/src/Modal/GenericModal/GenericModal.tsx
@@ -14,6 +14,8 @@ export type GenericModalProps = {
   onEscapeKeyup?: (event: KeyboardEvent) => void
   onOutsideModalClick?: (event: React.MouseEvent) => void
   onAfterLeave?: () => void
+  /** Override default focus to the modal title and shift focus to this element after the modal is opened */
+  onOpenFocusTo?: HTMLElement | null
 }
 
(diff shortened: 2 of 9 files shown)


mcwinter07:
Noting here that we'd probably want a story for guidance, maybe a test to validate focus shift - but wanted to check-in with folks on the idea before going further into the weeds

gyfchong:
So we can track what this was a fix for, can you record a video of a before and after and attach it here and the ticket?

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -14,6 +14,8 @@ export type GenericModalProps = {
   onEscapeKeyup?: (event: KeyboardEvent) => void
   onOutsideModalClick?: (event: React.MouseEvent) => void
   onAfterLeave?: () => void
+  /** Override default focus to the modal title and shift focus to this element after the modal is opened */
+  onOpenFocusTo?: HTMLElement | null
```
jakubriedl, line 18: wouldn't be better to pass just `onOpen` callback and allow the consumer to implement any behaviour they want, including focus? And we can have a story example how to use that for focus.

dougmacknz: Would much prefer to have a sensible default for accessibility with an escape hatch for the odd case. 9 times out of 10 we want focus to go to the heading, but most engineers & designers don't know that and will think it should just go to the first input because they don't understand the screen reader implications.

jakubriedl: totally we should have sensible default. Would having generic `onOpen` interface prevent that?

packages/components/src/Modal/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -39,6 +39,9 @@ export type ConfirmationModalProps = {
   title: string
   onConfirm?: () => void
   onDismiss: () => void
+  /** A callback that is triggered after the modal is opened. */
+  onAfterEnter?: () => void
```
mcwinter07, line 43: I've gone with `onAfterEnter` instead of `onAfterOpen` to keep in line with the convention in this component  since we have a `onAfterEnterHandler` which is then passed into `afterEnter` in the Transition component

packages/components/src/Modal/GenericModal/GenericModal.tsx
mcwinter07: I've updated the description of the PR but to quickly cover it here / hopefully allay concerns, I've gone with providing a callback that a consumer can user. If its called there's now a check that run in the `focusOnAccessibleLabel` function to see if focus has been shifted within the modal layer, if not shift focus to the accessible title as default. Going for a best of both worlds scenario here :)

jakubriedl (approved):

dougmacknz (approved):
nice

PR #4662, opened 2024-05-06
Expose ref for IconButton

vietanhtran16:

## Why

In conversations-ui, we are using `Menu` `MenuItem` and `IconButton` for our meatball. This meatball allows user to switch to edit mode and once user have either decided to cancel or submit the edit, we want to return their focus back to the IconButton of the Menu
image
image

## What

To enable the above, we are exposing the ref of IconButton so that we can focus on it

### DIFF
--- a/.changeset/angry-ants-know.md
+++ b/.changeset/angry-ants-know.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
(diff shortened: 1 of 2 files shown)


mcwinter07 (approved):
Thanks for the contribution 👍  I'll trigger a release an holler when that's ready to go

PR #4665, opened 2024-05-08
fix(RTE): duplicate id, added optional id

MrBaptastic:

## Why
While reviewing a11y on RTE, I noticed a duplicate id issue, It was also difficult to target the RTE without being able to control the id, so I have added an optional id props to override the generated Id inside the RTE component, this will allow to use `describedBy` or `labelledBy` on components that are related to the RTE

## What
- Fixed duplicate Id on RTE
- Added optional id prop

Before :
Screenshot 2024-05-08 at 12 34 55 pm

After :
Screenshot 2024-05-08 at 12 33 18 pm

### DIFF
--- a/.changeset/new-adults-rescue.md
+++ b/.changeset/new-adults-rescue.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+fixed duplicate id on RTE, added optional id to RTE
--- a/packages/components/src/RichTextEditor/RichTextEditor/RichTextEditor.tsx
+++ b/packages/components/src/RichTextEditor/RichTextEditor/RichTextEditor.tsx
@@ -87,12 +87,12 @@ export const RichTextEditor = ({
   status = "default",
   ...restProps
 }: RichTextEditorProps): JSX.Element => {
-  const reactId = useId()
+  const generatedId = useId()
   const [schema] = useState<ProseMirrorModel.Schema>(
     createSchemaFromControls(controls)
   )
 
-  const [editorId] = useState<string>(id || reactId)
+  const editorId = id || generatedId
   const labelId = labelledBy || `${editorId}-rte-label`
   const validationMessageAria = validationMessage
     ? `${editorId}-rte-validation-message`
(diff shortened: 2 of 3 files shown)


.changeset/new-adults-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
dougmacknz, line 2: Patch for fixes, minor for new features

```suggestion
"@kaizen/components": patch
```

dougmacknz: Oh, I guess you're adding the ability to add an id, but lets just do a patch

packages/components/src/RichTextEditor/RichTextEditor/RichTextEditor.tsx
```diff
@@ -89,8 +91,9 @@ export const RichTextEditor = ({
   const [schema] = useState<ProseMirrorModel.Schema>(
     createSchemaFromControls(controls)
   )
-  const [labelId] = useState<string>(labelledBy || reactId)
-  const [editorId] = useState<string>(reactId)
+
+  const [editorId] = useState<string>(id || reactId)
```
dougmacknz, line 95: As you mentioned, no need to have this in state anymore

dougmacknz, line 95: While you're here, `reactId` is a bad var name, mind renaming to `editorId` or similar?

MrBaptastic: Cant use editorId since its already the name of the state, I was contemplating renaming it `generatedId`

dougmacknz: You don't need the state, so you just can do `editorId` instead of `reactId` and remove this line.

MrBaptastic: The `editorId` can either be the generated Id from useId() or the prop Id, so we still need both

dougmacknz: You can just do `const editorId = id || useId()` I believe, but this is fine

dougmacknz (approved):

PR #4667, opened 2024-05-09
consumer to pass in plugin presets themselves

gyfchong:
Kinda makes it clearer to consumers?

### DIFF
--- a/packages/package-bundler/src/presets/default/plugins-default.ts
+++ b/packages/package-bundler/src/presets/default/plugins-default.ts
@@ -1,0 +1,34 @@
+import alias from "@rollup/plugin-alias"
+import { babel, getBabelOutputPlugin } from "@rollup/plugin-babel"
+import { InputPluginOption } from "rollup"
+import ignore from "rollup-plugin-ignore"
+import nodeExternals from "rollup-plugin-node-externals"
+
+export const presetDefault = [
+  nodeExternals({
+    devDeps: true,
+  }),
+  // This call to alias plugin will be additional to the above alias plugin call
+  alias({
+    entries: [
+      // i18n-react-intl package attempts to import locales from this path.
+      // When rollup attempts to import from the 'find' path, it will be
+      // redirected to import from the replacement path (same as storybook webpack config).
+      {
+        find: "__@cultureamp/i18n-react-intl/locales",
+        replacement: "locales",
+      },
+    ],
+  }),
+  // These libraries aren't used in KAIO, and require polyfills to be set up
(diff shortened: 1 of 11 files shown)


packages/package-bundler/src/presets/default/plugins-default.ts
```diff
@@ -0,0 +1,40 @@
...
+      {
+        find: "__@cultureamp/i18n-react-intl/locales",
+        replacement: "locales",
+      },
+    ],
+  }),
+  // These libraries aren't used in KAIO, and require polyfills to be set up
+  // in consuming repos. Ignoring them here removes the need for extra setup in
+  // consuming repos.
+  ignore(["stream", "http", "https", "zlib"]),
```
jakubriedl, line 32: should we include also other Nodejs modules similarly as we do for [example here](https://github.com/cultureamp/unified-home/blob/master/packages/next-storybook/src/addons/sensibleDefaults/preset.ts#L53-L66)?

packages/package-bundler/src/rollup.ts
```diff
@@ -1,70 +1,25 @@
...
-      ignore(["stream", "http", "https", "zlib"]),
-      babel({ babelHelpers: "bundled" }),
-      getBabelOutputPlugin({
-        plugins: [
-          "@babel/plugin-transform-react-pure-annotations",
-          "babel-plugin-pure-static-props",
-        ],
-      }),
-    ],
+    plugins: [...config.plugins, ...presetDefault({ alias: config.alias })],
```
jakubriedl, line 22: any reason not to pass whole config to preset?
```suggestion
    plugins: [...config.plugins, ...presetDefault(config)],
```

gyfchong: Not technically, but I don't want to pass anything it doesn't need. Admittedly the lines are blurring here, cause if we send in all config, then we may as well pull all the config into this file again?

jakubriedl: My take is that it's matter of separating concerns/coupling
- you want the default preset to be in seperate file so it's more self sufficient
- also you don't want to assume from here what that file needs from the config

jakubriedl: also when I think about it, the default preset should be inside the ui library preset. So it allows consumer to eject from default if they need to

gyfchong: Hm yeah makes sense, so there's 3 options:
- UI Library preset (inherits default preset)
- Default Preset
- None

jakubriedl: yes

PR #4668, opened 2024-05-09
Kzn 2334/table a11y fixes

mcwinter07:
## Why
This PR addresses several accessibility issue with the existing table implementation in Kaizen, reported in the last round of Intopia and during development of these fixes. While we accept that the Table in Kaizen is an imperfect solution, this PR seeks address as many of these accessibility issues without the need for a rebuild or breaking change.

## a11y issues
- Focus indicators on table headings and row links are browser default.
- Wherever there is a table with checkboxes, the checkboxes aren’t linked to anything, resulting in inputs with no label linked.
- Expandable pattern creates invalid HTML and likely issues for screen readers - consideration for the Tier 3.
- Icon headings passing the labetText into the SVG title, which was not accessible to screen readers

## Fixes
- Added default focus ring widths and colors to interactive  variants of the `TableHeaderRowCell` and interactive table cards
- Added a `checkboxLabel` prop as a mean to resolving the unlinked checkbox label in the `checkable` variant of the `TableHeaderRowCell`
  - I opted not to use the `labelText` as a default for this since it would not communicate to a user what it's intention was. This I think should explicitly be called out, i.e: `Select all *Resource name*`. This way translations and intent can be handled by the consumer.
- Updated the `TableHeaderRowCell` to pass labelText in the an aria-label for the `icon` variant
- Update the documentation with some clearer guidance on the APIs and sub components
  - I left these all as a one pager but am not appose to splitting them into individual pages, I just didn't want to spend too long fussing over the IA of this old component

### DIFF
--- a/packages/components/src/Table/Table.module.scss
+++ b/packages/components/src/Table/Table.module.scss
@@ -185,6 +185,28 @@ $row-height-data-variant: 48px;
     will-change: box-shadow, border-color, margin, padding, width;
   }
 
+  &:focus-visible {
+    outline: none;
+    position: relative;
+    &::after {
+      // This offset provide enough gap on on reverse for contrast ratios
+      $focus-ring-offset: calc(#{$border-focus-ring-border-width} + 2px);
+
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      width: calc(100% + #{$focus-ring-offset});
+      height: calc(100% + #{$focus-ring-offset});
+      content: "";
+      position: absolute;
+      background: transparent;
+      border-color: $color-blue-500;
+      border-width: $border-focus-ring-border-width;
+      border-style: $border-focus-ring-border-style;
+      border-radius: inherit;
+    }
+  }
+
   &.well {
     margin-top: $spacing-sm;
   }
--- a/packages/components/src/Table/Table.module.scss
+++ b/packages/components/src/Table/Table.module.scss
@@ -20,10 +20,6 @@ $row-height-data-variant: 48px;
   &:focus {
     text-decoration: none;
   }
-
-  &.headerRowCellButtonReversed {
-    color: $color-white;
-  }
 }
 
 // Special Table-only button reset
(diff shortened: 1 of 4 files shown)


mcwinter07:
One thing to call out that's missing from here is a sticker sheet and chromatic tests for the UI states like focus, hover, etc. Given this PR addresses some of those I'm curious if folks think we should have them here - noting that during the migration of Table to Kaio we decided not to invest time into it given the age and dubious status of this component

gyfchong:
Probably worth it now that we have them working, kinda like writing tests to me.

packages/components/src/Table/Table.module.scss
```diff
@@ -20,10 +20,6 @@ $row-height-data-variant: 48px;
   &:focus {
     text-decoration: none;
   }
-
-  &.headerRowCellButtonReversed {
```
mcwinter07, line 24: This never existed so icons would have been dark purple on a purple bg - does make me question if this was ever a problem or even if tables are used on reverse backgrounds? Noting they can also set color on the icon itself so maybe was never a problem?

packages/components/src/Table/Table.module.scss
```diff
@@ -80,19 +77,55 @@ $row-height-data-variant: 48px;
 .headerRowCell .headerRowCellTooltip {
   display: flex;
   align-items: stretch;
+  max-width: 100%;
 }
 
-// overflow has to be set at this level as well as on the heading for some reason ¯\_(ツ)_/¯
```
mcwinter07, line 85: the overflow hidden on the tooltip layer was obscuring the focus ring - this is a less intrusive way to let the width of the container be used to set the ellipsis for long labelText

packages/components/src/Table/Table.tsx
```diff
@@ -56,6 +57,14 @@ export type TableHeaderRowProps = {
 const ratioToPercent = (width?: number): string | number | undefined =>
   width != null ? `${width * 100}%` : width
 
+export type TableHeaderRowCellCheckboxProps = {
```
mcwinter07, line 60: Given how many checkbox related type there were, made sense to group these together. I've still left the `checkboxLabel` as optional though so as to provide a more accessible approach but not throw errors in existing implementations

packages/components/src/Table/Table.tsx
```diff
@@ -111,13 +124,8 @@ export const TableHeaderRowCell = ({
   wrapping = "nowrap",
   align = "start",
   tooltipInfo,
-  // If set, this will hide the tooltip exclamation icon. Useful in situations where
```
mcwinter07, line 114: Turned useful parts of these comments into type docs

packages/components/src/Table/_docs/Table.mdx
```diff
@@ -23,31 +23,61 @@ A table displays rows of data, including all data in a set, making it efficient
 
 <Canvas of={TableStories.Playground} />
 
-## API
+## TableContainer API
```
mcwinter07, line 26: Keeping this as a one pager but I've sectioned the props descriptions to each table component for easier digestion

packages/components/src/Table/_docs/Table.stickersheet.stories.tsx
```diff
@@ -0,0 +1,701 @@
+import React from "react"
```
mcwinter07, line 1: It's not technically using the sticker sheet since we'd be fighting against the styles, but this is still used to show the many states so I figured I'd keep with the parlance

gyfchong (approved):
Looking great!

PR #4679, opened 2024-05-15
chore: clean up tsconfig and add types to package-bundler

HeartSquared:
## Why

Our root tsconfig was shared with packages, often giving false errors as our packages used ESM while the root (for Storybook) used CommonJS.

## What

- Separate tsconfig for packages to be more self-contained
- Delete root tsconfig.dist
- Update root tsconfig to use module ESNext and moduleResolution Bundler
- Add TS linting per package
- Include types in dist for package-bundler


jakubriedl (approved):

gyfchong:
https://cultureamp.atlassian.net/browse/KZN-2362

packages/package-bundler/tsconfig.base.json
```diff
@@ -9,6 +9,7 @@
     "jsx": "react",
     "noImplicitAny": true,
     "noImplicitReturns": true,
-    "strict": true
+    "strict": true,
+    "noEmit": true
```
gyfchong, line 13: How come this needs to be declared at the same time as the flag?

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "lib": ["es2019", "dom"],
-    "module": "commonjs",
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
```
gyfchong, line 5: Fancy~

packages/package-bundler/tsconfig.base.json
HeartSquared: The bundler is consumed by others, so we want to ensure it has the flag by default :)
Our flags are a coincidence per se, but put in to ensure that linting doesn't generate any files 😬

gyfchong: Oh wrong place to comment this, pretty sure I saw this in our other packages, ah well no worries

PR #4680, opened 2024-05-15
Update tooltip guidance

mcwinter07:
## Why
The existing pattern for the Kaizen Tooltip's has some known issues when wrapping components or elements that lack a semantic `role` or interactivity (i.e: `link`, `button`, `onclick`). 

In these instances the `Tooltip` content is not accessible to screen reader and is only visually available to mouse or keyboard users.

## What
- Adds guidance on Tooltip compatibility and common pitfalls wrapping Kaizen components


packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
```diff
@@ -12,6 +21,13 @@ const meta = {
   args: {
     text: "Example tooltip text.",
   },
+  decorators: [
+    Story => (
+      <div className="flex mt-[50px] justify-center gap-12">
+        <Story />
+      </div>
+    ),
+  ],
```
jakubriedl, line 30: this isn't the right way to center story. Doing it messes up with screenshots, responsiveness, some docs addons and other things. You can use `layout: "centered"`  instead. https://storybook.js.org/docs/configure/story-layout#component-layout

packages/components/src/Tooltip/_docs/Tooltip.mdx
```diff
@@ -61,3 +61,73 @@ Again, try to keep the contents of your children as simple as possible, ideally
 
 When your direct child is not semantic, you will get a console warning like this:
 `<Tooltip /> is not directly wrapping a semantic element, screen reader users will not be able to access the tooltip info.`
+
+
+### Tooltips, Kaizen components and avoiding common pitfalls
+
+While technically any Kaizen component can be wrapped by a `Tooltip`, as called out previously, only those with accessible roles will be readable to assistive technologies.
```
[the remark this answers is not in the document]
mcwinter07: Sure can certainly link out to the MDN docs since that has a pretty comprehensive list

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
mcwinter07: ah nice, was just updating the original one that used inline styles so this should reduce the bloat further :) Cheers for the tip 👍

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
```diff
@@ -12,6 +21,16 @@ const meta = {
   args: {
     text: "Example tooltip text.",
   },
+  decorators: [
+    Story => (
+      <div className="flex mt-[60px] gap-12">
+        <Story />
+      </div>
+    ),
+  ],
```
jakubriedl, line 30: you can remove the decorator now when you've set layout

jakubriedl (approved):

packages/components/src/Tooltip/_docs/Tooltip.stories.tsx
mcwinter07: We still need the decorator to add some whitespace of each story so the Tooltip floats to the top instead of the sides. Could probably have the flex and gap specific to the button story but figured there was no harm since it won't impact any of the other stories that have only one element

PR #4683, opened 2024-05-17
Calendar 400% zoom fixes

mcwinter07:
## Issue
- At 400% zoom the popover for date picker renders calendar dates below the fold, which could not be seen or scrolled to with mouse or keyboard interactions. This causes the component to fails the [WCAG spec](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) for content reflow

## What was tried
- CSS only fixes 
	- Didn't work because the `fixed` position made it hard to calc the 
- PopperJS  PreventOverflow and DetectOverflow utils
	- Would reliably work without significant custom window calculation

## The fix?
We solved a very similar problem to this in the MultiSelect rebuild using the `size` middleware available in the `floatingUI` library. Migrating was relatively painless and are able to do this without changing the API for our consumers.

### Video walkthrough
https://cultureamp.slack.com/archives/C02NUQ27G56/p1716182427118769?thread_ts=1715925784.709049&cid=C02NUQ27G56

## Changes
- Migrate `CalendarPopover` from `popperJS` to `floatingUI`
	- The popover will now set a maximum height and width based on the available white space in the viewport. This will share a similar behavior to our other modern popovers using `floatingUI`.
	- The popover will also have a minimum height of `155px` to allow for clear indication of scroll-able content
		-  Screenshot 2024-05-17 at 1 20 29 pm
	- The popover will attempt to fill the page with the maximum width but scroll if enough space is not available
	  -  Screenshot 2024-05-21 at 11 34 36 am

- The `CalendarPopover`'s padding is reduced below medium screen breakpoints to provide more real-estate
- Expose the `floatingUI` config object on the DatePicker
  -  This could be helpful for folks needing to tweak the config, i.e: changing the strategy from `fixed` to `absolute` or adding middle ware options


mcwinter07:
Creating PR to get a chromatic deploy up,  still a WIP 😉

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -73,6 +74,10 @@ export type DatePickerProps = {
    * A descriptive message for the 'status' states.
    */
   validationMessage?: DateInputFieldProps["validationMessage"] | undefined
+  /**
+   * Configure the floatingUI options that controls the CalendarPopover
+   */
+  floatingOptions?: Partial<UseFloatingOptions>
```
HeartSquared, line 80: I don't think we should be exposing this? What might be the use case for it?

packages/components/src/Calendar/CalendarPopover/CalendarPopover.tsx
```diff
@@ -1,14 +1,20 @@
...
 import classnames from "classnames"
-import { usePopper } from "react-popper"
 import { OverrideClassName } from "~types/OverrideClassName"
 import styles from "./CalendarPopover.module.scss"
 
 export type CalendarPopoverProps = {
   children: React.ReactNode
   referenceElement: HTMLElement | null
-  popperOptions?: Partial<Options>
+  floatingOptions?: Partial<UseFloatingOptions>
```
HeartSquared, line 17: This would be a breaking change (therefore changeset would need to change) - are we sure no one uses this component outside our repo?

mcwinter07: Yep no repo's outside of Kaizen use it and we only use it for DatePicker and DateRangePicker where we don't expose this

packages/components/src/DatePicker/DatePicker.tsx
mcwinter07: Main case I can think of would be for the `strategy` (fixed or absolute) or the preference on the `placement` of the popover. I'd say its not essential but could be a nice to have

mcwinter07: Axed this prop so should be good to go @HeartSquared

.changeset/odd-eggs-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
HeartSquared, line 2: Probably a patch? 🤔

packages/components/src/Calendar/CalendarPopover/CalendarPopover.tsx
```diff
@@ -18,37 +24,45 @@ export type CalendarPopoverProps = {
...
-          options: {
-            offset: [0, 15],
-          },
+  const { floatingStyles } = useFloating({
+    placement: "bottom-start",
+    elements: {
+      reference: referenceElement,
+      floating: floatingElement,
+    },
+    strategy: "fixed",
```
HeartSquared, line 41: Reminder: As discussed, we ideally would remove this as it causes different issues, but we'll address it separately

packages/components/src/Calendar/CalendarPopover/_docs/CalendarPopover.stickersheet.stories.tsx
```diff
@@ -42,14 +44,11 @@ const CalendarPopoverExample = ({
...
         referenceElement={referenceElement}
-        popperOptions={{
-          modifiers: [
-            {
-              name: "offset",
-              options: {
-                offset: [24, 0],
-              },
-            },
+        floatingOptions={{
```
HeartSquared, line 47: I think instead of making a separate Responsive one, we should have one base only. If changing the middleware makes a big difference, then it should be reflected. Ideally we don't do strange things in our stories which might not match actual implementation.

packages/components/src/DatePicker/_docs/DatePicker.mdx
```diff
@@ -70,3 +70,17 @@ There are multiple props to disable days within the Calendar.
 
 <Canvas of={DatePickerStories.DisabledDays} />
 <Controls of={DatePickerStories.DisabledDays} />
+
+### Responsive behaviour
+
+To ensure calendar content is accessibility across multiple viewports and satisfy the [content reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) spec in WCAG guidelines, the `DatePicker`'s popover will scroll when not enough space is available. The main case for this is to assist users on 400% zoom and ensure content is not visually lost below the fold.
```
HeartSquared, line 76:
```suggestion
To ensure calendar content is accessible across multiple viewports and satisfy the [content reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) spec in WCAG guidelines, the `DatePicker`'s popover will scroll when there is not enough space available. The main case for this is to assist users on 400% zoom and ensure content is not visually lost below the fold.
```

packages/components/src/DatePicker/_docs/DatePicker.stories.tsx
```diff
@@ -249,3 +249,73 @@ export const Validation: Story = {
...
+  name: "Limited viewport height",
+  args: {
+    labelText: "Calendar with reduced space below",
+  },
+  decorators: [
+    Story => (
+      <div className="mb-[150px]">
+        <Story />
+      </div>
+    ),
```
HeartSquared, line 282: Is it possible to one-line the decorators?

```suggestion
    Story => <div className="mb-[150px]"><Story /></div>,
```

packages/components/src/Calendar/CalendarPopover/CalendarPopover.tsx
mcwinter07: Yep, created a new ticket for that :)

.changeset/odd-eggs-exercise.md
mcwinter07: Ah yess, forgot to update that after removing the `floatingOptions`

packages/components/src/DatePicker/_docs/DatePicker.stories.tsx
mcwinter07: Sadly it lints to a multi-line

packages/components/src/Calendar/CalendarPopover/_docs/CalendarPopover.stickersheet.stories.tsx
mcwinter07: Fair point, I'll have to tweak a couple of things to add it to the top one, it'll still need to be absolute otherwise it won't test well

mcwinter07: Interdesting, looks like the snapshots are being take before the calendar fills our the available space 🤔

mcwinter07: So more height means less height... makes sense
Screenshot 2024-05-22 at 4 51 43 pm

mcwinter07:
The addition of new interaction tests in the DatePicker have uncovered a separate issue with regarding an unlabelled dialogue. Instead of bundling the change that could be required there I think the best move is add an ignore to the rule for now and tackle this in a subsequent PR rather than bloat this one any further.

Note: this isn't just an issue on this PR but also on main - we might not have caught it because we didn't have a test that covers the accessible label for dialogue when it the CalendarPopover is open in the DatePicker.

packages/components/src/Calendar/CalendarPopover/_docs/CalendarPopover.stickersheet.stories.tsx
mcwinter07: So the set viewport resolved this issue (Hurray!) I suspect that this is all due to the way Chromatic snaps a page with fixed or sticky elements. Seemingly no amount of additional time or space added to the bottom of the document if it thinks it has to scroll 🤷 Probably a good simulation of real world behaviour (which we test in the responsive sticker sheet) but not what we want in this particular story

packages/components/src/DatePicker/_docs/DatePicker.stories.tsx
```diff
@@ -249,3 +250,118 @@ export const Validation: Story = {
 export const DisabledDays: Story = {
   parameters: { controls: { include: /^disabled/ } },
 }
+
+export const LimitedWindowWidth: Story = {
```
HeartSquared, line 254: We actually have a `Component.spec.stories.tsx` naming convention for any stories that are not used in docs.
Perhaps to review later whether we want to keep the pattern, but that's what we currently have if you want to take a look :)

packages/components/src/DatePicker/_docs/DatePicker.stories.tsx
```diff
@@ -249,3 +250,118 @@ export const Validation: Story = {
...
+          name: "Viewport at 400%",
+          styles: {
+            width: "320px",
+            height: "350px",
+          },
+        },
+      },
+      defaultViewport: "ViewportAt400",
+    },
+    a11y: { disable: true }, // accessible label fix to be addressed in a separate PR
```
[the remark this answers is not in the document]
mcwinter07: Funnily enough disabling one cause a cascade of other issues that seem to relate to no title, so figured I'd just plop this on here since it'll be a fast follow fix :)

.changeset/odd-eggs-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Migrate CalanderPopover and DatePicker to floatingUI
```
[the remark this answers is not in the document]
mcwinter07: Makes sense since this'll be the main components our consumers use

packages/components/src/DatePicker/_docs/DatePicker.stories.tsx
mcwinter07: Makes sense to keep it in the same convention, at least for now. Will update that :)

HeartSquared (approved):

PR #4695, opened 2024-05-21
Fix EditableRichTextContent minimum contrast requirement

mcwinter07:
## Why
The `EditableRichTextContent` doesn’t meet colour contrast requirements for an interactive item and doesn't communicate its interactive behaviour effectively.

## What
- Adds a border to imply the interactive nature of the hidden button
- Update focus state to standard focus ring
- Adds missing label margin to RTE's 
- Adds example on how to replicate a read-only like UI using the `RichTextContent`
- Updates and adds missing documentation and examples of the RTE


gyfchong (approved):
Code looks fine, as long as you have design approvals!

packages/components/src/RichTextEditor/RichTextEditor/RichTextEditor.tsx
```diff
@@ -159,8 +159,13 @@ export const RichTextEditor = ({
 
   return (
     <>
-      {!labelledBy && labelText && <Label id={labelId} labelText={labelText} />}
-      {/* TODO: add a bit of margin here once we have a classNameOverride on Label */}
```
mcwinter07, line 163: I've added the margin bottom to the styles so we can remove this

packages/components/src/RichTextEditor/RichTextContent/_docs/RichTextContent.stories.tsx
```diff
@@ -22,3 +24,22 @@ export const Playground: Story = {
...
   },
 }
+
+export const ReadOnly: Story = {
+  parameters: {
+    chromatic: { disable: false },
+  },
+  render: args => (
+    <>
+      <Label classNameOverride="block mb-6" id="sb--rich-text-content-label">
```
mcwinter07, line 34: I'm dubious on actually using the label component here in this example as a `content editable="false"` is akin to a regular div `RichTextContent` essentially is, what do folks thing about using a heading instead? CC: @dougmacknz  if you have any thoughts here on this?

packages/components/src/RichTextEditor/EditableRichTextContent/_docs/EditableRichTextContent.mdx
```diff
@@ -15,15 +15,17 @@ import * as EditableRichTextContentStories from "./EditableRichTextContent.stori
...
+To render the contents of your [RichTextEditor](/docs/components-richtexteditor-richtexteditor--docs) using the same structure for both components.
 
 <Canvas of={EditableRichTextContentStories.Playground} />
 <Controls of={EditableRichTextContentStories.Playground} />
 
 ## Usage
 
-The general idea is to use this component in combination with the <LinkTo pageId="components-richtexteditor">RichTextEditor</LinkTo> to toggle between a read only state
-and an editable state.
+The `EditableRichTextContent` indicates interactivity similar to a `text` or `textarea` input and should be used in combination with the [RichTextEditor](/docs/components-richtexteditor-richtexteditor--docs) to toggle between inactive and editable states.
```
mcwinter07, line 25: Following a catch up with Jess, it was made clear we need to clarify the purpose of the RTE-like components. Any feedback on the framing of each of these docs updates would be ace :)

packages/components/src/RichTextEditor/RichTextContent/_docs/RichTextContent.stories.tsx
dougmacknz: Yeah true label probably isn't correct, but it did make it really easy to match the styling once the real RTE comes in.

Heading doesn't really make sense IMO. Could just be a div I guess?

mcwinter07: Yeah that's fair. If that's the case then I should probably nix the aria-labelledby hey

dougmacknz: aria-labelledby pointing to a div is totally fine

gyfchong (approved):

PR #4705, opened 2024-05-23
fix(bundler): remove adding PURE comment to styleInject

HeartSquared:
## What

Fix style injection by removing PURE comment.
Also inject CSS layers declaration whenever used so that if the injected styles are done in an opposite way (eg. global styles loaded after Heading), then the layers are not affected.

## Why

Builds in Storybook and Next fail when the comment exists.


packages/components/src/Heading/Heading.module.scss
```diff
@@ -1,5 +1,6 @@
 @import "~@kaizen/design-tokens/sass/color";
 @import "~@kaizen/design-tokens/sass/typography";
+@import "../../styles/foundation/_layers.css";
```
gyfchong, line 3: Ohhh tricky one, nice find

gyfchong (approved):

gyfchong (approved):

PR #4730, opened 2024-05-29
Update RTE LinkModal validation response

mcwinter07:
## Why
- The current LinkModal in the RichTextEditor does not provide enough context for the user as to the validation method.

## Context
  - HTTP and HTTPS is currently required to ensure executable Javascript is not passed into the link
    - It is a simple validation function that we could potentially open up as a prop folks need to build upon it / find it too limiting, but at time of development this was considered enough to ensure security
   

## Note
We haven't been able to get confirmation on the wording from the content team, but we're going to forge ahead with this since its been sitting with us for a while and some clarity is better than none

## What
- Updates the link validation message
- Updates the description
- Add simple stories to cover the render states of the link modal


packages/components/src/RichTextEditor/utils/plugins/LinkManager/components/LinkModal/LinkModal.tsx
```diff
@@ -45,14 +46,22 @@ export const LinkModal = ({
       onDismiss={onDismiss}
       onAfterLeave={onAfterLeave}
       onAfterEnter={() => inputRef.current?.focus()}
+      data-testid="modal-test"
```
gyfchong, line 49: Could this be added in the test instead of the component?

mcwinter07: Oh I should actually remove this 😆 its a left over when I was trying to find a good way to test this

gyfchong (dismissed):

gyfchong (approved):

PR #4731, opened 2024-05-30
Add accessible title to DatePicker and DateRangerPicker dialogs

mcwinter07:
## Why
- During the a11y uplift for the DatePicker and DateRange Picker it was discovered that we have an unlabelled dialogue. This PR seeks to address that by passing the label text into the dialogue title for a more contextual description for assistive technologies

### Before
Screenshot 2024-05-30 at 1 55 06 pm

### After
**DatePicker**
Screenshot 2024-05-30 at 1 55 31 pm

**DateRangePicker**
Screenshot 2024-05-30 at 1 55 51 pm

## What
- sets aria-labelledby on `DatePicker` and `DateRangePicker` to give the `CalendarPopover` and accessible title
  - Adds new `DateRangePicker` stories for axe test coverage


packages/components/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -33,7 +33,7 @@ describe("<DatePicker />", () => {
 
   it("should have an empty input value when a date is not provided", () => {
     render(<DatePickerWrapper />)
-    const input = screen.getByLabelText("Input label")
+    const input = screen.getByLabelText("Input label", { selector: "input" })
```
mcwinter07, line 36: I've added the selector since the tests seemed to pick up on a partial match for the dialogue accessible title since it is concatenated to produce `Choose date for: {label}` - funnily enough even with `exact: true` passed into the options 🤷

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -305,17 +306,25 @@ export const DatePicker = ({
...
-            defaultMonth={defaultMonth}
-            weekStartsOn={weekStartsOn}
-            disabled={disabledDays}
-            locale={locale}
-            onDayClick={handleCalendarDayChange}
-            onMount={handleCalendarMount}
-          />
+        <CalendarPopover
+          referenceElement={containerRef.current}
+          aria-labelledby={`${id}-calendar-label ${id}-input-label`}
```
mcwinter07, line 311: I originally started with label but decided to add the "choose date for:" screen reader text to make it clearer for the user when their focus is shifted. Somewhat inspired by the [WCAG example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/).

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -305,17 +306,25 @@ export const DatePicker = ({
...
+        >
+          <>
+            <VisuallyHidden id={`${id}-calendar-label`}>
+              Choose date for:
```
dougmacknz, line 315: Small suggestion to give a bit more info that it's a selection from a calendar. Thoughts?

```suggestion
              Select date from calendar for:
```

packages/components/src/DateRangePicker/_docs/DateRangePicker.stories.tsx
```diff
@@ -58,3 +58,99 @@ export const Playground: Story = {
     },
   },
 }
+
+const selectedDateRange = {
+  from: new Date(2022, 2, 6),
+  to: new Date(2022, 2, 15),
+}
+
+export const AboveIfAvailable: Story = {
```
mcwinter07, line 67: Same kind of stories I added for the DatePicker - this should cover us for testing the open state against the axe tool. These are also included in the MDX as a reference so I haven't bumped them to a spec.stories file.

packages/components/src/DatePicker/DatePicker.tsx
mcwinter07: Good point, it's a nice way to delineate between the "choose date" button and the calendar

packages/components/src/DateInput/DateInput/DateInput.tsx
```diff
@@ -27,6 +27,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
     <div className={classnames(styles.dateInput, classNameOverride)}>
       <Label
         htmlFor={id}
+        id={`${id}-input-label`}
```
dougmacknz, line 30: How hard would it be to avoid the loose coupling here and instead init this string once and use it in the two places?

mcwinter07: It's slightly awkward but I thought it was a better alternative than introducing and drilling a new prop down through the DateInputWithIconButon then into the DateInput since `id` is already in use

mcwinter07: To you're point though it is something we can easily update in the `DateRangePicker` since it sits all the same level

gyfchong (approved):

PR #4733, opened 2024-05-31
Pin PNPM version in CI

gyfchong:


jakubriedl (commented):
can you please add PR description to explain why we need to do this? will help with future reference

mcwinter07 (approved):

PR #4744, opened 2024-06-05
Tooltip v3

jakubriedl:
## Why
We're heading in new direction with Kaizen Design System. This PR is the first component in the new way where we test and introduce few interconnected patterns.

- Building blocks pattern
- Standard library (React Aria Components) to build out building blocks
- Versioned entrypoints to be able to flexibly add and maintain blocks including breaking changes
- Grouped entrypoints to allow for much larger number of components that are inherent aspect of building blocks
- Unified documentation in Storybook including usage guidelines and API docs

## Reviewing
While this PR may seem huge at the first glance, vast majority of line changes is lock file and file changes from moving around. To review this PR pay most attention to `src/__overlays__/Tooltip/*` and inside `Button`.


packages/components/src/__future__/Tooltip/Focusable.tsx
```diff
@@ -0,0 +1,37 @@
...
+}: FocusableProps): JSX.Element => {
+  const ref = useRef<HTMLDivElement>(null)
+  const { focusableProps } = useFocusable(props, ref)
+
+  return (
+    <div
+      ref={ref}
+      className={classnames(styles.focusableWrapper, className)}
+      {...focusableProps}
+      data-inline-hidden-content
```
gyfchong, line 25: How does this work?

packages/components/src/__future__/Tooltip/Focusable.tsx
```diff
@@ -0,0 +1,37 @@
...
+      // We want the div to be focusable for keyboard users,
+      // but screen readers will have the VisuallyHidden content
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={0}
+      // Negate the aria description (added by RAC) as we have the VisuallyHidden content
+      aria-describedby={undefined}
```
gyfchong, line 31: If I understand this this setting (+ the `inline-hidden-content`) it's to provide visually hidden content rather than rely on `aria-describedby` to describe its contents. Is this always the case though? Or is it more of a tooltip/KAIO thing, eg. would this be applicable to Table Rows, Card or Data point on a graph etc. As a utility I'd expect it to work on anything and everything, even more so if I'm trying to create a bespoke component which no one's ever thought of.

So I think we need to decide, is this a public facing utility or not;
- if so, then it might be better to make these optimisations an opt-in API to cater for both DST use cases and make it easier for consumers (and us) to both understand the utility and document its use cases.
- if not, then this should probably remain an internal utility, and if the consumer needs something like this, they can read RAC docs on how to compose something similar. (in which we need a place to put internal utils...)

Although outside of these two options, I don't quite understand the value of this utility to consumers.

packages/components/src/__future__/Tooltip/ToggleTip/ToggleTip.tsx
```diff
@@ -0,0 +1,46 @@
...
+      className={classnames(styles.toggleTip, className)}
+      data-inline-hidden-content
+      onKeyDown={e => {
+        if (e.key === "Enter" || e.key === " ") toggle()
+      }}
+      onPointerDown={toggle}
+      onBlur={() => state.close(true)}
+      {...props}
+      // Cannot use react-aria's Button component as it prevents adding aria-hidden
+      aria-hidden="true"
```
gyfchong, line 41: This seems like a critical issue that we should address, because it sounds like we are saying that `ToggleTip` needs to be a snowflake button and as such will not behave consistently with our main `Button` component (or inherit any updates).

I also think this might be unnatural in our consumer's workflow, we'd need to check with Jess, but I imagine designers will simply drag a Button and a Tooltip onto their canvas and say "this should toggle the Tooltip", without even thinking to search for a specific component -- given how many ToggleTips/Tooltips are present in the existing product, this could cause some friction around:

- Button features not being present in `ToggleTip` in which devs/design will pull us up on that
- Devs struggling to assemble the `Tooltip` and `Button` only to either ask or realise late in the process that there's a `ToggleTip` component, which then brings up the first point.

But given we cannot actually attach `Tooltip` to the official `Button`, both of these points potentially have a high chance of resulting in a bespoke component being made, slowing down consumers for a very common pattern in  our products.

So I think we need to discuss this one and consider what we might do to avoid these scenarios since it will undoubtedly affect adoption rates.

packages/components/src/__future__/Tooltip/Focusable.tsx
jakubriedl: yeah that's right, it's used to provide VisuallyHidden content instead of describedBy tooltip. Basically the `inline-hidden-content` on a tooltip (and other) triggers tells the `Tooltip` to render the content also into VisuallyHidden and so the `describedby` isn't needed.

That is quite specific tooltip a11y problem/solution but I'm not sure how it will work for other future components. I can imagine this being reusable, but not sure if there's reason to.

packages/components/src/__future__/Tooltip/ToggleTip/ToggleTip.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { AriaButtonOptions, useTooltipTrigger } from "react-aria"
+import { TooltipTriggerStateContext } from "react-aria-components"
+import styles from "./ToggleTip.module.scss"
+
+export type ToggleTipProps = {
+  children: ReactNode
+  className?: HTMLAttributes<HTMLButtonElement>["className"]
+} & AriaButtonOptions<"button">
+
+export const ToggleTip = ({
```
gyfchong, line 12: We should make sure the usage/design guidelines in docs highlight the specific use case this was built for ie. just an icon+tooltip -- to set the expectation that it doesn't support all `Button` cases

packages/components/src/__overlays__/Tooltip/v3/_docs/Tooltip.mdx
jakubriedl, line 1: this page needs updating

packages/components/src/KaizenProvider/KaizenProvider.tsx
```diff
@@ -27,13 +28,13 @@ export const KaizenProvider = ({
 
   return (
     <OptionalIntlProvider locale={locale}>
-      <>
+      <I18nProvider locale={locale}>
```
gyfchong, line 31: This is a bit confusing, maybe rename it to something specfic to React Aria so that we don't mix this up with `@cultureamp/i18n-react-intl` 

And consider putting it together in the `OptionalIntlProvider` but not make it optional

gyfchong:
Did we want to cap the width of the content to match the images?
image

packages/components/src/__overlays__/Tooltip/v3/Tooltip.tsx
```diff
@@ -0,0 +1,101 @@
...
+    | "end"
+    | "end top"
+    | "end bottom"
+}
+
+/**
+ * A tooltip displays a description of an element on hover or focus.
+ */
+export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
+  ({ children, className, ...props }, ref): JSX.Element => {
```
gyfchong, line 47: Realised this is `className` instead of `classNameOverride`, I'm good with the rename to `className`, but I think we should keep it consistently `classNameOverride` for now and go through the deprecation process to change all the components at once.

packages/components/src/__overlays__/Tooltip/v3/ToggleTip/ToggleTip.tsx
```diff
@@ -0,0 +1,47 @@
...
+  className,
+  ...props
+}: ToggleTipProps): JSX.Element => {
+  const ref = useRef<HTMLButtonElement>(null)
+  const state = useContext(TooltipTriggerStateContext)!
+
+  // registers ref to triggerRef for positioning the tooltip
+  // it also registers event handler for escape key to close the tooltip
+  // other handlers like onKeyDown or onPointerDown we provide own
+  useTooltipTrigger(props, state, ref)
```
gyfchong, line 23: Where did we land on replacing this with Popover?

packages/components/src/__overlays__/Tooltip/v3/Tooltip.tsx
jakubriedl: Yes this is intentional change. With v3 (and whole new gen of components) we aim to cleanup a lot of custom interfaces and patterns that are across the components and fully clean them up. No more moods, no more `classNameOverride` and other. 
The consistency goal is that this will be pattern for all new/v3+ components. There's a bit of question what we do when we start v4 and backfill v3 from v2 if we rename it everywhere or just move, but that's bit future problem.

packages/components/src/__overlays__/Tooltip/v3/ToggleTip/ToggleTip.tsx
jakubriedl: We've landed that it's 2 similar yet different things. The popover will be likely styled differently, will be in general bigger & heavier and will have different focus considerations. So in general intended for larger amount of content.

gyfchong (approved):

packages/components/src/Button/Button/Button.module.scss
```diff
@@ -33,47 +33,50 @@
...
+      &::after {
+        border-color: $color-blue-300;
       }
     }
   }
 
-  &:disabled,
+  &.disabled,
   &.working {
-    pointer-events: none;
```
dougmacknz, line 75: We were relying on this as a way of actually stopping the button from triggering its `onClick` when it's in the `working` state, this introduces a bug

packages/components/src/Button/Button/Button.module.scss
```diff
@@ -33,47 +33,50 @@
...
+    cursor: default;
     opacity: 30%;
+
+    &:focus-visible {
+      &::after {
+        border-style: $border-dashed-border-style;
+        border-color: $color-gray-400;
```
dougmacknz, line 77: This looks way more washed out than the dashed borders we did for the disabled buttons on RTE (they used purple 800).

If you have local running: http://localhost:6006/?path=/story/components-richtexteditor-subcomponents-toggleiconbutton--sticker-sheet-default

Just checking this is on purpose?

jakubriedl: good spot, @HeartSquared if I recall correctly you've made this change. I guess it's based on a figma, can you please share the link to it?

storybook/global-a11y-rules.ts
```diff
@@ -25,4 +25,10 @@ export const globalA11yRules = [
     id: "aria-allowed-role",
     selector: ':not(button[name="day"][role="gridcell"])',
   },
+  {
+    // In some edgecases like ToggleTip we want to be able to disable the rule and hide the button from screen readers
```
dougmacknz, line 29: If you care: this could update to reference react aria Select's usage instead, in case someone comes along and thinks they can delete this now

packages/components/src/Button/Button/Button.module.scss
HeartSquared: I don't recall unfortunately. I may have copied it from here with the plan to check later? Feel free to change it.
https://github.com/cultureamp/kaizen-design-system/blob/main/packages/components/src/Filter/FilterMultiSelect/subcomponents/SelectionControlButton/SelectionControlButton.module.scss#L68

packages/components/src/__utilities__/Focusable/v3/Focusable.tsx
```diff
@@ -0,0 +1,39 @@
...
+      ref={ref}
+      className={classnames(styles.focusableWrapper, className)}
+      {...focusableProps}
+      data-inline-hidden-content
+      // We want the div to be focusable for keyboard users,
+      // but screen readers will have the VisuallyHidden content
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={0}
+      // Negate the aria description (added by RAC) as we have the VisuallyHidden content
+      aria-describedby={undefined}
```
dougmacknz, line 31: Nit pick but you could add to the comment above that aria-describedby on a div does nothing, so that's why we use visually hidden instead.

packages/components/src/Button/Button/Button.module.scss
jakubriedl: good context, I've disabled `onClick` and `onPointerDown` on the working button. Kept `onBlur` and `onFocus` and similar if that makes sense

packages/components/src/__overlays__/Tooltip/v3/ToggleTip/ToggleTip.tsx
jakubriedl: After more a11y complications that appeared yesterday we've decided to remove ToggleTip for now and keep it only as Popover. If there's significant demand in future we can rethink but for now it doesn't seems to be worth it

packages/components/src/__utilities__/Focusable/v3/Focusable.tsx
```diff
@@ -0,0 +1,40 @@
...
+      className={classnames(styles.focusableWrapper, className)}
+      {...focusableProps}
+      data-inline-hidden-content
+      // We want the div to be focusable for keyboard users,
+      // but screen readers will have the VisuallyHidden content
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={0}
+      // aria-describedby on div doesn't do anthing so we instead render the content in VisuallyHidden from tooltip
+      // but because RAC adds it as it assumes it's interactive element we remove it here
+      aria-describedby={undefined}
```
dougmacknz, line 32: Given this is a util that could be used for anything, it doesn't seem ideal that something specific to tooltip is baked in here. Are there any other options?

packages/components/src/utils/mergeClassNames.ts
```diff
@@ -0,0 +1,27 @@
+import classnames from "classnames"
+
+type ClassNameValue = string | number | boolean | undefined | null
+
+export const mergeClassNames = <
```
dougmacknz, line 5: Trying to understand why this was created over using classnames?

jakubriedl: it was because RAC has [handy feature](https://react-spectrum.adobe.com/react-aria/styling.html#render-props) where you can pass function that receives state to `classname`. This is super useful when working with tailwind and we want to keep it. And because of that we need a helper that can deal correctly with the function call and merging.
I'll add a comment here to explain it for anyone coming to this later

dougmacknz (dismissed):
(looks like you're still making the final touches but fyi I've finished reviewing and lgtm)

packages/components/src/__overlays__/Tooltip/v2/index.ts
```diff
@@ -0,0 +1 @@
+export * from "../v1/Tooltip"
```
dougmacknz, line 1: Looks like this results in deprecation warnings in VS Code for v2 imports as well, which I'm assuming wasn't intended.

dougmacknz: Sorry never mind, I was looking at my implementation and I had done it slightly differently

storybook/components/DocsContainer/DefaultDocsContainer.tsx
```diff
@@ -53,6 +47,11 @@ export const DefaultDocsContainer = ({
   ...props
 }: DocsContainerProps & { children: React.ReactNode }): JSX.Element => (
   <DocsContainer {...props}>
-    <DocsLayout>{children}</DocsLayout>
+    <DocsLayout>
+      {/* <Title />
+      <KAIOInstallation exportNames="Tooltip" family="overlays" version="1" /> */}
+      <div className="test">{children}</div>
```
jakubriedl, line 53: I guess this is leftover from testing, right?

```suggestion
      {children}
```

mcwinter07: Ah yeah that can go for now.

dougmacknz (approved):

dougmacknz (approved):

PR #4769, opened 2024-06-17
Migrate future tooltip to versioned entires

mcwinter07:
## Why
To update the Tooltip POC to the new folder structure to support versioned entry points and logic groups.

See [Miro](https://miro.com/app/board/uXjVK_gakNs=/?moveToWidget=3458764592102748095&cot=14) for documented IA

## What
- adds a /v1, /v2 and /v3 folder structure
- updates rollup config to bundle overlays
- migrates Tooltip future to v2
- migrate Focusable and ReversedColor into __utilities__ group
  - Once we have docs then we should create entry points for the utilities logical group


mcwinter07:
There's some issue with the build but I think this needs to be tackled in the original PR

packages/components/src/__overlays__/Tooltip/v2/ReversedColors.tsx
jakubriedl, line 1: `Focusable` and `ReversedColors` should be probably in some `utilities` namespace/family as those are generic helpers that we'll reuse across all families

packages/components/src/__overlays__/v3.ts
jakubriedl, line 1: This feels bit confusing. If we're saying that upcoming version is v3, then the new tooltip is v3 and I don't think we should publish it on v2.

Basically simplified mental model how I think about it
v1 = current components
v2 = either old "future" components or we could skip entirely
v3 = new generation of components

mcwinter07: Ok, so he idea that any idea is that any part of the Aria component rebuild will be only be on v3?

A couple of other q's to get some clarification:
- If v2 will be for things that are the current future components like `Tag`, `Select` and `Workflow`, will this entry only have 3 components? 
- Are we actually publishing a v1 entry point or will this just be the current major of `@kaizen/components`?

jakubriedl: > Ok, so he idea that any idea is that any part of the Aria component rebuild will be only be on v3?
yeah I think so

> Are we actually publishing a v1 entry point or will this just be the current major of @kaizen/components?
yeah we're publishing it to be clear what version it is and enable people to think in the incremental versions. having it on root would be confusing. But we want to do that in seperate PR, because to move to that we'll need code mod

with the v2 I'm not completely sure what's the best approach. We do want to squash versions after deprecating them. We also want to cleanup and get rid of these 3 outliers. And we want to give consumers time to move. Maybe the approach could be that we publish current as v2, current aliased + future as v2 (but without the old workflow and tag, Select as SimpleSelect). But let's talk about it

packages/components/src/__overlays__/Tooltip/v2/ReversedColors.tsx
mcwinter07: `ReversedColors` seems easy enough, but the `Focusable` looks like it's interconnected with Tooltip context so we'd need to genericise it before moving it. 

Are you thinking these utils are something we distribute or is just a internal use?

jakubriedl: ouch, you're right, but by my mistake. I've pushed experiment I had locally that has created the connection. I'll revert it

packages/components/src/Button/GenericButton/GenericButton.tsx
```diff
@@ -225,9 +225,9 @@ const renderCustomComponent = (
           props["aria-describedby"] === null
             ? undefined
             : classnames(
-              contextProps["aria-describedby"],
-              linkProps["aria-describedby"]
-            )
+                contextProps["aria-describedby"],
+                linkProps["aria-describedby"]
+              )
```
mcwinter07, line 230: These are just linting fixes no actual change from what was already on the tooltip branch

packages/components/src/__overlays__/Tooltip/v3/Focusable.tsx
jakubriedl, line 1: This was moved to utilities so can be deleted from here

packages/components/src/__overlays__/Tooltip/v3/_docs/Tooltip.stories.tsx
jakubriedl, line 1: I've updated this file a bit in the original branch. You may need to pull and resolve conflict

packages/components/src/__utilities__/Focusable/v1/Focusable.tsx
jakubriedl, line 1: `Focusable` and `ReversedColors` are from v3 generation of components so we should have them there. And wouldn't export them in v1/v2 as they are not applicable there at all

jakubriedl (changes requested):

packages/components/src/__utilities__/Focusable/v1/Focusable.tsx
mcwinter07: Makes sense 👍 I've updated so we only have v3 at the moment. There may be some things that could go into a v1 or 2 (Like VisuallyHidden) when we migrate everything across but figure we can just add those entry points then

packages/components/src/__overlays__/Tooltip/v3/_docs/Tooltip.stories.tsx
mcwinter07: Done and done 👍 @jakubriedl ready for a re-review when you get a moment

jakubriedl (commented):
Looks good, just only now I've noticed that you've copied the old Tooltip to `v1` instead of moved. I guess we don't want to duplicate it right? Should we link the old index to the `v1` and delete the old folder?

mcwinter07:
> Looks good, just only now I've noticed that you've copied the old Tooltip to `v1` instead of moved. I guess we don't want to duplicate it right? Should we link the old index to the `v1` and delete the old folder?

Can't see any issue as long and when we release it is as a major 👍

jakubriedl:
> Can't see any issue as long as when we release it is as a major 👍

if we reexport that in the barrel file then it shouldn't be braking change right? And so we don't need major

mcwinter07:
> > Can't see any issue as long as when we release it is as a major 👍
> 
> if we reexport that in the barrel file then it shouldn't be braking change right? And so we don't need major

So remove the old folder from /components/src and export from the v1 version in `components/src/index.ts` barrel?

packages/components/src/__overlays__/Tooltip/v3/_docs/Tooltip.stories.tsx
```diff
@@ -32,57 +31,27 @@ export default meta
 type Story = StoryObj<typeof meta>
 
 export const OnButton: Story = {
-  render: ({ defaultOpen, isOpen, ...args }) => (
-    <TooltipTrigger defaultOpen={defaultOpen} isOpen={isOpen}>
+  render: args => (
```
jakubriedl, line 34: this looks like git conflict resolved the other way. The red is what we want in this file

jakubriedl (commented):
It looks good now, just the one last remaining conflict in stories

packages/components/src/__overlays__/Tooltip/v3/_docs/Tooltip.stories.tsx
mcwinter07: Ah nice catch, will clear that one up 👍

mcwinter07: Reverted back to the original state, should be good to merge if all is hunky dory

jakubriedl (approved):
nice work

PR #4772, opened 2024-06-18
feat(TitleBlockZen): update title font to match composable header

HeartSquared:
## What

Update `TitleBlockZen`'s title font to match the composable header.

## Why

Match the font family of the composable header.

After discussion with Dave, removed breadcrumb breakpoint title font scaling given:
- We don't know the actual use case it was solving
- The change of font changes the width of the text - therefore may already be solving initial reason for it
- The breakpoint being at 1644 means most users never actually see the proper visual hierarchy, which isn't ideal

### DIFF
--- a/.changeset/forty-games-battle.md
+++ b/.changeset/forty-games-battle.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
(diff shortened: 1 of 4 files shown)


dougmacknz (approved):
LGTM

PR #4781, opened 2024-06-24
Fef 1446 devboxify kaizen design system

ckychris:
## Why
Set up Devbox for local development

## What

- Devbox init
- Removed `.tool-versions`
- Updated readme
- Updated root commands to have turbo to run required builds
- Changed the storybook default port
- GitHub workflows now use Corepack
- GitHub workflow cache refactor
- Remove buildkite custom Docker images in favour of standard images
- Added concurrency group to ensure the correct order of storybook deployment versions


devbox.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/main/.schema/devbox.schema.json",
+  "packages": ["nodejs@18.19.0"],
```
gyfchong, line 3: Seems a bit old?

ckychris: Yeah, some projects are now on Node 20. When migrating projects to Devbox, I tend to leave the Node version as is. This ensures that my PR won't cause any errors.

gyfchong (approved):

PR #4788, opened 2024-06-27
feat: translate FilterBar and DatePicker components

HeartSquared:
## Why

[KZN-2450](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2450)
Customer request to support `fr-CA` for DatePicker.

[KZN-2484](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2484)
Request to translate hard-coded strings within FilterBar.

## What

- Add `fr-CA` support in `DatePicker`, `FilterDatePicker`, `FilterDateRangePicker`
	- Translate hard-coded text in above components
- Translate hard-coded text within `FilterBar` and subcomponents
- Remove jest mock for i18n `useIntl` as we now need more complex checks within tests
	- Fix all tests which have translations

[KZN-2450]: https://cultureamp.atlassian.net/browse/KZN-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KZN-2484]: https://cultureamp.atlassian.net/browse/KZN-2484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/DatePicker/_docs/controls/datePickerLocaleControls.ts
```diff
@@ -1,6 +1,6 @@
 import { ArgTypes } from "@storybook/react"
 
-const DATE_PICKER_SUPPORTED_LOCALES = ["en-US", "en-AU"]
+const DATE_PICKER_SUPPORTED_LOCALES = ["en-US", "en-AU", "fr-CA"]
```
jakubriedl, line 3: what is the reason we historically don't support all CA supported locales? If I understand the ask from Lisa right, she needs all locales that perf supports

HeartSquared: i18n wasn't implemented into Kaizen at the time, so these were just implementations of the library.
We can probably add all LTR languages into this list if you want, but at the time we were building the components, we wanted to keep the scope small and avoid figuring out whether a language is RTL to flip the element (or not).

We had also wanted to see if we could use the i18n library to help prevent needing to set the prop, but I think given my quick testing locally gave me `en` when I was expecting something more specific (given our initial feature request was `en-AU` and `en-US`), it isn't something I can easily do here. So if we want to look into it, we'll have to do the discovery as a separate task.

mcwinter07: Historically I think it either built prior to rolling out i18n or we were limiting the scope for the initial build?

This specific task was to address the French Canadian [client request](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1717715050909919) from Jo, who's keen to get this in. @HeartSquared correct me if I'm wrong but Lisa's request relate to the filter bar?

.changeset/large-melons-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Add locale `fr-CA` and translate text in `DatePicker`, `FilterDatePicker`, `FilterDateRangePicker`.
```
mcwinter07, line 5: Could we also callout the FilterBar updates here?

HeartSquared: Oh yes missed it! Thanks!

packages/components/src/DatePicker/_docs/controls/datePickerLocaleControls.ts
jakubriedl: Ok cool good to know, i18n is [a known mess across the company](https://cultureamp.atlassian.net/wiki/spaces/Prod/pages/2994635071/State+of+localization) and every time we can make it a little better we should

There was a lot of French canadian requests across the company because of one particular customer. It was across many teams. So that's why the request was phrased like that. But I think if we can add all LTR locales we would fix more issues even before some customer raises them.

You can use our i18n-react-intl lib to map our locales to region specific by using [LOCALE_REGIONS](https://github.com/cultureamp/unified-home/blob/master/packages/i18n-react-intl/src/constants.ts#L62) which should be exported.

packages/components/tsconfig.json
```diff
@@ -4,13 +4,15 @@
...
+      "~tests/*": ["./__tests__/*"],
       "~types/*": ["./src/types/*"],
       "~utils/*": ["./src/utils/*"],
       "~components/*": ["./src/*"]
     },
     "types": ["@testing-library/jest-dom"]
   },
   "files": ["types.d.ts"],
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "__tests__/**/*"],
```
mcwinter07, line 16: So this is just to include the `renderWithIntl` util if our consumers need it?

packages/components/src/DatePicker/_docs/controls/datePickerLocaleControls.ts
HeartSquared: That list doesn't have a way to differentiate between `en-AU` and `en-US`, so I think if we want to investigate this, we'll have to do it separately.

I don't want to delay this PR by spending time investigating which locales are LTR, so if you could provide me the list then I can add them in. Otherwise, I think it's best to create a follow up task instead.

packages/components/tsconfig.json
HeartSquared: Nope! This is for internal usage only. Code for our consumers is based on `tsconfig.dist.json` :)

packages/components/src/DatePicker/_docs/controls/datePickerLocaleControls.ts
jakubriedl: fair point, it's because we don't have `en-AU` in our platform (users can't set to). Which is different can of worms. 
I think in long term we'll need to work with product (Rog) to figure out what to do with this `en-AU`. But for now I think adding all the locale regions we support to this list would be fix issues for every locale.

mcwinter07 (approved):
Nice, I think that covers the gamut of translations needed for DatePicker, functionality seems hunky dory to support fr-ca for the client request

packages/components/src/DatePicker/_docs/controls/datePickerLocaleControls.ts
HeartSquared: I'll create a separate branch to add the other locales to the switch to save from going through CI again, given it's self-contained.

PR #4789, opened 2024-07-01
Update tooltip with new design doc content

mcwinter07:
## Why
- update Tooltip docs following Margaret's second pass on on the [design documentation
](https://www.figma.com/design/IoUwr7gnxryCvrBuZa5fjE/Tooltips?node-id=56-2774&t=Ze7t5GSudtUNIGeD-4)

## What
- converts image to dos and don'ts stories
- adds a reusable dos and don'ts doc components
- update some text content

## Notes
- I've left out this example (below) since the link in the tooltip is more of a popover pattern, which I don't think should be included since we don't have ToggleTip to facilitate anything like that.
  - We can redirect to that at a different point once the new popover has been built
Screenshot 2024-07-01 at 11 03 25 am

- I've also left in the placements stories and some text content for the dos and don'ts that feel relevant
  - would also be worth checking in on why they were removed with Margaret later


storybook/components/DosAndDonts/DoesAndDonts.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: HTMLAttributes<HTMLDivElement>): JSX.Element => (
+  <div
+    className={classnames(className, "grid grid-cols-1 gap-16 lg:grid-cols-2")}
+    {...otherProps}
+  >
+    {children}
+  </div>
+)
+
+DosAndDonts.displayName = "DosAndDonts"
```
jakubriedl, line 20: nit: setting displayName manually is not needed anymore with modern react & tools. It detects that automatically.

jakubriedl (approved):

storybook/components/DosAndDonts/DoesAndDonts.tsx
mcwinter07: Ah cool, good to know. We should probs update our plop templates going forward and drop this pattern from any rebuilds in the versioned releases

PR #4794, opened 2024-07-03
feat: add all supported locales to date components

HeartSquared:
## Why

[KZN-2486](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2486)

Extension of support request which added `fr-CA`

## What

- Add all supported locales to `DatePicker`, `FilterDatePicker`, `FilterDateRangePicker`
- Update default locale object from `en-AU` to `en-GB`
	- They are visually the same, but `en-AU` is not in [our list of locales](https://github.com/cultureamp/unified-home/blob/master/packages/i18n-react-intl/src/constants.ts#L62)
- Update `Calendar`s to infer starting weekday based on provided locale

[KZN-2486]: https://cultureamp.atlassian.net/browse/KZN-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/DatePicker/utils/getLocale.ts
+++ b/packages/components/src/DatePicker/utils/getLocale.ts
@@ -57,7 +57,6 @@ export const getLocale = (locale: DatePickerSupportedLocales): Locale => {
       return da
     case "nl":
       return nl
-    case "en":
     case "en-US":
       return enUS
     case "fr":
@@ -88,6 +87,7 @@ export const getLocale = (locale: DatePickerSupportedLocales): Locale => {
       return ptBR
     case "uk":
       return uk
+    case "en":
     case "en-GB":
     case "en-AU":
       return enGB
--- a/packages/components/src/DatePicker/utils/getLocale.ts
+++ b/packages/components/src/DatePicker/utils/getLocale.ts
@@ -1,21 +1,152 @@
 import type { Locale } from "date-fns"
-import { enAU, enUS, frCA } from "date-fns/locale"
-import { StringSuggestions } from "~types/StringSuggestions"
+import {
+  arSA,
+  bg,
+  cs,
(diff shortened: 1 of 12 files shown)


packages/components/src/DatePicker/utils/getLocale.ts
```diff
@@ -1,21 +1,152 @@
...
-      return enAU
+    case "zh":
+      return zhCN
+    case "zh-TW":
+      return zhTW
+    case "da":
+      return da
+    case "nl":
+      return nl
+    case "en":
```
jakubriedl, line 60: This would change current behaviour for "en" which now uses AU and would start using US. Can we map "en" to "en-GB"?

jakubriedl (changes requested):
Looks great, just one mapping to avoid change of behavior for "en"

packages/components/src/DatePicker/utils/getLocale.ts
HeartSquared: Hmm...I suppose that's true, but would it not be strange for `en` to be `en-US` for everything except the datepickers, therefore making that previously a bug?
I'll change it anyway; it will be easy to drop it again if we want.

jakubriedl: Yeah it is strange, but it is how it is today. We're intentionally keeping the current unexpected behaviour instead of replacing it with different unexpected behaviour (for AU users). We'll get to this when we properly plan for uplifting the DatePickers or i18n as whole.

jakubriedl (approved):

PR #4796, opened 2024-07-05
feat(Brand): update asset for collective intelligence

HeartSquared:
## Why

[KZN-2432](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2432)

Accessibility improvements.

## What

- Update Collective Intelligence source svg
- Update prettier config to lint `mjs` files

[KZN-2432]: https://cultureamp.atlassian.net/browse/KZN-2432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Brand/Brand.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { HTMLAttributes } from "react"
 import { assetUrl } from "@kaizen/hosted-assets"
 import { OverrideClassName } from "~types/OverrideClassName"
+import collectiveIntelligenceUrl from "./BrandCollectiveIntelligenceLogo.svg"
```
gyfchong, line 4: What made you opt for using the SVG directly over making it a React component?

HeartSquared: Hmm... didn't really think about it. Just tried not changing the render structure too much 😬 I did a quick search for the usage of `Brand` and `classNameOverride` was being put on some of them, so wanted to avoid potentially breaking those.

Also attributes for `picture` may differ from `svg`, so the spread props will be in a funny place.

I don't mind changing it if you'd prefer? Would certainly remove the need for all the extra SVG config.

.changeset/wise-plums-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
```
dougmacknz, line 2: Sounds like a patch would be more suitable?

dougmacknz (dismissed):
LGTM

.changeset/wise-plums-lay.md
gyfchong:
```suggestion
"@kaizen/components": patch
```

gyfchong (approved):
Waiting on Figma changes to merge

PR #4805, opened 2024-07-10
Button versioning and v3 RAC build

mcwinter07:
## Why
- Start the rebuild of the default Button rebuilt with RAC

## What
- migrates original Button actions and add v1 and v2 entry
  - re-export v1 from components barrels so not to cause a breaking change in the import statement
- creates v3 of Button built in using RAC's Button
  - Add basic usage guidelines
  - Add API specification documentation calling out changes in the API
    - Add disclaimer on both that this is intended to be used with Menu only - pending the Button RFC

## Important
Do not merge until Menu rebuild PR is merged in

## Notes
- As was [discussed here](https://cultureamp.slack.com/archives/C05D8V4GALX/p1720660691396679), this is not at full parity of the current button
  -  The intention is to support the Menu rebuild with the default variant
  - Once the Button RFC is complete and the variants/ styles are signed off we can resume work on feature parity


mcwinter07:
This required branching off Menu to test the Button implementation would work. As discussed this morning in standup, we will merge this into main separately instead of into the Menu Branch - This should not be merge in till then but is ready for review

packages/components/src/Filter/FilterBar/subcomponents/AddFiltersMenu/AddFiltersMenu.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react"
 import { useIntl } from "@cultureamp/i18n-react-intl"
-import { Button } from "~components/Button"
 import { AddIcon } from "~components/Icon"
-import { Menu, MenuList, MenuItem } from "~components/Menu"
+import { Menu, MenuList, MenuItem } from "~components/__actions__/v1"
```
gyfchong, line 4: Why not v2?

packages/components/src/Icon/subcomponents/SVG/SVG.tsx
```diff
@@ -15,6 +15,7 @@ type DecorativeIcon = {
 
 export type BaseSVGProps = {
   inheritSize?: boolean
+  shouldFlipOnRtl?: boolean
```
gyfchong, line 18: Probably don't want to expose this as a prop, since this might imply that any icon can be flipped for RTL. 

Maybe we give the icons a unique class name and target them that way

mcwinter07: Ah woops, sorry that's actually just part of the exploration I did when I was looking at components - I'll remove that from the PR 👍

packages/components/src/Filter/FilterBar/subcomponents/AddFiltersMenu/AddFiltersMenu.tsx
mcwinter07: This would be part of the Menu PR that's brought across since I needed to branch of Douglet's work 😉 but valid comment, cc: @dougmacknz

packages/components/src/TitleBlockZen/types.ts
```diff
@@ -1,7 +1,7 @@
 import { CompanyAvatarProps, GenericAvatarProps } from "~components/Avatar"
-import { ButtonProps, CustomButtonProps } from "~components/Button"
-import { MenuItemProps } from "~components/Menu"
 import { Select } from "~components/Select"
+import { MenuItemProps } from "~components/__actions__/v1"
+import { ButtonProps, CustomButtonProps } from "~components/__actions__/v2"
```
gyfchong, line 4: Lol wild

packages/components/src/Filter/FilterBar/subcomponents/AddFiltersMenu/AddFiltersMenu.tsx
dougmacknz: Updated now on my branch

packages/components/src/__actions__/Button/v3/Button.module.scss
```diff
@@ -0,0 +1,109 @@
...
+
+// spacing-40 does not currently exist. Holding on adding this until the button rfc has been signed off
+$button-small-x-y-min: 2.5rem;
+$button-y-padding: calc(#{$spacing-8} - #{$border-solid-border-width});
+
+// $spacing-22 (1.375rem) doesn't exist in design tokens. Since this will be changes in the rebuild we shouldn't add it
+$button-x-padding: calc($spacing-12 - #{$border-solid-border-width});
+$button-x-padding-small: calc(#{$spacing-8} - #{$border-solid-border-width});
+
+.button {
```
gyfchong, line 17: I'm wondering if during these rebuilds we introduce a layers to improve our specificity control 🤔 cc/ @jakubriedl

packages/components/src/TitleBlockZen/types.ts
mcwinter07: Wonder if we'll see these in the wild :P I think Doug's updated these so will fix these imports when that's merged in and I rebase off main :)

packages/components/src/__actions__/Button/v3/Button.module.scss
mcwinter07: Actually, while we're talking about potential changes to line up with rebuilds, do we want to slowly ween ourselves off SCSS and write new component as CSS? / are there any blockers to doing so. I know it was discussed a long time ago, but wondering if that's still on the cards for us?

packages/components/src/Filter/FilterBar/subcomponents/AddFiltersMenu/AddFiltersMenu.tsx
mcwinter07: Noice ✋ 5️⃣  I'll correct all those when I rebase off main

packages/components/src/__actions__/Button/v3/Button.module.scss
gyfchong: We need Storybook's config to separate the process of tailwind and `modules.css` or something around that space needs to fix.

gyfchong:
Should we merge this into Menu first?

mcwinter07:
> Should we merge this into Menu first?

@gyfchong I think the plan was to merge them in separately but wait till Menu had been merged in first

packages/components/src/__actions__/Button/v3/Button.module.scss
jakubriedl: all good points, yes and yes.
Let's talk about them on next opportunity how we can plan them and get them done. But for this PR let's not baloon the scope.

gyfchong (approved):

gyfchong (approved):

PR #4832, opened 2024-07-16
fix(Menu): Ensure focus returns to correct menu button on close

dougmacknz:
## Why
Fixes an issue where there's multiple menus on a page, when closing a menu, the focus goes to a different menu button

React focus lock is getting confused about which trigger button to return focus to. It happens when focus is on one menu button when another one is opened, as shown in the video below.

https://github.com/user-attachments/assets/d48a68b7-9b39-4327-a08e-808d7043e89e

## What
Override React focus lock's default focus return, hard coding to always return to the menu's trigger button.

### DIFF
--- a/.changeset/selfish-ears-lick.md
+++ b/.changeset/selfish-ears-lick.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix bug where menu focus goes to a different menu button on close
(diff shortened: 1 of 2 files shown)


sentience (approved):
A straightforward change. Nice work, Dug!

gyfchong (approved):

PR #4837, opened 2024-07-16
fix rest props spread in Brand

mcwinter07:
## Why
- A quick fix for the spread issue Baptiste [called out here](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1721101717140679).
  - tldr: `classNameOverride` was being passed directly into the `<picture>` element, which was causing a hydration error on consumer end

## What
- reassigns spread of `restProps` to `otherProps` in Brand.tsx


packages/components/src/Brand/Brand.tsx
```diff
@@ -58,11 +58,11 @@ export const Brand = ({
     )
   }
 
-  const { alt, classNameOverride } = restProps
+  const { alt, classNameOverride, ...otherProps } = restProps
   const brandTheme = reversed ? "-reversed" : "-default"
 
   return (
-    <picture className={classNameOverride} {...(restProps as PictureProps)}>
+    <picture className={classNameOverride} {...(otherProps as PictureProps)}>
```
HeartSquared, line 65: Either my IDE isn't showing me errors, otherwise this should be okay now :)

```suggestion
    <picture className={classNameOverride} {...otherProps}>
```

HeartSquared (approved):

PR #4839, opened 2024-07-16
ci: fix turbo falsely caching

HeartSquared:
## Why

Storybook tests have been falsely failing in CI due to the Storybook build being cached despite changes having been made.

## What

Update turbo config.
Note: Apparently when a flag is detected after the turbo command, cache will always not be found. Given we only have one Storybook, it's less of an issue for now not having it cached often.


package.json
```diff
@@ -17,9 +17,10 @@
     "storybook": "storybook dev -p ${STORYBOOK_PORT:-4479} -c storybook",
     "storybook:preview": "NODE_ENV=production pnpm storybook --docs",
     "storybook:clean": "rm -rf storybook/public",
-    "storybook:build:root": "storybook build -c storybook -o storybook/public",
-    "storybook:build:test:root": "pnpm storybook:build:root --test",
-    "storybook:build:prod:root": "pnpm storybook:build:root --docs",
+    "storybook:build": "storybook build -c storybook -o storybook/public",
+    "storybook:build:root": "pnpm turbo storybook:build",
```
HeartSquared, line 21: Partly wondered whether I should call it `storybook:build:full` or `:all` but that sounded pretty weird so I just kept it as `:root`. Open to clearer names if anyone can think of one.

gyfchong (approved):

PR #4840, opened 2024-07-16
Fix: Right-aligned ContextModal heading text for RTL languages

mschimk1:

## Why

This PR fixes an issue with the `ContextModal` component heading text alignment for RTL languages. Please view storybook screenshots attached below for a visual of the change.

## What

The `ContextModal` component heading text is left-aligned for both ltr and rtl directions. It should be aligned to the right for RTL text direction.

**Before**

Screenshot 2024-07-17 at 7 53 22 AM

**After**

Screenshot 2024-07-17 at 7 53 57 AM


gyfchong:
Thanks for the PR!

Given that the close button sits above the title in the LTR version, you'll need to move it to the top right corner in your RTL version.

packages/components/src/Modal/ContextModal/ContextModal.module.scss
```diff
@@ -34,6 +34,10 @@
   color: $color-purple-800;
   text-align: left;
```
gyfchong, line 35: I think you can achieve the same affect by using `start` here, rather than implementing a `dir`

```suggestion
  text-align: start;
```

mschimk1: Okay, thanks for your feedback @gyfchong. I will make this change.

gyfchong (approved):

PR #4844, opened 2024-07-18
DO NOT MERGE: trigger ErrorPage chromatic snapshot

HeartSquared:
## DO NOT MERGE

Testing only. See if the ErrorPage time out is flaky or consistent.


HeartSquared:
Tested:
1. A change in `ErrorPage` stickersheet with Turbosnap still enabled - Passed
2. Disabling Turbosnap overall - Passed

It seems more like what happened in `main` was likely a flaky timeout, so we'll just keep an eye on whether it happens again in the future.

PR #4845, opened 2024-07-18
fix(Menu): Pass onClick through even when rendered as a link

dougmacknz:
## Why
No reason to block this, it's needed in some cases where you want to do something before navigating, like tracking. Best to still keep the element as a link in the DOM in this case.

## What
Currently, `onClick` is not being passed through in the case that MenuItem has an `href` and renders a link

### DIFF
--- a/.changeset/lazy-crabs-pay.md
+++ b/.changeset/lazy-crabs-pay.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Pass onClick through to MenuItem even when it is rendered as a link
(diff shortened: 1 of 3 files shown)


Moskvyak (approved):

Moskvyak (approved):

mcwinter07 (commented):
@dougmacknz I think Cass was saying if we rebase off main and force push to the branch this will (hopefully) pass chromatic. Holler if you want me to give that a crack :)

mcwinter07:
> @dougmacknz I think Cass was saying if we rebase off main and force push to the branch this will (hopefully) pass chromatic. Holler if you want me to give that a crack :)

nvm, commented just as you did 😆

mcwinter07 (approved):
Looks hunky dory - I think the story flagged in chromatic wasn't picked up when it got merged last time so probs why its in the diff

PR #4852, opened 2024-07-19
ci(chromatic): upgrade chromatic GH action to v11

HeartSquared:
## Why

Reached out to Chromatic support for recent requests. They mentioned we were 10 major versions behind.

## What

Upgrade Chromatic GH action to v11


gyfchong (dismissed):
No `latest`? Too safe! :P

gyfchong (approved):

PR #4862, opened 2024-07-22
feat: accessibility colour update

HeartSquared:
## Why

[KZN-2525](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2525)

Updating UI as per [RFC](https://cultureamp.atlassian.net/wiki/spaces/DES/pages/3745153025/Improving+color+contrast+and+simplifying+semantic+colours+Design+Request+for+Comment+RFC).

## Notes
- To check: `Tile` back is always blue regardless of the variant. Border set to blue.
- Adding new variants for `Card` and `Well` will be addressed in a following PR (OOS here).

## What

- Added new design-token `$border-width-1`
- Card:
  - Add border to variants
- Collapsible:
  - Add border to `Collapsible` and `CollapsibleGroup`
  - Update border for `ExpertAdviceCollapsible`
  - Fix border-radius between header and content
- EmptyState:
  - Add border to variants
  - Change `action` (to later be deprecated) to match `negative` variant color
- FieldMessage:
  - Update cautionary icon color
- GuidanceBlock:
  - Add border to variants
  - Change `assertive` (to later be deprecated) to match `negative` variant color
- Modals:
  - Update `ConfirmationModal` cautionary icon color
  - Change `assertive` (to later be deprecated) to match `negative` variant color
- Notifications:
  - Update border for variants
  - Update cautionary icon color
- Popover:
  - Add border to variants
  - Update cautionary icon color
- Tag (Legacy):
  - Update cautionary icon color
- Tiles:
  - Add border to variants
  - Change `assertive` (to later be deprecated) to match `negative` variant color
- Tooltip (v1):
  - Update border for variants
- Well:
  - Update border for variants

[KZN-2525]: https://cultureamp.atlassian.net/browse/KZN-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Collapsible/Collapsible/Collapsible.module.scss
```diff
@@ -39,6 +39,7 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
 .container {
   background-color: $color-white;
   box-shadow: $shadow-small-box-shadow;
+  border: $border-width-1 $border-solid-border-style $color-gray-500;
   border-radius: $border-borderless-border-radius;
```
HeartSquared, line 43: As per Jess, the Figma designs have a straight line (no border-radius) between the title and the content when it is open.

To implement the style fix.

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.module.scss
```diff
@@ -60,36 +62,40 @@ $tilePaddingTop: $spacing-xl;
...
   justify-content: space-between;
+  border-color: $color-gray-500;
   padding: $tilePaddingTop $spacing-md $spacing-md;
 }
 
 .faceBack {
   padding-top: $tilePaddingTop;
   transform: rotateY(180deg);
+  border-color: $color-blue-400;
   background-color: $color-blue-100;
```
HeartSquared, line 73: To investigate the usage of Tiles with a back and let Jess know, as her expectations (without knowing context) would be that the back would either match the front or be the default colour (white).

packages/components/src/Collapsible/Collapsible/Collapsible.module.scss
```diff
@@ -39,6 +39,7 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
 .container {
```
HeartSquared, line 39: To investigate usages of the `clear` variant or lack of variant (not the same as `default`) and let Jess know.
Will create a follow up task for alignment pending the investigation (implementation OOS here).

packages/components/src/Card/Card.module.scss
```diff
@@ -4,32 +4,40 @@
 
 .wrapper {
```
HeartSquared, line 5: Delete the reversed stickersheet for Card as there are no such variants.

gyfchong:
Might be worth considering updating the variable syntax to CSS vars

HeartSquared:
> Might be worth considering updating the variable syntax to CSS vars

Same question as in the other PR - what's the strategy for this rollout?

1. Update only the changed values with new syntax
  - Pros: Easy boyscouting
  - Cons: A single file will have both new and old syntax mixed together
2. Update the whole file with new syntax
  - Pros: Clean file
  - Cons: Harder to review as there are more changes to review, and only some require a value change (becomes easy to miss if not changed when it should have)
3. Do not update syntax now, but instead update separately
  - Pros: Keeps this PR clean and easy to review
  - Cons: Might not get around to it any time soon as it requires justification to do the work

Ideally I wouldn't like to mix changes to values with syntax refactors, so I'd lean towards option 3, but I also can accept option 1 if we really want it.

gyfchong:
Option 1. We improve as we go even at the cost of unclear PR's.

But we could mitigate this with comments around the main changes.

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.module.scss
HeartSquared: Leaving this for a later design review.

packages/components/src/Collapsible/Collapsible/Collapsible.module.scss
```diff
@@ -39,25 +35,17 @@ $border-color: rgba($color-gray-600-rgb, 0.2);
...
   background-color: $color-white;
   box-shadow: $shadow-small-box-shadow;
+  border: var(--border-width-1) var(--border-solid-border-style)
```
mcwinter07, line 38: We'll need to calc the border width from the padding

mcwinter07 (commented):
Capture the bizarre experience we found while pairing for action in Empty state. Step to replicate: Hard refresh on the reversed variant Screenshot 2024-07-25 at 5 01 05 pm

Screenshot 2024-07-25 at 5 07 43 pm

packages/components/src/Collapsible/Collapsible/Collapsible.module.scss
HeartSquared: As per Jess, to review all the border changes and adjust the padding so the height is the same as it was prior

HeartSquared:
> Capture the bizarre experience we found while pairing for action in Empty state.

Spoke to Jess and we will update the asset to the same one as the Negative variant

packages/components/src/EmptyState/EmptyState.module.scss
```diff
@@ -3,21 +3,22 @@
 @import "~@kaizen/design-tokens/sass/border";
 @import "./mixins";
 
+$border-width: var(--border-width-1);
```
gyfchong, line 6: Isn't this a bit unnecessary?

HeartSquared: Hmm true. I want a scoped variable for calculations, and sometimes they ended up in different classes. But if not, should be doable in CSS :) Will play around with it

packages/components/src/Notification/subcomponents/GenericNotification/_mixins.scss
```diff
@@ -5,9 +5,8 @@
 @import "~@kaizen/design-tokens/sass/border";
 @import "~@kaizen/design-tokens/sass/animation";
 
-$notification-vertical-padding: calc(
-  #{$spacing-6} - #{$border-solid-border-width}
-);
+$border-width: var(--border-width-1);
```
gyfchong, line 8:
```suggestion
--border-width: var(--border-width-1);
```

packages/components/src/Popover/Popover.module.scss
```diff
@@ -44,24 +44,32 @@ $large-width: 450px;
...
   border-radius: $border-solid-border-radius;
   color: $color-purple-800;
   text-align: left;
-  padding: $spacing-sm 1rem 1rem;
+  padding-block: calc(var(--spacing-12) - var(--border-width))
+    calc(var(--spacing-16) - var(--border-width));
+  padding-inline: calc(var(--spacing-16) - var(--border-width));
 }
 
+$default-box-border-color: var(--color-gray-500);
```
gyfchong, line 59:
```suggestion
--default-box-border-color: var(--color-gray-500);
```

packages/components/src/Popover/Popover.module.scss
```diff
@@ -100,15 +108,17 @@ $negative-box-border-color: $color-red-300;
   @include arrow($color-red-100, $negative-box-border-color);
 }
 
+$cautionary-box-border-color: var(--color-yellow-700);
```
gyfchong, line 111:
```suggestion
--cautionary-box-border-color: var(--color-yellow-700);
```

gyfchong (approved):
Minor inconsistencies with CSS vars usage, but merge if you want!

packages/components/src/Notification/subcomponents/GenericNotification/_mixins.scss
HeartSquared: As discussed, this one is a bit special since it's used in another scss variable, and it is set globally in the file. I've only been creating the css vars in scoped contexts to avoid name clashes.

gyfchong (approved):

mcwinter07 (approved):

PR #4864, opened 2024-07-24
KZN-2534-dependency-updates

gyfchong:


packages/components/package.json
```diff
@@ -118,7 +118,8 @@
   "devDependenciesComments": {
     "postcss": "Installed in root",
     "postcss-cli": "Installed in root",
-    "postcss-preset-env": "Installed in root"
+    "postcss-preset-env": "Installed in root",
+    "react-day-picker": "v9 causes typescript errors"
```
gyfchong, line 122: 🎏 Pinning to v8

gyfchong:
```suggestion
    "react-day-picker": "Pinned to v8, v9 causes typescript errors"
```

packages/components/src/RichTextEditor/utils/core/createRichTextEditor.ts
```diff
@@ -31,6 +31,8 @@ export const createRichTextEditor = ({
   attributes,
   isEditable = () => true,
 }: EditorArgs): EditorAPI => {
+  let editorView: ProseMirrorEditorView | null = null
```
gyfchong, line 34: This seems to have fixed the flakiness

packages/components/package.json
```diff
@@ -118,7 +118,8 @@
   "devDependenciesComments": {
     "postcss": "Installed in root",
     "postcss-cli": "Installed in root",
-    "postcss-preset-env": "Installed in root"
+    "postcss-preset-env": "Installed in root",
+    "react-day-picker": "Pinned to v8, v9 causes typescript errors"
```
HeartSquared, line 122: 1. Already a comment in L95 :)
2. When did it get unlocked to begin with 😂 Nonetheless, we should actually be trying to upgrade to v9 and fixing our code because they're meant to have fixed the a11y issues they introduced prior.

packages/components/src/RichTextEditor/utils/core/createRichTextEditor.ts
HeartSquared: 🤞🏻

packages/components/package.json
gyfchong: In another PR

HeartSquared (approved):
LGTM

PR #4867, opened 2024-07-26
Rename mood to color

gyfchong:
## Why
In our effort to replace all `mood` props with a more consistent API 

## What
- Deprecation of `mood`
- Introduction of `color`

### DIFF
--- a/packages/components/src/BrandMoment/BrandMoment.tsx
+++ b/packages/components/src/BrandMoment/BrandMoment.tsx
@@ -58,7 +58,7 @@ export const BrandMoment = ({
     <div
       className={classnames(
         styles.body,
-        mood ? styles[mood] : styles[color],
+        color ? styles[color] : styles[mood],
         classNameOverride
       )}
       {...restProps}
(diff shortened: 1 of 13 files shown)


packages/components/src/BrandMoment/BrandMoment.tsx
```diff
@@ -10,7 +10,9 @@ import { useMediaQueries } from "~utils/useMediaQueries"
 import styles from "./BrandMoment.module.scss"
 
 export type BrandMomentProps = {
-  mood: "informative" | "positive" | "negative"
+  // @deprecated use `color` instead
+  mood?: "informative" | "positive" | "negative"
+  color: "blue" | "green" | "red"
```
HeartSquared, line 15: Making this required makes it a breaking change 😅 You might need a discriminated union to enforce one or the other as required

gyfchong: Oh I got this the other way round, maybe make `color` optional instead?

PR #4868, opened 2024-07-26
KZN-2526/deprecate-modal-moods

gyfchong:
## Why
Part of the component-wide deprecation of moods

## What
- ConfirmationModal: Deprecated `mood` in favour of `variant`
- Codemod for ConfirmationModal to transition from `mood` to `variant`
- InputEditModal: Deprecate `mood` in favour of a single default style
- Codemod for removing the `mood` prop from InputEditModal

## Notes
There is no Codemod to transition consumers that use `destructive` mood, as data shows that very few consumers use the `destructive` option on `InputEditModal` and for those few the `footerActions` suffice as a workaround should they need to keep their `destructive` style button.

Data also shows that only one consumer app uses the `ModalFooter` externally from the `InputEditModal`, and it's used from the legacy packages, so again no Codemod was created to help that transition.


packages/components/src/Modal/ConfirmationModal/ConfirmationModal.module.scss
```diff
@@ -35,14 +34,87 @@ $dt-assertive-header-background: $color-orange-100;
...
+  @media (max-width: $layout-breakpoints-medium) {
+    padding: $spacing-md;
+  }
+}
+
+// --------------------------------
+// @deprecated moods use variant styles instead
+// --------------------------------
+
+$dt-positive-header-background: $color-green-100;
```
mcwinter07, line 110: Just confirming we're only keeping these because they'll be axed in the next major?

.changeset/heavy-badgers-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Deprecate mood for ConfirmationModal and InputEditModal
```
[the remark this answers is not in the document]
gyfchong: Nah, don't think it's necessary

packages/components/src/Modal/ConfirmationModal/ConfirmationModal.module.scss
gyfchong: Yep, on the chopping block

mcwinter07 (approved):

PR #4869, opened 2024-07-26
GuidanceBlock v2

gyfchong:
## Why
To reduce the amount of variants in GuidanceBlock

## What
- `variant` only accepts `default` and `expert-advice`
- Moved GuidanceBlock to the v2 entrypoint


packages/components/src/index.ts
```diff
@@ -20,7 +20,7 @@ export * from "./ErrorPage"
 export * from "./FieldGroup"
 export * from "./FieldMessage"
 export * from "./Filter"
-export * from "./GuidanceBlock"
```
mcwinter07, line 23: Do we want to leave this here for to allow folks to transition to the versioned structure?

mcwinter07: nvm, missed the bottom import

mcwinter07 (approved):

PR #4877, opened 2024-07-29
fix(Modal): Remove focus ring on title when focus-visible

dougmacknz:
## What
Adjusting the focus styling when first opening a modal with a keyboard, to wrap the modal container instead of the title

Before:
image

After:
image

## Why
Looks nicer

### DIFF
--- a/packages/components/src/Modal/GenericModal/GenericModal.module.scss
+++ b/packages/components/src/Modal/GenericModal/GenericModal.module.scss
@@ -54,6 +54,20 @@
   @include ca-media-tablet-and-up {
     width: 100%;
   }
+
+  // wrap the modal container with a focus ring when the title has focus
+  // don't copy this over on the rebuild
+  // rather than putting the focus on the title, put it on the role=dialog element itself
+  &:has(:global([class*="modalLabel"]):focus-visible)::after {
+    content: "";
+    position: absolute;
+    background: transparent;
+    border-radius: $border-focus-ring-border-radius;
+    border-width: $border-focus-ring-border-width;
+    border-style: $border-focus-ring-border-style;
+    border-color: $color-blue-300;
+    inset: -4px;
+  }
 }
 
 // WAITING FOR TEXTFIELD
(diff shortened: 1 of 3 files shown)


packages/components/src/Modal/GenericModal/GenericModal.module.scss
```diff
@@ -54,6 +54,20 @@
   @include ca-media-tablet-and-up {
     width: 100%;
   }
+
+  // wrap the modal container with a focus ring when the title has focus
+  // don't copy this over on the rebuild
+  // rather than putting the focus on the title, put it on the role=dialog element itself
+  &:has(:global([class*="modalLabel"]):focus-visible)::after {
```
gyfchong, line 61: 🧙‍♂️

gyfchong: I don't see a ring when I go to ContextModal stories 🙈

gyfchong: Maybe it's being clipped somehow
image

But Chromatic shows it!

HeartSquared: It's `focus-visible` so you'll need to not use the mouse 😉

gyfchong: Ohhh duh

gyfchong (approved):

packages/components/src/Modal/GenericModal/GenericModal.module.scss
dougmacknz: fyi I tried a couple of things to avoid getting this hacky:
- tried moving the focus to the dialog instead of the title
  - unfortunately the styling is put on an inner child of the dialog, so that would require more work
- tried giving this element an extra class from the GenericModal.module.scss file so that I could target that instead of using `:global` with a wildcard, but it didn't work for some reason.

but it's gonna be legacy code so i'm saying good enough

PR #4922, opened 2024-08-13
updated codemod to match only GuidanceBlock

gyfchong:
## Why
`GuidanceBlock` is placed in a new entrypoint

## What
- Updated the existing codemod to modify the import statment with the correct family and version


mcwinter07 (approved):
👨‍🍳 Nice fix

PR #4931, opened 2024-08-16
Support css modules in storybook

gyfchong:
## Why
Storybook's existing config to process styles is outdated and cannot support both SCSS and CSS modules at the same time.

## What
- Created a new storybook config using the latest Vite which supports the functionality out of the box and provides a significant speed boost.

## Notes
The DST has had conversations about using `next-storybook` to remain consistent with the rest of the business however the DS goals require much less config (and support for different config eg. MDX) than the rest of the business, which could lead to slower development time.

Also on the note of config, the `next-storybook` makes heavy use of custom plugins, due to the complexity it has been difficult to upgrade it to newer versions and config syntax, namely the move to ESM which will take some time to resolve all dependencies which don't support ESM. So if Kaizen were to adopt this technology, we'd potentially lose out on benefits which could improve our productivity.

That isn't to say we shouldn't ever adopt `next-storybook` just that this isn't the right time to do it as we don't gain very much for the time invested in potentially upgrading it ourselves. Eventually I can see the `next-storybook` becoming simpler as it no longer needs to cater for things like excluding/including packages for processing styles.


.changeset/strong-pigs-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": minor
+---
+
+Exported types from LikertScale
```
HeartSquared, line 5: This doesn't seem right 😛

gyfchong: It was never exported and Storybook needs it

.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 yarn-*.log
 /storybook/public/
```
gyfchong, line 3:
```suggestion
```

.changeset/strong-pigs-return.md
HeartSquared: Ohhh. So it's actually not a mistake 🤯

package.json
```diff
@@ -12,19 +12,19 @@
     "clean": "pnpm turbo clean && pnpm storybook:clean && rimraf node_modules",
     "test": "pnpm turbo test",
     "test:ci": "pnpm turbo test:ci",
-    "storybook": "storybook dev -p ${STORYBOOK_PORT:-6006} -c storybook",
+    "storybook": "pnpm -F storybook storybook",
```
HeartSquared, line 15: Whilst I understand how this works, it does look a bit weird. Perhaps worth making a `pkg:storybook` like we did with `pkg:aio`?

package.json
```diff
@@ -12,19 +12,19 @@
...
     "storybook:preview": "NODE_ENV=production pnpm storybook --docs",
```
HeartSquared, line 16: You missed this one :)

.storybook/main.ts
```diff
@@ -0,0 +1,85 @@
...
+ * It is needed in projects that use Yarn PnP or are set up within a monorepo.
+ */
+function getAbsolutePath(value: string): any {
+  return dirname(require.resolve(join(value, "package.json")))
+}
+
+const config: StorybookConfig = {
+  stories: [
+    "../packages/**/*.mdx",
+    "../packages/**/*.stories.tsx",
```
HeartSquared, line 16: Hmm...so this is actually still the dev env for all packages then?
In that case, the dist stylesheets imported into the preview files will not be required given that Storybook will be building off the source code (would be nice to be able to test the dist styles, but can be considered a separate task).

jakubriedl:
> The DST has had conversations about using next-storybook to remain consistent with the rest of the business however the DS goals require much less config than the rest of the business, which could lead to slower development time.

> Also on the note of config, the next-storybook makes heavy use of custom plugins, due to the complexity it has been difficult to upgrade it to newer versions and config syntax, namely the move to ESM (historically, 1 PE, 2 Staff, 2 Senior engineers have tried over the past 3months, each taking avg of a 1 month with no success). So if Kaizen were to adopt this technology, we'd potentially lose out on benefits which could improve our productivity.

> That isn't to say we shouldn't ever adopt next-storybook just that this isn't the right time to do it as we don't gain very much for the time invested in potentially upgrading it ourselves. Eventually I can see the next-storybook becoming simpler as it no longer needs to cater for things like excluding/including packages for processing styles.

I understand why not to update to`next-storybook` now, but I don't think the facts and perception here is right. The upgrade issues were not because of addons, and we haven't spent 3 months on it. I've spent a day or two tops. There were just other priorities.

.storybook/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@storybook/addon-links": "^8.2.9",
+    "@storybook/addon-onboarding": "^8.2.9",
+    "@storybook/blocks": "^8.2.9",
+    "@storybook/react": "^8.2.9",
+    "@storybook/react-vite": "^8.2.9",
+    "@storybook/test": "^8.2.9",
+    "@types/react": "^18.3.3",
+    "@types/react-dom": "^18.3.0",
+    "@vitejs/plugin-react": "^4.3.1",
+    "eslint": "^9.9.0",
```
HeartSquared, line 33: Probably better to make the eslint match the version we are using for all the other things (unless we upgrade them all)

.storybook/pages/basic-component.md
HeartSquared, line 1: I don't think this was ever in Storybook?

.storybook/pages/code-standards.md
HeartSquared, line 1: Same with this - not in Storybook.

.storybook/pages/tokens/color/color-tokens.stories.tsx
```diff
@@ -190,10 +190,6 @@ export const ColorTokens: StoryFn = () => (
           <Heading variant="heading-3">White</Heading>
           <ColorDemo name="White" color={tokens.color.white} />
         </Stack>
-        <Stack gapSize={24}>
```
HeartSquared, line 193: This should probably come back :)

.storybook/preview.css
```diff
@@ -1,10 +1,3 @@
-@import "../../packages/components/styles/global.css";
-@import "highlight.js/styles/a11y-light.css";
```
HeartSquared, line 2: You may need the highlight styles back for code snippets (maybe can be moved into preview.tsx?)

.storybook/utils/backgrounds.ts
```diff
@@ -1,22 +1,22 @@
...
-  { name: "Gray 100", value: defaultTheme.color.gray["100"] },
-  { name: "Gray 300", value: defaultTheme.color.gray["300"] },
-  { name: "Purple 700", value: defaultTheme.color.purple["700"] },
-  { name: "Blue 500", value: defaultTheme.color.blue["500"] },
-  { name: "Green 500", value: defaultTheme.color.green["500"] },
-  { name: "Yellow 500", value: defaultTheme.color.yellow["500"] },
-  { name: "Orange 500", value: defaultTheme.color.orange["500"] },
-  { name: "Red 500", value: defaultTheme.color.red["500"] },
-]
+// export const backgrounds = [
```
HeartSquared, line 6: Can just delete these :)

packages/components/src/Brand/Brand.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { HTMLAttributes, SVGAttributes } from "react"
-import { assetUrl } from "@kaizen/hosted-assets"
-import { OverrideClassName } from "~types/OverrideClassName"
+import { OverrideClassName } from "~components/types/OverrideClassName"
+import { assetUrl } from "~components/utils/hostedAssets"
```
HeartSquared, line 3: Potentially OOS: Should we deprecate the package? Or just leave it?

packages/components/package.json
HeartSquared, line 1: If we no longer use `hosted-assets`, we can remove the dep

packages/design-tokens/less/color.less
```diff
@@ -208,7 +208,3 @@
 @color-white-rgb: var(--color-white-rgb, 255, 255, 255);
 @color-white-rgb-id: --color-white-rgb;
 @color-white-id: --color-white;
-@color-black: var(--color-black, #000000);
```
HeartSquared, line 211: Let's put this back :)

packages/design-tokens/less/spacing.less
```diff
@@ -8,7 +8,6 @@
 @spacing-8: var(--spacing-8, 0.5rem);
 @spacing-12: var(--spacing-12, 0.75rem);
 @spacing-16: var(--spacing-16, 1rem);
-@spacing-20: var(--spacing-20, 1.25rem);
```
HeartSquared, line 11: ↩️

packages/design-tokens/package.json
```diff
@@ -17,7 +17,9 @@
     "js",
     "dist"
   ],
-  "main": "dist/index.js",
+  "main": "dist/cjs/index.cjs",
+  "module": "dist/esm/index.mjs",
+  "types": "dist/types/index.d.ts",
   "private": false,
   "license": "MIT",
   "sideEffects": false,
```
HeartSquared, line 25: Might need to review the sideEffects of this package if this now supports tree-shaking

packages/design-tokens/sass/color.scss
```diff
@@ -208,7 +208,3 @@ $color-white: var(--color-white, #ffffff);
 $color-white-rgb: var(--color-white-rgb, 255, 255, 255);
 $color-white-rgb-id: --color-white-rgb;
 $color-white-id: --color-white;
-$color-black: var(--color-black, #000000);
```
HeartSquared, line 211: ↩️

packages/design-tokens/sass/spacing.scss
```diff
@@ -8,7 +8,6 @@ $spacing-6: var(--spacing-6, 0.375rem);
 $spacing-8: var(--spacing-8, 0.5rem);
 $spacing-12: var(--spacing-12, 0.75rem);
 $spacing-16: var(--spacing-16, 1rem);
-$spacing-20: var(--spacing-20, 1.25rem);
```
HeartSquared, line 11: ↩️

packages/design-tokens/src/__tests__/sass-tokens.spec.ts
```diff
@@ -47,8 +47,7 @@ ${typography.toString()}
       const newColorsWithHeartNames = newColors.filter(
         n =>
           heartColorNamePattern.test(n.prop) ||
-          n.prop.startsWith("$color-white") ||
-          n.prop.startsWith("$color-black")
```
HeartSquared, line 51: ↩️

packages/design-tokens/src/js/tokens.ts
```diff
@@ -112,7 +112,6 @@ export const tokens: TokenStructure = {
       600: "#524e56",
     },
     white: "#ffffff",
-    black: "#000000",
```
HeartSquared, line 115: ↩️

packages/design-tokens/src/types.ts
```diff
@@ -141,7 +141,6 @@ export type TokenStructure = {
       "600": Hex
     }
     white: Hex
-    black: Hex
```
HeartSquared, line 144: ↩️

packages/design-tokens/tokens/color.json
```diff
@@ -223,10 +223,6 @@
     "white": "var(--color-white, #ffffff)",
     "white-rgb": "var(--color-white-rgb, 255, 255, 255)",
     "white-rgb-id": "--color-white-rgb",
-    "white-id": "--color-white",
-    "black": "var(--color-black, #000000)",
```
HeartSquared, line 227: ↩️

packages/design-tokens/tokens/spacing.json
```diff
@@ -8,7 +8,6 @@
     "8": "var(--spacing-8, .5rem)",
     "12": "var(--spacing-12, .75rem)",
     "16": "var(--spacing-16, 1rem)",
-    "20": "var(--spacing-20, 1.25rem)",
```
HeartSquared, line 11: ↩️

packages/hosted-assets/CHANGELOG.md
```diff
@@ -81,7 +81,7 @@ Updates the package API to export the asset URL functions from the root of the m
 ```js
 // TS/JS
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { assetUrl } from "@kaizen/hosted-assets"
+import { assetUrl } from "~utils/hostedAssets"
```
HeartSquared, line 84: This is a changelog :P

packages/hosted-assets/README.md
```diff
@@ -26,7 +26,7 @@ asset-url("some/blob.png") // -> "https://<origin>/some/blob.png"
 3. Import the `@kaizen/hosted-assets` package and you can access the image through the `assetUrl`/`asset-url` function:
 
 ```js
-import { assetUrl } from "@kaizen/hosted-assets";
+import { assetUrl } from "~utils/hostedAssets";
```
HeartSquared, line 29: You probably want to leave this as before

packages/hosted-assets/index.ts
```diff
@@ -8,6 +8,8 @@ const ORIGIN_BASE_URL = "https://d1e7r7b0lb8p4d.cloudfront.net"
  * assetUrl("some/blob.png") -> "https://<origin>/some/blob.png"
  *
  * @see https://github.com/cultureamp/kaizen-design-system-assets/
+ *
+ * @deprecated import from `@kaizen/components` instead
```
HeartSquared, line 12: For external consumers? Does it come through from `@kaizen/components` or from a subpath?

packages/tailwind/package.json
```diff
@@ -19,12 +20,21 @@
...
+  "main": "dist/cjs/index.cjs",
+  "module": "dist/esm/index.mjs",
+  "types": "dist/types/index.d.ts",
   "author": "",
   "private": false,
   "license": "MIT",
   "dependencies": {
-    "@kaizen/design-tokens": "workspace:*"
+    "@kaizen/design-tokens": "workspace:*",
+    "@kaizen/components": "workspace:*"
```
HeartSquared, line 31: I'm a bit wary of this as a dep. KAIO has this package as a dep too doesn't it? So it would become circular.

packages/tailwind/src/_docs/pages/configuration.mdx
```diff
@@ -1,5 +1,5 @@
 import { Meta } from "@storybook/blocks"
-import { InlineNotification } from "~components/Notification"
+import { InlineNotification } from "@kaizen/components"
```
HeartSquared, line 2: Given it is only docs, and this creates a possible circular dependency, it might be safer to have this as a local alias

packages/tailwind/src/_docs/pages/modifiers/media-queries.stories.tsx
```diff
@@ -1,7 +1,6 @@
 import React from "react"
 import { Meta, StoryFn } from "@storybook/react"
-import { Heading } from "~components/Heading"
-import { Text } from "~components/Text"
+import { Heading, Text } from "@kaizen/components"
```
HeartSquared, line 3: Same comment as the other import - might be better as alias

packages/tailwind/src/_docs/pages/working-with-tailwind.stories.tsx
```diff
@@ -1,14 +1,14 @@
...
   CustomSurvey,
   CustomUnattributedSurvey,
-} from "~components/Illustration"
-import { Skirt } from "~components/Skirt"
-import { Button } from "~components/__actions__/v2"
-import { TailwindStoryTemplate } from "~storybook/components/TailwindStoryTemplate"
+  Skirt,
+  Heading,
+} from "@kaizen/components"
+import { Button } from "@kaizen/components/v2/actions"
```
HeartSquared, line 10: Ah, I recall this caused funny errors with alias at first. Perhaps a devDep may be safe enough? We'd need to investigate.

HeartSquared:
>The DST has had conversations about using `next-storybook` to remain consistent with the rest of the business however the DS goals require much less config than the rest of the business, which could lead to slower development time.

It wasn't just that we require much less config, but we also have other config requirements that do not exist in it (eg. MDX support for docs).

.storybook/main.ts
gyfchong: https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4112122241/Idea+Introduce+a+production+flag+to+Stories

packages/components/src/Brand/Brand.tsx
gyfchong: Leave it, doesn't hurt anyone

packages/design-tokens/package.json
gyfchong: All JS tokens are in a single file anyways, nothing to treeshake... yet...

packages/hosted-assets/index.ts
gyfchong: Comes through `@kaizen/components`

packages/tailwind/src/_docs/pages/configuration.mdx
gyfchong: Works for MDX but fails for the tsx stories, I think it works fine though

.storybook/components/DosAndDonts/DoesAndDonts.tsx
HeartSquared, line 1: Maybe fix the name of the component while you're here :P

.storybook/components/StickerSheet/types.ts
```diff
@@ -1,5 +1,5 @@
 import { StoryObj } from "@storybook/react"
-import { ColorSchema } from "~components/LikertScaleLegacy/types"
+import { ColorSchema } from "~components/LikertScaleLegacy"
```
HeartSquared, line 2:
```suggestion
import type { ColorSchema } from "~components/LikertScaleLegacy"
```

.storybook/main.ts
```diff
@@ -0,0 +1,85 @@
...
+          {
+            "~storybook": path.resolve(__dirname),
+            "~types": path.resolve(
+              __dirname,
+              "../packages/components/src/types"
+            ),
+            "~utils": path.resolve(
+              __dirname,
+              "../packages/components/src/utils"
+            ),
```
HeartSquared, line 62: Are these still valid?

.storybook/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "build": "storybook build -c ./ -o ./public",
+    "test": "test-storybook -c ./ --skipTags='skip-test' --url http://127.0.0.1:6006",
+    "clean": "rimraf ./public && rimraf ./node_modules"
+  },
+  "dependencies": {
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
+  },
+  "devDependencies": {
+    "@eslint/js": "^9.9.0",
```
HeartSquared, line 20: Does this need a downgrade or is it actually separate from `eslint`? 🤔

.storybook/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "jsx": "react",
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "paths": {
+      "~storybook/*": ["./*"],
+      "~components/*": ["../packages/components/src/*"]
```
HeartSquared, line 9: Perhaps fill in the rest of the aliases? :)

docs/basic-component.md
HeartSquared, line 1: Wait, where did the file go? O_O

docs/code-standards.md
HeartSquared, line 1: This file has gone missing too

package.json
```diff
@@ -7,92 +7,69 @@
...
+    "@storybook/addon-essentials": "^8.2.9",
+    "@storybook/addon-interactions": "^8.2.9",
+    "@storybook/addon-links": "^8.2.9",
+    "@storybook/blocks": "^8.2.9",
+    "@storybook/components": "^8.2.9",
+    "@storybook/core-events": "^8.2.9",
+    "@storybook/react": "^8.2.9",
+    "@storybook/test": "^8.2.9",
+    "@storybook/theming": "^8.2.9",
+    "@stylistic/eslint-plugin": "^2.6.2",
```
HeartSquared, line 60: Can these be removed now that you've moved it into a workspace?

packages/components/package.json
```diff
@@ -22,7 +22,8 @@
     "dist",
     "icons",
     "locales",
-    "src"
+    "src",
+    "styles"
```
HeartSquared, line 26: This shouldn't be in the dist right? 😬

packages/design-tokens/_docs/pages/color-tokens.stories.tsx
```diff
@@ -190,10 +190,6 @@ export const ColorTokens: StoryFn = () => (
           <Heading variant="heading-3">White</Heading>
           <ColorDemo name="White" color={tokens.color.white} />
         </Stack>
-        <Stack gapSize={24}>
-          <Heading variant="heading-3">black</Heading>
-          <ColorDemo name="Black" color={tokens.color.black} />
-        </Stack>
```
HeartSquared, line 196: Restore please :)

packages/design-tokens/package.json
```diff
@@ -17,37 +17,46 @@
...
     "test": "FORCE_COLOR=1 jest",
     "test:ci": "pnpm test -- --ci",
     "build": "pnpm clean && pnpm build:json && pnpm build:less && pnpm build:sass && pnpm build:ts",
     "build:json": "tsx ./bin/buildCssVarTokens.ts",
-    "build:ts": "tsc --project tsconfig.dist.json",
+    "build:ts": "pnpm package-bundler build",
     "build:less": "json-to-flat-sass './tokens/*.json' 'less' --extension 'less' --caseType 'kebab' && prettier less/* --write",
     "build:sass": "json-to-flat-sass './tokens/*.json' 'sass' --extension 'scss' --caseType 'kebab' && prettier sass/* --write",
+    "lint:scss": "pnpm stylelint './src/**/*.scss' --config .stylelintrc-scss.json",
+    "lint:styles": "pnpm lint:css && pnpm lint:scss",
```
HeartSquared, line 36: I can't seem to see a `lint:css`

packages/design-tokens/tsconfig.dist.json
HeartSquared, line 1: I reckon you'll need some extra changesets given you've changed the build process for all these packages

packages/tailwind/package.json
```diff
@@ -19,12 +20,20 @@
...
+  "module": "dist/esm/index.mjs",
+  "types": "dist/types/index.d.ts",
   "author": "",
   "private": false,
   "license": "MIT",
-  "dependencies": {
-    "@kaizen/design-tokens": "workspace:*"
+  "devDependencies": {
+    "@kaizen/components": "workspace:*",
+    "@kaizen/design-tokens": "workspace:*",
```
HeartSquared, line 31: I think design-tokens is still a dep for TW right? Or a peerDep if you'd prefer.

packages/tailwind/src/_docs/pages/modifiers/media-queries.stories.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react"
 import { Meta, StoryFn } from "@storybook/react"
+
```
HeartSquared, line 3: :P
```suggestion
```

packages/tailwind/src/_docs/pages/working-with-tailwind.stories.tsx
```diff
@@ -8,7 +8,8 @@ import {
 } from "~components/Illustration"
 import { Skirt } from "~components/Skirt"
 import { Button } from "~components/__actions__/v2"
-import { TailwindStoryTemplate } from "~storybook/components/TailwindStoryTemplate"
+
```
HeartSquared, line 11: :bye:
```suggestion
```

.storybook/package.json
gyfchong: Nope, can be removed since all linting is done from root

packages/components/package.json
gyfchong: Solved through alises

packages/design-tokens/package.json
gyfchong: Oops, leftover from an experiment with getting the packages to individually lint

package.json
gyfchong: > "@storybook/*": "Required for Storybook stories across all packages",

packages/components/src/utils/useResizeObserver.spec.tsx
```diff
@@ -26,22 +27,26 @@ jest.mock("resize-observer-polyfill", () => ({
 describe("useResizeObserver", () => {
   it("Calls the callback with the expected entries", async () => {
     const callback = jest.fn().mockImplementation(value => value)
-    const { result, waitForNextUpdate, unmount } = renderHook(() =>
-      useResizeObserver(callback)
-    )
+    const { result, unmount } = renderHook(() => useResizeObserver(callback))
     await act(async () => {
```
gyfchong, line 31: This seems to be the problem now @HeartSquared
```
Warning: The current testing environment is not configured to support act(...)
```

Unsure about this fix 🤔 
https://github.com/testing-library/react-testing-library/issues/1061#issuecomment-1119435440

gyfchong: Deprecated in favour of native DOM API:
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

.github/workflows/test-stories.yaml
```diff
@@ -109,17 +109,18 @@ jobs:
...
         with:
           token: ${{ github.token }}
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
-          storybookBuildDir: "./storybook/public"
+          storybookBuildDir: "./docs/storybook-static"
           onlyChanged: "!(main)"
           externals: |
             **/!(*.module).scss
+            **/!(*.module).css
             packages/components/**/*.css
```
HeartSquared, line 124: You'll want to take this out then, because it will match `.module.css` files within KAIO :)

.stylelintignore
```diff
@@ -1,4 +1,9 @@
 packages/design-tokens/**/*.scss
 packages/design-tokens/**/*.css
 packages/components/locales/*
-storybook/styles/*
+packages/components/dist/*
+packages/design-tokens/dist/*
+packages/tailwind/dist/*
```
HeartSquared, line 6: I wonder if just `dist` would work 🤔 But no need to change

gyfchong: It maybe kinda didn't work, but only components needs to be here cause the others don't contain CSS

HeartSquared: Haha drat! Whatever works :)

HeartSquared (approved):

PR #4939, opened 2024-08-20
feat(Avatar): allow control over alt text

dougmacknz:
## Why
It's fairly common to put an avatar next to someone's name. In these cases, we want the alt text to be blank, otherwise it ends up doubling up the same information.

E.g. this common usage in a MultiCombobox:
image

The only way that this can currently be achieved would be to blank out the `fullName` prop, but then you'd miss out on the initials feature.

## What
Add an `alt` prop to the `Avatar` component so that consumers can have control of it separate from `fullName`.

### DIFF
--- a/packages/components/src/Avatar/Avatar.spec.tsx
+++ b/packages/components/src/Avatar/Avatar.spec.tsx
@@ -23,4 +23,16 @@ describe("<Avatar />", () => {
       expect(screen.getByText("JWAVLN")).toBeInTheDocument()
     })
   })
+
+  describe("alt text", () => {
+    it("uses alt prop over full name when using initials", () => {
+      render(<Avatar fullName="Jane Doe" alt="alt override" />)
+      expect(screen.getByTitle("alt override")).toBeInTheDocument()
+    })
+
+    it("uses alt prop over full name when using fallback img", () => {
+      render(<Avatar fullName="Jane Doe" alt="alt override" disableInitials />)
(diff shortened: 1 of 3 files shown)


packages/components/src/Avatar/Avatar.tsx
```diff
@@ -118,6 +124,7 @@ export const Avatar = ({
   disableInitials = false,
   isCompany = false,
   isCurrentUser = true,
+  alt = fullName || "",
```
HeartSquared, line 127: Does this work? O_O
Might be worth adding a test to make sure.

packages/components/src/Avatar/Avatar.spec.tsx
```diff
@@ -23,4 +23,16 @@ describe("<Avatar />", () => {
       expect(screen.getByText("JWAVLN")).toBeInTheDocument()
     })
   })
+
+  describe("alt text", () => {
```
HeartSquared, line 27: I think there needs to be some extra test cases. I see 4 different places you've changed but only have 2 tests :)

dougmacknz: I've built the tests out more. There's 3 possible renders:
* img src supplied
* full name provided with initials feature turned on
* fall back icon with no img or initials

So for each of these I've made sure it:
* uses alt text as the top priority
* uses full name in lieu of alt
* hides the elements if neither is provided

HeartSquared (approved):

PR #4941, opened 2024-08-21
chore(Popover): deprecate `variant` and `customIcon` props

HeartSquared:
## What

Deprecate variant for Popover.
Also fixed the RTL styles.

## Why

[KZN-2528](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2528)
Part of the a11y contrast improvements initiative.

[KZN-2528]: https://cultureamp.atlassian.net/browse/KZN-2528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Popover/Popover.module.scss
```diff
@@ -127,7 +127,7 @@ $cautionary-box-border-color: var(--color-yellow-700);
   display: flex;
   align-items: center;
   white-space: nowrap;
-  padding-right: $spacing-md;
+  padding-inline-end: $spacing-md;
```
mcwinter07, line 130: Nice, could potentially update this to spacing-24, or even the css var if you're keen :)

mcwinter07 (commented):
Should we have a codemod as part of this release or is that a separate task we want to do?

HeartSquared:
> Should we have a codemod as part of this release or is that a separate task we want to do?

Oh good point. I suppose it does make sense to have a codemod that removes the props.

Looks like `customIcon` is never used by anyone either, so it would be pretty safe!
https://github.com/search?q=org%3Acultureamp%20customIcon&type=code

mcwinter07 (approved):
Nice abstraction of the `removeProps` transformer

PR #4943, opened 2024-08-22
Toast notification fix

aaronknol:

## Why

Using the `addToastNotification` function inside a useEffect was causing 4 toast's to appear instead of just one. The same toast id was being used, so I expected only one toast to get added. This is how we were triggering the toasts to be added:

```
const { addToastNotification } = useToastNotification()

  useEffect(() => {
    if (isFetchNextPageError) {

      addToastNotification({
        id: "add-note-error",
        title: ...
        type: "negative",
        message: ...
      })
    }
  }, [addToastNotification, formatMessage, isFetchNextPageError])
  ```

Each time the same id is used, yet that didn't prevent more than one version of the toast to be created.
By setting a breakpoint on the `notifications.find` in ToastNotificationContext and running in browser, we saw that the `notifications` array was empty when the page loaded, (stopping 4 times at this breakpoint with notifcations being empty) and then notifications had 4 items in it - and nothing in between. This led us to looking at how the `notificationsExist` was working out if the notification was in the array yet or not

### CODE REVIEW

File: packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx [hunk 1]
```diff
@@ -44,18 +44,23 @@ export const ToastNotificationProvider = ({
...
-
-      const notificationExists = notifications.find(
-        ({ id }) => id === notification.id
-      )
-
-      if (!notificationExists) {
-        setNotifications(existing => [...existing, notificationWithId])
-      }
-    }
+    useCallback(
```
aaronknol, line 47: Wrapping useCallback around this because in the consumer it could be used as a dependancy of a useEffect

packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx
```diff
@@ -44,18 +44,23 @@ export const ToastNotificationProvider = ({
...
+      notification => {
+        const uuid = uuidv4()
+        const notificationWithId = { id: uuid, ...notification }
+
+        setNotifications(existing => {
```
aaronknol, line 52: From our testing, `notifications` is always empty, but due to our retry logic this function is called very quickly whilst notifications hasn't been updated yet. Due to that, we moved the "checking" logic inside the callback for setNotifications- as at that point, `existing` will have the correct value

HeartSquared: Hey Aaron, could you add a test for this fix please? :)

.changeset/kind-eggs-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Fix toast notification
```
HeartSquared, line 5: Since this description goes into the changelog, we would prefer it to have some extra details (eg. what are you fixing?) :)

HeartSquared (commented):
It would be helpful to have some extra context around the problem within the PR description too 👍🏻

packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx
```diff
@@ -44,18 +44,23 @@ export const ToastNotificationProvider = ({
...
+          const notificationExists = existing.find(
```
HeartSquared, line 53: Given the problem statement where you mention that checking the existing notifications isn't working, I'm surprised that the solution isn't something within this `existing.find` check but instead is the whole `addToastNotification`.

I'm not against the fix itself if it works to be able to unblock you, but it sounds like it might be in the wrong place.

packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx
aaronknol: Thanks @HeartSquared ! I'm not sure where I can add a test for this- should it be an interactive story? I can't see a spec file for toast's anywhere either

aaronknol:
> It would be helpful to have some extra context around the problem within the PR description too 👍🏻

Yep sure, I've added some more to the description, hope that helps @HeartSquared

packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx
aaronknol: We found that we need to use the previous state syntax of setting state, because at the time that the check was being made, the notifications array hadn't been updated yet. By using `setNotifications( existing () => ...)` we are waiting for the previous state to have been set

packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx
HeartSquared: Storybook is fine 😄 It's a bit of an awkward one to try and test in a spec file (given this was mostly legacy).

As a side note: Are you able to provide some steps to replicate the problem inside our repo (so that we can be sure we've actually fixed it)?

aaronknol: Thanks @HeartSquared I've added an interactive story for this

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
```diff
@@ -76,6 +77,58 @@ export const CreateNotification: Story = {
...
+              message: "There should only be one notification",
+            })
+          }}
+        />
+        <Button
+          label="Clear notifications"
+          onClick={() => clearToastNotifications()}
+        />
+      </>
+    )
```
HeartSquared, line 115: I would probably suggest you replicate your implementation with the `useEffect` instead if possible 🤞🏻 Only because this doesn't seem like someone would actually write this code, so it's easy to consider it an irrelevant test and delete it in the future

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
```diff
@@ -76,6 +77,58 @@ export const CreateNotification: Story = {
...
+  play: async context => {
+    const { canvasElement } = context
+    const { getByRole, findAllByText } = within(canvasElement.parentElement!)
+
+    const createNotificationsButton = getByRole("button", {
+      name: "Create notifications",
+    })
+    userEvent.click(createNotificationsButton)
+
+    const div = await findAllByText("There should only be one notification")
```
HeartSquared, line 126:
```suggestion
    const notifications = await findAllByText("There should only be one notification")
```

HeartSquared (approved):

PR #4944, opened 2024-08-22
fix storybook dir in buildkite

gyfchong:
Probably not that important, and could be removed instead since we don't use Buildkite to deploy storybook anymore. But that's a decision for another day


.github/workflows/test-stories.yaml
```diff
@@ -117,6 +117,7 @@ jobs:
           token: ${{ github.token }}
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           storybookBuildDir: "./docs/storybook-static"
+          storybookBaseDir: "./docs"
```
gyfchong, line 120: So I think the storybook config _needs_ to be in a `.storybook/` directory, and I kinda like it haha

HeartSquared: I don't think it _needs_ to be :P But it is nicer anyway so I approve the change 👍🏻

HeartSquared (approved):

PR #4968, opened 2024-08-26
fix(Future Select): Remove role=dialog from listbox container

dougmacknz:
Future Select uses the MultiSelect's Popover component, which has role=dialog built in.

Select shouldn't have this, so I've moved the role=dialog stuff to the MultiSelect implementation of Popover only.

Had to adjust tests a bit because a few of them were written to expect the dialog element and in a specific place. Deleted one because it wouldn't be possible to test without a test id and wasn't super valuable.

### DIFF
--- a/packages/components/src/__future__/Select/Select.tsx
+++ b/packages/components/src/__future__/Select/Select.tsx
@@ -195,7 +195,6 @@ export const Select = <Option extends SelectOption = SelectOption>({
             id={popoverId}
             portalContainer={portalContainer}
             refs={refs}
-            data-testid="kz-select-popover"
             focusOnProps={{
               onEscapeKey: state.close,
               onClickOutside: state.close,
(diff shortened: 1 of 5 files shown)


packages/components/src/__future__/Select/Select.tsx
```diff
@@ -195,6 +195,7 @@ export const Select = <Option extends SelectOption = SelectOption>({
             id={popoverId}
             portalContainer={portalContainer}
             refs={refs}
+            data-testid="kz-select-popover"
```
HeartSquared, line 198: Any chance we can avoid baking in a test id? 😅 
Also, I thought you mentioned it should have `role="listbox"` or am I imagining things? 😬

dougmacknz: Not unless we add a prop to Select which isn't better.

Listbox is included but it's many layers away.

We could also just delete the tests because they're not really testing anything with any user impact.

dougmacknz: I've removed it now by removing the test that checks the default behaviour, and adjusting the last one to just check that it still renders somewhere when passed a bad id.

HeartSquared (approved):

PR #4971, opened 2024-08-26
KZN-2614-jest-to-vitest

gyfchong:
## Why
Jest bad, vitest good

## What
Everything changed

## Notes
- Tests that generally rely on classname matching are also left to `jest` as `vitest`'s class name resolution is the final hashed CSS modules version, which makes it hard to target. So these will need to be rewritten to avoid


packages/components/__tests__/reactIntlMock.ts
```diff
@@ -0,0 +1,37 @@
...
+  return str.replace(regex, (match, capturedValue) => {
+    if (capturedValue) {
+      return value[capturedValue]
+    }
+    // Handle other captured values here if needed
+    return match
+  })
+}
+
+export const reactIntlMock = (): any => {
```
gyfchong, line 15: @HeartSquared I resolved the react intl problem with a global mock, but did you try this before doing the custom render? Was there a reason why this didn't work for Jest?

gyfchong:
Moved to https://github.com/cultureamp/kaizen-design-system/pull/4982

PR #4974, opened 2024-08-27
Fix-storybook-issues

gyfchong:
- build commands are clearer
- docs mode fixed
- use the build commands in the correct context


pnpm-lock.yaml
```diff
@@ -1932,6 +1932,11 @@ packages:
       '@cultureamp/next-head-hook': '>=1 || ~0.0.0'
       next: '>=13.5.0'
       react: '>=16.14.0'
+    peerDependenciesMeta:
```
HeartSquared, line 1935: Strange that you have lock file changes 😅

HeartSquared (approved):

PR #4977, opened 2024-08-27
feat(Future Workflow): Change footer to white bg

dougmacknz:
## Why
Design update for better contrast

## What
- Change footer background to white
- Remove reversed from all buttons in story examples
- Adjust font and icon colours for the new background
- Change all 'next' buttons to primary buttons (previously only the final step's button was primary)


dougmacknz:
Creating a v3 version instead

PR #4978, opened 2024-08-27
feat(Workflow): Create v3 Workflow

dougmacknz:
## Why
Design update for better colour contrast on the new v3 version

## What
- Move Workflow into the new grouped component categories, creating a `layout` category
- Copy the v2 component into a new v3 version
- Make the following changes: 
  - Change footer background to white
  - Remove reversed from all buttons in story examples
  - Adjust font and icon colours for the new background
  - Change all 'next' buttons to primary buttons (previously only the final step's button was primary)
  - Change the page title to use `composable-header-title`
  - Convert all sass to css
  - Change the status prop to take a ReactNode, removing the tight coupling to v1 Tag

See https://github.com/cultureamp/kaizen-design-system/pull/4977/files for an easy way to see what has changed between v2 and v3 (though a little outdated now)


mcwinter07:
Visually this is all looking hunky dory. @HeartSquared and I were talking about the folder titles recently since the folks names don't really indicate much in chromatic, i.e: if there's a bunch of changes across multiple components you'd see v1, v2, v3 listed constantly.
Screenshot 2024-08-27 at 3 59 11 pm

Maybe not the best example since there's only a few stories there, but we were thinking the meta should look a little more like:

```
const meta = {
  title: "Layout/Workflow/Workflow (v1)/...otherFolders",
```

mcwinter07 (commented):
Two things we'll also need to add since this is a new logical group:

1. We need to [create the package.json](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4132896968/Package+versioning+in+kaizen+components#Create-a-new-version-directory-and-add-versioned-entry-points) that points to the dist for layers.

Screenshot 2024-08-27 at 4 15 31 pm

2. Update the [pnpm workspace yml](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4132896968/Package+versioning+in+kaizen+components#Update-pnpm-workspace) to include layout

There's a bunch to do in these so I started writing up a [checklist like doc](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4132896968/Package+versioning+in+kaizen+components#Create-a-new-version-directory-and-add-versioned-entry-points) for this kind of migration to the versioned entry points. I think you've clocked all the other changes thought aside from the ones I've called out :)

dougmacknz:
> Update the [pnpm workspace yml](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4132896968/Package+versioning+in+kaizen+components#Update-pnpm-workspace) to include layout

This one seems like it doesn't need anything because there's nothing in there for the other groups?

mcwinter07:
> > Update the [pnpm workspace yml](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4132896968/Package+versioning+in+kaizen+components#Update-pnpm-workspace) to include layout
> 
> This one seems like it doesn't need anything because there's nothing in there for the other groups?

Oh yeah, sorry that was only if we were creating a new version (ie: 4) jumped the gun there 😆

packages/components/src/__layout__/Workflow/v3/subcomponents/Header/components/Titles/Titles.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { HTMLAttributes, ReactNode } from "react"
+import { Heading } from "~components/Heading"
+import { VisuallyHidden } from "~components/VisuallyHidden"
+import { OverrideClassName } from "~components/types/OverrideClassName"
+import styles from "./Titles.module.scss"
+
+export type WorkflowTitlesProps = {
+  workflowName: string
+  stepName: string
+  status?: ReactNode
```
mcwinter07, line 10: What are you're thoughts on either update this to `statusTag` / `headingTag`? My main thinking here is that now this can take a React Node, we may need to provide more clarity in terms of what we are expecting. Could even be just a type description if we want to leave the props the same.

dougmacknz: yeah good call, i like `statusTag` the most because even though yes it's open to anything, we're still expecting a status tag there

mcwinter07: Brill, sound good to me

mcwinter07:
Looks good, will give a tick once the mdx for v1 is updated to point to v3  👍

mcwinter07 (approved):

PR #4982, opened 2024-08-29
add vitest

gyfchong:
## Why
Jest bad, vitest good

## What
Everything changed


mcwinter07 (commented):
Looks hunky dory, though we have these two workflow steps that seems like they're still `require`?
Screenshot 2024-09-02 at 11 52 00 am
I assume this is because ci.yml is now lint.yml and the job for test-stories has been broken into shards.

packages/components/package.json
```diff
@@ -40,8 +40,7 @@
     "build:global-styles": "postcss ./styles/global.css --output dist/global.css",
     "build:combine-styles": "concat-cli -f ./dist/global.css ./dist/styles.css -o ./dist/styles.css && rm ./dist/global.css",
     "build:styles": "pnpm build:global-styles && pnpm build:combine-styles",
-    "test": "FORCE_COLOR=1 jest",
```
HeartSquared, line 43: FYI this `FORCE_COLOR=1` is for us to see colour in the terminal when we use turbo to run these commands (colour gets stripped otherwise). Maybe just check if it's still relevant or not :)

packages/components/package.json
```diff
@@ -107,13 +106,17 @@
...
+    "@testing-library/jest-dom": "^6.5.0",
+    "@types/jest-axe": "^3.5.9",
     "@types/lodash.debounce": "^4.0.9",
     "@types/react-highlight": "^0.12.8",
     "@types/react-textfit": "^1.1.4",
     "@types/uuid": "^10.0.0",
     "autoprefixer": "^10.4.20",
     "concat-cli": "^4.0.0",
     "identity-obj-proxy": "^3.0.0",
+    "jest-axe": "^9.0.0",
```
HeartSquared, line 119: Do we still need this or were you able to fix all the tests?

packages/components/src/DatePicker/subcomponents/DateInputField/DateInputField.spec.tsx
```diff
@@ -1,17 +1,17 @@
 import React, { useRef } from "react"
-import { screen, waitFor } from "@testing-library/react"
+import { screen, waitFor, render } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { enUS } from "date-fns/locale"
-import { renderWithIntl } from "~tests"
```
HeartSquared, line 5: Nice 🎉

packages/components/src/ErrorPage/ErrorPage.spec.tsx
HeartSquared, line 1: Please put these back :') This was the replacement for the stickersheet that kept timing out

packages/components/src/ErrorPage/_docs/ErrorPage.stickersheet.stories.tsx
```diff
@@ -38,6 +38,30 @@ const StickerSheetTemplate: StickerSheetStory = {
           <StickerSheet.Row>
             <ErrorPage code="400" />
           </StickerSheet.Row>
+          <StickerSheet.Row>
```
HeartSquared, line 41: Please revert: These kept timing out in Chromatic every few runs, so we moved it to a unit test (given the code only changes contents).

packages/components/src/Filter/FilterDateRangePicker/_docs/FilterDateRangePicker.stickersheet.stories.tsx
```diff
@@ -168,7 +195,7 @@ export const StickerSheetDefault: StickerSheetStory = {
 export const StickerSheetRTL: StickerSheetStory = {
   ...StickerSheetTemplate,
   name: "Sticker Sheet (RTL)",
-  parameters: { textDirection: "rtl" },
+  parameters: { textDirection: "rtl", delay: 800 },
```
HeartSquared, line 198: Is this delay in general, or just for Chromatic (you'll need to change it - assertion would probably be the best)?
https://www.chromatic.com/docs/delay/

packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+      expect(element).toBeInTheDocument()
+      expect(hiddenState).toHaveTextContent("Shown")
+    })
+
+    await userEvent.click(canvas.getByTestId("close-button"))
+
+    await waitFor(() => {
+      setTimeout(() => {
+        expect(hiddenState).toHaveTextContent("Hidden")
+        expect(element).not.toBeInTheDocument()
```
HeartSquared, line 56: FYI: I'm not sure if it exists in SB testing, but `toBeVisible` may or may not help with not needing the timeout.

packages/components/src/RichTextEditor/utils/commands/fixtures/helpers.ts
```diff
@@ -5,14 +5,15 @@ import {
   Transaction,
 } from "prosemirror-state"
 import { findChildrenByType } from "prosemirror-utils"
+//  eslint-disable-next-line import/no-extraneous-dependencies
```
HeartSquared, line 8: Does this comment need to exist? Doesn't seem to apply to other files

packages/components/svgoUtils.ts
```diff
@@ -1,7 +1,10 @@
 // Certain attributes work in html, but not React.
 // If an attribute needs to be converted to its React counterpart,
+
+import { RootItem, SVGOItem } from "./svgo.spec"
```
HeartSquared, line 4: Please revert this :') As we sadly found that SVGO only likes JS

packages/components/svgo.spec.ts
```diff
@@ -35,7 +40,7 @@ describe("removeRootSVGElement", () => {
   }
 
   it("Replaces the <svg> element with its children", () => {
-    const rootItem = {
+    const rootItem: RootItem = {
```
HeartSquared, line 43: Maybe revert this file since I'll deal with it when I move it to v1

tsconfig.json
```diff
@@ -19,8 +19,7 @@
       "~tailwind/*": ["./packages/tailwind/src/*"],
       "~tests": ["./packages/components/__tests__/index"],
       "~tests/*": ["./packages/components/__tests__/*"]
-    },
-    "types": ["@testing-library/jest-dom"]
```
HeartSquared, line 23: Looks like the library is still around. We definitely don't need this here?

vite.config.ts
```diff
@@ -0,0 +1,45 @@
+/** @type {import('vite').UserConfig} */
+import path from "path"
+import { mergeAlias } from "vite"
+
+export default {
```
HeartSquared, line 5: If we have this file, does that mean we can remove the config that's inside Storybook (which should be the same)?

package.json
```diff
@@ -12,13 +12,10 @@
...
     "test": "pnpm turbo test",
-    "test:ci": "pnpm turbo test:ci",
-    "jest:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --no-cache",
-    "eslint-config": "pnpm eslint -c .eslintrc.js --max-warnings=0",
     "lint": "pnpm lint:js && pnpm lint:md && pnpm lint:format && pnpm lint:styles",
     "lint:ts": "pnpm turbo lint:ts",
-    "lint:js": "pnpm eslint-config '**/*.{ts,tsx,mjs,js}'",
-    "lint:md": "pnpm eslint-config '**/*.{md,mdx}'",
+    "lint:js": "pnpm eslint '**/*.{ts,tsx,mjs,js}' --max-warnings=0",
+    "lint:md": "pnpm eslint '**/*.{md,mdx}' --max-warnings=0",
```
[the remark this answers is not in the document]
gyfchong: Yep, the naming of `.eslintrc.js` means the command defaults to it.

packages/components/package.json
gyfchong: We still need it for Storybook's test-runner... But admittedly shouldn't be in this `package.json`

packages/components/src/Filter/FilterDateRangePicker/_docs/FilterDateRangePicker.stickersheet.stories.tsx
gyfchong: Mistake, it was to try getting the snapshot to maybe adjust itself after some delay. But didn't even work. Removed

packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.stories.tsx
gyfchong: I think cause the DOM node gets removed

packages/components/src/RichTextEditor/utils/commands/fixtures/helpers.ts
gyfchong: I think it's cause it's not a spec file that `eslint` is complaining? Not sure

packages/components/package.json
gyfchong: Doesn't make a difference any more :'D'

packages/components/svgo.spec.ts
gyfchong: If I revert this, Vitest can't parse it hmmm

```  FAIL  svgo.spec.ts [ svgo.spec.ts ]
@kaizen/components:test: Error: require() of ES Module /Users/geoff.chong/Code/kaizen-design-system/packages/components/svgoUtils.js from /Users/geoff.chong/Code/kaizen-design-system/packages/components/svgo.spec.ts not supported.```

File: tsconfig.json (hunk 10 above)
Comment (gyfchong) [reply tsconfig.json:23]: It's only relevant in the tsconfig for writing tests, but since none of the packages extend off base any more, having it here doesn't make a diff. The package still exists in root for the `vitest.setup` config

File: packages/components/svgo.spec.ts (hunk 9 above)
Comment (gyfchong) [reply packages/components/svgo.spec.ts]: Might need to skip this one and just wait for your fix

File: .github/workflows/test-unit.yml [hunk 13]
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  unit-tests:
+    if: github.head_ref != 'changeset-release/main'
+    name: "vitest"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - name: Unit tests
```
HeartSquared, line 29: Optional: You can put the shard number in the step names :) We do this for the stories

gyfchong: Might as well, since I need to start my branch again

vite.config.ts
```diff
@@ -0,0 +1,46 @@
...
+        // redirected to import from the replacement path (Same as KAIO rollup config).
+        "__@cultureamp/i18n-react-intl/locales": path.resolve(
+          __dirname,
+          "./packages/components/locales"
+        ),
+      }
+    ),
+  },
+  test: {
+    // TODO: Remove this when @HeartSquared fixes icons
```
HeartSquared, line 40: Hahaha whaaaat 😂 I don't think we'd be removing the script until we actually delete the old icon, so it would still be relevant here

mcwinter07: lul, doxed in code comments 😆

HeartSquared (approved):

PR #4986, opened 2024-08-30
fix(Workflow v3): design adjustments

dougmacknz:
## Why
Forgot to do a design review

## What
- Remove box shadow from header
- Add border top to footer and border bottom to header
- Tidy up the story example a bit


mcwinter07 (approved):
Noice 👍

PR #5018, opened 2024-09-04
docs(Collapsible): fix docs

HeartSquared:
## Why

image

## What

Remove deleted story reference from Collapsible docs.


gyfchong (approved):
Thanks, my bad 😅

PR #5022, opened 2024-09-05
Move GuidanceBlock to root

gyfchong:
Since we're looking to deprecate this component and push for alternative solutions, I'm going to silently delete the new v2 `GuidanceBlock`

```
containers/GuidanceBlock/v1 -> GuidanceBlock
containers/GuidanceBlock/v2 -> Delete
```


HeartSquared:
I think you'll need a codemod? Or at least modify the one you previously made

dougmacknz (approved):

packages/components/codemods/updateGuidanceBlockVariantProp/index.ts
HeartSquared, line 1: You'll want to remove it from the README too :)

HeartSquared (approved):

PR #5024, opened 2024-09-05
chore(storybook): update nav sorting

HeartSquared:
## Why

We want to surface docs above stories.

## What

- Usage Guidelines will be at the top, followed by API Specification
- Nested docs (eg. future versions) will be surfaced above the docs of the parent*
	- *TBC if we want this or if we prefer it below, but it's easy to revert
- Replicates existing structure:
	- For Tailwind docs
	- Kaizen Provider is at the top of components


docs/.storybook/preview.tsx
```diff
@@ -110,32 +107,123 @@ const preview = {
...
+        const groupB = titleB.split("/")[0]
+
+        const groups = [
+          "Introduction",
+          "Guides",
+          "Actions",
+          "Containers",
+          "Illustrations",
+          "Layout",
+          "Overlays",
```
gyfchong, line 182: May as well put all the groups planned here

gyfchong (approved):

docs/.storybook/preview.tsx
HeartSquared: I didn't want to put them in until we actually use them 😬 It would be easy to add anyway when the time comes

PR #5043, opened 2024-09-09
refactor(Tooltip): move Tooltip to future structure

HeartSquared:
## Why

Remove family groups from code - Step 1: Simplify internal structure.
Removal of entrypoint will be done separately.

## What

```
overlays/Tooltip/v3 -> future/Tooltip
overlays/Tooltip/v1 -> Tooltip
// Put as a subcomponent for now as there is no foreseeable usage outside of Tooltip. We can extract it later if we need to.
overlays/Focusable/v3 -> future/Tooltip/subcomponents/Focusable
```


dougmacknz (approved):

dougmacknz:
too many conflicts now :o

PR #5044, opened 2024-09-09
Remove Workflow v1 and v2, make v3 current

dougmacknz:
## Why

v1 of Workflow isn't used anywhere

Only 4 repos using v2 Workflow, which shouldn't be too hard to PR updates for ourselves so that we can simplify this down to just one version.

## What

Delete Workflow v1 and v2, make v3 the current version.


HeartSquared:
The docs structure still says `Workflow/Workflow (v3)`, so you'll want to update that and the import statements :)

mcwinter07 (approved):
Looks good, could you give Baps a holler in [this thread](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1725844968506789) once its released :)

PR #5046, opened 2024-09-09
ProgressStepper spacing margin fix 

mcwinter07:
## Why
I've remove the Text component from the Progress stepper as a fix to the CSS layer issue that Baptise was experiencing with the Workflow v3. I think this is a relatively safe move since the stepper is more a visual artifact than text and just needed the a few tokens.

## What
- replace Text usage with tokens

### DIFF
--- a/.changeset/popular-cherries-sparkle.md
+++ b/.changeset/popular-cherries-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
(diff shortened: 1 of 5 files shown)


mcwinter07:
I wonder if it is still important that we do the `@layer` update here for `Text` or if we should look more holistically at this issue?

```
@layer reset {
  .heading {
    margin: 0;
  }
}
```

dougmacknz (approved):

PR #5052, opened 2024-09-12
Winter/small fixes to progress stepper

mcwinter07:
## Why
- adding a couple of small fixes following us dropping the versioned entry for this

## What
- removes ProgressStepper v2 and layout folder
- updates storybook meta titles to remove nesting


.changeset/eighty-dolphins-admire.md
```diff
@@ -0,0 +1,5 @@
+---
```
HeartSquared, line 1: I think the changeset can be empty since it was already omitted from the dist :)

mcwinter07: Good point 🤔

HeartSquared (approved):

PR #5068, opened 2024-09-20
Clean up obsolete docs

sentience:
## Why
We're publishing a job ad with a link to the Kaizen site, so we want to tidy up any broken bits to put our best foot forward.

## What
- Fix broken link and obsolete content on Welcome page


.changeset/lucky-laws-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@docs/storybook": patch
```
HeartSquared, line 2: No need for a changeset since we don't actually release anything :) (change to an empty one).
Docs get published as soon as the PR is merged.

It would be good if we could control when we release the docs instead of it possibly being ahead of our releases, but we haven't really found a way as of yet - definitely known though.

HeartSquared (approved):

PR #5069, opened 2024-09-20
Fix removeToastNotification to use previousState

aaronknol:

## Why
When we had multiple toast's on a page, all with different id's, calling `removeToastNotification` would remove all toast's _after_ the one that was being deleted.
The toast's all have a button inside them that when clicked will call removeToastNotification.
This is a similar problem to what we were having with `addToastNotification` which we fixed here

### New ###
By using the previousState syntax of setState, the `notifications` in the `removeToastNotification` is always updated and correct.
```
setNotifications(prev => prev.filter(({ id }) => id !== notificationID))
```

### Old ###
It appears that the copy of the state that was being created wasn't correct, it hadn't been updated yet before being sliced.
```
const copy = notifications.slice()
    copy.splice(notificationIndex, 1) // Mutation
    setNotifications(copy)
```


mcwinter07 (approved):
Nice, it even looks cleaner than the original method of splicing it out 👍 Cheers for the contribution @aaronknol

PR #5076, opened 2024-09-24
InlineNotification: allow ref to be passed in and add focus styling

dougmacknz:


packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.tsx
```diff
@@ -49,89 +56,102 @@ export type GenericNotificationVariant = {
...
+      classNameOverride,
+      ...restProps
+    },
+    ref
+  ): JSX.Element | null => {
+    const [isHidden, setIsHidden] = useState<boolean>(true)
+    const [isRemoved, setIsRemoved] = useState<boolean>(false)
 
-  const containerRef = useRef<HTMLDivElement>(null)
+    const containerRef = isRefObject(ref) ? ref : useRef<HTMLDivElement>(null)
```
HeartSquared, line 83: Something like this since hooks shouldn't be called conditionally.

```suggestion
    const fallbackRef = useRef<HTMLDivElement>(null)
    const containerRef = isRefObject(ref) ? ref : fallbackRef
```

HeartSquared (commented):
I think we have some existing tests for fallback ref behaviour. Maybe add that in (I wonder if we could abstract the pattern into a custom hook?).

HeartSquared (approved):

PR #5077, opened 2024-09-24
Filter select accessible label and focus jump fix

mcwinter07:
## Why
Addresses the two issues raised by Kayd regarding the FilterSelect's Listbox labelledby and the focus jump issue that has also been reported in the future Select.

## Context on the focus jump issue
This was raised a while back by Harry whereby opening a select with the keyboard would force the page to refocus to the top of the screen. TLDR  of the solution:  we pushed teams to use the portal container to ensure the page jump would not occur.

With this update we solve this issue for both FilterSelect and future Select. Note that I would still consider this a bandaid on top of our implementation that has too many components that contribute to focus management. This makes it hard to understand which component contributes to which concern. In the long run I would advise we refactor these, but this fix should make for an improve DX without having to change the API.

## What
- Update the props spreading to pass correct labelledby id to the FilterSelect's listbox
- Updates the Listbox and Overlay implementation to manually focus to the list elements


mcwinter07:
So I looked at the incomplete issues in the a11y tag and as far as I can tell they're false positives

packages/components/src/Filter/FilterSelect/FilterSelect.tsx
```diff
@@ -90,19 +96,23 @@ export const FilterSelect = <Option extends SelectOption = SelectOption>({
...
         classNameOverride={classNameOverride}
       >
-        <FilterContents classNameOverride={styles.filterContents}>
-          <SelectProvider<Option> state={state}>
-            <SelectPopoverContents menuProps={menuProps}>
-              {children}
-            </SelectPopoverContents>
-          </SelectProvider>
-        </FilterContents>
+        <>
```
mcwinter07, line 105: remove this fragment

packages/components/src/Filter/FilterSelect/FilterSelect.tsx
```diff
@@ -79,6 +79,12 @@ export const FilterSelect = <Option extends SelectOption = SelectOption>({
 
   const { buttonProps } = useButton(triggerProps, triggerRef)
 
+  // The id is being remapped be the button props id point to nowhere. This should ideally be refactored but this should solve the a11y warnings we have been receiving
```
mcwinter07, line 82: better this english good

packages/components/src/Filter/FilterSelect/FilterSelect.spec.tsx
```diff
@@ -107,7 +107,7 @@ describe("<FilterSelect>", () => {
         render(<FilterSelectWrapper isOpen />)
         expect(screen.queryByRole("listbox")).toBeVisible()
         await waitFor(() => {
-          expect(screen.queryByRole("listbox")).toHaveFocus()
+          expect(screen.queryAllByRole("option")[0]).toHaveFocus()
```
mcwinter07, line 110: change to getAllByRole

packages/components/src/__future__/Select/_docs/Select.mdx
```diff
@@ -98,8 +98,6 @@ Set `isFullWidth` to `true` to have the Select span the full width of its contai
 
 By default, the Select's popover will attach itself to the `body` of the document using React's `createPortal`.
 
-You can change the default behaviour by providing a `portalContainerId` to attach this to different element in the DOM. This can help to resolve issues that may arise with `z-index` or having a Select in a modal.
+You can change the default behavior by providing a `portalContainerId` to attach this to different element in the DOM. This can help to resolve issues that may arise with `z-index` or having a Select in a modal.
 
 <Canvas of={SelectStories.PortalContainer} />
```
mcwinter07, line 103: Now this isn't primarily needed to showcase the focus jumping issue we should just update this to be in a modal to reflect the original solve

packages/components/src/__future__/Select/subcomponents/ListBox/ListBox.tsx
```diff
@@ -22,13 +38,37 @@ export const ListBox = <Option extends SelectOption>({
...
+    {
+      ...menuProps,
+      disallowEmptySelection: true,
+      autoFocus: false,
+    },
     state,
     ref
   )
 
+  useEffect(() => {
```
mcwinter07, line 53: split into two separate useEffects

dougmacknz (approved):
Deliciously hacky

🚢 🚢 🚢

PR #5084, opened 2024-09-26
Update typography heading tokens

mcwinter07:
## Why
This is part of the heading uplift work that was looked into a few sprints back. This is the partial update heading tokens where we will be addressing only font-weight.

Screenshot 2024-09-26 at 1 54 48 PM

## What
- updates heading typography tokens to use updated weights
  -     display: 800
  -     h1: 500
  -     h2: 600
  -     h3: 600
  -     h4: 600
  -     h5: 600
  -     h6: 600


mcwinter07:
So I've gone through and checked all the visual changes in Chromatic. The ones I've left unchecked are the visual changes within component. They all look fine to me but having Jess do a once over of those in case there are any changes we should decouple from headings (i.e: embed styles into the component and not use the heading tokens) or design may need to update on their end.

If that's all hunky dory this PR should be ready to go and be merged in next week on Monday :)

HeartSquared (approved):

PR #5114, opened 2024-10-02
chore(deps): Upgrade rac to ^1.4.0

NatalieLoyzaga:

## Why

New release has `isPending` as a prop for the Button https://react-spectrum.adobe.com/releases/2024-09-30.html

### DIFF
--- a/.changeset/tall-steaks-sin.md
+++ b/.changeset/tall-steaks-sin.md
@@ -1,0 +1,5 @@
+---
+"@kaizen/components": patch
+---
+
+Upgrade react aria component to version ^1.4.0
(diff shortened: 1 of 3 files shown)


NatalieLoyzaga:
Is an empty changeset the right approach? I noticed on other PRs where we were upgrading deps an empty changeset was commited

HeartSquared:
> Is an empty changeset the right approach? I noticed on other PRs where we were upgrading deps an empty changeset was commited

Nope 😄 Empty changesets are only for internal dep updates like Storybook because we don't need to make a release. But for this dep, since we'd want to release it, we'd need a proper changeset :)

PR #5120, opened 2024-10-02
chore(ci): split Storybook tests from Chromatic

HeartSquared:
## Why

We want to run non-Chromatic Storybook tests whilst still in draft.

## What

Split CI workflows for Storybook tests from Chromatic tests


mcwinter07 (approved):
Nice :) will be good to have the tests trigger on our drafts now 👍

PR #5122, opened 2024-10-03
Filter select aria label fix

mcwinter07:
## Why
This PR addresses the a11y issues that Kayd raised with the Filter Select - this was split from the PR that had to be reverted.

## What
- remaps the id to draw the accessible label from the trigger

### DIFF
--- a/packages/components/src/Filter/FilterSelect/FilterSelect.tsx
+++ b/packages/components/src/Filter/FilterSelect/FilterSelect.tsx
@@ -80,7 +80,7 @@ export const FilterSelect = <Option extends SelectOption = SelectOption>({
   const { buttonProps } = useButton(triggerProps, triggerRef)
   const renderTriggerButtonProps = {
     ...buttonProps,
-    "aria-labelledBy": undefined,
+    "aria-labelledby": undefined,
     "aria-controls": menuProps.id,
   }
   return (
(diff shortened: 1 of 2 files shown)


packages/components/src/Filter/FilterSelect/FilterSelect.tsx
```diff
@@ -90,19 +94,24 @@ export const FilterSelect = <Option extends SelectOption = SelectOption>({
...
       >
-        <FilterContents classNameOverride={styles.filterContents}>
-          <SelectProvider<Option> state={state}>
-            <SelectPopoverContents menuProps={menuProps}>
-              {children}
-            </SelectPopoverContents>
-          </SelectProvider>
-        </FilterContents>
+        <>
+          {/* <VisuallyHidden id={menuProps.id}>{label}</VisuallyHidden> */}
```
HeartSquared, line 104: Should this be here?

mcwinter07: hahahha nope 😆 consider it 🪓 'd

mcwinter07: Done and done 🩸🪓

HeartSquared (dismissed):

HeartSquared (approved):

PR #5148, opened 2024-10-07
Add: `xs` size to LoadingSpinner

mcwinter07:
## Why
This is part of the prework for the new Button to enable content dense layouts.

## What
- Adds `xs` size to `LoadingSpinner`
- Migrate from SCSS to CSS variables and stylesheet
- Remove width declaration from svg and set it on the container
  - This can now be controlled via the `--loading-spinner-size` variable in the `classNameOverride` if required
- Update docs examples


packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.mdx
```diff
@@ -29,14 +29,17 @@ When inside a button, it is intended to have the same color as the label text.
...
 Override styles such as the colour of the spinner using `classNameOverride`.
 
 <Canvas of={LoadingSpinnerStories.ClassNameOverride} />
 
-### Size
-
-Generally use `"md"` (default value) unless spinner is inside a form field.
-
-<Canvas of={LoadingSpinnerStories.Size} />
+You can also use this to target the `--loading-spinner-size` CSS variable if required. Note that this will not affect the stroke width of the svg so use with caution.
```
mcwinter07, line 45: I'm a little dubious in terms of how much we want to promote this so curious for thoughts @dougmacknz @caseysinclair. Happy to leave it out of here and just have it as a thing folks can do if they need an escape hatch :)

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -2,16 +2,98 @@ import React, { HTMLAttributes } from "react"
 import classnames from "classnames"
 import { VisuallyHidden } from "~components/VisuallyHidden"
 import { OverrideClassName } from "~components/types/OverrideClassName"
-import styles from "./LoadingSpinner.module.scss"
+import styles from "./LoadingSpinner.module.css"
+
+type LoadingSpinnerSize = "xs" | "sm" | "md"
```
mcwinter07, line 7: Anyone think this is worth exporting?

dougmacknz: I can't see anyone wanting this so yeah I'd leave it private unless/until someone asks for it

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.mdx
dougmacknz: Yeah I'd leave this out

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
HeartSquared: Wouldn't even make a type for it tbh 😅

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -2,16 +2,98 @@ import React, { HTMLAttributes } from "react"
...
   accessibilityLabel: string
   /**
-   * Generally use "md" unless spinner is inside a form field
+   * Generally use "md" unless spinner is inside a form field.  @default "md"
    */
-  size?: "sm" | "md"
+  size?: LoadingSpinnerSize
 } & OverrideClassName<Omit<HTMLAttributes<HTMLDivElement>, "children">>
 
+const renderLoadingSpinner = (size: LoadingSpinnerSize): JSX.Element => {
```
HeartSquared, line 17: I'd convert these to individual components instead of a function that returns a JSX.Element

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -24,60 +106,16 @@ export const LoadingSpinner = ({
 }: LoadingSpinnerProps): JSX.Element => (
   <div
     data-automation-id="loading-spinner"
```
HeartSquared, line 108: You reckon we could remove this? 👀

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.module.css
```diff
@@ -0,0 +1,28 @@
...
+  height: var(--loading-spinner-size);
+}
+
+.loadingSpinner.xs {
+  --loading-spinner-size: var(--spacing-16);
+}
+
+.loadingSpinner.sm {
+  --loading-spinner-size: var(--spacing-24);
+}
```
HeartSquared, line 15: We should try to keep specificity low where possible, so I'd remove the `.loadingSpinner` part. CSS modules will prevent name clashing.

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.module.css
```diff
@@ -0,0 +1,28 @@
+.loadingSpinner {
+  --loading-spinner-size: var(--spacing-48);
```
HeartSquared, line 2: If we expect that size will always exist, we could probably safely move this into a class so it's clear that it's for `md`. 
That being said, I could see the argument for missing a default I guess 🤷🏻‍♀️  I'll let you decide :)

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.module.css
```diff
@@ -0,0 +1,28 @@
...
+
+  display: flex;
+  width: var(--loading-spinner-size);
+  height: var(--loading-spinner-size);
+}
+
+.loadingSpinner.xs {
+  --loading-spinner-size: var(--spacing-16);
```
HeartSquared, line 10: Given spinners are closer to images and icons, we should probably actually use `px` values for them as they don't need to scale with increased font size (have been considering that spacing probably makes sense to be too in most cases, but that's a different problem).

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.mdx
HeartSquared: Agreed, we should leave this out because we aren't encouraging the use of manipulating the size (unlike new Icon).

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stickersheet.stories.tsx
```diff
@@ -18,16 +18,28 @@ const StickerSheetTemplate: StickerSheetStory = {
...
           <LoadingSpinner accessibilityLabel="Loading" size="md" />
           <LoadingSpinner
             accessibilityLabel="Loading"
             classNameOverride="text-purple-400"
           />
+          <LoadingSpinner
+            accessibilityLabel="Loading"
+            size="xs"
+            classNameOverride="[--loading-spinner-size:10px]"
+          />
```
HeartSquared, line 42: You can leave this one out :)

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stories.tsx
```diff
@@ -25,12 +25,27 @@ export const Playground: Story = {
 }
 
 export const ClassNameOverride: Story = {
-  args: { classNameOverride: "text-green-400" },
+  render: args => (
+    <>
+      <LoadingSpinner {...args} classNameOverride="text-green-400" />
```
HeartSquared, line 30: Curious about the need to increase these examples?

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.mdx
mcwinter07: Cool 👍

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stories.tsx
mcwinter07: I was aligning this to the examples Jess has in Figma, figured while we probably don't want to introduce a color prop we can at least show the one that match

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.module.css
mcwinter07: Makes sense to me, will make the change 👍

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.module.css
mcwinter07: My thinking is to declare the CSS var here as the default since I remember coming into some issue leaving one undefined on the base selector. Seemed a little hat on a hat to re-declare it in a .md class but could see that making sense for clarity

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: Sure could break them into small subcomponents for cleanliness. In terms of rendering them out, I suppose I could do a ternary (though two conditions are a little ick to me), otherwise just a function that have a switch? i.e:

```
    {() => {
      switch (size) {
        case "sm":
          return <SmallSpinner />
        case "xs":
          return <ExtraSmallSpinner />
        case "md":
          return <MediumSpinner />
        default:
          return null
      }
    }}
    {size === "sm" ? (
      <SmallSpinner />
    ) : size === "xs" ? (
      <ExtraSmallSpinner />
    ) : (
      <MediumSpinner />
    )}
```

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: Likely, I did a quick scan and the [closest usages](https://github.com/search?q=org%3Acultureamp+%22loading-spinner+NOT+repo%3Acultureamp%2Fkaizen-design-system+NOT+repo%3Acultureamp%2Fkaizen-legacy&type=code) I found don't seem like they should actually be impacted

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: I suppose the ternary isn't too bad, a less code needed and we likely won't need any more than 3 sizes (likely)

HeartSquared: Or something like this :)

```suggestion
const Spinner = (size: LoadingSpinnerSize): JSX.Element => {
  if (size === "xs") return <ExtraSmallSpinner />
  if (size === "sm") return <SmallSpinner />
  return <MediumSpinner />
}
```

mcwinter07: Right, I getcha 😉

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stories.tsx
HeartSquared: Hmm...not really necessary imo to completely match the design examples unless our API is strictly limited to their selection (eg. variants). This story is showcasing the `classNameOverride` prop, and actually in this case, you're preventing the controls from working 😅

packages/components/src/Loading/LoadingSpinner/LoadingSpinner.tsx
mcwinter07: I've moved all the spinners into a `SpinnerIcon` subcomponent, it was starting to look a little bloated and messy in the top of the component

HeartSquared: As long as you don't export it at the entrypoint then it should be fine :) Because currently `SpinnerIcon` exists so it would conflict 😬

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stories.tsx
mcwinter07: Isn't this story more to showcase in our documentation, ie:

Screenshot 2024-10-10 at 11 35 15 AM

To our consumer the controls never did anything (at least on the docs page).

I suppose the question I'm wondering is, are we meant to have standard spinner colors, since they have been specifically designed?

mcwinter07: Maybe a separate story to delineate the difference between just classNameOverride and colors if they're needed? My main concern stems from have mild color variances in spinners

HeartSquared: I'd say that standard colours aren't needed. Using `classNameOverride` is the correct pattern.

The docs page doesn't show the controls for that story, but the individual story does :)

But yeah, I guess looking at that screenshot, I suppose showcasing a few more is nice 🤔 Perhaps change the order so that the one that can be changed by controls is the first one? Or move the spread args to the end so they can all be overridden by the controls?

That being said, color isn't the only use case for this (eg. margin, display), so I wonder if it would seem too specific and confusing only showing multiple colours 🤔 I'll leave the decision to you :)

mcwinter07: mmm I might check in with Jess to see how important those colors are to document, it could be a null issue in which case 🤷 but yeah to your point maybe leaving the classNameOverride story clean is good since it doesn't technically just to color

mcwinter07: Cool, so just chatted with Jess. They do have a set of colors that they use, but I said I wouldn't add scope to the ticket. I've create an additional story to be a little more explicit on the colors but kept it separate from the classNameOverride story

packages/components/src/Loading/LoadingSpinner/_docs/LoadingSpinner.stories.tsx
```diff
@@ -25,12 +25,40 @@ export const Playground: Story = {
...
+  render: args => (
+    <LoadingSpinner {...args} classNameOverride="text-green-400" />
+  ),
+  decorators: [
+    Story => (
+      <div className="flex gap-24">
+        <Story />
+      </div>
+    ),
+  ],
```
HeartSquared, line 37: Less lines if we revert 😉 
```suggestion
args: { classNameOverride: "text-green-400" },
```

mcwinter07: Now that's usin' the ol' noggin 👍  will do

HeartSquared (approved):
LGTM 🚀

mcwinter07:
Holding off on merging this till Monday so we can align with design publishing Figma

PR #5150, opened 2024-10-09
Create future Tabs

dougmacknz:
## Why
Move off the Reach UI which is no longer being actively maintained and use RAC instead.

## What

What changed from the current Tabs:
- Adjusted the props API to match RAC (omitting some stuff that we won't support, at least for now)
- Changed SCSS to CSS
- Added migration notes from current to future to MDX file

...and that's it, no design changes!


packages/components/src/__future__/Tabs/_docs/Tabs.mdx
```diff
@@ -0,0 +1,50 @@
...
+  exportNames={["Tabs", "TabList", "Tab", "TabPanel"]}
+  isFuture
+/>
+
+## Overview
+
+<Canvas of={TabsStories.Playground} />
+<Controls of={TabsStories.Playground} />
+
+## Migration from current KAIO Tabs
```
HeartSquared, line 26: Maybe make the migration path a separate file (like what we did with future Tag) :)

packages/components/src/__future__/Tabs/subcomponents/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./Tab"
+export * from "./TabList"
+export * from "./TabPanel"
+export * from "./Tab"
```
HeartSquared, line 4: Tab exported twice?

packages/components/src/__future__/Tabs/subcomponents/Tab.module.css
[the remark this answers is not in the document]
dougmacknz: No strong opinions here, your preference is to do it?

packages/components/src/__future__/Tabs/_docs/Tabs--api-specification.mdx
```diff
@@ -0,0 +1,40 @@
...
+## Overview
+
+<Canvas of={TabsStories.Playground} />
+<Controls of={TabsStories.Playground} />
+
+## Uncontrolled vs controlled
+
+This component is uncontrolled by default. You can specify a default active tab on load with `defaultSelectedKey`.
+
+If you need to control the state of the active tabs on the consuming side, use the `selectedKey` prop instead of `defaultSelectedKey`.
```
HeartSquared, line 30: Nit: Might be nice to have some code examples to showcase uncontrolled vs controlled.

packages/components/src/__future__/Tabs/_docs/Tabs.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <Button
+          label="Switch to tab 2"
+          onClick={(): void => setSelectedKey("two")}
+        />
+      </>
+    )
+  },
+}
+
+export const UsageInCard: Story = {
```
HeartSquared, line 77: Slightly confused about this story since it isn't showcased in the docs nor used in tests. Could we delete it?

packages/components/src/__future__/Tabs/subcomponents/Tab.module.css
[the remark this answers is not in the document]
HeartSquared: My preference is consistency 😛 If it's not too much trouble, then yes please. Otherwise happy to leave it for a later discussion.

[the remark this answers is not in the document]
dougmacknz: ez

packages/components/src/__future__/Tabs/_docs/Tabs.stories.tsx
dougmacknz: Good point, deleted

packages/components/src/__future__/Tabs/_docs/Tabs--api-specification.mdx
dougmacknz: coming up!

HeartSquared (approved):

PR #5178, opened 2024-10-17
docs: move components into logical groups

HeartSquared:
## Why

So that we're not in a half/half structure.

## What

Update Storybook nav structure by grouping components under headings based off [logical groups proposal](https://miro.com/app/board/uXjVK_gakNs=/).


HeartSquared:
Went the other way for now instead

PR #5194, opened 2024-10-21
FilterBar: manage focus on add/remove filters

dougmacknz:
## Why
Avoid losing focus for a11y purposes

## What
* Move focus to the filter button when adding a filter
* Move focus to 'Add filter' button when removing a filter

### DIFF
--- a/packages/components/src/Filter/FilterBar/context/types.ts
+++ b/packages/components/src/Filter/FilterBar/context/types.ts
@@ -32,6 +32,7 @@ export type FilterBarState<ValuesMap extends FiltersValues> = {
   activeFilterIds: Set<keyof ValuesMap>
   values: Partial<ValuesMap>
   dependentFilterIds: Set<keyof ValuesMap>
+  focusId?: keyof ValuesMap | undefined
 }
 
 export type ActiveFiltersArray<ValuesMap extends FiltersValues> = Array<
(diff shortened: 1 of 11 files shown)


packages/components/src/Filter/FilterBar/context/types.ts
```diff
@@ -32,6 +32,7 @@ export type FilterBarState<ValuesMap extends FiltersValues> = {
   activeFilterIds: Set<keyof ValuesMap>
   values: Partial<ValuesMap>
   dependentFilterIds: Set<keyof ValuesMap>
+  focusId?: keyof ValuesMap | undefined
```
mcwinter07, line 35: is the "?" here enough to assert that focusId could be undefined?

packages/components/src/Filter/FilterBar/subcomponents/AddFiltersMenu/AddFiltersMenu.tsx
```diff
@@ -14,13 +15,22 @@ export const AddFiltersMenu = (): JSX.Element => {
...
+  const { getInactiveFilters, showFilter, focusId, setFocus } =
+    useFilterBarContext()
   const inactiveFilters = getInactiveFilters()
 
+  useEffect(() => {
+    if (focusId === "add_filter") {
+      buttonRef.current?.focus()
+      setFocus(undefined)
+    }
+  }, [focusId])
```
mcwinter07, line 27: No notes but just to to talk throw my understanding so far, when the add_filter event is dispatched this will refocus to the button and set the focus in in the FilterBarContnent as part of some clean up?

dougmacknz: yeah exactly, it sets the focus to undefined to reset it back to its original state now that it's done the job we wanted

packages/components/src/Filter/FilterBar/context/types.ts
dougmacknz: Yep nice catch

mcwinter07 (approved):

PR #5195, opened 2024-10-22
fix(TextArea): Refactor autogrow to remove visual jank

dougmacknz:
Refactor TextArea `autogrow` feature to remove a pretty horrible visual jank.

In the future we will be able to a new CSS property `field-sizing` to do this natively, but for now we still need to use black magic. Method taken from: https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/

Bonus:
- Adjusted the focus ring implementation to remove the focus ring element and use outline instead
- Converted TextArea.module.scss to CSS
- Remove placeholder styling that isn't needed (tested and it still inherits the font etc, plus we're not supporting placeholders here anymore anyway).


dougmacknz:
(just running chromatic)

mcwinter07 (commented):
Just to confirm the expected behaviour with auto grow, the field should grow as the user add more content below the fold?  I tested on chrome and firefox and the auto grow story doesn't seems to extend the inputs height. Holler if I'm doing something wrong or misunderstood the fix :)

https://github.com/user-attachments/assets/0c98bada-6a56-4232-b946-adf3e9a058d9

dougmacknz:
Replaced by https://github.com/cultureamp/kaizen-design-system/pull/5197

PR #5197, opened 2024-10-24
fix(TextArea): Refactor autogrow to remove visual jank (take 2)

dougmacknz:
Refactor TextArea `autogrow` feature to remove a pretty horrible visual jank.

In the future we will be able to a new CSS property `field-sizing` to do this natively, but for now we still need to use black magic. Method taken from: https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/

Bonus:
- Adjusted the focus ring implementation to remove the focus ring element and use outline instead
- Converted TextArea.module.scss to CSS
- Remove placeholder styling that isn't needed (tested and it still inherits the font etc, plus we're not supporting placeholders here anymore anyway).


mcwinter07 (approved):
Nice work 😉 It'll be awesome when we get to use the native functionality *stares intensely at Safari and Firefox* but this looks like its just the ticket for now  🚀

PR #5245, opened 2024-11-08
chore(deps): upgrade to eslint v9

HeartSquared:
## Why

Eslint 8 is deprecated. Keeping on top of dep updates.

## What

- Upgrade to eslint 9
- Revised linting rules to be mostly based on current recommended rules
- Removed `eslint-plugin-ssr-friendly` as it is not compatible with Eslint flat config (the only option in v9+)
- Removed `@stylistic/eslint-plugin` as it conflicts too much with `prettier`
	- It does actually give us better control and nicer looking code, but we can revisit later if we are happy to remove prettier


HeartSquared:
Replaced with https://github.com/cultureamp/kaizen-design-system/pull/5339

PR #5269, opened 2024-11-12
fix(Pagination): Tighten margins and reduce pages shown to make usable on 320px

dougmacknz:
## Why
Pagination currently overflows 320px

image

## What
- Tighten up some margins and padding
- Reduce the 'sibling' page buttons that are shown down to 0 from 1 on small viewport

https://github.com/user-attachments/assets/91c8c4fa-ddb7-42a7-b890-1277288c9d5c


packages/components/src/Pagination/Pagination.module.scss
```diff
@@ -51,15 +51,3 @@
     }
   }
 }
-
-// Truncate indicator
-.truncateIndicatorWrapper {
```
dougmacknz, line 56: Wasn't being used at all, It got split out into it's own component and I guess they forgot to delete this

HeartSquared (approved):

PR #5277, opened 2024-11-13
fix: prevent EmptyState horizontal overflow in panels

yvonneyeh:
Fixed horizontal scrollbar appearing in `EmptyState` component when rendered within panels. The component previously overflowed its container at smaller viewport sizes, causing unwanted horizontal scrolling.

## Why
[CON-295](https://cultureamp.atlassian.net/browse/CON-295)

`EmptyState` overflows to the right and the border is rendered out of frame

EmptyState overflow

### Changes:
- Add box-sizing and proper width constraints to prevent content overflow
- Improve responsive layout with better flex properties
- Ensure illustrations and text stay within container bounds

Fixed EmptyState

[CON-295]: https://cultureamp.atlassian.net/browse/CON-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/EmptyState/EmptyState.module.scss
```diff
@@ -7,19 +7,22 @@
...
   width: 100%;
+  box-sizing: border-box;
+  justify-content: space-around;
   padding: calc(2% - var(--empty-state-border-width))
     calc(var(--spacing-24) - var(--empty-state-border-width));
   color: $color-purple-800;
   border: var(--empty-state-border-width) solid var(--empty-state-border-color);
   border-radius: var(--border-solid-border-radius);
   background-color: var(--empty-state-background-color);
+  overflow: hidden;
```
HeartSquared, line 19: We'd prefer to avoid this because it might end up truncating content unintentionally, do we definitely need it?

HeartSquared (commented):
Ideally I would suggest changing to use container queries, but given you've already made the changes and they seem to be working, I won't ask you to make the changes.

Although we'd definitely like to avoid hiding overflow, so hopefully it's something unnecessary and can be removed :)

packages/components/src/EmptyState/EmptyState.module.scss
yvonneyeh: Removed it!

HeartSquared (approved):
<3

PR #5286, opened 2024-11-18
Convert GuidanceBlock stylesheet to CSS

dougmacknz:
Cleaned this up before copying over some of it to a repo

### DIFF
--- a/packages/components/src/GuidanceBlock/GuidanceBlock.module.css
+++ b/packages/components/src/GuidanceBlock/GuidanceBlock.module.css
@@ -16,7 +16,7 @@
 
   display: flex;
   flex-direction: column;
-  gap: var(--spacing-sm);
+  gap: var(--spacing-12);
   min-width: 320px;
   max-width: var(--layout-breakpoints-large);
   border: var(--border-width) var(--border-solid-border-style);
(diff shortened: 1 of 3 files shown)


packages/components/src/GuidanceBlock/GuidanceBlock.module.css
```diff
@@ -166,51 +151,12 @@ $scene-illustration-size: 300px;
...
-  opacity: 0%;
+  opacity: 0;
   margin-bottom: 0;
 }
 
 .headingWrapper {
-  margin-bottom: $spacing-md;
-}
-
-.cancel {
```
dougmacknz, line 177: Not used anywhere in React

packages/components/src/GuidanceBlock/GuidanceBlock.module.css
```diff
@@ -34,60 +12,67 @@ $scene-illustration-size: 300px;
 
 .banner {
   --border-width: var(--border-width-1);
+  --banner-padding: var(--spacing-lg);
```
HeartSquared, line 15: Probably worth updating the spacing variables to the current ones

HeartSquared (approved):
Would be nice to use the up-to-date spacing tokens, but otherwise all good :)
I'll let you choose whether to update them or not, since it's not really a blocker.

packages/components/src/GuidanceBlock/GuidanceBlock.module.css
dougmacknz: I've done it for all of them except for this one, because there's no equivalent to `lg`

HeartSquared (approved):

PR #5287, opened 2024-11-18
refactor(EmptyState): update styles using container queries

HeartSquared:
## Why

The recent changes fixed an overflow issue with EmptyState, but accidentally introduced image shrinking in its place.
image

## What

- Update EmptyState styles to respond using container queries
- Add extra visual tests for container breakpoints
- Deprecate `layoutContext` prop as it no longer does anything
- Created a new StickerSheet component using CSS grid as container queries inside a table do not respond well
	- StickerSheet: Rename `heading` prop to `title`
	- StickerSheet: Add `headers` prop - replaces usage of `StickerSheet.Header`
	- StickerSheetRow: Rename `rowTitle` to `header`
- Deleted .spec file as the tests are not useful/covered by Storybook tests


docs/components/_future/StickerSheet/StickerSheet.module.css
```diff
@@ -0,0 +1,12 @@
+.stickerSheetSectionHeading {
+  margin-bottom: var(--spacing-12);
+
+  .stickerSheetContainer + .stickerSheetContainer & {
+    margin-top: var(--spacing-lg);
```
HeartSquared, line 5: Oops should change this spacing token

packages/components/src/EmptyState/EmptyState.module.css
```diff
@@ -0,0 +1,114 @@
...
+.content {
+  justify-content: center;
+  display: grid;
+  grid-template-columns: minmax(var(--empty-state-illustration-min-width), auto) minmax(
+      auto,
+      var(--empty-state-text-container-max-width)
+    );
+  grid-template-rows: minmax(auto, var(--empty-state-illustration-max-height)) auto;
+  padding: var(--spacing-24);
+  grid-column-gap: var(--spacing-32);
```
mcwinter07, line 66: Can't screenshot this cause figma is dumb but I think the original spacing was 36px gap between the image in the designs - small nit but just want to confirm the update was intended / hunky dory

mcwinter07: (noting that this makes sense to me since it aligns to our new spacing token values)

packages/components/src/EmptyState/EmptyState.tsx
```diff
@@ -66,6 +67,7 @@ export const EmptyState = ({
   id,
   illustrationType,
   variant = "informative",
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   layoutContext = "sidebarAndContent",
```
mcwinter07, line 71: Understand why we're pulling in out of the spread, but do we need a default if we're not going to use it?

packages/components/src/EmptyState/EmptyState.module.css
HeartSquared: Haha you got me! Totally forgot about this.
I changed it because we don't actually have a `spacing-36` 👀 There's only 32 and 40. Let me check with Jess.

docs/components/_future/StickerSheet/components/StickerSheetRow/StickerSheetRow.module.css
```diff
@@ -0,0 +1,5 @@
+.stickerSheetRow {
+  display: grid;
+  grid-template-columns: subgrid;
+  grid-column: 1 / -1;
```
mcwinter07, line 4: Huh, TIL that -1 in the grid-column will indent the content - funky syntactically but this is a cool blind spot to find 😉

mcwinter07 (commented):
Empty state change looks pretty good outside of the few little comments. Love the sticker sheet componentry moving away from the a `table` 👍

docs/components/_future/StickerSheet/components/StickerSheetRow/StickerSheetRow.module.css
HeartSquared: ...indent? 🙃 
I think it's actually saying "From the first column until the last column"? At least that was my understanding 😂

packages/components/src/EmptyState/EmptyState.tsx
HeartSquared: Nope we do not! I'll remove it :)

packages/components/src/EmptyState/EmptyState.module.css
HeartSquared: Confirmed with Jess that 32 is the reasonable one to use 😄

mcwinter07 (approved):

docs/components/_future/StickerSheet/types.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { StoryObj } from "@storybook/react"
+import type { ColorSchema } from "~components/LikertScaleLegacy"
+
+export type StickerSheetArgs = {
+  isReversed?: boolean
+  colorSchema?: ColorSchema
```
HeartSquared, line 7: Whut -_- Will kill in the follow up PR

PR #5294, opened 2024-11-20
fix(upgradeIconV1): update some icons to default to filled

HeartSquared:
## Why

As per designer, exceptions for circle icons which should be defaulted to filled.

## What

- Update codemod to default the following icons to filled:
	- check_circle
	- info
	- error
	- help
	- privacy_tip
- Update README with more details of what `upgradeIconV1` does


mcwinter07 (approved):
Looks  hunky dory to me (~~ -_-)~~ merge on!

PR #5323, opened 2024-11-26
Kzn 2730/update info button label

Zystix:
## Why
Provide users with a way to have contextual accessible labels for the Information Button.

https://cultureamp.atlassian.net/jira/software/c/projects/KZN/boards/634?selectedIssue=KZN-2730

## What
Adds a prop to the GenericTile component to allow control of the Information button's label.

### DIFF
--- a/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
+++ b/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
@@ -9,7 +9,6 @@ const meta = {
     title: "Title",
     metadata: "Side A",
     footer: <>Example Footer</>,
-    information: "Side B",
   },
   argTypes: {
     mood: { control: false },
@@ -27,12 +26,6 @@ export const Playground: Story = {
         sourceState: "shown",
       },
     },
-  },
-}
-
-export const InfoLabelTest: Story = {
-  args: {
-    infoButtonLabel: "Test",
   },
 }
 
@@ -58,6 +51,13 @@ export const Variants: Story = {
 
 export const Information: Story = {
   args: {
+    information: "Side B",
+  },
+}
+
+export const TileWithCustomInfoLabel: Story = {
+  args: {
+    infoButtonLabel: "Test",
     information: "Side B",
   },
 }
--- a/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
+++ b/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
@@ -32,8 +32,8 @@ export const Playground: Story = {
 
 export const InfoLabelTest: Story = {
   args: {
-    infoButtonLabel: "Test"
-  }
+    infoButtonLabel: "Test",
+  },
 }
 
 export const Variants: Story = {
--- a/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
+++ b/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
@@ -9,6 +9,7 @@ const meta = {
     title: "Title",
     metadata: "Side A",
     footer: <>Example Footer</>,
+    information: "Side B",
   },
   argTypes: {
     mood: { control: false },
@@ -27,6 +28,20 @@ export const Playground: Story = {
       },
     },
   },
+}
+
+export const InfoLabelTest: Story = {
+  render: args => (
+    <>
+      <InformationTile
+        {...args}
+        variant="default"
+        title="default"
+        footer="default"
+        infoButtonLabel="Testy test"
+      />
+    </>
+  ),
 }
 
 export const Variants: Story = {
--- a/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
+++ b/packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
@@ -32,7 +32,7 @@ export const Playground: Story = {
 
 export const InfoLabelTest: Story = {
   args: {
-    infoButtonLabel: "Super Testy test"
+    infoButtonLabel: "Test"
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
```diff
@@ -22,6 +22,7 @@ export type GenericTileProps = {
   titleTag?: AllowedHeadingTags
   metadata?: string
   information?: TileInformation | React.ReactNode
+  infoButtonLabel?: string
```
mcwinter07, line 25: Would be good to add a property description here with the `@default` value called out. Something variant prop below :)

packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
```diff
@@ -29,6 +30,20 @@ export const Playground: Story = {
   },
 }
 
+export const InfoLabelTest: Story = {
+  render: args => (
```
mcwinter07, line 34: One handy feature of the modern Storybook is that whenever we're just updating prop values in a story, we can pass in the new `args` instead of redefining the render, ie:

```tsx
export const InfoLabelTest: Story = {
  args: {
    infoButtonLabel: "Testy test"
  }
}
```

Will product the same result as what we had but with less lines of code 😄 

What that's doing under the hood is passing all the storybooks args into the `InformationTile` component.

Non-essential watching but this [video walk through](https://storybook.js.org/docs/api/csf) helped me get a general understanding of what's happening in the format :)

mcwinter07 (commented):
Nice work on the first PR @Zystix 👍  I've left a  few comments for some addition improvement. It's probably worth also addressing the IconButton's label on the reverse side ([ln 143](https://github.com/cultureamp/kaizen-design-system/pull/5323/files#diff-d747b3ff0acc22418c27b202ea5e87a145b03ffe197dbd93538f7e57644940fbR143)). We could probably use string interpolation to construct a label using the `title` prop, something like `return to ${title}`, though this will also have to be translated before we merge in the change. Happy to chat more about this tomorrow too

packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
Zystix: Done!

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
Zystix: Done!

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
```diff
@@ -46,6 +47,7 @@ export const GenericTile = ({
   titleTag = "h3",
   metadata,
   information,
+  infoButtonLabel = "Information",
```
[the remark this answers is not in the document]
Zystix: Gave it a whirl, lemme know if it looks right

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
```diff
@@ -46,13 +49,23 @@ export const GenericTile = ({
...
   variant = "default",
   footer,
   classNameOverride,
   ...restProps
 }: GenericTileProps): JSX.Element => {
   const [isFlipped, setIsFlipped] = useState<boolean>(false)
+  const { formatMessage } = useIntl()
+
+  infoButtonLabel = infoButtonLabel
+    ? formatMessage({
```
mcwinter07, line 63: This one we'll want to flip the logic a little. When the `infoButtonLabel` is undefined we'll provide a translatable default, but when it is defined, we can assume that our users will provide a translated string.

There's also some small things we'll need to tweak on the description, but we can talk about that in our pairing catch up :)

packages/components/src/Tile/InformationTile/_docs/InformationTile.stories.tsx
```diff
@@ -9,6 +9,7 @@ const meta = {
     title: "Title",
     metadata: "Side A",
     footer: <>Example Footer</>,
+    information: "Side B",
```
mcwinter07, line 12: I reckon we move this into the args in ln 35 just to make it explicit that it relates to this story

Zystix: We have it in both places atm, is it needed by all the tests then?

Zystix: Done!

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
Zystix: Done!

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
Zystix: Done!

packages/components/src/Tile/subcomponents/GenericTile/GenericTile.tsx
```diff
@@ -80,7 +94,7 @@ export const GenericTile = ({
       {information && (
         <div className={styles.informationBtn}>
           <IconButton
-            label="Information"
+            label={`${translatedInfoLabel} ${title}`}
```
mcwinter07, line 97: I just thought of one problem that we'll come into here. If a user does provide a custom label, we're going to be passing in the Tiles `title` regardless 🤔 That could mean they do something like:

```tsx
<InformationTile 
  {...otherProps}
  title="Performance Insights"
  infoButtonLabel="Reveal more information for Performance Insights"
/>
```

That would create an accessible label saying:

"Reveal more information for Performance Insights Performance insights"

Zystix: Made a small change

Zystix: Made a change, what do you think?

mcwinter07 (approved):
Awesome work on your first PR 👍

PR #5325, opened 2024-11-28
Add deprecated isReversed prop to Button

mcwinter07:
## Why
The current method to set reversed styles on Button v3 is to use the `ReverseColors` provider that we export from `"@kaizen/components/v3/utilities"`.

This presents a risk for the codemod, as each instance of Button that was reversed would likely need to be wrapped in a Provider, whereas the intent behind `ReverseColors` is to wrap a parent element to reduce the number of props needed to style reversed color themes.

As we can’t predictably say how this could impact renders or their performance we think a safer methods of gradual migration is to add a deprecated `isReversed` prop. This will allow the codemod to run safely and provide type warning to our consumers which is a clearer way to describe the new pattern.

## What
- Adds `isReversed` prop and JSdoc deprecated flag to Button v3


packages/components/src/__actions__/Button/v3/Button.tsx
```diff
@@ -45,11 +51,12 @@ export const Button = forwardRef(
       isPending,
       hasHiddenPendingLabel: propsHasHiddenPendingLabel = false,
       pendingLabel,
+      isReversed = false,
       ...restProps
     }: ButtonProps,
     ref: React.ForwardedRef<HTMLButtonElement>
   ) => {
-    const isReversed = useReversedColors()
+    const isReversedVariant = useReversedColors() || isReversed
```
HeartSquared, line 59: I think hooks need to be called as they are.
Also, the prop should override the context imo, given it's explicit whereas the context is more magical.

packages/components/src/__actions__/Button/v3/Button.tsx
```diff
@@ -26,6 +26,12 @@ type ButtonBaseProps = Omit<RACButtonProps, "children"> & {
   iconPosition?: "start" | "end"
   /** Controls if the button inherits width from its parent. @default "false" */
   isFullWidth?: boolean
+  /**
+   * Use the [ReverseColors](https://cultureamp.design/?path=/docs/actions-button-button-v3-api-specification--docs#variants) Provider instead.
+   * @default false
+   * @deprecated Controls the reversed style of Button. This is here to support gradual migration to the ReverseColors Provider.
```
HeartSquared, line 32: Ideally I think we should swap the prop description with the deprecation message.

dougmacknz (dismissed):

packages/components/src/__actions__/Button/v3/Button.tsx
```diff
@@ -45,11 +50,14 @@ export const Button = forwardRef(
...
       pendingLabel,
+      isReversed,
       ...restProps
     }: ButtonProps,
     ref: React.ForwardedRef<HTMLButtonElement>
   ) => {
-    const isReversed = useReversedColors()
+    const shouldUseReverse = useReversedColors()
+    const isReversedVariant =
+      isReversed !== undefined ? isReversed : shouldUseReverse
```
HeartSquared, line 60: You can do this shorthand (null-coalescing) :)
```suggestion
    const isReversedVariant = isReversed ?? shouldUseReverse
```

HeartSquared (dismissed):
Just one nit. This will fail linting in the future, so easier if we just fix it now, but I won't block the PR for it.

packages/components/src/__actions__/Button/v3/Button.tsx
mcwinter07: Oh even better 😉

HeartSquared (approved):

PR #5336, opened 2024-12-01
Fix Tile flip focus

Zystix:
## Why
A11Y fix ensuring focus is returned to the info button when the tile is flipped. This makes Tiles more accessible for screen reader users.

https://cultureamp.atlassian.net/browse/KZN-2306

## What
Added refs to the info buttons on both sides of the tile and refocused on them based on the flip state of the tile.


mcwinter07 (approved):
Nice work. I think this is a good improvement on the focus management on the Tile flip ☁️5️⃣

PR #5338, opened 2024-12-02
chore(deps): upgrade eslint to v9

HeartSquared:
## Why

Eslint 8 is deprecated. Keeping on top of dep updates.

## What

- Upgrade to eslint 9
- Revised linting rules to be mostly based on current recommended rules
- Removed `eslint-plugin-ssr-friendly` as it is not compatible with Eslint flat config (the only option in v9+)
- Removed `@stylistic/eslint-plugin` as it conflicts too much with `prettier`
	- It does actually give us better control and nicer looking code, but we can revisit later if we are happy to remove prettier

## Notes

You may need to update your VSCode to use the flat config
image

## Related branches

1. Release branch: https://github.com/cultureamp/kaizen-design-system/pull/5339
2. Config: This branch
3. Auto fixes: https://github.com/cultureamp/kaizen-design-system/pull/5340
4. Manual fixes: https://github.com/cultureamp/kaizen-design-system/pull/5341


mcwinter07 (approved):
Scanned through the config and that looks good to me. We'll likely have lots of changes in code but this is to be expected 👍

PR #5341, opened 2024-12-02
style: manually fix linting errors

HeartSquared:
## What

Manually fixed linting.

1. Release branch: https://github.com/cultureamp/kaizen-design-system/pull/5339
2. Config: https://github.com/cultureamp/kaizen-design-system/pull/5338
3. Auto fixes: https://github.com/cultureamp/kaizen-design-system/pull/5340
4. Manual fixes: This branch


packages/components/src/RichTextEditor/utils/core/createRichTextEditor.ts
```diff
@@ -58,9 +58,7 @@ export const createRichTextEditor = ({
   })
 
   return {
-    destroy: () => {
-      editorView && editorView.destroy()
-    },
+    destroy: () => editorView.destroy(),
```
mcwinter07, line 61: This looks like it checks out. Just to regurgitate my understanding: because we're assigning a value to editorView on ln 53, we don't (and guess we never did) need the `&&` so we wont need  `editorView?.destroy()`

HeartSquared: No idea. I just removed the first part and it didn't complain about the second, so I left it 😅 I can put the question mark back in if you're worried the types are actually inaccurate?

packages/components/src/ToggleSwitch/ToggleSwitchField/ToggleSwitchField.tsx
```diff
@@ -31,7 +31,7 @@ export const ToggleSwitchField = ({
   ...restProps
 }: ToggleSwitchFieldProps): JSX.Element => {
   const reactId = useId()
-  const id = propsId || reactId
+  const id = propsId ?? reactId
```
mcwinter07, line 34: Small nit, feel free to ignore but do we want to update this to fallbackId to keep with the convention of the other changes?

packages/components/src/RichTextEditor/utils/core/createRichTextEditor.ts
mcwinter07: I mean it makes sense based on the code thats there so if types aren't complaining that should be pretty safe

packages/components/src/ToggleSwitch/ToggleSwitchField/ToggleSwitchField.tsx
HeartSquared: Done :)

mcwinter07 (approved):

PR #5347, opened 2024-12-04
chore(deps): remove @tanstack/react-query

HeartSquared:
## What

Remove `@tanstack/react-query`.
Converted the story to a code example.

## Why

Found it throwing errors in the Storybook build. Given it was only around for documentation and was essentially only for showcasing one way of writing consumer code, it's not worth keeping.

image


mcwinter07 (approved):
Nice, lets purge this from our repo since we don't actually use it 🔥

PR #5355, opened 2024-12-05
Tabs (future): add carousel functionality when tabs overflow container

dougmacknz:
## What
Add carousel to Future Tabs when tabs start overflowing the width of the viewport

https://github.com/user-attachments/assets/a98a1d97-b6ef-428d-ac53-9e87ad3ebffc

## Why
Nicer than having heaps of tabs wrap to a new line

### DIFF
--- a/packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
+++ b/packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
@@ -1,13 +1,12 @@
 import React from 'react'
 import { Meta, StoryObj } from '@storybook/react'
-import { within, userEvent } from '@storybook/test'
+import { within, userEvent, expect } from '@storybook/test'
 import { Text } from '~components/Text'
 import { Tab, TabList, TabPanel, Tabs } from '../index'
 
 const meta = {
-  title: 'Components/Tabs/Tabs (Future)/ Tabs (Future) (Tests)',
+  title: 'Components/Tabs/Tabs (Future)/Tests',
   parameters: {
-    chromatic: { disable: false },
     controls: { disable: true },
   },
 } satisfies Meta
@@ -16,7 +15,7 @@ export default meta
 
 type Story = StoryObj<typeof meta>
 
-export const OverflowCarousel: Story = {
+export const ArrowsShowingAndHiding: Story = {
   render: () => {
     return (
       <div style={{ maxWidth: '500px' }}>
@@ -58,16 +57,22 @@ export const OverflowCarousel: Story = {
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement.parentElement!)
 
+    expect(canvas.queryByTestId('kz-tablist-left-arrow')).not.toBeInTheDocument()
+
     const rightArrow = await canvas.findByTestId('kz-tablist-right-arrow')
 
     await userEvent.click(rightArrow)
-    await new Promise((r) => setTimeout(r, 200))
-
-    await userEvent.click(rightArrow)
-    await new Promise((r) => setTimeout(r, 200))
+    await new Promise((r) => setTimeout(r, 500))
 
     const leftArrow = await canvas.findByTestId('kz-tablist-left-arrow')
-    await userEvent.click(leftArrow)
-    await new Promise((r) => setTimeout(r, 200))
+
+    expect(leftArrow).toBeInTheDocument()
+    expect(rightArrow).toBeInTheDocument()
+
+    await userEvent.click(rightArrow)
+    await new Promise((r) => setTimeout(r, 500))
+
+    expect(leftArrow).toBeInTheDocument()
+    expect(rightArrow).not.toBeInTheDocument()
   },
 }
--- a/packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
+++ b/packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
@@ -20,7 +20,7 @@ export const ArrowsShowingAndHiding: Story = {
     return (
       <div style={{ maxWidth: '500px' }}>
         <Tabs>
-          <TabList aria-label="Tabs">
+          <TabList aria-label="Tabs" data-testid="sb-arrows">
             <Tab id="one">Tab 1</Tab>
             <Tab id="two">Tab 2</Tab>
             <Tab id="three" badge="3">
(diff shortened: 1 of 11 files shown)


dougmacknz:
(putting in review for chromatic)

packages/components/src/__future__/Tabs/subcomponents/TabList/TabList.tsx
```diff
@@ -20,13 +21,114 @@ export type TabListProps = {
...
+    <div className={styles.container} id={tabListId}>
+      {leftArrowEnabled && (
+        // making a conscious decision to use <div onClick> over <button> here, because:
+        // - <button> would add pointless noise for a screen reader user
+        // - keyboard only user can toggle through tabs with left/right arrow keys already
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+        <div
+          onClick={() => handleArrowPress('left')}
+          className={styles.leftArrow}
+          data-testid="kz-tablist-left-arrow"
```
HeartSquared, line 109: Potentially a nit, but I think it would be better to avoid hard-coding a testid into the elements since if you have multiple on a page, you'll find multiple elements. Perhaps if there was something you could combine it with (maybe even just prefix it with `data-testid` from the spread props?)

packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { within, userEvent, expect } from '@storybook/test'
+import { Text } from '~components/Text'
+import { Tab, TabList, TabPanel, Tabs } from '../index'
+
+const meta = {
+  title: 'Components/Tabs/Tabs (Future)/Tests',
+  parameters: {
+    controls: { disable: true },
+  },
+} satisfies Meta
```
HeartSquared, line 12: I'd stick to the pattern for the normal stories so you can utilise the meta for the rest of the stories. The StickerSheet should be the only exception because we give it some special args.

```suggestion
} satisfies Meta<typeof Tabs>
```

packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement.parentElement!)
+
+    expect(canvas.queryByTestId('kz-tablist-left-arrow')).not.toBeInTheDocument()
+
+    const rightArrow = await canvas.findByTestId('kz-tablist-right-arrow')
+
+    await userEvent.click(rightArrow)
+    await new Promise((r) => setTimeout(r, 500))
```
HeartSquared, line 65: `await waitFor` the left arrow should hopefully allow you to not need a timeout :)

packages/components/src/__future__/Tabs/subcomponents/TabList/TabList.module.css
```diff
@@ -1,8 +1,60 @@
+.container {
+  position: relative;
+}
+
 .tabList {
   border-bottom: 1px solid rgba(var(--color-gray-600-rgb), 0.1);
-  padding: var(--spacing-xs) var(--spacing-md) 0;
+  padding: var(--spacing-xs) 0 0;
```
HeartSquared, line 7: Could we use the new equivalent variable? :)

packages/components/src/__future__/Tabs/subcomponents/TabList/TabList.tsx
```diff
@@ -20,13 +21,114 @@ export type TabListProps = {
...
+
+    return () => {
+      firstTabObserver.disconnect()
+      lastTabObserver.disconnect()
+    }
+  }, [isDocumentReady, containerElement, isRTL])
+
+  const handleArrowPress = (direction: 'left' | 'right'): void => {
+    if (tabListRef.current) {
+      const scrollAmount = 120
```
dougmacknz, line 91: I should move this out into a constant

HeartSquared (commented):
Main thing to fix that we found - we need to scroll the active tab into view if an external event changes the active tab.

packages/components/src/__future__/Tabs/_docs/Tabs.spec.stories.tsx
dougmacknz: For the first one yes, but then I'll still need the timeout when there's no arrow appearance/disappearance to hook into. So is it worth doing it, or just keep it consistent with the timeout

HeartSquared: Checking that the element is not in the dom when you expect it to disappear probably should work?
I'd just try to avoid timeouts when you have an explicit result you're waiting for, because it's still potentially flaky. It only makes sense to use it when the thing you're testing has functionality that relies on timeouts. It's okay as a last resort if nothing else works though.

dougmacknz: The timeout is actually more robust than that. Your idea works okay when it's testing for disappearance, but hooking into appearance actually breaks this. 

The problem with testing for appearance/disappearance is that that doesn't mean the animation is done. The sequence goes:
  * Press right -> animation starts -> left arrow appears -> animation finishes. 

Even with disappearance, it's closer, but it still doesn't mean the animation is done because of the threshold of 75% (it disappears before the element is completely out of view).

So testing for left arrow appearing ends up trying to press right again before the animation is done and it breaks.

There's another issue as well as mentioned in my last comment: if the example had more tabs in it, or if we reduce the scroll amount,  there's not always an arrow appearing or disappearing on each press (only the first and last presses). There would be times where both arrows are showing before the press, and continuing to show after press.

The main thing I don't love at the moment is the magical number on the timeout.

dougmacknz: If we want to go deep on this, we could see if there's a way to hook into/run a callback when the scrolling is finished.

e.g.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event

Not supported on Safari tho :(

packages/components/src/__future__/Tabs/constants.ts
```diff
@@ -0,0 +1 @@
+export const SCROLL_AMOUNT = 120
```
HeartSquared, line 1: Nit: Haha, probably didn't need to make a separate file for this since we only use it in the one place. But no matter.

packages/components/src/__future__/Tabs/subcomponents/TabList/TabList.tsx
```diff
@@ -1,6 +1,13 @@
-import React, { ReactNode } from 'react'
+import React, { ReactNode, useContext, useEffect, useId, useRef, useState } from 'react'
 import classnames from 'classnames'
-import { TabList as RACTabList, TabListProps as RACTabListProps } from 'react-aria-components'
+import {
+  TabList as RACTabList,
+  TabListProps as RACTabListProps,
```
HeartSquared, line 5:
```suggestion
  type TabListProps as RACTabListProps,
```

HeartSquared (approved):
Just some nits. Feel free to merge if you wish :)

PR #5358, opened 2024-12-05
fix(TextArea): Only use internal value for autogrow data property

dougmacknz:
## Why
https://cultureamp.slack.com/archives/C0189KBPM4Y/p1733288289604299

## What
There's an internal version of the TextArea value held to support the autogrow feature. It was being used to feed into the TextArea in the case that value isn't provided, but that's not actually needed.

### DIFF
--- a/packages/components/src/TextArea/TextArea.tsx
+++ b/packages/components/src/TextArea/TextArea.tsx
@@ -33,11 +33,9 @@ export const TextArea = ({
 }: TextAreaProps): JSX.Element => {
   const [internalValue, setInternalValue] = useState<
     string | number | readonly string[] | undefined
-  >(autogrow && !value ? defaultValue : undefined)
-  // ^ holds an internal state of the value so that autogrow can still work with uncontrolled textareas
-  // essentially forces the textarea into an (interally) controlled mode if autogrow is true and mode is uncontrolled
+  >(value ?? defaultValue ?? '')
+  // ^holds an internal state of the value, used for the autogrow feature if uncontrolled (no `value` provided)
 
-  const controlledValue = value ?? internalValue
   const fallbackRef = useRef(null)
   const textAreaRef = propsTextAreaRef ?? fallbackRef
 
@@ -51,7 +49,7 @@ export const TextArea = ({
       className={classnames(styles.wrapper, {
         [styles.wrapperAutogrow]: autogrow,
       })}
-      data-value={autogrow ? controlledValue : undefined}
+      data-value={value ?? internalValue}
     >
       <textarea
         className={classnames(
@@ -65,8 +63,8 @@ export const TextArea = ({
         )}
         rows={rows}
         onChange={onChange}
-        value={controlledValue}
-        defaultValue={controlledValue ? undefined : defaultValue}
+        value={value}
+        defaultValue={value ? undefined : defaultValue}
         // ^ React throws a warning if you specify both a value and a defaultValue
         ref={textAreaRef}
         disabled={disabled}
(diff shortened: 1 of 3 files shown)


packages/components/src/TextArea/TextArea.tsx
```diff
@@ -33,11 +33,9 @@ export const TextArea = ({
 }: TextAreaProps): JSX.Element => {
   const [internalValue, setInternalValue] = useState<
     string | number | readonly string[] | undefined
-  >(autogrow && !value ? defaultValue : undefined)
-  // ^ holds an internal state of the value so that autogrow can still work with uncontrolled textareas
-  // essentially forces the textarea into an (interally) controlled mode if autogrow is true and mode is uncontrolled
+  >(value ?? defaultValue ?? '')
```
HeartSquared, line 36: According to your comment below, `value` shouldn't exist right?
```suggestion
  >(defaultValue ?? '')
```

dougmacknz: We could get away with that yeah. It wouldn't matter for the actual behaviour, because value takes priority over this for autogrow, but it would create a situation where this internal value is not accurate when value is provided. I did this just for what I considered "correctness" but now that I think of it, there still could be a situation where this becomes incorrect if the consumer adjusts `value` without `onChange` running.

dougmacknz: I was thinking it would be ugly to only populate/update this state when needed, but it's probably worth it to save potential re-renders and what not so I've changed it now.

packages/components/src/TextArea/_docs/TextArea.stickersheet.stories.tsx
```diff
@@ -24,20 +25,30 @@ export default {
...
+        autogrow
+      />
```
HeartSquared, line 50: Nit: Another option :)

```suggestion
      <StickerSheet title="Autogrow">
      <StickerSheet.Row>
      <TextArea
        rows={1}
        value="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque mattis nisi sit amet consectetur ultricies. Vestibulum porta arcu vitae erat egestas pulvinar. Proin tincidunt mi ultricies risus ultrices semper. Donec consequat, mi at tincidunt mattis, felis metus semper metus, a imperdiet justo odio ac elit. Nulla et lacinia enim. Donec placerat, arcu a hendrerit iaculis, massa ante venenatis urna, vitae vestibulum massa orci et erat. Sed in cursus arcu. Fusce fringilla urna tincidunt risus sodales mollis. Ut sit amet mi vitae justo aliquam congue eget eu nisl. Mauris sit amet dolor fermentum, rutrum orci eget, feugiat mi. Etiam feugiat auctor augue, non volutpat nisi aliquet non. Praesent dignissim, lacus id iaculis porttitor, purus libero fermentum urna, in faucibus massa enim sed dui. Maecenas et nisi in nulla condimentum dictum. Maecenas tincidunt turpis non consequat pharetra. Suspendisse in auctor erat, vel ullamcorper elit. Nullam rutrum pharetra turpis, id luctus nisi sollicitudin ac. Maecenas sodales malesuada orci. Integer ultrices, nisi non blandit commodo, turpis enim iaculis ante, vel blandit diam sapien eget justo. Mauris scelerisque euismod quam, nec accumsan lorem venenatis vel. Sed rhoncus purus turpis, vel efficitur metus placerat et."
        autogrow
      />
      </StickerSheet.Row>
      </StickerSheet>
```

HeartSquared (approved):

HeartSquared (approved):

PR #5368, opened 2024-12-08
Mark up Tiles in TileGrid as List

Zystix:
## Why

Marking up TileGrid items in a list so that screen readers can report on how many tiles are in a list, so that context about the TileGrid is more accessible.

https://cultureamp.atlassian.net/browse/KZN-2865

## What

Converted TileGrid to contain its children in an unordered list and wrapped each tile in a list item when passed as props to the TileGrid.

### DIFF
--- a/packages/components/src/Tile/TileGrid/_docs/TileGrid.stories.tsx
+++ b/packages/components/src/Tile/TileGrid/_docs/TileGrid.stories.tsx
@@ -78,3 +78,43 @@ export const FlipOneNotOthers: Story = {
     })
   },
 }
+
+// Does it work with one tile
+export const OneTile: Story = {
+  args: {
+    children: (
+      <InformationTile
+        title="Title A"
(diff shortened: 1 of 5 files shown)


packages/components/src/Tile/TileGrid/_docs/TileGrid.stories.tsx
```diff
@@ -78,3 +78,85 @@ export const FlipOneNotOthers: Story = {
     })
   },
 }
+
+// @Zystix: Add Chromatic flag to tests or add to stickersheet.
+
+// Does it work with one tile
```
HeartSquared, line 84: Maybe just clean up by removing the comments (and L107). Code should be self-documenting, so usually comments are reserved for business logic/hard to read code (which may be caused by consuming a library) :)

HeartSquared (approved):

PR #5370, opened 2024-12-10
Fix Broken Workflow Style in RTL

Zystix:
## Why

The stepDivider styling for the Workflow componenent was misbehaving in RTL

https://cultureamp.atlassian.net/browse/KZN-2884

## What

Reinstated RTL specific styling with the :dir(rtl) selector instead.


HeartSquared (commented):
Awesome! Could you please add a Chromatic test so we ensure we don't miss this in the future? :)

HeartSquared (approved):
👏🏻

PR #5375, opened 2024-12-13
Initial Conversion of SCSS to CSS

Zystix:
## Why

Converting SCSS to CSS

https://cultureamp.atlassian.net/browse/KZN-2877

## What

Converts Avatar, AvatarGroup, Badge, Brand, BrandMoment to CSS


packages/components/src/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,163 @@
+.wrapper {
+  --avatar-sm: 1.25rem;
+  --avatar-md: 3rem;
+  --avatar-lg: 4.5rem;
+  --avatar-xl: 6rem;
+  --avatar-xxl: 7.75rem;
```
mcwinter07, line 6: Looking at this and noticed we actually slim down the number of variables we have here to a single one, something like `--avatar-x-y`.

We can then pass it into the width and height properties in the base class, then re-assign the value in all the modifiers like `.small`.

packages/components/src/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,163 @@
...
+
+  background: var(--color-gray-300);
+  border-radius: 100%;
+  box-sizing: border-box;
+  overflow: hidden;
+  align-items: center;
+  display: flex;
+  justify-content: center;
+
```
mcwinter07, line 15: What I'm thinking is something like:
```
width: var(--avatar-x-y);
height: var(--avatar-x-y);
```
here and then in the sizes we can just update the value of the CSS variable, ie:
```
  &.small {
    --avatar-x-y: 1.25rem;
  }
```

packages/components/src/BrandMoment/BrandMoment.module.css
```diff
@@ -38,67 +34,73 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
...
 // --------------------------------
 // Reusable container
 // --------------------------------
+*/
 .container {
-  max-width: $layout-content-max-width;
+  max-width: var(--layout-content-max-width);
   margin: 0 auto;
-  padding: 0 $spacing-md;
+  padding: 0, var(--spacing-24);
```
mcwinter07, line 45: When placing in multiple values into a a shorthand property in CSS like padding we don't need to add commas - it might actually break the CSS on the other end as it could interpret `,` as a design value.

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,105 @@
+.badge {
+  --dt-color-background-default: var(--color-gray-300);
+  --small: var(--spacing-24);
+  --badge-duration-timing: (var(--animation-duration-slow) / 2)
```
mcwinter07, line 4: I don't think this is used anywhere? Looks like we could probs axe it

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,105 @@
+.badge {
+  --dt-color-background-default: var(--color-gray-300);
```
mcwinter07, line 2: Looking at the code below instead of creating a variable that gets used only once I recon we update this so we set this on `.badge` class and then update that value in the modifier classes like `active`.

Also its good opportunity to use a default value in your CSS variable instead of having to declare it at the top of the stylesheet.

CSS variable is pretty handy that we can do something like:

```
.badge {
    <!-- other properties -->
    background-color: var(--badge-background-color, var(--color-gray-300))
}
```

Then later in the stylesheet, we just reassign the CSS variable as usual in the modifier class, ie:

```
.active {
    --badge-background-color: var(--color-blue-500)
}
```

packages/components/src/AvatarGroup/AvatarGroup.module.css
```diff
@@ -0,0 +1,70 @@
+.AvatarGroup {
+  --avatar-sm: 1.25rem;
+  --avatar-md: 3rem;
+  --avatar-lg: 4.5rem;
```
mcwinter07, line 4: I think there's the same opportunity here to reduce this down to a single variable and update that value in the modifier classes

mcwinter07 (commented):
I've left a couple of comment about some opportunities we have to clean up the variable usage in the CSS.

With the ButtonGroup I can see the a CSS module but a SCSS module deletion in the PR diff / an update to the CSS import.

mcwinter07:
Ooo, also I'd checkout the UI tests for the Avatar, a little bit of funkiness see to be happening in the Avatar stickersheets 😉

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,103 @@
...
+  padding: 2px 1.875rem;
+  width: 24px;
+}
+
+.default {
+  color: var(--color-purple-800);
+}
+
+.active {
+  background-color: var(--color-blue-500);
```
HeartSquared, line 36: It looks like you've created a CSS var on L17. Since you've done that, I suggest you change this (and the other) declarations to overriding the variable as opposed to overriding the attribute.

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    &.active {
+      background-color: var(--color-blue-500);
+      color: var(--color-white);
+    }
+
+    &.dot {
+      background-color: var(--color-green-300);
+    }
```
HeartSquared, line 101: Looks like `.dot` never changes background colour no matter what variation. I think you can simplify the styles by deduping.

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,103 @@
...
+
+  .badge {
+    transition:
+      transform var(--badge-duration-timing),
+      background-color var(--badge-duration-timing);
+    transform-origin: center;
+    transform: scale3d(1, 1, 1);
+  }
+
+  &.animationOn .badge {
```
HeartSquared, line 87: You should be able to take this layer of nesting out.

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,103 @@
...
+
+.reversed.dark {
+  background-color: var(--color-purple-700);
+  color: var(--color-white);
+}
+
+.reversed {
+  background-color: rgba(var(--color-white-rgb), 0.1);
+  color: var(--color-white);
+}
```
HeartSquared, line 74: This has a lower specificity than the combined classes above. I suggest you move this up, as reading the file isn't reflective of the actual cascade.

Another thing - for the classes that change styles when combined with `.reversed`, you can use nesting to make it a little more readable.

packages/components/src/BrandMoment/BrandMoment.module.css
```diff
@@ -38,67 +34,73 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
...
-    padding: 0 $spacing-xxl;
+    padding: 0 3.75rem;
   }
 }
 
-@media (min-width: $breakpoint-xl) {
+@media (width >= 1366px) {
   .container {
-    padding: 0 $spacing-xxxxl;
+    padding: 0 5.25rem;
```
HeartSquared, line 56: It's possibly a little confusing since it's not a proper spacing token migration guide, but this has been approved by Jess for the old spacing tokens to the new ones.

https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/4459692600/KAIO+Adoption+Guide#Box-migration

You'll need to do a little extra step to get the conversion (Box was using another size system), but essentially grab the old value (in rem) then *16 (to get px), then use the equivalent new token.
In your case, `spacing-xxxxl` = `5.25rem` = `n * 16` = `84px` = `spacing-80`.

You will get updates in Chromatic when you make these changes - these are to be expected, and you'll want to communicate these to consumers too on release.

packages/components/src/BrandMoment/BrandMoment.module.css
```diff
@@ -38,67 +34,73 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
...
 // --------------------------------
 // Main
 // --------------------------------
+*/
 .main {
   flex: 1 0 auto;
   display: flex;
   align-items: center;
-  margin-block-end: $spacing-xxl;
+  margin-block-end: 3.75rem;
```
HeartSquared, line 79: As above

packages/components/src/BrandMoment/BrandMoment.module.css
```diff
@@ -38,67 +34,73 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
...
 }
 
 .mainInner {
   display: grid;
   grid-template-columns: 1fr;
-  gap: $spacing-xl $spacing-lg;
+  gap: var(--spacing-48) 2.25rem;
```
HeartSquared, line 85: As above (and more below this point - I won't comment on them all)

packages/components/src/BrandMoment/BrandMoment.module.css
```diff
@@ -160,20 +162,22 @@ $breakpoint-xl: 1366px; // TODO: create an xl breakpoint in design-tokens and re
   }
 }
 
+/*
 // --------------------------------
 // Footer
 // --------------------------------
+*/
```
HeartSquared, line 169: I suggest you just simplify these to a smaller block, or at least make the comment symbols match haha. Or even delete them. They actually aren't required.

```suggestion
/*
 * Footer
 */
```

```suggestion
/* Footer */
```

packages/components/src/ButtonGroup/ButtonGroup.module.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.child:not(.firstChild) {
+  border-inline-start: 1px solid var(--color-blue-300);
+}
+
+.child:not(.firstChild, .lastChild) {
+  &:focus-visible {
+    &::after {
+      inset-inline: var(--focus-ring-offset-inner) var(--focus-ring-offset-inner);
```
HeartSquared, line 23: Maybe?
```suggestion
      inset-inline: var(--focus-ring-offset-inner);
```

HeartSquared (commented):
Let me know if you want to chat through any comments, or if you think any of it is too much work :) Most comments are suggestions for how to compose the CSS better (figured it's a good learning opportunity), but I'm open to descoping some of it if you have other priorities, since this is an improvement regardless.

packages/components/src/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,146 @@
...
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  width: var(--avatar-x-y);
+  height: var(--avatar-x-y);
+  box-shadow: var(--shadow-small-box-shadow);
+  border: 3px solid var(--color-white);
+}
+
+.wrapper.small {
```
HeartSquared, line 17: Haha oh no! I must not have been clear enough. Here's an example:

```suggestion
.small {
```

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,101 @@
...
+
+.default {
+  color: var(--color-purple-800);
+}
+
+.active {
+  --badge-background-color: var(--color-blue-500);
+
+  color: var(--color-white);
+}
```
HeartSquared, line 54: Hopefully this helps a little in understanding the structure?

```suggestion
}

.reversed {
  --badge-background-color: rgba(var(--color-white-rgb), 0.1);

  color: var(--color-white);
}

// TODO: .dark.reversed moved into .dark

.large {
  display: inline-flex;
  justify-content: center;
  border-radius: var(--spacing-48);
  font-size: var(--typography-data-medium-font-size);
  line-height: var(--typography-data-medium-line-height);
  letter-spacing: var(--typography-data-medium-letter-spacing);
  padding: 2px 1.875rem;
  width: 24px;
  
  // TODO:
  // &.dot { ... } can probably go here, but it does mean that .dot must go above it.
  // this .large can likely go further down in the file after the variants
}

.default {
  color: var(--color-purple-800);
}

.active {
  --badge-background-color: var(--color-blue-500);

  color: var(--color-white);
  
  &.reversed {
    --badge-background-color: var(--color-green-300);

  color: var(--color-purple-800);
  }
}
```

packages/components/src/Avatar/Avatar.module.css
Zystix: 😭

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,101 @@
...
+  font-size: var(--typography-paragraph-extra-small-font-size);
+  font-weight: var(--typography-paragraph-bold-font-weight);
+  letter-spacing: var(--typography-paragraph-extra-small-letter-spacing);
+  line-height: var(--typography-paragraph-extra-small-line-height);
+  padding: 1px var(--spacing-6);
+  min-width: 8px;
+  text-align: center;
+  background-color: var(--badge-background-color, var(--color-gray-300));
+
+  .reversed {
```
mcwinter07, line 16: Similar to what we were talking about in Avatar. Nesting the class like this will target the children of the element with the badge class:
Screenshot 2025-01-09 at 2 02 53 PM

Whereas the original SCSS reversed class is at the root the badge stylesheet:
Screenshot 2025-01-09 at 2 03 53 PM.

A handy thing to lean on here is hovering over the class name in VSCode within the stylesheet, which will bring up the prompt that shows how that class will target elements on the page

mcwinter07: We could use the `&` when nesting in CSS like this, which will increase the specificity of the styles.

Screenshot 2025-01-09 at 2 11 26 PM

However where possible we probably want to leave this un-nested so its specificity is relatively low so folks can override the styles if necessary

packages/components/src/ButtonGroup/ButtonGroup.module.css
```diff
@@ -0,0 +1,48 @@
+.buttonGroup {
```
mcwinter07, line 1: Well've also need to update ButtonGroup to use this and remove the old SCSS module for buttonGroup

packages/components/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,101 @@
...
+    background-color var(--badge-duration-timing);
+  transform-origin: center;
+  transform: scale3d(1, 1, 1);
+}
+
+.animationOn .badge {
+  transform: scale3d(1.35, 1.35, 1.35);
+}
+
+.animationOn .badge .dark {
```
mcwinter07, line 93: I think this should have the badge and dark together. 
At the moment this is looking for a element that is a child of `.badge` that has the class name `.dark` Screenshot 2025-01-13 at 10 28 44 AM

Whereas before this was looking for a `.badge` that also have the class name `.dark`: Screenshot 2025-01-13 at 10 28 49 AM

If you see no space between the `&` and `.class`, then we can assume that this is specifying both classes :)

packages/components/src/AvatarGroup/AvatarGroup.module.css
```diff
@@ -0,0 +1,61 @@
...
+
+.medium .AvatarCounter {
+  font-size: var(--typography-heading-5-font-size);
+}
+
+.small .AvatarCounter,
+.medium .AvatarCounter {
+  font-family: var(--typography-heading-5-font-family);
+  font-weight: var(--typography-heading-5-font-weight);
+  letter-spacing: var(--typography-heading-5-letter-spacing);
```
mcwinter07, line 49: One small optimisation and reduction in code is to move these styles (ln47 - 49) into the base avatarCounter class since they are pretty much the base styles for the counter. 

Because the large styles is defiend at the bottom of the doc and have a higher specificity then they should win out if its a large avatar

mcwinter07 (approved):
👍 Merge on good buddy

PR #5392, opened 2024-12-16
1:1 Future Tags hover state incorrect

Zystix:
## Why

Hovering over the Future Tabs turns your cursor into the text input variant instead of the default. This has been changed to be consistent with the current Tab component. 

https://cultureamp.atlassian.net/browse/KZN-2916

## What

Set the cursor to default on hover when tab is disabled and not disabled.

### DIFF
--- a/packages/components/src/__future__/Tabs/subcomponents/Tab/Tab.module.css
+++ b/packages/components/src/__future__/Tabs/subcomponents/Tab/Tab.module.css
@@ -31,6 +31,10 @@
     opacity: 0.3;
   }
 
+  &[data-disabled]:hover {
+    cursor: default;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/components/src/__future__/Tabs/subcomponents/Tab/Tab.module.css
```diff
@@ -31,13 +31,18 @@
     opacity: 0.3;
   }
 
+  &[data-disabled]:hover {
+    cursor: default;
```
dougmacknz, line 35: You can actually just chuck `cursor: default` on the `.tab` and you don't need to add this

dougmacknz (approved):

PR #5393, opened 2024-12-16
docs(bundler): update documentation for package-bundler

HeartSquared:
## What & Why

Improve documentation for package-bundler


packages/package-bundler/README.md
```diff
@@ -8,165 +8,52 @@ ESM code is tree-shakeable.
...
-
-- `postcss`
-- `postcss-preset-env`
-- `rollup`
-- `tslib` (peerDep of `rollup`)
-- If using Tailwind:
-  - `tailwindcss`
-
-### Required files
+If your consumers are all using Node 12+ (including for their TypeScript [`moduleResolution`](https://www.typescriptlang.org/tsconfig/#moduleResolution)), you should be able to use the [`exports`](https://nodejs.org/api/packages.html#exports) field if you have multiple entrypoints.
```
brandonwinch, line 23: I wonder if this needs to be more specific about which configuration of `moduleResolution` supports `exports`? 🤔

packages/package-bundler/README.md
```diff
@@ -8,165 +8,52 @@ ESM code is tree-shakeable.
...
-    'tailwindcss/nesting': 'postcss-nesting',
-    'tailwindcss': {},
-    'autoprefixer': {},
-    'cssnano': {},
-  },
-}
-```
-
-### Rollup
+Eg. If we want to create an entrypoint at `@kaizen/package-bundler/future`:
```
brandonwinch, line 31: Do we even let people have anything under Node v10? Aren't there security rules or something we breach if we run on a version this low? 🤔

HeartSquared: Probably 🤔 But it seems most of our repos use Node10 module resolution (KAIO currently included because we want to ensure we catch issues which our consumers might be seeing, even though it would make the most sense for us to use `bundler`). Unless everyone moves away from that, this is still relevant 🤷🏻

I suppose technically it's just an inaccurate TypeScript setting? 😂 Might be okay to remove this if we can get everyone using `bundler`?

What do you reckon would make the most sense?

brandonwinch: I think with the changes it's fine - it just wasn't clear what was meant by 'Node10' - but it's been updated to include the specific `moduleResolution` names now. 👍

brandonwinch (approved):

.changeset/pink-cars-walk.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
brandonwinch, line 2: Sweet changeset! :P

PR #5394, opened 2024-12-16
Build RAC Popover into Menu to simplify API

dougmacknz:
## Why
Currently v3 Menu requires consumers to pull `Popover` from `react-aria-components` every time.

The only downside is that it would no longer allow someone to add a Menu _not_ inside a Popover. e.g.
image

If we really wanted to support that, we could create a subcomponent and expose that, but I can't imagine any use case for this.

## What
Bake RAC `Popover` into `Menu` so to simplify the API down

### Before

```
<MenuTrigger>
  <Button />
  <Popover>
    <Menu />
  </Popover>
</MenuTrigger>
```

### After
```
<MenuTrigger>
  <Button />
  <Menu />
</MenuTrigger>
```

### DIFF
--- a/.changeset/flat-taxis-cross.md
+++ b/.changeset/flat-taxis-cross.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+Build RAC Popover into Menu to simplify API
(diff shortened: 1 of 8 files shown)


dougmacknz:
Closing because I forgot consumers may want access to `Popover` props like `isOpen`

PR #5402, opened 2024-12-17
Button v3 Disabled Styles Update

Zystix:
## Why

Disabled styles have changed for Button v3

https://cultureamp.atlassian.net/browse/KZN-2919

## What

Updated the disabled colours for the different variants of button v3

Chromatic: https://www.chromatic.com/review?appId=60a1e4a102f0cb003b5d19d6&number=5402&type=linked


packages/components/src/__actions__/Button/v3/Button.module.css
```diff
@@ -206,6 +230,12 @@
     --button-bg-color: rgba(var(--color-white-rgb), 0.1);
     --button-border-color: transparent;
   }
+
+  &:disabled {
```
mcwinter07, line 234: Small quirk of how this component is composed since we're using RAC button as the underlying technology, I'd recommend styling this instead with isDisabled class instead of the pseudo selector. This guarentee that this style will be applied when buttons internal states to isDisabled :)

mcwinter07 (dismissed):
Nice, ta for removing the old code :) Once Jess gives the thumbs up we can merge this in 🕺

mcwinter07 (approved):

PR #5403, opened 2024-12-17
chore(eslint): enable consistent-type-imports

HeartSquared:
## What

Update linting:
- Enable `@typescript-eslint/consistent-type-imports`
- Sort imports alphabetically per import


mcwinter07 (approved):
Updated eslint config looks good. I don't think I need to check all 600+ imports but the first block I looked at seem super fine so will give the ✅

PR #5407, opened 2024-12-19
refactor(codemods): clean up codemods

HeartSquared:
## What and Why

Update all codemods to use the same transformer structure and delete deprecated functions.


mcwinter07 (approved):
Nice work ✋5️⃣

PR #5423, opened 2024-12-22
Update inline notification icon color

Zystix:
## Why

Inline notification icon color (blue-400) is lighter than else where

https://cultureamp.atlassian.net/browse/KZN-2923

## What

Changed color to blue-500

### DIFF
--- a/packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.module.scss
+++ b/packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.module.scss
@@ -124,7 +124,7 @@
   border-color: var(--color-blue-400);
   color: $color-purple-800;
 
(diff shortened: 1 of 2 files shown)


dougmacknz (dismissed):

packages/components/src/Notification/subcomponents/GenericNotification/GenericNotification.module.scss
```diff
@@ -124,7 +124,7 @@
   border-color: var(--color-blue-400);
```
HeartSquared, line 124: Do you need to update this too?

Zystix: Quite possibly!

HeartSquared (approved):

PR #5425, opened 2024-12-23
"Add Filters" -> "Add filters" in AddFiltersMenu

Zystix:
## Why

AddFilterMenu's label didn't match styling convention

## What

Made the "F" in "Add Filters" lowercase.


.changeset/fluffy-candles-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+Update AddFiltersMenu's default label to match guidelines
```
dougmacknz, line 5: This can be improved:

```suggestion
FilterBar: 'Add Filter' button label changed to 'Add filter' to follow guidelines
```

Firstly, establishing that the change is in the `FilterBar` component. `AddFiltersMenu` is not something that consumers use directly so won't be familiar for them.

Secondly, this describes the change more specifically in case it affects them. E.g. in the case that someone had a test that was matching on a 'Add Filter' label, they'll easily be able to tell from this release note that they just need to adjust that to `Add filter`

dougmacknz (approved):

PR #5426, opened 2024-12-23
Simplify KaizenProvider

vietanhtran16:

## Why

We have built a shared component as part of `@cultureamp/conversations-shared-ui` which makes use of `useToastNotification`. This hook relies on `KaizenProvider` being set up by the consumer and most importantly, the consumer needs to have the exact same version as the required `peer` dependency that we specify in `@cultureamp/conversations-shared-ui`. That is not always gonna be the case though and it will be fragile hoping for consumers to always have the same version.

Due to that, I am setting up `KaizenProvider` in our component alongside with other providers to ensure that we dont have to rely on the consumers and prevent it breaking in production (it will throw error when the context is not found)
image

When I do the above though, it threw [this error](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1734046695545609?thread_ts=1733869168.781349&cid=C0189KBPM4Y) which Michael encoutered as well so decided to take a look at `KaizenProvider` and `ToastNotification`


packages/components/src/KaizenProvider/KaizenProvider.tsx
```diff
@@ -11,21 +11,19 @@ export type KaizenProviderProps = {
...
-  const [notificationsList, setNotificationsList] = useState<JSX.Element>()
 
   useEffect(() => {
     // SSR does not have a document, which is required for ToastNotificationsList.
     // Await document render before rendering the component.
     if (document !== undefined) {
-      setNotificationsList(<ToastNotificationsList />)
       setDocumentIsAvailable(true)
     }
-  }, [documentIsAvailable])
```
vietanhtran16, line 23: it doesn't need to be a dependency, nothing inside useEffect is referring to it

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
```diff
@@ -175,26 +175,27 @@ export const NoDuplicatesWithSameId: Story = {
   render: () => {
     const { addToastNotification } = useToastNotification()
 
-    useEffect(() => {
```
vietanhtran16, line 178: [this story was failing](https://github.com/cultureamp/kaizen-design-system/actions/runs/12459144607/job/34775641714) before I made this change due to the toast was not visible.

Not sure why that was the case but I think using useEffect here might have cause some issue as I suspect [addToastNotification is not being wrapped inside a `useCallBack`](https://github.com/cultureamp/kaizen-design-system/blob/2fcf1ccee78c0808630e19bafbd8330897277843/packages/components/src/Notification/ToastNotification/context/ToastNotificationContext.tsx#L44) so having it inside `useEffect` dependency will cause it to be called many times

vietanhtran16: creating a button and using `userEvent.click` is a much more efficient way to do it though

packages/components/src/Notification/ToastNotification/_docs/ToastNotification.stories.tsx
```diff
@@ -175,26 +175,27 @@ export const NoDuplicatesWithSameId: Story = {
...
-      addToastNotification({
-        id: 'id--clear-example-1',
-        title: 'First',
-        type: 'positive',
-        message: 'There should only be one notification',
-      })
-    }, [addToastNotification])
-
-    return <div>Irrelevant content</div>
+    return (
```
HeartSquared, line 178: The main issue I'd have with you changing this test is that I think some teams actually use `addToastNotification` in a useEffect in a similar fashion to this. If it fails here, then it will start throwing errors for consumers too.

I suggest you make a canary and test against `performance-review-cycles-ui` and see if everything passes or not.
Slack thread of the SSR issue when ToastNotificationProvider was introduced: https://cultureamp.slack.com/archives/C0189KBPM4Y/p1700782380804669

vietanhtran16: let me test it out. I reckon all these `addToastNotification` `updateToastNotification` `removeToastNotification` `clearToastNotifications` functions should be wrapped inside an `useCallback` tbh as they can be used as dependency in `useEffect`

vietanhtran16: actually I will need to test a canary from `1.67.0` version instead as the latest version has [some issue](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1734919522238699?thread_ts=1733869168.781349&cid=C0189KBPM4Y)

vietanhtran16: I always get this error with pnpm storybook even after pnpm reset so unable to run locally and test it
image

packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
```diff
@@ -1,36 +1,35 @@
...
+  const [toastContainer, setToastContainer] = useState<Element | null>(null)
 
-  const containers = document.querySelectorAll('[data-testid="toast-notifications-list"')
-
-  if (containers) {
-    // Remove any duplicate instances
-    // (eg. Storybook docs page has multiple stories each with their own context)
-    containers.forEach((c, i) => {
-      if (i === 0) return
-      c.remove()
```
vietanhtran16, line 17: this `.remove()` was the culprit when there are multiple KaizenProvider on the same page

mcwinter07 (approved):
Nice work @vietanhtran16! Tested out the canary in perf-review-cycle-ui too and it looks like it goes off without a hitch. Really appreciate the investigation into this

PR #5429, opened 2024-12-23
Update Chromatic Externals

Zystix:
## Why

We are transitioning to module.css files, Chromatic must be updated to detect changes in these files

## What

Added module.css files to Chromatic's externals

### DIFF
--- a/.github/workflows/chromatic-pr.yaml
+++ b/.github/workflows/chromatic-pr.yaml
@@ -58,6 +58,7 @@ jobs:
           externals: |
             **/!(*.module).scss
             packages/components/**/*.css
+            packages/components/**/*.module.css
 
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic-pr.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
           externals: |
             **/!(*.module).scss
             packages/components/**/*.css
+            packages/components/**/*.module.css
```
HeartSquared, line 61: Close! But we actually want the opposite 😂 We want Turbosnap to not be disabled when `.module.css` files are updated, because they are directly imported.

```suggestion
            **/!(*.module).css
```

dougmacknz (approved):

PR #5435, opened 2025-01-02
fix LinkButton stylesheet import path

mcwinter07:
## Why
Failed file resolution reported in in the release [thread](https://cultureamp.slack.com/archives/CETLTFG9G/p1734656795682039) by Viet and Brendan [here](https://github.com/cultureamp/calibrations-ui/actions/runs/12527289914/job/34940725042?pr=679).

This is due to the import of the shared button styles using the `~` alias import instead of a relative import, which does not get compiled correctly.

## What
Update import of the Button stylesheet to use a relative path.

### DIFF
--- a/.changeset/kind-spies-refuse.md
+++ b/.changeset/kind-spies-refuse.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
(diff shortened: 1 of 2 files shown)


mcwinter07:
Passes all the chromatic tests and is a pretty safe update so I'm going to merge this in

PR #5440, opened 2025-01-07
Update FilterSelects CSS specificity

mcwinter07:
## Why
This is a interim fix while we resolve the issue of our dist's styles.css being compiled out of order, which seems to primarily impact the Filter Select component due to the low specificity of its styles.

## What
- converts FilterSelect and Filter SCSS to CSS
- updates FilterSelect's classNameOverride to double up on specificity


packages/components/src/Filter/FilterSelect/FilterSelect.tsx
```diff
@@ -93,7 +94,7 @@ export const FilterSelect = <Option extends SelectOption = SelectOption>({
           })
         }
         onMount={setTriggerRef}
-        classNameOverride={classNameOverride}
+        classNameOverride={classnames(styles.filterSelect, classNameOverride)}
```
dougmacknz, line 97: This `filterSelect` class doesn't seem to exist in CSS

mcwinter07: Ah true, this is when I was testing to see if this something tethered this to a higher specificity. Will axe it since it's not in use

mcwinter07: Removed and good for another squiz 👀  @dougmacknz

dougmacknz (approved):

PR #5444, opened 2025-01-10
Use padding instead of margin on Content

dougmacknz:
## Why

The new responsive TabList was interfering with this. See https://cultureamp.slack.com/archives/C0189KBPM4Y/p1736373685188149

## What

Adjust the `Content` component to use padding instead of margin for adding inline spacing.


mcwinter07 (approved):
Looks good in my book. We could convert it to CSS too but will leave that as an optional update 😉

ryanseddon (approved):
Nice

PR #5450, opened 2025-01-13
Content: adjust max-width after margin to padding change shrunk width

dougmacknz:
## Why
We changed our Page Layout Content to use padding instead of margin which shrunk the content width

## What
- Increase our page layout max width to cater for the change from margin to padding
- Update TitleBlockZen to also use padding because it uses the same design token for page width


packages/design-tokens/sass/layout.scss
```diff
@@ -1,6 +1,6 @@
 /** THIS IS AN AUTOGENERATED FILE **/
 
-$layout-content-max-width: 1392px;
+$layout-content-max-width: 1536px;
 $layout-content-max-width-with-sidebar: 1080px;
 $layout-content-side-margin: 72px;
```
ryanseddon, line 5: Should this be called `$layout-content-side-padding` now?

PR #5451, opened 2025-01-14
fix MenuItem and Tab RAC children type issue

mcwinter07:

## Why
Some I noticed when working through dep update was some type issue with Children in these components. As RAC can accept children either as a function with render props or a React node there's were throwing a type error. I've made this small change to fix this, which should account for both scenarios. If we don't want this as a feature from RAC then we should update the children type :) You've probably got more context on this @dougmacknz so if you get a squiz, otherwise keen for any other feedback

Screenshot 2025-01-14 at 1 38 23 PM

## What
Update child rendering in MenuItem and Tab rc

### DIFF
--- a/.changeset/gorgeous-numbers-beg.md
+++ b/.changeset/gorgeous-numbers-beg.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+Update RC MenuItem and Tab child render to fix types and allow for RAC render props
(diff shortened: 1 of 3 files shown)


mcwinter07:
Actually this is blocking the typescript update so I might move this change into [this PR](https://github.com/cultureamp/kaizen-design-system/pull/5386) so I can merge it into rollup

PR #5452, opened 2025-01-14
update build storybook script

mcwinter07:
## Why
Storybook is currently hanging on our Build Storybook step in Buildkite. This causes our build to time out after 15m depsite looking like it should be finished (all other build logs seem to indicate that its fine). This PR is to test out if adding a success (0) or fail exit code (1) will impact this.

## What
- Add some additional echos in the script for easier logging
- Add an exit code


mcwinter07:
This was solved in [this PR](https://github.com/cultureamp/kaizen-design-system/commit/5d320983235e0f637523b0a473c37e3440fc022e)

PR #5454, opened 2025-01-16
Fix circular depenacies across multiple component types

mcwinter07:
## Why
This is part of an attempt to reduce or remove the hang time for our build in CI. After reading this post on the impact of [circular dependencies with Vite](https://provash.dev/blog/avoid-circular-dependency-in-vite/), this could be the cause of our unknown hang time.

## Finding the circular deps
To ID the problem files I ran `madge` against out components src folder which returned the following:

Screenshot 2025-01-16 at 2 46 46 PM

If you want to run this you will need install or run madge via npx then run this from the root our repo:
`npx madge --extensions ts,tsx --ts-config=tsconfig.json packages/components/src -c`

I've left the scss function unchange as I wasn't sure how this was causing a circular dep. the -mui helper are part of legacy vendor support and ideally should be removed / shifted into kaizen legacy.

## What
- Move some types when necessary to avoid circular deps


dougmacknz (approved):

packages/components/src/Brand/Brand.tsx
```diff
@@ -7,9 +7,6 @@ import styles from './Brand.module.css'
 
 type MeaningfulSVG = { 'role': 'img'; 'aria-label': string; 'alt'?: never }
 type DecorativeSVG = { 'role': 'presentation'; 'aria-label'?: never; 'alt'?: never }
```
HeartSquared, line 9: You've defined this in `types.ts` now, so better to export and import from there rather than have the duplication :)
Alt move all the types in there 😬

mcwinter07:
Closing this PR since this was pre-reorg of our package structure, if we want to tackle this probably best to do so in a clean branch

PR #5456, opened 2025-01-17
Fix Storybook margin on small desktop views

Zystix:
## Why

Fixathon. Storybook pages can lose their margin at a certain screen size.

https://cultureamp.atlassian.net/browse/KZN-2941

## What

Used media selectors to reapply a margin at the screen size in question.


.changeset/short-jobs-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@docs/storybook': patch
```
mcwinter07, line 2: Since this is just a docs update and doesn't actually need to trigger any releases we can just leave this as an empty changeset, which looks like this with no content:

```
---
---
```

This will allow it to pass our branch protection policies (that require a changeset's for everything) without needing to trigger a release step.

mcwinter07 (commented):
The change looks like it does the trick, just a minor tweak to changeset and if you could update PR title to something more declarative like "Fix storybook margin on small desktop views" that'd be sweet :)

mcwinter07 (approved):

PR #5477, opened 2025-01-20
Fix FilterSelect jumping issue

mcwinter07:
## Why
This is a second (third?) attempt at resolving the Select / FilterSelect jump issue. 

## Steps to replicate (prior to this fix)
In chrome
1. Go to a page with the FilterSelect that **does not** have a selected value (see a somewhat janky example with the stickersheets [here](https://cultureamp.design/?path=/docs/components-filterselect--docs))
2. use keyboard navigation to open the FilterSelect
3. The view should jump up to the top of the page

## Why we think this happens

This component is composed of many primitive that seek to control focus management. The suspected culprit is the React Aria listbox keyboard behavior.

We've had to try a co-opt its focus management manually but due to how it renders, seemingly its start position is above the page, causing the jump on focus.

Screenshot 2025-01-20 at 11 07 20 AM

## What
I've introduced a hook that checks for a stable y position that should ensure that the focus resolve after it is a single value

### DIFF
--- a/packages/components/src/__rc__/Select/hooks/useHasCalculatedListboxPosition.ts
+++ b/packages/components/src/__rc__/Select/hooks/useHasCalculatedListboxPosition.ts
@@ -5,7 +5,7 @@ import { useEffect, useState } from 'react'
  * This now polls to check if the element's position is stable by comparing the first and last position.
  */
 export const useHasCalculatedListboxPosition = (ref: React.RefObject<HTMLElement>): boolean => {
-  const [isStable, setIsStable] = useState(false)
+  const [hasStablePosition, setHasStablePosition] = useState(false)
   const [lastYPosition, setLastYPosition] = useState<number | null>(null)
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


mcwinter07:
Validated this would in survey-reporting-ui with the canary `0.0.0-canary-fix-filter-select-jump-issue-useHasStableYPosition-20250120225815`, will check with the two teams that reported this to see if it works ok for them

packages/components/src/__rc__/Select/hooks/useHasCalculatedListboxPosition.ts
```diff
@@ -0,0 +1,31 @@
...
+  const [lastYPosition, setLastYPosition] = useState<number | null>(null)
+
+  useEffect(() => {
+    const checkPosition = (): void => {
+      if (ref.current) {
+        const { y } = ref.current.getBoundingClientRect()
+        if (lastYPosition === null) {
+          setLastYPosition(y)
+        } else if (y === lastYPosition && y >= 0) {
+          setIsStable(true)
```
dougmacknz, line 18: Would it make sense, and is possible to clear the interval here as well?

mcwinter07: Nice spot on this, the interval was re-running constantly while the popover was open. I've updated the interval to call itself only while is position is not stable, so we're going from this:

Screenshot 2025-01-30 at 9 17 39 am

To this:
Screenshot 2025-01-30 at 9 29 34 am

mcwinter07:
Closing old PR

PR #5492, opened 2025-02-02
feat: set exports field `@kaizen/components`

brandonwinch:
## Why
The `@kaizen/components` package currently doesn't have any of the `export` fields set for what it outputs. Historically this has been okay, but when Vitest tries to consume it, it defaults to consuming CJS files since these aren't set. More details in the 🧵 here: https://cultureamp.slack.com/archives/C2ZNME08P/p1738299498099589

## What
I've set the `export` fields to mirror what's in each `package.json` file. I've also tested this via a canary.


ryanseddon (approved):
lgtm

PR #5498, opened 2025-02-03
Ryan/hanging storybook builds

ryanseddon:
## Why

Our buildkite builds would sometimes just hang and timeout due to storybook never exiting the `build:docs` task.

## What

The culprit was `react-docgen-typescript` as mentioned in this [issue](https://github.com/storybookjs/storybook/issues/30424) commenting out the entire `typescript` block in `docs/.storybook/main.ts` no longer made `build:docs` hang and it would exit correctly.

Digging into the storbook src it mentions that in v8 they've moved from `react-docgen-typescript` to `react-docgen` and that is the recommendation going forward.

https://github.com/storybookjs/storybook/blob/ba47d3e3aea378a2da911219b78e36e76bfff930/code/lib/cli-storybook/src/automigrate/fixes/react-docgen.ts#L49-L54

So I just edited the `reactDocgen` key to use that and the build no longer hung!

### DIFF
--- a/.changeset/stale-toys-brake.md
+++ b/.changeset/stale-toys-brake.md
@@ -2,4 +2,4 @@
 '@docs/storybook': patch
 ---
 
-Move from react-docgen-typescript to react-docgen
+Overriding transient dependency @storybook/react-vite package to fix issue
--- a/.changeset/stale-toys-brake.md
+++ b/.changeset/stale-toys-brake.md
@@ -1,0 +1,5 @@
+---
+'@docs/storybook': patch
+---
+
+Move from react-docgen-typescript to react-docgen
--- a/package.json
+++ b/package.json
@@ -104,6 +104,9 @@
         "react-dom": "18",
         "eslint": "9"
       }
+    },
+    "overrides": {
+      "@joshwooding/vite-plugin-react-docgen-typescript": "0.5.0"
     }
   }
 }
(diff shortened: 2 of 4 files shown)


mcwinter07 (commented):
Interesting, I just had a squiz at comparing the branch preview for button v3 against the current cultureamp.design of the same page and it looks like there's some props not being pulled into the controls table (ie: isPending and pendingLabel). Not sure if that means we'll have to be more explicit about what going in there but I can see any of the pending props.

Also some minor differences in types?
Screenshot 2025-02-03 at 5 05 24 pm

ryanseddon:
Good catch, comparing this to prod I'm seeing lots of differences where the types are missing or heavily simplified compared to what the ts compiler extracts out. Might have to go back to the drawing board.

package.json
```diff
@@ -104,6 +104,9 @@
         "react-dom": "18",
         "eslint": "9"
       }
+    },
+    "overrides": {
+      "@joshwooding/vite-plugin-react-docgen-typescript": "0.5.0"
```
ryanseddon, line 109: This is a temporary fix until storybook updates the dependency in their react-vite framework package.

.changeset/stale-toys-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@docs/storybook': patch
+---
+
+Move from react-docgen-typescript to react-docgen
```
mcwinter07, line 5: @ryanseddon should we update this to just say we're just overriding the transitive dep in storybook?

mcwinter07 (approved):

PR #5499, opened 2025-02-03
feat: export dist styles from kaio

brandonwinch:
This is a follow-up from my earlier work here: https://github.com/cultureamp/kaizen-design-system/pull/5492

I had missed that consumers are also importing the built out CSS. I think ideally we don't want to expose the `./dist` directory, but in order to avoid a breaking change, I figured it would be best to keep the same path for now.

I've tested this via a Canary and the problem looks to be resolved.

### DIFF
--- a/.changeset/unlucky-shrimps-relate.md
+++ b/.changeset/unlucky-shrimps-relate.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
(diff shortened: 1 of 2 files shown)


ryanseddon:
I think exposing the `dist` dir for this instance makes sense and is a necessary evil.

ryanseddon (approved):

PR #5501, opened 2025-02-04
feat: set style condition in exports so styles can be imported from a CSS file.

brandonwinch:
A follow-up to https://github.com/cultureamp/kaizen-design-system/pull/5499. We'll get there eventually!

While it's not encouraged, some of our packages & apps are importing the `dist/styles.css` file from a `.css` file. In these cases, we need to set the `style` condition for that to resolve properly.

I've tested a canary and it looks sweet. 🚀


ryanseddon (approved):

lisatassone:
Legend! Thanks for looking into this.

PR #5502, opened 2025-02-04
Link Component

Zystix:
## Why

Creating a new Link component.

https://cultureamp.atlassian.net/browse/KZN-2593

## What

Adds the Link component, storybook tests, sticker sheets, API spec and usage guidelines.


docs/pages/client-side-routing.mdx
```diff
@@ -0,0 +1,100 @@
...
+    // index 2 is the types for the pages routerOptions
+    routerOptions: NonNullable<Parameters<NextRouter['push']>[2]>
+  }
+}
+
+function App({ Component, pageProps, router }: AppProps) {
+  return (
+    <FrontendServices {...config}>
+      {/* application code */}
+      <RacRouterProvider navigate={(href, opts) => router.push(href, undefined, opts)}>
```
jakubriedl, line 33: can we make this provider part of [next-services](https://github.com/cultureamp/frontend-foundations/tree/master/packages/next-services) so it works for the consuming teams out of the box and without extra config?

mcwinter07: Hey @jakubriedl did you mean before this release goes out? 

I certainly think that makes sense as a future / subsequent goal, but don't think it should block this specific release since the component is agnostic to the routing solution and standard hrefs will work out of the box.

Potentially a good opportunity for @Zystix once this is done (noting we only have him till the end of Feb), but will also raise with the team to see when we can prioritize the work

jakubriedl: yeah totally agree this shouldn't be blocker. But I'd add it there at the same time as it's just small change and write the documentation in the expects it to be there. It'll save you a lot of explaining to teams what they have to do, just say in the docs "have up-to-date next-services" and you can skip this whole section. Easier for you, and for the consumer as well

docs/pages/client-side-routing.mdx
```diff
@@ -0,0 +1,100 @@
+<Meta title="Guides/Client Side Routing" />
```
mcwinter07, line 1: Think you might be missing some of the imports at the top of the document on this page 
Screenshot 2025-02-06 at 2 57 58 pm

packages/components/src/Link/Link.module.css
```diff
@@ -0,0 +1,106 @@
...
+
+[class*='extra-small'] .isInline > *,
+.extra-small {
+  --link-font-size: var(--typography-paragraph-extra-small-font-size);
+  --link-line-height: var(--typography-paragraph-extra-small-line-height);
+  --icon-font-size: 0.875rem;
+  --text-icon-gap: var(--spacing-2);
+}
+
+[class^='small'] .isInline > *,
```
ryanseddon, line 68: What's the reason for the attribute selector use here? Can it just be `.small`

packages/components/src/Link/_docs/Link.doc.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+export const Playground: Story = {}
+
+export const LinkVariants: Story = {
+  render: (props) => (
+    <>
+      <Link {...props} variant="primary" />
+      <br />
+      <Link {...props} variant="secondary" />
+      <br />
```
ryanseddon, line 43: Can probably delete this br

packages/components/src/Link/_docs/Link.stickersheet.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+            <StickerSheet.Row key={size + variant} header={`${variant} (${size})`}>
+              <Link
+                variant={variant}
+                size={size}
+                href={href}
+                isUnderlined={false}
+                isReversed={isReversed ? isReversed : false}
+                isInline={false}
+                icon={<Icon name="add" isPresentational />}
+                iconPosition={'start'}
```
ryanseddon, line 40: Don't need the curly brackets if it's just a string can just do `iconPosition="start"` same for the rest of the examples below.

packages/components/src/Link/Link.module.css
Zystix: Because of the hashing of classes it can't "see" the .small class of the <Text> component its wrapped in when inline

packages/components/src/Link/Link.module.css
```diff
@@ -0,0 +1,106 @@
+.link {
+  color: var(--link-text-color, var(--color-blue-500));
+  font-family: var(--typography-paragraph-body-font-family);
+  font-size: var(--link-font-size);
+  line-height: var(--link-line-height);
+  font-weight: 400;
+  text-decoration: none;
+  position: relative;
```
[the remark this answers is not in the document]
Zystix: Probably an experiment from trying to get the icon positioned correctly. Removed :)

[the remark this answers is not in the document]
Zystix: nvm, it makes the focus outline work correctly

packages/components/src/Link/Link.module.css
```diff
@@ -0,0 +1,106 @@
...
+  outline: 0;
+}
+
+.link[data-focus-visible]::after {
```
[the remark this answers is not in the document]
Zystix: Doesn't seem so

packages/components/src/Link/Link.module.css
ryanseddon: The css module will apply the correct hash as part of the postcss plugin so you can just reference it as a normal class

packages/components/src/Link/Link.module.css
```diff
@@ -0,0 +1,106 @@
...
+  background: transparent;
+  border-color: var(--color-blue-500);
+  border-radius: 0;
+  border-width: var(--border-focus-ring-border-width);
+  border-style: var(--border-focus-ring-border-style);
+  inset: calc(-1 * (var(--border-focus-ring-border-width) * 2) - 1px);
+}
+
+.isUnderlined {
+  border-bottom: var(--spacing-1) solid var(--link-text-color, var(--color-blue-500));
```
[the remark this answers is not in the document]
mcwinter07: If I recall the original issue was it created a gap between the icon and the text because they were separate elements, not sure if this is still the case 🤔

packages/components/src/Link/Link.module.css
Zystix: Didn't appear to work when we tried it? Does it look for the hash for link.module.css file?

ryanseddon: Another note is that the `.[size] .isInline > *` selector is never applied due to your condition of not applying the styles when isInline is false.

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const isReversedVariant = isReversed ?? shouldUseReverse
+    const childIsFunction = typeof children === 'function'
+
+    return (
+      <RACLink
+        ref={ref}
+        className={mergeClassNames(
+          styles.link,
+          isDisabled && styles.isDisabled,
+          !isInline && styles[size],
```
ryanseddon, line 68: This is the condition I'm referring to, your class selector is never applied due to the fact that the size classes aren't added when the component has isInline set to true

packages/components/src/Link/Link.module.css
Zystix: Using the syntax we currently have appears to work in grabbing the parent components class

packages/components/src/Link/Link.module.css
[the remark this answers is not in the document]
Zystix: Indeed, applying the text decoration to the anchor tag only underlines the text. Is it worth trying to get this working instead of the border we already have?

packages/components/src/Link/Link.tsx
Zystix: I think this is the desired functionality, we still allow the size to be passed cos we couldn't get the discriminated union to behave. Would be good to get that sorted though

packages/components/src/Link/Link.module.css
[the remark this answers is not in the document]
mcwinter07: Random thought on how we could achieve this if that's still the case / we didn't want to use border bottom - we could potentially give the icon a pseudo element before or after (depending on iconPosition) and use letter spacing to simulate a continuous line, ie:

```
 .icon::before {
    letter-spacing: 15px;
    content: "\00a0";
    text-decoration: underline;
  }
  ```

Then again that may be more work that border bottom 🤷

File: packages/components/src/Link/Link.module.css [hunk 10]
```diff
@@ -0,0 +1,106 @@
+.link {
+  color: var(--link-text-color, var(--color-blue-500));
+  font-family: var(--typography-paragraph-body-font-family);
+  font-size: var(--link-font-size);
+  line-height: var(--link-line-height);
+  font-weight: 400;
```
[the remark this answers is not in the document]
Zystix: @mcwinter07 What do you think, is it possible that css var changes in future but we still want link to weight 400?

packages/components/src/Link/_docs/Link--api-specification.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+While client side routing is possible, the `Link` is agnostic to the routing technology chosen. Refer to our general set up guide to get started with [client side routing](#client-side-routing).
+
+#### Opening new tabs and accessibility considerations
+
+The general recommendation is to limit the number of links that open a new tab or window on a single page. While there are valid scenarios that can help avoid loss of data and or progress, as with links in forms, opening new tabs can be disorienting for users - especially for those who have difficulty perceiving visual content.
+
+In order to provide advance warning to all users, it is recommended that links using `target="_blank"` be accompanied by a visual indicator and audible warning. As shown in the following example, additional context can be provided via a visually hidden element within the `children` of the component.
+
+<Canvas of={exampleStories.LinkOpensInNewTab} />
```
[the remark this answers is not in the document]
Zystix: @mcwinter07 Do you know what this visual indicator and audible warning would be?

ryanseddon:
Something else to consider (I don't know if both a relevant anymore) but you might want to look at seeing if adding `rel="noreferrer"` is something we want on any Link elements that go off to an external url. 

Seems having `target="_blank"` no automatically does the `rel="noopener"` behaviour.

https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener

packages/components/src/Link/Link.module.css
ryanseddon: Right I see, I missed the usage of `<Text>` passing down the size. That means that the class below is never applied as its hash is local to the Link component and not the Text component.

You can see in the below screenshot that the `.extra-small` class is converted to `._extra-small_1sik3_61` but the one applied to the text element is `._extra-small_1fj6b_35` as the size is passed down from Text -> Link.

packages/components/src/Link/_docs/Link--api-specification.mdx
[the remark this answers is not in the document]
mcwinter07: @Zystix I reckon we use the same as in the guidance for LinkButton, I believe it the "open_as_new" icon 
Screenshot 2025-02-11 at 9 22 38 am

packages/components/src/Link/Link.module.css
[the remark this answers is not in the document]
mcwinter07: For now it's probably best that we treat the default as the paragraph body so no qualms with updating this to use the token. If we were to change what that was we'd probably want it to roll down to the Link any way :)

packages/components/src/Link/Link.module.css
Zystix: This is the desired functionality, this is how we're achieving grabbing the parent Text's size class when the Link `isInline`. Is there anything you'd change here?

packages/components/src/Link/_docs/Link.doc.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+export const LinkWithIconEnd: Story = {
+  render: (props) => (
+    <Link {...props} icon={<Icon name="add" isPresentational />} iconPosition="end" />
+  ),
+}
+
+export const LinkOpensInNewTab: Story = {
+  render: (props) => (
+    <Link {...props} target="_blank" icon={<ExternalLinkIcon role="presentation" />} />
```
mcwinter07, line 68: I'd also recommend we use the `end` icon position with this one

packages/components/src/Link/_docs/Link.doc.stories.tsx
```diff
@@ -0,0 +1,184 @@
+import React from 'react'
+import { type Meta, type StoryObj } from '@storybook/react'
+import { ExternalLinkIcon, FlagOffIcon } from '~components/Icon'
```
mcwinter07, line 3: We need to replace with these usages with the RC icon Icon component - this is why the external link icons look larger than they should be

packages/components/src/Link/Link.module.css
ryanseddon: Yeah for all the selectors there you can drop the comma separated class e.g. 

```css
[class*='extra-small'] .isInline > *,
.extra-small <-- don't need this 
```

Just do 
```css
[class*='extra-small'] .isInline > * {
  ...
}
```

ryanseddon: I'll also add that the`class^='small'` will never actually match (at least not in strybook) as the class begins with a `_` so it nver finds a class starting with `small`

packages/components/src/Link/Link.module.css
```diff
@@ -0,0 +1,106 @@
...
+.secondary[data-pressed] {
+  --link-text-color: var(--color-black);
+}
+
+[class*='extra-small'] .isInline > *,
+.extra-small {
+  --link-font-size: var(--typography-paragraph-extra-small-font-size);
+  --link-line-height: var(--typography-paragraph-extra-small-line-height);
+  --icon-font-size: 0.875rem;
+  --text-icon-gap: var(--spacing-2);
```
ryanseddon, line 65: This doesn't work the way you expect it as the variable is defined within the selector range so the --text-icon-gap is always the fallback 0.5rem

Here you can see the variable is not defined in the scope of where it's used
Screenshot 2025-02-11 at 11 37 10 am

Here it's defined (it just happens to be the same value as the fallback)
Screenshot 2025-02-11 at 11 37 24 am

packages/components/src/Link/_docs/Link.doc.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+export const LinkOpensInNewTab: Story = {
+  render: (props) => (
+    <Link {...props} target="_blank" icon={<ExternalLinkIcon role="presentation" />} />
+  ),
+}
+
+export const WithText: Story = {
+  render: (props) => (
+    <>
+      <Text variant="extra-small">
```
ryanseddon, line 75: Can we hook up the storybook controls to this value so we can adjust it within the story and see how it changes? At the moment the controls don't do anything

Screenshot 2025-02-11 at 11 42 27 am

packages/components/src/Link/Link.module.css
ryanseddon: I feel like this is a bit of code smell relying on a parent component class to make changes, can we do this within code and know the size being passed down?

packages/components/src/Link/Link.module.css
ryanseddon: In fact all of these variables do not have any effect as they are scoped to this selector and the variables are used up the selector scope

packages/components/src/Link/Link.module.css
mcwinter07: We can inherit the text size, the issue I think lies more in the icon sizing and spacing between the text and icon. We initially thought we might be able to leverage `em` to calculate a ratio based on the size but map to a cleanly between each font size.

Or were you thinking more like JS to look at the parent node and get the computed style to determine a size based off that?

packages/components/src/Link/Link.module.css
Zystix: Screenshot 2025-02-11 at 1 21 40 PM

It appears to work as intended in the stickersheet? Thats the body size

Zystix: Ah, doesn't work when `isInline`, I'm pretty sure it used to so I might have broken something when cleaning up.

packages/components/src/Link/_docs/Link.doc.stories.tsx
Zystix: Size doesn't do anything if the Link `isInline`, which is bad I know. It would be ideal if we didn't have Size as a prop if `isInline` but we couldn't get the discriminated union to work. We added sample paragraphs for every Link size though.

packages/components/src/Link/Link.module.css
ryanseddon: Thinking this over I don't think there is a better way to approach this.

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Link as RACLink, type LinkProps as RACLinkProps } from 'react-aria-components'
+import { type TextProps } from '~components/Text'
+import { mergeClassNames } from '~components/utils/mergeClassNames'
+import { LinkContent } from './subcomponents/LinkContent'
+import styles from './Link.module.css'
+
+export type LinkProps = (UnderlinedLink | NonUnderlinedLink) & (InlineLink | NonInlineLink)
+
+type BaseLinkProps = {
+  variant?: 'primary' | 'secondary' | 'white'
```
mcwinter07, line 11: Can we add some type descriptions for our types with the default values where relevant?

Doesn't have to be too long, but something like this will provide some good feedback via intellisense 😉 
```
  /** Controls the visual style of a link. @default 'primary' */
```

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,80 @@
...
+  iconPosition?: 'start' | 'end'
+} & Omit<RACLinkProps, 'children'> & {
+    /** Used as the label for the Link. */
+    children: RACLinkProps['children']
+  }
+
+export type UnderlinedLink = BaseLinkProps & {
```
mcwinter07, line 20: One thing I noticed is that we're create a new intersection with `BaseLinkProps` in each of these types, then in the `LinkProps` we're combining all of these together again.

Screenshot 2025-02-17 at 2 03 14 pm

 I'm not sure the potential impact on the final type but we might be able to reduce the times that `BaseLinkProps` is added into our types and intersect it in the `LinkProps` instead

packages/components/src/Link/subcomponents/LinkContent.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export type LinkContentProps = {
+  children: ReactNode
+  icon?: JSX.Element
+  iconPosition?: 'start' | 'end'
+  isUnderlined: boolean
+}
+
+const LinkIcon = ({ icon }: { icon: JSX.Element }): JSX.Element => (
+  <span className={mergeClassNames(styles.icon)}>{icon}</span>
```
mcwinter07, line 13: We can probably remove the `mergeClassNames` since we have one class :)

docs/pages/client-side-routing.mdx
```diff
@@ -0,0 +1,102 @@
...
+      <LinkButton href={`${router.pathname}/path-2`} routerOptions={{ scroll: false }}>
+        Link with routerOptions
+      </LinkButton>
+    </>
+  )
+}
+```
+
+Additional config options for Next.js can be found in the React Aria's documentation on the [RouterProvider](https://react-spectrum.adobe.com/react-aria/routing.html#nextjs), including the alternative setup for the [App router](https://react-spectrum.adobe.com/react-aria/routing.html##app-router).
+
```
mcwinter07, line 68: Can we also add a small note here that a fast follow to this will be a contribution to the `next-services` package, something like this:

> A fast follow for this work is to contribute a `RacRouterProvider` to the `next-services` packages to reduce the custom code required by consumers.

Just in case we don't have time in your last week to do this 😉

packages/components/src/Link/_docs/Link.doc.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { Icon } from '~components/__rc__/Icon'
+import { Link } from '../Link'
+
+const meta = {
+  title: 'Components/Link',
+  component: Link,
+  args: {
+    children: 'Link',
+    href: 'https://www.google.com',
+    variant: 'primary',
```
mcwinter07, line 13: since `variant`, `isUnderlined` and `isInline` all have default values, we can probably remove them here so they don't show up in our stories code snippets :) ie: 

### With default args in the meta (current)

Screenshot 2025-02-24 at 3 20 21 pm

### No default args in the meta

Screenshot 2025-02-24 at 3 14 30 pm

mcwinter07 (commented):
You'll also need to export the Link from our barrel file to ensure that it gets bundled and distributed  😉 

Screenshot 2025-02-24 at 4 01 24 pm

docs/pages/client-side-routing.mdx
Zystix: Does this still work since this doco isn't directly connected to link? We could add it in the Link doco after we link to this page?

mcwinter07: It would be the same config for both Link and LinkButton since they use the same underlying React Aria Link component :)

mcwinter07 (dismissed):

devbox.lock
```diff
@@ -1,6 +1,9 @@
 {
   "lockfile_version": "1",
   "packages": {
+    "github:NixOS/nixpkgs/nixpkgs-unstable": {
```
ryanseddon, line 4: We'll have to make sure everyone is on 0.14.0 of devbox as I don't know what happens if you haven't updated?

ryanseddon (commented):
Just some spelling fixes

ryanseddon (approved):

mcwinter07 (approved):

PR #5503, opened 2025-02-04
Kzn 2885/refactor select to use rac

mcwinter07:
## Why
I've been unable to resolve the focus jumping issue with the Select / FilterSelect in its current due to the many layers of focus management without obtuse workaround that are just getting more brittle. Given this, I have attempted to just refactor and try to simplify the Select and FilterSelect to use RAC Popover and Diolgue components so we don't need to rely on FocusScope or the FocusOn package.

As it currently stands there is similar functionality achieve here with minimal API changes that would likely impact consumers. There is probably more that could be done in terms of stripping out the useSelect hook entirely, but at that point we might as well hold off until the Picker rebuild.

## Components this would change
- Select (future)
- Filter
  - FilterSelect (uses Filter)
  - FilterDateRangePicker (uses Filter)
  - FilterDatePicker (uses Filter)

## Current state vs differences
Currently we have a couple of tests failing that warrants further investigate.
- defaultOpen doesn't not shift focus
- if `isNonModal` is used then click outside  does not close the modal. I suspect we need to use `shouldCloseOnInteractOutside` to manually close it (unvalidated)

## differences
- RAC default behaviour for Popovers and dialogs is to trap focus and disable scroll until a selection of interaction outside of the element is done.
  -  This can be updated via `isNonModal` but creates the seems to need to handle click outside interactions to close the popover
  - scrolling will close the modal

## What
- Remove the use of custom Overlay, FocusScope / FocusOn components in various Selects


packages/components/src/__rc__/Select/Select.tsx
```diff
@@ -173,13 +177,53 @@ export const Select = <Option extends SelectOption = SelectOption>({
...
+              <div ref={popoverContentsRef}>
+                <SelectProvider<Option> state={state}>
+                  <SelectPopoverContents menuProps={menuProps}>{children}</SelectPopoverContents>
+                </SelectProvider>
+              </div>
+            </Dialog>
+          </RacPopover>
+        )}
+      </div>
+      {/* {state.isOpen && (
```
mcwinter07, line 208: This is an alternative that I have briefly looked at - which would be to handle this like we do the MultiSelect

packages/components/src/__rc__/Select/hooks/useHasCalculatedListboxPosition.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState } from 'react'
+
+/** This is a helper util to resolve the focus jumping issue with future Select and FilterSelect.
+ * Due to the floating element's position starting as a negative value on render and then jumping to the correct position, this caused the focus to jump to the top of the page.
+ * This now polls to check if the element's position is stable by comparing the first and last position.
+ */
+export const useHasCalculatedListboxPosition = (ref: React.RefObject<HTMLElement>): boolean => {
```
mcwinter07, line 7: This will no longer be needed if the refactor is hunky dory :)

mcwinter07:
Hmmm dang, tried to work out the using the `isNonModal` prop to allow for scrolling, but that directly inhibits the ability to close the modal when a user click outside anywhere on the page... I found a very irate engineer in this thread that shared a very similar [frustration that this is intended](https://github.com/adobe/react-spectrum/issues/6819)

ryanseddon:
Seems odd to lock scrolling when native selects don't do this nor does it say anything about it in the aria authoring practices even the close on scroll isn't a thing.

mcwinter07:
Closing this PR since this solution didn't pan out

PR #5507, opened 2025-02-06
removes disabled hover state for ToggleSwitchField

Kitty-Al:
## Why

Bug fix for `ToggleSwitchField` component
https://cultureamp.atlassian.net/browse/KZN-2988

## What

- removes colour change on disabled hover state for `ToggleSwitchField`
- updates the cursor to be consistently `default` and not a mix of `default` and `text` over the `ToggleSwitch`
- applies disabled state opacity to `ToggleSwitch` 

Video of hover state before:
https://www.loom.com/share/79edbb648b6f4d1dbb8353f374830784?sid=01bd2fb2-1834-479f-b8c8-556a04b02511

Video of hover state after: 
https://www.loom.com/share/4d0ee52c50b94589895f57fd56c46250?sid=72a576b4-7dfa-441e-bb3e-01dcfead71bc

Before:
old

After:
new

### DIFF
--- a/packages/components/src/ToggleSwitch/ToggleSwitch/ToggleSwitch.module.scss
+++ b/packages/components/src/ToggleSwitch/ToggleSwitch/ToggleSwitch.module.scss
@@ -125,6 +125,11 @@ $focus-ring-offset: 1px;
   }
 }
 
+// When the ToggleSwitch is used as part of the ToggleSwitchField, the disabled state opacity is set on the Label component
+label .disabled.track {
+  opacity: 100%;
+}
+
 .disabled.track {
   opacity: 30%;
 }
--- a/packages/components/src/ToggleSwitch/ToggleSwitch/ToggleSwitch.module.scss
+++ b/packages/components/src/ToggleSwitch/ToggleSwitch/ToggleSwitch.module.scss
@@ -36,7 +36,7 @@ $focus-ring-offset: 1px;
   }
 
   .checkbox:not(.disabled) + &:hover,
-  label:hover .checkbox + & {
+  label:hover .checkbox:not(.disabled) + & {
     background-color: $color-purple-500;
     background-color: $color-gray-600;
   }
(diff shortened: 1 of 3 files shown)


packages/components/src/ToggleSwitch/ToggleSwitch/ToggleSwitch.module.scss
```diff
@@ -118,10 +119,18 @@ $focus-ring-offset: 1px;
...
     .checkbox:not(.disabled) + .track:hover,
-    label:hover & .checkbox + .track {
+    label:hover & .checkbox:not(.disabled) + .track {
       background-color: rgba($color-white-rgb, 0.65);
     }
   }
 }
 
+label .disabled.track {
+  opacity: 100%;
```
dougmacknz, line 129: If you wanted to avoid this, because it's a bit confusing why you're having to set it to 100% from the scope of this file, you could just remove the `disabled` prop being passed to the `Label` component from `ToggleSwitchField.tsx`

The `disabled` prop on `Label` doesn't do anything other than set opacity to 30%

Kitty-Al: Thanks for this, I've removed this styling and made it so that the `Label` component handles the styling of the `ToggleSwitchField`. Not passing the `disabled` prop to the `ToggleSwitch` when it's used in `ToggleSwitchField`

packages/components/src/ToggleSwitch/ToggleSwitchField/ToggleSwitchField.tsx
```diff
@@ -59,7 +59,6 @@ export const ToggleSwitchField = ({
         >
           <ToggleSwitch
             id={`${id}-field-toggle`}
-            disabled={disabled}
```
dougmacknz, line 62: This one will be required because it gets passed through to the `input`

It's the `disabled` prop on `<Label>` above that you'll want to get rid of

dougmacknz (approved):

PR #5517, opened 2025-02-10
Adds share icon to default set

Kitty-Al:
## Why
https://cultureamp.atlassian.net/browse/KZN-2953

## What

Exposes the share icon via the default set


ryanseddon (approved):
Just need to merge main and we can release this with my PR

PR #5541, opened 2025-02-23
Create next entry point and update docs

dougmacknz:
- Introduce `next` entry point
- Update docs to remove references to `future` and `v1/v2/v3` and suggest importing from `next`
- Add (deprecated) to current components with a next replacement, with a notice on the docs pages
- Make the deprecated and next versions the same level in Storybook side nav, rather than next being nested inside current.
- Move the ReversedColors util to the main entry point, keeping backwards support for now
- Some extra curricular work to organise form components a bit better in the Storybook sidebar


mcwinter07 (commented):
Wondering what your thought are on labelling primitives for thing that don't have a docs page but can be used like generic tile or modal?
**Generic Tile**
Screenshot 2025-02-26 at 9 17 08 am 
**Generic Modal**
025-02-26 at 9 22 29 am

dougmacknz:
Yeah I reckon that's a good idea, I've marked those as primitive now as well

mcwinter07 (approved):
Nice work - looking forward to the consolidated entry points! 🚢 :shipit:

PR #5552, opened 2025-03-03
React 19 upgrade

vietanhtran16:

## Why

To prepare for our [Nextjs 15 and React 19 upgrade](https://cultureamp.slack.com/archives/C2ZNME08P/p1737618033802909), I created this PR to test it out. My findings has been documented [here](https://cultureamp.slack.com/archives/C2ZNME08P/p1741140687795789?thread_ts=1737618033.802909&cid=C2ZNME08P)

## What

- Upgrade to react and react-dom 19 for devDependencies
- Allow react 19 and react dom 19 as peer dependencies
- Released as canary and [tested in conversations-ui](https://github.com/cultureamp/conversations-ui/pull/1658)


packages/components/package.json
```diff
@@ -115,7 +115,7 @@
   },
   "dependencies": {
     "@floating-ui/react-dom": "^2.1.2",
-    "@headlessui/react": "<=1.5.0",
```
vietanhtran16, line 118: I am not sure why this was restricted to 1.5.0 or lower. Checked the commit history but didnt see any comments about it

packages/components/src/AvatarGroup/AvatarGroup.tsx
```diff
@@ -29,7 +29,7 @@ export type AvatarGroupProps = {
   avatars: AvatarList
 } & OverrideClassName<HTMLAttributes<HTMLUListElement>>
 
-const renderCounter = (remainingAvatars: number): JSX.Element | void => {
```
vietanhtran16, line 32: void is a not a valid type and causing type error with react 19

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -136,7 +136,7 @@ export const GenericModal = ({
       afterEnter={onAfterEnterHandler}
       afterLeave={onAfterLeaveHandler}
       data-generic-modal-transition-wrapper
-      onClick={(e: React.MouseEvent): void => e.stopPropagation()}
+      // onClick={(e: React.MouseEvent): void => e.stopPropagation()}
```
vietanhtran16, line 139: this is no longer a props in newer version of headless-ui/react. I think worth confirming whether it is still needed

vietanhtran16:
A bunch of tests are failing. I reckon it might be because of mismatch react version from possibly storybook? I also noticed react and react dom as dependency in the root package.json - are they needed there?

packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
```diff
@@ -43,6 +43,7 @@ $cell-border-radius: 3px;
   justify-content: space-between;
   width: 100%;
   color: $color-purple-800;
+  top: -0.25rem;
```
ryanseddon, line 46: This is needed as the original rdp nav would actually jump out of its containing block.

v8 on the left and this branch on the right, this fixes that. Due to how the markup has changed the cleanest way was to pull the nav outside of it's containing block again.

Screenshot 2025-05-14 at 3 43 06 pm

mcwinter07 (commented):
Changes look good but given how much the deps update for react day picker impact the behavior of FilterDateRangePicker and DateRangePicker, can we update or add another changeset with a patch to call out this update out specifically. It'll help if we ever need to refer to the changelog

ryanseddon:
@mcwinter07 absolutely was planning to get everything settled and then do a proper changeset

mcwinter07 (approved):

PR #5555, opened 2025-03-04
Modal: Update directional CSS to logical properties/values for RTL support

dougmacknz:
Updates all hardcoded directional properties and values to logical properties and values for RTL support.

Also removes `localeDirection` prop from `InputEditModal` which only has legacy usages in murmur, so should've been gotten rid of in the move to KAIO.


ryanseddon:
We should add an eslint check here to recommend or even enforce logical properties

ryanseddon (approved):

PR #5558, opened 2025-03-06
Create Upcoming Releases page

dougmacknz:


docs/pages/upcoming-releases.mdx
```diff
@@ -0,0 +1,343 @@
...
+      <td>`ModalFooter`</td>
+      <td>☝️</td>
+      <td>-</td>
+      <td>☝️</td>
+    </tr>
+    <tr>
+      <td colSpan={4}>Others</td>
+    </tr>
+  </tbody>
+</table>
```
mcwinter07, line 262: Just a note, I may have some more props that will be deprecated with all this button internal work. I reckon it'd probs be best to hold until the following release give the shorter lead time but happy to hear thought :)

mcwinter07: Also on that note, do you think it'll be ok re: the removal of v1 Button / IconButton to only remove it as a exported component? We may need to hang onto this until all internal uses are gone in its entirety.

mcwinter07 (approved):
Looks good to me. I've left some comments and tangentle quandaries in the `Deprecated props being removed` relating to internal Button work, but more me seeking an opinion 😉

docs/pages/upcoming-releases.mdx
dougmacknz: Oops just realised I forgot to add the 'others' here

dougmacknz: >  I may have some more props that will be deprecated with all this button internal work

Yeah I reckon any Button stuff (e.g. `reversed`) should wait for the next cycle so that this one can just focus on getting people over to the new button.

> Also on that note, do you think it'll be ok re: the removal of v1 Button / IconButton to only remove it as a exported component? 

Ya that seems like it would be fine as far as I can tell

mcwinter07 (approved):

PR #5569, opened 2025-03-17
Deprecate actions prop, add actionsSlot and create supporting codemod

mcwinter07:
## Why
This is part of the work to phase out the old v1 button component internally.

While I initially looked into a remap so props did not need to be change, this presented a risk if we were to try to roll this out to all component. Instead I opted to create a new prop that can take JSX elements and provide a codemode that get user _most_ of the way there. Similar to the actual Button codemod it will still create scenario where a type error will occur, which can manually be updated by most users.

### actionsSlot
I opted for a single prop to keep the API similar to the concept of `actions`, but instead of defining a new object, I've just allowed it to be a JSX element, ie:

```jsx
<GuidanceBlock
  actions={{
    primary: {
      label: "Hello",
      onClick: () => {}
    },
    secondary: {
      label: "Hello 2",
      onClick: () => {}
    },
  }}
  {...otherProps}
/>
```
**after**
```jsx
<GuidanceBlock
    actionsSlot={<>
      <Button variant="secondary" onPress={() => {}}>Hello</Button>
      <Button variant="tertiary" onPress={() => {}}>Hello 2</Button>
    </>}
  {...otherProps}
/>
```

The reason I have opted for is two fold:
- To ensure a true mapping of the GuidanceBlock action we would need to account for handling every thing include tooltips. This would create a needlessly complex codemod that needs to handle both the migration of button and the Tooltip (one hasn't been created for that).
- This is an old component and a pattern we want to sunset. There are not that many uses and the type errors the codemod should create (ie: passing tooltip into a new Button) will act as a forcing function to fix the component.
  - Only use of [tooltip](https://dx-insights.usw2.dev-us.cultureamp.io/question#eyJuYW1lIjoiS0FJTyBJY29uQnV0dG9uIHByb3BzIHVzYWdlIiwiZGVzY3JpcHRpb24iOiJQcm9wcyB0aGF0IG1heSBhZmZlY3QgdGhlIGNvZGVtb2QgZnJvbSBJY29uQnV0dG9uIHRvIG5ldyBCdXR0b24iLCJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjM5LCJhZ2dyZWdhdGlvbiI6W1siY291bnQiXV0sImJyZWFrb3V0IjpbWyJmaWVsZCIsMzA1LHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLFsiZmllbGQiLDMwNix7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSxbImZpZWxkIiwzMDQseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sWyJmaWVsZCIsMzAyLHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dXSwiZmlsdGVyIjpbImFuZCIsWyJjb250YWlucyIsWyJmaWVsZCIsMzAxLHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLCJAa2FpemVuL2NvbXBvbmVudHMiLHsiY2FzZS1zZW5zaXRpdmUiOmZhbHNlLCJlZmZlY3RpdmUtdHlwZSI6InR5cGUvQm9vbGVhbiJ9XSxbImNvbnRhaW5zIixbImZpZWxkIiwzMDUseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sIkd1aWRhbmNlQmxvY2siLHsiY2FzZS1zZW5zaXRpdmUiOmZhbHNlfV0sWyJjb250YWlucyIsWyJmaWVsZCIsMzA2LHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLCJhY3Rpb25zIix7ImNhc2Utc2Vuc2l0aXZlIjpmYWxzZX1dLFsiY29udGFpbnMiLFsiZmllbGQiLDMwNCx7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSwiVG9vbHRpcCIseyJjYXNlLXNlbnNpdGl2ZSI6ZmFsc2V9XV19fSwiZGlzcGxheSI6InRhYmxlIiwiZGlzcGxheUlzTG9ja2VkIjp0cnVlLCJwYXJhbWV0ZXJzIjpbXSwidmlzdWFsaXphdGlvbl9zZXR0aW5ncyI6eyJ0YWJsZS5waXZvdF9jb2x1bW4iOiJyZXBvX2Z1bGxfbmFtZSIsInRhYmxlLmNlbGxfY29sdW1uIjoicmVwb19hbmRfcGF0aCIsInRhYmxlLmNvbHVtbl93aWR0aHMiOltudWxsLG51bGwsMTAwN119LCJvcmlnaW5hbF9jYXJkX2lkIjoyMjYsInR5cGUiOiJxdWVzdGlvbiJ9)

### Alternatives
We could also do individual props like `primaryAction` and `secondaryAction` if we are concerned about the number of elements consumers may put in the `actionsSlot`

## What
- marks the `actions` prop as deprecated
- marks `secondaryDismiss` as deprecated
  -  there is not wide enough use to suggest we need to carry this across - [only usage of dismiss set to false](https://dx-insights.usw2.dev-us.cultureamp.io/question#eyJuYW1lIjoiS0FJTyBJY29uQnV0dG9uIHByb3BzIHVzYWdlIiwiZGVzY3JpcHRpb24iOiJQcm9wcyB0aGF0IG1heSBhZmZlY3QgdGhlIGNvZGVtb2QgZnJvbSBJY29uQnV0dG9uIHRvIG5ldyBCdXR0b24iLCJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjM5LCJhZ2dyZWdhdGlvbiI6W1siY291bnQiXV0sImJyZWFrb3V0IjpbWyJmaWVsZCIsMzA1LHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLFsiZmllbGQiLDMwNix7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSxbImZpZWxkIiwzMDQseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sWyJmaWVsZCIsMzAyLHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dXSwiZmlsdGVyIjpbImFuZCIsWyJjb250YWlucyIsWyJmaWVsZCIsMzAxLHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLCJAa2FpemVuL2NvbXBvbmVudHMiLHsiY2FzZS1zZW5zaXRpdmUiOmZhbHNlLCJlZmZlY3RpdmUtdHlwZSI6InR5cGUvQm9vbGVhbiJ9XSxbImNvbnRhaW5zIixbImZpZWxkIiwzMDUseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sIkd1aWRhbmNlQmxvY2siLHsiY2FzZS1zZW5zaXRpdmUiOmZhbHNlfV0sWyJjb250YWlucyIsWyJmaWVsZCIsMzA2LHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLCJkaXNtaXNzIix7ImNhc2Utc2Vuc2l0aXZlIjpmYWxzZX1dXX19LCJkaXNwbGF5IjoidGFibGUiLCJkaXNwbGF5SXNMb2NrZWQiOnRydWUsInBhcmFtZXRlcnMiOltdLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7InRhYmxlLnBpdm90X2NvbHVtbiI6InJlcG9fZnVsbF9uYW1lIiwidGFibGUuY2VsbF9jb2x1bW4iOiJyZXBvX2FuZF9wYXRoIn0sIm9yaWdpbmFsX2NhcmRfaWQiOjIyNiwidHlwZSI6InF1ZXN0aW9uIn0=)
- adds `actionsSlot` that can take jsx element
- adds codemod to enable easier migration

## Test branch
The latest canary has been released in `0.0.0-canary-canary-gb-codemod-test-2-20250318010353` and I have been testing against the unified-home app [here](https://github.com/cultureamp/unified-home/commit/712cebf192cc5e86e7eb4899c5ba9b9b4df37512). This codemod should covers all instances in this repo and most in others. Some manual intervention will be required in others, as with Tooltips, but we will document how to migrate this in stories.


packages/components/codemods/migrateGuidanceBlockActionsToActionsSlot/migrateGuidanceBlockActionsToActionsSlot.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+const transformGuidanceBlock = (sourceFile: TransformSourceArgs['sourceFile']): string => {
+  const kaioTagNamesMap = getKaioTagNamesMapByComponentName(sourceFile, ['GuidanceBlock'])
+  return transformSource({
+    sourceFile,
+    transformers: [migrateGuidanceBlockActionsToActionsSlot(kaioTagNamesMap!)],
+  })
+}
+
+describe('transformActionsToButtonNext()', () => {
+  it('imports Button from the /next path and transforms all button-like actions prop into a Button component', () => {
```
ryanseddon, line 19: Would be good to add in a tooltip one that shows the expected non behaviour of this codemod for the older tooltips

packages/components/codemods/migrateGuidanceBlockActionsToActionsSlot/transformActionsToActionsSlot.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+              size="medium"
+              isPending pendingLabel='Loading...'
+              hasHiddenPendingLabel
+              data-custom-attr='custom-attr'
+              variant="secondary"
+            >Learn more</Button></>}/>
+      `)
+
+    expect(transformInput(inputAst).replace(/\n+/g, ' ').replace(/\s+/g, ' ')).toBe(
+      printAst(outputAst).replace(/\n+/g, ' ').replace(/\s+/g, ' '),
```
ryanseddon, line 136: `\s+` is equivalent to looking for any of these `\r\n\t\f\v` so you only need to do the one global replace

```suggestion
    expect(transformInput(inputAst).replace(/\s+/g, ' ')).toBe(printAst(outputAst).replace(/\s+/g, ' '),
```

packages/components/codemods/migrateGuidanceBlockActionsToActionsSlot/transformActionsToActionsSlot.spec.ts
```diff
@@ -0,0 +1,196 @@
...
```
ryanseddon, line 193:
```suggestion
    expect(transformInput(inputAst).replace(/\s+/g, ' ')).toBe(printAst(outputAst).replace(/\s+/g, ' '),
```

packages/components/codemods/utils/transformV1ButtonPropsToButtonOrLinkButton.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export const migrateV1ButtonSize = (
+  oldValue: Exclude<V1ButtonProps['size'], undefined>,
+): Exclude<NextButtonProps['size'], undefined> => {
+  switch (oldValue) {
+    case 'small':
+      return 'medium'
+    case 'regular':
+      return 'large'
+  }
```
ryanseddon, line 20: Should we handle undefined here? Probably better off not using a switch too.

```suggestion
  const buttonSizeMap = {
    'small': 'medium',
    'regular': 'large',
  };

  return buttonSizeMap[oldValue] || undefined;
```

packages/components/src/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -242,6 +239,20 @@ export const GuidanceBlock = ({
             )}
           </Media>
         )}
+        {!actions && actionsSlot && (
```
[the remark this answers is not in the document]
mcwinter07: Good point, looks a lot nicer to read 🤤

packages/components/codemods/migrateGuidanceBlockActionsToActionsSlot/transformActionsToActionsSlot.spec.ts
mcwinter07: Ah that sweet regex magic I always forget 🪄 🐇

packages/components/src/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -242,6 +239,20 @@ export const GuidanceBlock = ({
...
+          <Media query="(max-width: 767px)">
+            {(isMobile: boolean): JSX.Element => (
+              <div
+                className={classNames({
+                  noRightMargin: isMobile || componentIsMobile,
+                  rightMargin: !(isMobile || componentIsMobile) && layout === 'default',
```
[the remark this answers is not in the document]
mcwinter07: I might actually remove this after review the designs. None of them have this additional margin and given it didn't work before I can't imagine we'll see any diffs. Also means we'll be able to remove some of the other old media query stuff so there's an additional win :) 

Screenshot 2025-03-21 at 9 34 47 am

[The design](https://www.figma.com/design/eZKEE5kXbEMY3lx84oz8iN/%F0%9F%92%9C-Heart-UI-Kit?m=auto&node-id=50167-151665) uses 36px padding consistently across screen sizes

packages/components/src/GuidanceBlock/GuidanceBlock.tsx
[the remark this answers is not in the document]
mcwinter07: I opted to remove the Media component and replace it with a useMediaQuery util that we have. Pretty confident that there'll be no loss in functionality and one less dep to keep :)

packages/components/codemods/utils/transformV1ButtonPropsToButtonOrLinkButton.ts
mcwinter07: Given the simplicity of this util and condition I've opted to just handle this in the`transformButtonProp` function within the `size` case. Keen to hear thoughts :)

packages/components/src/GuidanceBlock/GuidanceBlock.tsx
```diff
@@ -193,55 +166,43 @@ export const GuidanceBlock = ({
...
-                      icon={
-                        withActionButtonArrow ? (
-                          <Icon name="arrow_forward" isPresentational shouldMirrorInRTL />
-                        ) : undefined
-                      }
-                      iconPosition="end"
-                      {...actions.primary}
-                      fullWidth={isMobile || componentIsMobile}
-                    />
+        {actions?.primary || actionsSlot ? (
```
ryanseddon, line 169: Is this correct if the component has actions.primary and actionsSlot you will use the old deprecated actions instead?

ryanseddon (approved):

PR #5578, opened 2025-03-20
Use outline instead of pseudo elements for focus rings

dougmacknz:
These pseudo element focus rings were implemented because `outline` didn't support rounded borders until more recently. Now that our supported browsers support this, we can adjust these to use outlines which is much more simple and less potential for issues.

This isn't comprehensive, I've left a bunch for components that are going away in the 2.0 release, or things that I don't want to touch (TitleBlock, Table)

### DIFF
--- a/.changeset/smart-falcons-look.md
+++ b/.changeset/smart-falcons-look.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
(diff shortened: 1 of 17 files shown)


mcwinter07 (approved):
Cool, I've review the focus border against safari, which has been the most recent hold out on the radius inheritance so this looks all hunky dory in my books 👍

PR #5601, opened 2025-03-25
TimeField: fix spacing after dep update

dougmacknz:
## Why
A recent dep update of the `@react-stately/datepicker` package caused some extra padding in the component

Before:
image

After:
image

On further investigation, there were some new invisible 'segment' characters added to ensure correct RTL behaviour:
image

Currently, every segment has padding added by us.

## What
Avoid adding padding to these new invisible character segments.

### DIFF
--- a/.changeset/vast-gifts-know.md
+++ b/.changeset/vast-gifts-know.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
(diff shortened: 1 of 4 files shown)


mcwinter07 (approved):
Another stealth update for RA sneaks up on us :shakes-fist: Good catch on the fix

PR #5609, opened 2025-03-27
Modal refactor to remove headlessui

dougmacknz:
## What
- Replaces the dependency on `headlessui/react` with `motion` for our modal animation and tapping into when the animation is complete.
  - Using `motion` also allows us to remove a bunch of legacy Zurb CSS code for animation that we had forked into our repo.
- Rejigs a bunch of things in GenericModal and its subcomponents (ModalHeader, ModalBody, ModalFooter) to move us toward a much better set of components for our consumers to build custom modals, and also sets us up to migrate the buttons in a follow up.
  - GenericModal now comes with a base set of styles.
  - ModalFooter now accepts children and has its `actions` prop deprecated
- Deletes a lot of duplication and dead code.
  - e.g. paddings for modal footers was repeated in each modal component. This is now consolidated into the one ModalFooter component. 
- Uplifts a bunch of legacy Sass with our modern approach to CSS
- Replaces the `react-focus-lock` usage with `react-focus-on` and uninstalls `react-focus-lock` because it was our last usage.
- Removes bespoke code in GenericModal for the escape press handler and scroll lock to the `react-focus-on` library.

## Why
- HeadlessUI dep has been a pain for us, it randomly broke our UIs in a point release (leading to us pinning the version). Now when I try to update to latest, the animation for the modal no longer works
- Building custom modals is a big pain for our consuming teams, requiring them to duplicate a bunch of styles from our generic modal stylesheet.


packages/components/src/Modal/GenericModal/animations.ts
```diff
@@ -0,0 +1,34 @@
+export const backdropMotion = {
+  hidden: {
+    opacity: 0,
+  },
+  visible: {
+    opacity: 0.5,
+    animationTimingFunction: 'cubic-bezier(0.485, 0.155, 0.24, 1.245)', // $animation-easing-function-bounce-in
```
dougmacknz, line 7: todo: use the animation vars

packages/components/src/Modal/GenericModal/subcomponents/ModalFooter/ModalFooter.tsx
```diff
@@ -26,44 +21,45 @@ export type ModalFooterProps = {
    * @deprecated Please use data-testid
    */
   automationId?: string
-  alignStart?: boolean
```
mcwinter07, line 29: So a quick squiz on metabase and it looks like this isn't use, which is why we're removing it yeah?

dougmacknz: Oh yeah it wasn't doing anything as well

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -1,71 +1,45 @@
...
-  focusLockDisabled?: boolean
-  onEscapeKeyup?: (event: KeyboardEvent) => void
+  onEscapeKeyup?: (event: Event) => void
   onOutsideModalClick?: (event: React.MouseEvent) => void
   /** A callback that is triggered after the modal is opened. */
   onAfterEnter?: () => void
   /** A callback that is triggered after the modal is closed. */
   onAfterLeave?: () => void
   className?: string
+  size?: 'medium' | 'large'
```
mcwinter07, line 22: Could you add a type description for this just to outline it controls the visual size of the container

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -1,71 +1,45 @@
...
+import { backdropMotion, modalMotion } from './animations'
 import { ModalContext } from './context/ModalContext'
-import styles from './GenericModal.module.scss'
+import styles from './GenericModal.module.css'
 
 export type GenericModalProps = {
   id?: string
   isOpen: boolean
   children: React.ReactNode
-  focusLockDisabled?: boolean
```
mcwinter07, line 13: Looks like there's no [focusLockDisabled](https://dx-insights.usw2.dev-us.cultureamp.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOiJjYXJkX18zMDQiLCJmaWx0ZXIiOlsiYW5kIixbImNvbnRhaW5zIix7ImNhc2Utc2Vuc2l0aXZlIjpmYWxzZSwiZWZmZWN0aXZlLXR5cGUiOiJ0eXBlL0Jvb2xlYW4ifSxbImZpZWxkIiwiSW1wb3J0Iix7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSwiQ29udGV4dE1vZGFsIiwiTW9kYWwiXSxbImNvbnRhaW5zIixbImZpZWxkIiwiUGFja2FnZSBOYW1lIix7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSwiQGthaXplbi9jb21wb25lbnRzIix7ImNhc2Utc2Vuc2l0aXZlIjpmYWxzZX1dLFsiY29udGFpbnMiLFsiZmllbGQiLCJQcm9wIE5hbWUiLHsiYmFzZS10eXBlIjoidHlwZS9UZXh0In1dLCJmb2N1c0xvY2tEaXNhYmxlZCIseyJjYXNlLXNlbnNpdGl2ZSI6ZmFsc2V9XV19fSwiZGlzcGxheSI6InRhYmxlIiwidmlzdWFsaXphdGlvbl9zZXR0aW5ncyI6e30sIm9yaWdpbmFsX2NhcmRfaWQiOm51bGwsInR5cGUiOiJxdWVzdGlvbiJ9) so looks like its safe to axe it - also seems like a bad idea anyway

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -1,71 +1,45 @@
-import React, { useEffect, useId, useState } from 'react'
+import React, { useId, useRef } from 'react'
 import { createPortal } from 'react-dom'
-import { Transition } from '@headlessui/react'
-import FocusLock from 'react-focus-lock'
+import classNames from 'classnames'
+import { AnimatePresence, motion } from 'motion/react'
```
mcwinter07, line 4: Still need to test the first contentful paint but found [this page](https://motion.dev/docs/react-reduce-bundle-size) on the motion docs. If it becomes an issue we might be able to reduce the bundle size

packages/components/src/TextField/TextField.tsx
```diff
@@ -78,7 +78,7 @@ export const TextField = ({
         startIconAdornment={icon}
         endIconAdornment={
           status === 'success' && (
-            <div className={classnames(styles.icon, styles.success, disabled && styles.disabled)}>
+            <div className={classnames(styles.icon, styles.success)}>
```
mcwinter07, line 81: Was this meant to be in this PR?

dougmacknz: Yeah, strangely when I removed the animation thing from TextField, it must have been masking this issue where the icon was getting double 0.3 opacity applied to it.

packages/components/src/Modal/GenericModal/subcomponents/ModalFooter/ModalFooter.tsx
mcwinter07: groovy, sounds good!

packages/components/src/TextField/TextField.tsx
mcwinter07: Ah right, gotcha 🤔 I suppose this is on the icon anyhow so that was always weird

dougmacknz: Yeah so the container wrapping this has already got a disabled class which is getting 0.3 opacity applied

packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -78,108 +55,94 @@ export const GenericModal = ({
...
+                  if (latest === 'visible') {
+                    handleAfterOpen()
+                    return
+                  }
+                  handleAfterClose()
+                }}
+              >
+                <div
+                  ref={modalRef}
+                  className={classNames(styles.modal, size === 'large' && styles.large)}
```
mcwinter07, line 131: I reckon we move the `className` prop down to here since it won't do anything on the FocusOn container

packages/components/src/Modal/GenericModal/GenericModal.tsx
mcwinter07: I'm not the biggest performance wonk but I think based on my test in coversations UI locally I can't see any massive discernible difference in the contentful pain so I'm going to say this should be ok for the mo'. We can potentially optimize this later by using the hooks or LazyMotion as recommended in the link above. Evidence below:

**With Motion**
Screenshot 2025-04-03 at 1 38 45 pm
Screenshot 2025-04-03 at 1 37 10 pm

**Original GenericModal**
Screenshot 2025-04-03 at 1 33 12 pm
Screenshot 2025-04-03 at 1 36 53 pm

mcwinter07:
Going to close this since we opted to upgrade headless-ui instead of replace it. Will keep a link in a tech health ticket to this PR for when we tackle the Modal rebuild

dougmacknz:
Boo 👎 this was 100% ready to go

PR #5625, opened 2025-03-31
Add Well isAiLoading prop for AI Moment

mcwinter07:
## Why
This is part of our support of the new rollout of AI moments and a means of standardizing the AI moment styles within the Kaizen `Well`.

## What
- Updates `Well` docs to API spec and Usage Guidelines format
- Adds `useExitingAnimation` hook util to allow for the addition and removal of the exiting animation.
  -  This allows a graceful end to the current animation loop since we can't guarantee how long a request might take / the AI is thinkining
- Adds `sAiLoading` prop to the Well component
  - when `true` the Well will blur and subtly pulse and rotate background pseudo elements
  - when `false` only the simulate gradient border will be visible
  - when undefined, this will be a default Well
- Well can still inehirit a background color set with the `color` prop but will not change the border color or style white `sAiLoading` is `true` or `false`

### Before
Screenshot 2025-03-31 at 4 52 22 pm

### After
Screenshot 2025-03-31 at 4 57 07 pm
Screenshot 2025-03-31 at 4 53 24 pm

## Additional notes
- similar to the PRC implementation I had to add a extra element to the `Well` to ensure the simulated border could sit within the Wells container. Where this differs is that I have chosen to add the white background as a floating absolute element and not wrapped the children - This is to ensure that `classNameOverride`s will not be impacted by the new element, which was not a consideration needed for prc.
  - Screenshot 2025-03-31 at 4 45 24 pm
  - Happy for feedback or recommendations here :)


packages/components/src/Well/Well.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { type HTMLAttributes } from 'react'
 import classnames from 'classnames'
 import { type OverrideClassName } from '~components/types/OverrideClassName'
+import { useExitingAnimation } from '~components/utils/useExitingAnimation'
 import { type WellBorderStyleType, type WellColors, type WellVariantType } from './types'
-import styles from './Well.module.scss'
+import styles from './Well.module.css'
```
NJWong, line 6: I've got a super naive question - what's the motivation for using CSS modules over Tailwind for components like this?

My first guess is because the component predates Tailwind integration and there isn't a lot of benefit in rewriting it when it works. Or maybe because certain thing (like once-off animations) are easier to write in CSS?

packages/components/src/Well/Well.tsx
```diff
@@ -26,21 +29,37 @@ export const Well = ({
...
+
+  return (
+    <div
+      className={classnames(
+        styles.container,
+        styles[borderStyle],
+        styles[color],
+        variant && styles[variant],
+        noMargin && styles.noMargin,
+        classNameOverride,
```
NJWong, line 50: Would the consumer expect that their class overrides to also override the AI moment styles if they're using that variant? If that's the case, does `classNameOverride` need to be after the AI moment styles?

packages/components/src/Well/Well.tsx
```diff
@@ -26,21 +29,37 @@ export const Well = ({
...
+        isAiLoading !== undefined && styles.aiMoment,
+        isAiLoading !== undefined && animationClass,
+      )}
+      {...restProps}
+    >
+      <>
+        {/* This is a visual background only and should not children in case classNameOverride is used */}
```
NJWong, line 57: I don't fully understand this comment 😅

packages/components/src/Well/_docs/Well.mdx
```diff
@@ -1,41 +0,0 @@
-import { Canvas, Controls, Meta } from '@storybook/blocks'
```
NJWong, line 1: When we create or update components, we should create two types of documentation:
1. API Spec
2. Usage Guidelines

Is that right?

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from 'react'
+
+/** a utility hook for handling an the addition and removal of an animation and exit animation class */
+export const useExitingAnimation = (
+  shouldAnimate: boolean | undefined,
+  animatingClass: string,
+  exitingClass: string,
+  exitDuration: number = 1000,
+): string | undefined => {
```
NJWong, line 9:
```suggestion
): string => {
```

I don't think we need `undefined` here because `animationClass` is always a string

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [animationClass, setAnimationClass] = useState('')
+  const prevIsAiLoading = React.useRef<boolean | undefined>(undefined)
```
NJWong, line 11: It looks like this hook could be used in different components, so we should probably rename this so it's not specific to ai moment

```suggestion
  const prevIsLoading = React.useRef<boolean | undefined>(undefined)
```

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
NJWong: This also potentially brings up a larger can of worms about the `@typescript-eslint/explicit-function-return-type` rule - I'm not a big fan of explicit function return types because it's easy to lie to the TS compiler about what a function could return. In this case, the `animationClass` variable in `Well.tsx` was typed as `string | undefined` and it gets unnecessarily messy if we need to keep adding in `undefined` guards

packages/components/src/Well/Well.tsx
mcwinter07: Not a naive question at all! :) Definitely both of those but also a few of historical reasons behind this was that:
- We wanted our CSS as close to vanilla CSS as possible to reduce the config that we need to bundle and ship (also why we're drying to drop SCSS with each component update)
- As team we write we tend to write a lot of CSS and decided it was easier read and less verbose to write in Vanilla (especially the case with grid, media and container queries)
- We were also concerned about the global scope tailwind and wanted to ship components with isolated CSS with sensible defaults (this was before folks started using prefixes) 

Ultimately it was decided that, while we liked Tailwind for micro layouts, we didn't feel like it the safest or most efficient route for publishing UI components.

mcwinter07: Worth noting that we don't actually ship tailwind from any of our components, it's just used for storybook when we want to test `classNameOverride` / `className` or when adjust layouts of our stories

packages/components/src/Well/Well.tsx
mcwinter07: heheh yeah, probably more of a thought in progress there 😆 I think because in PRC they wrap the whole of children, but I wanted to do something that wouldn't impact anything that got added to the classname like display properties or padding and such - I'll remove the comment since it doesn't really need to be there

packages/components/src/Well/Well.tsx
NJWong: Awesome, thanks for that explanation!

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+
+.aiMoment {
+  --clr-1: rgba(100 210 215 / 100%) 0%;
+  --clr-2: rgba(103 199 175 / 100%) 20%;
+  --clr-3: rgba(100 210 215 / 100%) 41%;
+  --clr-4: rgba(200 136 226 / 100%) 100%;
+  --animated-border-width: 2px;
+
+  background-color: transparent;
+  position: relative;
```
ryanseddon, line 122: I think this is a good opportunity to do a better modern approach to stacking items using a stacked grid over positioning and inset.

You can do a stack like so
```css
.aiMoment {
  //...
  display: grid;
  
  // All direct children and pseudo classes have this applied to stack 
  & > *, &::before, &::after {
    grid-area: 1/1;
  }
}
```

You can still use z-index with grid so that still works as expected and the benefit of this is the document flow of the internal contents is kept in tact.

I tried this locally and all browsers are good

```diff
diff --git a/packages/components/src/Well/Well.module.css b/packages/components/src/Well/Well.module.css
index bcffce42a..59c7ea8a2 100644
--- a/packages/components/src/Well/Well.module.css
+++ b/packages/components/src/Well/Well.module.css
@@ -119,20 +119,21 @@
   --animated-border-width: 2px;
 
   background-color: transparent;
-  position: relative;
+  display: grid;
   border: 1px solid transparent;
   border-radius: calc(var(--border-borderless-border-radius));
+
+  & > *,
+  &::before,
+  &::after {
+    grid-area: 1/1;
+  }
 }
 
 /* This absolute element is used to give the illusion of a gradient border without expanding the size of the Well */
 .aiMomentBackground {
   margin: var(--animated-border-width);
   box-sizing: border-box;
-  position: absolute;
-  content: '';
-  z-index: -1;
-  inset: 0;
-  display: block;
   border-radius: inherit;
   border: 1px solid transparent;
   background-color: var(--background-color);
@@ -140,9 +141,8 @@
 
 .aiMoment::before,
 .aiMoment::after {
   content: '';
-  position: absolute;
-  inset: -1px;
   z-index: -2;
   background-image: linear-gradient(
     var(--gradient-angle),
```

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+    opacity: 0.3;
+  }
+
+  100% {
+    background-size: 100% 100%;
+  }
+}
+
+@keyframes rotation-loop-after {
+  0% {
```
ryanseddon, line 207: You can combine animation steps that do the same thing
```suggestion
  0%, 100% {
```

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+  opacity: 0.7;
+}
+
+.aiMomentExiting::after,
+.aiMomentExiting::before {
+  animation-iteration-count: 1;
+}
+
+@keyframes rotation-loop-before {
+  0% {
```
ryanseddon, line 189:
```suggestion
  0%, 100% {
```

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+    --gradient-angle: 270deg;
+
+    background-position: 0% 20%;
+    background-size: 200% 300%;
+    opacity: 0.3;
+  }
+
+  100% {
+    background-size: 100% 100%;
+  }
```
ryanseddon, line 203:
```suggestion
```

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+    --gradient-angle: 270deg;
+
+    background-position: 0% 20%;
+    background-size: 110% 130%;
+    opacity: 0.7;
+  }
+
+  100% {
+    background-size: 100% 100%;
+  }
```
ryanseddon, line 221:
```suggestion
```

packages/components/src/Well/Well.module.css
mcwinter07: Ah yeah, didn't think to use grid :) sounds like a swish idea

mcwinter07: The only concern I have is that we'd be now enforcing a display property on the Well container, which at the moment is not set. Potentially could try and use `aiMomentBackground` instead of the outer div?

mcwinter07: The scenario where this could be an issue is if a display property is assign in classNameOverride that isn't grid

packages/components/src/Well/Well.module.css
mcwinter07: TIL 🙇 Many thanks

packages/components/src/Well/Well.module.css
ryanseddon: True but the same could be said for `position`.

mcwinter07: There are a [number of instances in Metabase](https://dx-insights.usw2.dev-us.cultureamp.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjM5LCJmaWx0ZXIiOlsiYW5kIixbImNvbnRhaW5zIixbImZpZWxkIiwzMDEseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sIkBrYWl6ZW4vY29tcG9uZW50cyIseyJjYXNlLXNlbnNpdGl2ZSI6ZmFsc2UsImVmZmVjdGl2ZS10eXBlIjoidHlwZS9Cb29sZWFuIn1dLFsiY29udGFpbnMiLHsiY2FzZS1zZW5zaXRpdmUiOmZhbHNlLCJlZmZlY3RpdmUtdHlwZSI6InR5cGUvQm9vbGVhbiJ9LFsiZmllbGQiLDMwNSx7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSwiV2VsbCIsImNhcmQiXSxbImNvbnRhaW5zIixbImZpZWxkIiwzMDYseyJiYXNlLXR5cGUiOiJ0eXBlL1RleHQifV0sImNsYXNzTmFtZU92ZXJyaWRlIix7ImNhc2Utc2Vuc2l0aXZlIjpmYWxzZX1dLFsiY29udGFpbnMiLFsiZmllbGQiLDMwNCx7ImJhc2UtdHlwZSI6InR5cGUvVGV4dCJ9XSwiZmxleCIseyJjYXNlLXNlbnNpdGl2ZSI6ZmFsc2V9XV19fSwiZGlzcGxheSI6InRhYmxlIiwiZGlzcGxheUlzTG9ja2VkIjp0cnVlLCJwYXJhbWV0ZXJzIjpbXSwidmlzdWFsaXphdGlvbl9zZXR0aW5ncyI6eyJ0YWJsZS5waXZvdF9jb2x1bW4iOiJyZXBvX2Z1bGxfbmFtZSIsInRhYmxlLmNlbGxfY29sdW1uIjoicmVwb19hbmRfcGF0aCJ9LCJvcmlnaW5hbF9jYXJkX2lkIjoyMjYsInR5cGUiOiJxdWVzdGlvbiJ9) where Card and Well are passed flex that gives me pause 🤔 (There is also one instance I found of position absolute).

I'm not apposed to using grid for AI moment, but we will have to update our documentation - its potentially a little odd with the default not having a display property so I'd just want to mkae sure we call out that behaviour

mcwinter07: Just a quick example if flex overrides the display property
Screenshot 2025-04-07 at 4 39 14 pm

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
mcwinter07: Thanks for the spot 👓 I've removed the undefined since as you say, is always returned as a string 👍 As to the larger can of worms question around explicit returns and our ts rules, I'm not strong in favor for and against. I believe this rule was added to ensure we don't rub up against consumers that have stricter types - https://cultureamp.slack.com/archives/C02NUQ27G56/p1683244063168049?thread_ts=1683242725.889089&cid=C02NUQ27G56 - don't know if this is still relevant 🤷

packages/components/src/Well/Well.module.css
mcwinter07: The alternative is using some of that grid logic on the `aiMomentBackground` instead and use grid-area on the pseudo elements, ie:

```
  <div className="aiMoment" <--- has transparent bg and a position rel >
    <div className="AiMomentBackground" <--- has before an after pseudo assign to the a grid-area /> 
    {children}
  </div>
```

Doesn't change the fact we'd need position relative on the Well container but keen to hear thoughts @ryanseddon

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
NJWong: Yeah cool I don't think we need to take any action on the rule for now since this could just be a once-off thing - so we'll just keep it under observation 👀

packages/components/src/Well/Well.module.css
ryanseddon: Ok then this possibly introduces more problems then it solves, happy for you to use position over grid.

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
```diff
@@ -0,0 +1,33 @@
...
+        }, exitDuration)
+
+        return () => clearTimeout(timeout)
+      }
+      if (shouldAnimate === true) {
+        setAnimationClass(animatingClass)
+      }
+    }
+    previousIsLoading.current = shouldAnimate
+    return
```
ryanseddon, line 29: Why the empty return?

packages/components/src/Well/Well.module.css
```diff
@@ -0,0 +1,222 @@
...
+  border: 1px solid transparent;
+  border-radius: calc(var(--border-borderless-border-radius));
+}
+
+/* This absolute element is used to give the illusion of a gradient border without expanding the size of the Well */
+.aiMomentBackground {
+  margin: var(--animated-border-width);
+  box-sizing: border-box;
+  position: absolute;
+  content: '';
```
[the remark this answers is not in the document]
ryanseddon: You've still got this here is it needed?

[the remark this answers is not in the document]
mcwinter07: Ah yes, one last thing to remove 🪓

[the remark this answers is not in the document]
mcwinter07: Tis' gone 💥

packages/components/src/utils/useExitingAnimation/useExitingAnimation.tsx
mcwinter07: We have a tsConfig rule that suggest all code path have a return value. 
Screenshot 2025-04-11 at 11 43 04 am
I suppose its really need it should just be calling setAnimation or setTimeout. Do you think a more accurate type would be:
```tsx
  useEffect((): void | (() => void) => {
    if (shouldAnimate !== undefined) {
      if (previousIsLoading.current === true && shouldAnimate === false) {
        setAnimationClass(exitingClass)

        const timeout = setTimeout(() => {
          setAnimationClass('')
        }, exitDuration)

        return () => clearTimeout(timeout)
      }
      if (shouldAnimate === true) {
        setAnimationClass(animatingClass)
      }
    }
    previousIsLoading.current = shouldAnimate
  }, [shouldAnimate, animatingClass, exitingClass, exitDuration])
 ```

ryanseddon: Yeah, I think that makes more sense than an empty return.

mcwinter07: Shibby, I've updated the type and axed the empty return - thanks for picking up that code smell

NJWong (approved):
Nice one

PR #5628, opened 2025-04-01
fix(deps): Moves react-aria-components to a devDep and adds it as a peerDep

ryanseddon:
## Why

With the introduction of the [RouterProvider default into next-services](https://github.com/cultureamp/frontend-foundations/pull/1302) this has a dep on react-aria-components in order for us to avoid duplicate versions being installed and causing odd behaviour we need to move it to a peerDependency.

This is a breaking change as not all projects will have a hard dependency on react-aria-components and they may have to install and add react-aria-components as a dep.

We can also suggest that they change a [default config of pnpm to auto install peerDeps](https://truecoderguru.com/blog/pnpm/pnpm-auto-install-peer-dependencies) this would avoid having to add react-aria-components as a direct dependency to projects that only transiently rely on react-aria-components.
## Todo

 - [ ] Gather some data in metabase on who has and hasn't got a hard dep on react-aria-components
 - [ ] Gather info on who is using @kaizen/components/v3/react-aria-components re-exports

## What

- Remove react-aria-components as a direct dep of @kaizen/components
- Add as devDep
- Add peerDep range

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -151,7 +151,6 @@
     "prosemirror-view": "^1.37.1",
     "react-animate-height": "^3.2.3",
     "react-aria": "^3.38.1",
-    "react-aria-components": "^1.7.1",
     "react-day-picker": "8.10.1",
     "react-focus-lock": "^2.13.6",
     "react-focus-on": "^3.9.4",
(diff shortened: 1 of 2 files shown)


ryanseddon:
We no longer need to do this as we're standardising on KAIO being the place you import react-aria-components from

PR #5654, opened 2025-04-07
fix(GenericModal): update headlessui dep and Transition usage

mcwinter07:
## Why
- This was causing a strict mode issue originally reported [here](https://cultureamp.slack.com/archives/C2ZNME08P/p1743088483404099) which would result in the Modal not working when wrapped in StrictMode.

## What
- updates opinned headlessUI dep
- adds `escapeKeyHandler` to resolve potential memory leak
- update usage of Transition following breaking change in headlessUI 2.0

## Demo
I've release a canary and tested [here](https://codesandbox.io/p/devbox/modal-with-strictmode-fix-ht3rf8) - the modal now renders when the app is wrapped in strict mode.

Screenshot 2025-04-07 at 2 45 47 pm


ryanseddon (approved):
Nice, would we be able to turn on strict mode now in our storybook?

mcwinter07:
> Nice, would we be able to turn on strict mode now in our storybook?
@ryanseddon I think we may potentially have some other components that require changes before that. I did a quick search on some [tickets in Jira](https://cultureamp.atlassian.net/jira/software/c/projects/KZN/issues/?filter=doneissues&jql=project%20%3D%20%22KZN%22%20AND%20statusCategory%20%3D%20%22Done%22%20AND%20textfields%20~%20%22strict%22%20ORDER%20BY%20created%20DESC) that got de-prioritized a while back. 
Screenshot 2025-04-08 at 9 09 08 am

For now I'll merge and release this fix, but we could create some health tickets if we want to see if those component still present issues for strict mode

PR #5656, opened 2025-04-08
add changeset to release color-string pkg rollback

mcwinter07:
## Why
We missed adding a changeset for the design-tokens package on the color-string roll back.

## What
- adds changset for color-string rollback
- add dep comments so we don't accidentally bump the version

### DIFF
--- a/packages/design-tokens/package.json
+++ b/packages/design-tokens/package.json
@@ -35,8 +35,11 @@
     "clean:dist": "rimraf 'dist'"
   },
   "dependencies": {
-    "color-string": "^1.9.1",
+    "color-string": "1.9.1",
     "lodash.kebabcase": "^4.1.1"
+  },
+  "dependenciesComments": {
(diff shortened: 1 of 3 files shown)


packages/design-tokens/package.json
```diff
@@ -35,9 +35,12 @@
     "clean:dist": "rimraf 'dist'"
   },
   "dependencies": {
-    "color-string": "^1.9.1",
+    "color-string": "1.9.1",
     "lodash.kebabcase": "^4.1.1"
   },
+  "dependenciesComments": {
+    "color-string": "Version locked as 2.x as this will break implementation of older repos that are using the ThemeProvider (https://cultureamp.slack.com/archives/C0189KBPM4Y/p1742787520546679) - we will remove this in KAIO 2.0 so we can upgrade to 2.x then if still needed"
```
NJWong, line 42:
```suggestion
    "color-string": "Version locked at 1.9.1 as this will break implementation of older repos that are using the ThemeProvider (https://cultureamp.slack.com/archives/C0189KBPM4Y/p1742787520546679) - we will remove this in KAIO 2.0 so we can upgrade to 2.x then if still needed"
```

ryanseddon (dismissed):

NJWong (approved):

PR #5675, opened 2025-04-13
add AI moment props, styles and docs to Card

mcwinter07:
## Why
This is part of the AI moment uplift to Well and Cards.

## What
- add isAiMoment loading prop and AI styles to Card
- Updates documentation to new API Spec and Usage Guidelines format
- Updates stories and stickersheets


packages/components/src/Card/Card.module.css
```diff
@@ -88,3 +94,106 @@
...
+  --clr-1: oklch(80.15% 0.101 199.35deg) 0%;
+  --clr-2: oklch(76.41% 0.098 175.22deg) 20%;
+  --clr-3: oklch(80.15% 0.101 199.35deg) 41%;
+  --clr-4: oklch(72.34% 0.143 315.73deg) 100%;
+  --animated-border-width: 2px;
+
+  background-color: transparent;
+  position: relative;
+  border: 1px solid transparent;
+  border-radius: calc(var(--border-borderless-border-radius) + 1px);
```
ryanseddon, line 108: Other than this line it seems a copy of the same css from the Well component could we just move this out into it's own file and in both Well & Card hust import it?


import styles from './Card.module.css'

ryanseddon: Or the other option is to compose into the classNames so it's still defined in the respective styles e.g.

```
.aiMoment {
  composes: aiMoment from '../styles/utils/AIMoment.module.css';
}

Comment (mcwinter07) [reply packages/components/src/Card/Card.module.css:108]: Yeah good point, makes sense to centralize it since they don't really diverge (and probably won't in the future).

I might roll with the first pattern as we did run into some issues in the past using `composes`. I honestly can't remember what it was, outside of some CSS compilation issue, but we made a chose to remove them all at some point

File: packages/components/src/Card/Card.module.css [hunk 2]
```diff
@@ -88,3 +94,106 @@
...
+
+/* This absolute element is used to give the illusion of a gradient border without expanding the size of the Well */
+.aiMomentBackground {
+  margin: var(--animated-border-width);
+  box-sizing: border-box;
+  position: absolute;
+  z-index: -1;
+  inset: 0;
+  display: block;
+  border-radius: var(--border-borderless-border-radius);
```
NJWong, line 119:
```suggestion
  border-radius: calc(var(--border-borderless-border-radius) - 2px);
```

I don't want this to come off as too picky, but we can make the border radius look more natural if we do this!

Before:
Screenshot 2025-04-14 at 2 05 37 pm

After:
Screenshot 2025-04-14 at 2 05 44 pm

NJWong: I just played around with the `calc` to see what looked best, but here's some quick math for the reason why nested border radius need a bit of extra work to look good: https://www.30secondsofcode.org/css/s/nested-border-radius/

NJWong: A better solution I just saw is to use `--animated-border-width`

packages/components/src/Card/Card.module.css
mcwinter07: Hmmm getting some specificity issues with importing from a shared styles 🤔 tried `composes: aiMoment from '../styles/utils/AIMoment.module.css' ` to see if that'd resolve where it lands in the cascade but that didn't seem to even add the styles in. 

Will give the @layer rule a crack to see if that'll do the trick.

mcwinter07: This seems to work hunky dory in card.module.css:

```CSS
@layer base, aiMoment;
@import '../../styles/utils/aiMoment.module.css' layer(aiMoment);
@layer base {
  .wrapper {
// ... wrapper styles
```
I'm concerned that creating a layer just for aiMoment is a little too specific though 🤔   @ryanseddon  any thoughts on layer name conventions? Maybe something like

```CSS
@layer components, themes, overrides 
```

We haven't really leveraged them much in our DST so we don't have much of a convention outside of our foundation stuff

ryanseddon: yeah I think components, themes, overrides is a good idea.

mcwinter07: Cool :) think we should add them into out `layer.css` so we don't have to define them each time in the component?

packages/components/src/Card/Card.module.css
mcwinter07: Nono, I agree, this looks much better and really more of my mistake :) Long live pedantry in all things border widths and radii 🥳 `--animated-border-width` should do the trick nicely

packages/components/src/Card/Card.module.css
mcwinter07: Ok so I think I resolved this without needing to add a new layer by manually bumping the specificity (with: `.aiMoment.aiMoment`) in the shared stylesheet.

I'm still keen to chat with Kevin tomorrow about the use of layers since there may be cases like this where it could be relevant - though I don't want to muddy up this PR with layer changes if its not necessary for now / and is something Kevin is actively working on :)

In any case, keen to hear thoughts on this solution @ryanseddon

packages/components/src/Card/Card.module.css
mcwinter07: Cool those border should be nice and properly rounded now :) appreciate the eagle eyes 🐦‍⬛

packages/components/src/Card/Card.module.css
mcwinter07: Brill will do 👍

ryanseddon (approved):

PR #5677, opened 2025-04-15
chore(deps): upgrade Storybook to 8.6.12 

ryanseddon:
## Why

To upgrade to react 19 we need to get onto the latest storybook stable release

## What

Updated storybook to 8.6.x removed override for storybook hang issue as 8.6.x now has the version we set in the override. Setting override for @storybook/icons as it by default it installs 1.2.x as a transient dep. 

We'll have to keep an eye on storbook releases as I imagine we can remove this soon.


ryanseddon:
Not doing a changeset here as there will be one in the main branch that this is merging too.

Kitty-Al (approved):

PR #5678, opened 2025-04-15
[DO NOT MERGE] [KZN-3248] add data-testid back to ToastNotificationsList

NJWong:
## Why

This PR adds the `data-testid` back to `ToastNotificationsList`. There are some repos like `recognize-ui` that have UI tests that look for the toast container element using this test ID.

[Slack Support thread](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1744600059646219)

## What

There was a [refactor](https://github.com/cultureamp/kaizen-design-system/commit/7303979b102047ec73c8e29f9bafebd032768b7f#diff-2795f43ef06807da6239b83fda40ee1372dcd7d990b84d8e3ac93ebe55d2c9afL23) earlier this year which unintentionally removed this test ID, so this PR adds it back.

### DIFF
--- a/packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
+++ b/packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
@@ -15,6 +15,7 @@ export const ToastNotificationsList = (): JSX.Element => {
     if (!container) {
       container = document.createElement('div')
       container.setAttribute('id', toastNotificationListId)
+      container.setAttribute('data-testid', toastNotificationListId)
       container.setAttribute('role', 'status')
       container.className = styles.toastNotificationsList
       document.body.appendChild(container)


packages/components/src/Notification/ToastNotification/ToastNotificationsList/ToastNotificationsList.tsx
```diff
@@ -15,6 +15,7 @@ export const ToastNotificationsList = (): JSX.Element => {
     if (!container) {
       container = document.createElement('div')
       container.setAttribute('id', toastNotificationListId)
+      container.setAttribute('data-testid', toastNotificationListId)
```
mcwinter07, line 18: hmmm, this is an interesting one - we generally don't bake in test ids into our components, and was something we made an effort to remove between our old to new packages. The updated solution was to open up the data attributes in most of our components so folks can add their own 🤔

In this instance, and the problem that David is facing, there's no real way for him to do that since ToastNotificationList is added in through the KaizenProvider 🤔 

I guess hard coding it may have to be the way forward since getByRole("status") isn't distinct enough?

mcwinter07: He could potentially use a querySelector to get the id of the node if we don't want to add that additional attribute, ie:

```
    const toastlistContainer = canvasElement.querySelector('#toast-notifications-list')
    expect(within(toastlistContainer).getByText("some text")).toBeVisible()
```

But no strong feeling against adding this back for testing purposes to make it easier 🤔 @NJWong

NJWong: Ah makes sense why it was removed

Yeah the issue that he's having is that when he does `getByRole("status")` there's another component in certain storybook stories that have another element with that same role. In his case it's the notification container from legacy kaizen `toast-notification-manager-portal`

The workaround I gave him was to use `getAllByRole` and then just manually get the element he's looking for in that array. Maybe since he's working with legacy UI that might have to be his long term solution

mcwinter07: Ah Yeah, I suppose we'd always run the risk of having more than one role status on the page - this is very much a edge case in the testing environment but there's no real harm in adding a test-id this for convenience - at least that way they get to use getByTestId? 🤷

mcwinter07: If you want to add a changeset I'm happy to approve, otherwise if David's ok with the querySelector Route that's hunky dory too :)

NJWong: I think he'll be okay with the querySelector approach - I think it's just for the purpose of getting old UI tests to work

I messaged him in the Slack thread but he's out sick for the rest of today so I'll wait for him to get back to me. For now I'll put a "Do not merge" on this PR 🚫

NJWong:
Closing because the `querySelector` workaround works for David Li's use case

PR #5679, opened 2025-04-17
Wrap component styles in @layer kz-components

sentience:

## Why

To unblock adoption of Tailwind CSS v4 in consuming apps, Kaizen component styles must be wrapped in `@layer components` that can be given lower priority than Tailwind's new `@layer utilities`. Without this change, consuming apps are unable to override Kaizen component styles with their own Tailwind-generated styles, since Kaizen's "unlayered" styles will always take priority over TW4 "layered" styles.

 Ideally this would not be necessary, and consuming apps could simply import Kaizen's styles into a layer:

 ```
 @import url(~@kaizen/components/dist/styles.css) layer(components);
 ```

Unfortunately, this syntax is [not supported](https://github.com/vercel/next.js/issues/55763#issuecomment-2791540126) by the outdated version of webpack css-loader that is contained in Next.js. Until it is, we must publish styles for packages like Kaizen with a layer provided.

## What

Wrap all CSS modules in `@layer components { … }`.

### Isn't this a breaking change?

Although technically this lowers the priority of Kaizen's component styles below any consuming apps' "unlayered" styles, regardless of selector specificity or source order, we believe that Kaizen styles are already output before app styles in apps' CSS bundles to ensure app styles take priority; therefore, this should have no effect on the relative priority of styles in consuming apps.

This reduces their priority against styles in consuming applications, which should override Kaizen component styles. This is particularly necessary to enable those applications to upgrade to Tailwind v4, which generates all its utility styles inside of a layer. Without this change, those TW4-generated styles would be of lower priority than Kaizen's unlayered styles, and therefore could not override Kaizen styles.


ryanseddon:
When it comes time to review this make sure you turn on hide whitespace, you can add `w=1` to the url e.g. lttps://github.com/cultureamp/kaizen-design-system/pull/5679/files?w=1

sentience:
> When it comes time to review this make sure you turn on hide whitespace, you can add `w=1` to the url e.g. https://github.com/cultureamp/kaizen-design-system/pull/5679/files?w=1

GitHub also has a nice UI for this nowadays:

CleanShot 2025-04-29 at 10 04 58@2x

sentience:
Looks like I've got some visual diffs to debug.

sentience:
Seeing some flaky focus states in Chromatic, but it does appear that this PR introduces one visual diff on`TextField`'s "success" variant:

CleanShot 2025-04-29 at 13 17 02

It looks to me like those green checkmark icons _should_ be visible in the Hover and Default states for the control, but aren't visible on main for some reason. Is it possible I've fixed a bug here?

sentience:
@mcwinter07 I just checked Storybook on main and I'm not actually seeing the missing checkmarks in `TextField`. I think this is just a flaky animation state.

mcwinter07:
> @mcwinter07 I just checked Storybook on main and I'm not actually seeing the missing checkmarks in `TextField`. I think this is just a flaky animation state.

Ah yeah, thanks for checking @sentience :) can confirm this is just a flaky ui test so will approve on chromatic

sentience:
Here's a weird one. I installed a canary build of this branch into PRC to test it, and [I'm getting a TypeScript error](https://github.com/cultureamp/performance-review-cycles-ui/actions/runs/14723012576/job/41320163071?pr=3392) complaining about the `overrideClassName` prop on the `Text` component:

> Error: src/BaseTable/BaseTable.tsx(193,17): error TS2322: Type '{ children: string; tag: "span"; variant: "body"; classNameOverride: string; }' is not assignable to type 'IntrinsicAttributes & TextProps'.
>   Property 'classNameOverride' does not exist on type 'IntrinsicAttributes & TextProps'.

Looking at the bundled types in the package, I'm seeing a subtle change to the Text.d.ts file:

**On main:**
```ts
import { type HTMLAttributes } from 'react';
import { type OverrideClassName } from '../types/components/types/OverrideClassName';
⋮
export interface TextProps extends OverrideClassName<HTMLAttributes<HTMLElement>> {
    children: React.ReactNode;
    /**
     * HTML elements that are allowed on Text
     * @default "p"
     */
    tag?: AllowedTextTags;
    /**
     * Allowed text variants
     */
    variant: TextVariants;
    color?: AllowedTextColors;
}
```

**On this branch:**
```ts
import { type HTMLAttributes } from 'react';
import { type OverrideClassName } from '~components/types/OverrideClassName';
⋮
export interface TextProps extends OverrideClassName<HTMLAttributes<HTMLElement>> {
    children: React.ReactNode;
    /**
     * HTML elements that are allowed on Text
     * @default "p"
     */
    tag?: AllowedTextTags;
    /**
     * Allowed text variants
     */
    variant: TextVariants;
    color?: AllowedTextColors;
}
```

Notice the path for the `OverrideClassName` import has changed from `../types/components…` to `~components/…`.

Any idea why this might have changed?

mcwinter07:
@sentience Was having a dig through the dist and I found two that need a `@layer component` wrapped around them. 

Screenshot 2025-04-29 at 3 03 56 pm

Screenshot 2025-04-29 at 3 07 12 pm

Happy to add these two if you're strapped for time :)

sentience:
> @sentience Was having a dig through the dist and I found two that need a `@layer component` wrapped around them.

I deliberately excluded `Select` because its styles need to override those bundled with React Select, which are not in a layer. I'll add a comment to this effect.

I'm guessing AIMoment must've landed on main after I branched. I'll add it.

docs/.storybook/preview-head.html
```diff
@@ -1,3 +1,14 @@
+<style>
+  /* Despite this being at the top of the first import in preview.tsx,
+     some component styles are still output before it in the static
+     Storybook build, which causes visual issues due to the `components`
+     layer being applied first.
+
+     Re-declaring the layer order here ensures it happens before any
+     JavaScript-generated styles. */
+  @layer tokens, normalize, reset, components;
```
mcwinter07, line 9: Makes sense. Presumably if we wanted to define any nested layers, this should function fine should still function fine

.changeset/forty-kids-cut.md
```diff
@@ -0,0 +1,19 @@
...
+
+```
+@import url(~@kaizen/components/dist/styles.css) layer(components);
+```
+
+Unfortunately, this syntax is [not supported](https://github.com/vercel/next.js/issues/55763#issuecomment-2791540126) by the outdated version of webpack css-loader that is contained in Next.js. Until it is, we must publish styles for packages like Kaizen with a layer provided.
+
+#### Isn't this a breaking change?
+
+Although technically this lowers the priority of Kaizen's component styles below any consuming apps' "unlayered" styles, regardless of selector specificity or source order, we believe that Kaizen styles are already output before app styles in apps' CSS bundles to ensure app styles take priority; therefore, this should have no effect on the relative priority of styles in consuming apps.
```
mcwinter07, line 19: Is there any chance that vendor styles might start winning out the specificity wars / is something we should flag when this gets released? 

I recall we had an [issue in mumur](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1737687211324029) a while back with bootstrap supersede some of our Link styles but not sure if this would be an issue in the new world of CSS layers 🍰

mcwinter07 (commented):
Looks good to me :) I'll hold on a ✅ until the `@layer` is added for the AiMoment stylesheet.

Left one comment on the changeset re: a general wondering about vender styles, but if its not likely an issue then happy to just urge caution for folks upgrading

.changeset/forty-kids-cut.md
sentience: Hmm yes, it's definitely a risk. Un-layered Bootstrap styles in Murmur will have greater priority than any styles in Kaizen following this change. Are you aware of any specific styles in Kaizen that are there to override Bootstrap in Murmur? If so, we could consider pulling them out of the `components` layer.

sentience:
Any ideas about [this](https://github.com/cultureamp/kaizen-design-system/pull/5679#issuecomment-2837466505), @mcwinter07?

.changeset/forty-kids-cut.md
mcwinter07: To my knowledge we don't have anything specific (least not from that). I think in that instance it was recommended that the team remove the styles that impact link

sentience: If such styles needed to stay in the consuming app, the other fix would be for the consuming app to move them into the `components` layer too.

mcwinter07:
Changes look good, I'm just going to test out the recent canary in unified home and a couple of my go to repos 👍 if it all looks hunky dory happy to give this a ✅

We'll also want to update a couple of sections in the [Guides](https://cultureamp.design/?path=/docs/guides-app-starter--docs) just to call out the CSS layers for future folks readers but I'll whip up a separate ticket for that

sentience:
@mcwinter07 I've tested this pretty extensively as a canary on my prc-ui branch. That branch still has a bunch of unwanted visual diffs, but invariably they are due to non-Kaizen packages that need their CSS updated or breaking changes in TW4 that we need to adjust to.

Are you happy to release this?

mcwinter07:
> @mcwinter07 I've tested this pretty extensively as a canary on my prc-ui branch. That branch still has a bunch of unwanted visual diffs, but invariably they are due to non-Kaizen packages that need their CSS updated or breaking changes in TW4 that we need to adjust to.

@sentience I just tested it out in home, have eyeballed any major ui changes in storybook, which seems hunky dory, but there does appear to be an issue running local dev:

```
Syntax error: /home/runner/work/unified-home/unified-home/node_modules/.pnpm/@kaizen+components@0.0.0-canary-add-components-layer-20250430022335_@cultureamp+i18n-react-in_3npx42isy4wa2l7nbvqzeulmpy/node_modules/@kaizen/components/dist/styles.css `@layer components` is used but no matching `@tailwind components` directive is present. (2:1)
```

You can see the error in full [here too](https://github.com/cultureamp/unified-home/actions/runs/14747915218/job/41398687278)) but I suspect that's because I haven't actually bumped the version of Tailwind in my branch to 4. Presumably that causes some kind of issue with then importing Kaizen's distributed styles into the tailwind.css 🤔

Is the upgrade to tailwind 4 a hard dependency for this release?

sentience:
> @sentience I just tested it out in home, haven't eyeballed any major ui changes in storybook, which seems hunky dory, but there does appear to be an issue running local dev:

_Interesting._ So what's happening here is that in Tailwind 3, the Tailwind compiler pre-processes the native "@layer components { … }" syntax and outputs all styles that were wrapped in `@layer components` at the point in your source code where you write `@tailwind components` – effectively transforming CSS Cascade Layers into source order.

In order to do this, PostCSS needs to be able to process all of your CSS code at once – so that it can see both the `@tailwind components` declaration (in the consuming app's tailwind.css file) and all the uses of `@layer components { … }` (now in Kaizen's styles).

Right now, most of our consuming apps are running PostCSS (including Tailwind) separately over individual CSS files, and _then_ letting Next.js's webpack css-loader combine these processed files into a single CSS bundle. But it looks like [Tailwind 3 requires us](https://v3.tailwindcss.com/docs/using-with-preprocessors#build-time-imports) to add [**postcss-import**](https://github.com/postcss/postcss-import) to our PostCSS config to combine all our CSS files together _before_ running Tailwind once across all of them.

Can you add **postcss-import** to the top of unified-home's postcss.config.js (as shown [here](https://v3.tailwindcss.com/docs/using-with-preprocessors#build-time-imports)) and see if that works?

> Is the upgrade to tailwind 4 a hard dependency for this release?

It shouldn't be! I was hoping not to require _any_ changes of existing apps who are still on Tailwind 3, but it looks like this **postcss-import** change is necessary.

mcwinter07:
> > @sentience I just tested it out in home, haven't eyeballed any major ui changes in storybook, which seems hunky dory, but there does appear to be an issue running local dev:
> 
> _Interesting._ So what's happening here is that in Tailwind 3, the Tailwind compiler pre-processes the native `@layer components { … }` syntax and outputs all styles that were wrapped in `@layer components` at the point in your source code where you write `@tailwind components` – effectively transforming CSS Cascade Layers into source order.
> 
> In order to do this, PostCSS needs to be able to process all of your CSS code at once – so that it can see both the `@tailwind components` declaration (in the consuming app's tailwind.css file) and all the uses of `@layer components { … }` (now in Kaizen's styles).
> 
> Right now, most of our consuming apps are running PostCSS (including Tailwind) separately over individual CSS files, and _then_ letting Next.js's webpack css-loader combine these processed files into a single CSS bundle. But it looks like [Tailwind 3 requires us](https://v3.tailwindcss.com/docs/using-with-preprocessors#build-time-imports) to add [**postcss-import**](https://github.com/postcss/postcss-import) to our PostCSS config to combine all our CSS files together _before_ running Tailwind once across all of them.
> 
> Can you add **postcss-import** to the top of unified-home's postcss.config.js (as shown [here](https://v3.tailwindcss.com/docs/using-with-preprocessors#build-time-imports)) and see if that works?
> 
> > Is the upgrade to tailwind 4 a hard dependency for this release?
> 
> It shouldn't be! I was hoping not to require _any_ changes of existing apps who are still on Tailwind 3, but it looks like this **postcss-import** change is necessary.

Got [some new errors](https://github.com/cultureamp/unified-home/actions/runs/14767183139/job/41460837400) so that's progress 😆 Having look now to see what might be causing the issue resolving the styles sheet import path.

I have a hunch it has something to do with having both imports and tailwind directive in the same file and causing some kind of issue here?

> One important thing to note about postcss-import is that it strictly adheres to the CSS spec and disallows @import statements anywhere except at the very top of a file.

There's a couple of recommendation in the postcss-import section of the tailwind docs so will see if that bears fruit

🕵️ cc: @sentience

sentience:
Ugh yeah, our imports that reference packages with a `~` prefix are using a webpack css-loader exclusive feature. They won't work with postcss-import.

Something that's worth trying is converting those to `url()` syntax:

**Before:**
```css
@import "~@kaizen/components/dist/styles.css";
```

**After:**
```css
@import url(~@kaizen/components/dist/styles.css);
```

sentience:
Reading various docs it looks to me like postcss-import requires this syntax:

```css
@import "@kaizen/components/dist/styles.css";
```

instead of the syntax required by css-loader:

```css
@import "~@kaizen/components/dist/styles.css";
```

Can you try removing the tildes on CSS imports in unified-home and see if that makes it work with postcss-import?

mcwinter07:
> Reading various docs it looks to me like postcss-import requires this syntax:
> 
> ```css
> @import "@kaizen/components/dist/styles.css";
> ```
> 
> instead of the syntax required by css-loader:
> 
> ```css
> @import "~@kaizen/components/dist/styles.css";
> ```
> 
> Can you try removing the tildes on CSS imports in unified-home and see if that makes it work with postcss-import?

Build seems to work that time, but I don't think the kaizen stylesheet resolved correctly 🤔

Screenshot 2025-05-01 at 1 55 42 pm

Gotta bounce to a few meetings this arvo but I've pushed the latest to the [branch up here](https://github.com/cultureamp/unified-home/tree/winter/test-css-layer-canary) if you want to take a poke. Will loop back at the tail end of the day :)

sentience:
Thanks! Yeah that looks no bueno.

I probably won't get time to look at that today as I'm crunching through the remaining TW4 visual diffs in performance-review-cycles-ui. Once I get to the end of that list, however, I can come back and look at this issue if you haven't already gotten to the bottom of it.

mcwinter07:
Hey @sentience bit of an update and recapping my investigation after trying to implement this with v3 tailwind in `unified-home` and `survey-config-ui` repos.

I've been looking into the issue with introducing the postcss-import into the postcss.config.js. 
Screenshot 2025-05-05 at 2 11 23 pm

The issue this seemed to introduce was in resolving the import paths when this syntax is used `~@kaizen/components` (also `@component/x` imports in unified home)

Screenshot 2025-05-05 at 2 23 06 pm

It looks like we can get around this by manually resolving the paths, ie: if "~" is found

Screenshot 2025-05-05 at 2 24 50 pm.

This worked alright pre-canary on survey-reporting-ui, however if I bump it over to the canary I get the original issue with the components layer:

Screenshot 2025-05-05 at 3 00 46 pm

Out of curiosity I did dig into the dist styles and do a quick find and replace for all the `@layer components` with `@layer kz-components`. This seemed to allow the build to pass and seemed to render with the relevant styles. so potentially that's a path forward? (noting we'd then have an additional layer).

There might be a big blind spot I'm missing so if you have time free tomorrow / this week I'd be keen to pair to talk through this :)

sentience:
Thanks for the investigation, @mcwinter07!

If the canary build produced the error again after postcss-import was introduced, that tells me that we are importing Kaizen's CSS from JavaScript instead of with a `@import` inside of the app's top-level CSS file (where postcss-import can process that `@import` into inline CSS). Can you check the apps you're playing with for Kaizen CSS imports in JavaScript and move them into CSS?

Yes, avoiding Tailwind's layer names is definitely an option, but we already have a large number of layers in play so I'd prefer to avoid adding more if we can.

mcwinter07:
> Thanks for the investigation, @mcwinter07!
> 
> If the canary build produced the error again after postcss-import was introduced, that tells me that we are importing Kaizen's CSS from JavaScript instead of with a `@import` inside of the app's top-level CSS file (where postcss-import can process that `@import` into inline CSS). Can you check the apps you're playing with for Kaizen CSS imports in JavaScript and move them into CSS?
> 
> Yes, avoiding Tailwind's layer names is definitely an option, but we already have a large number of layers in play so I'd prefer to avoid adding more if we can.

Hey @sentience still no real luck digging into this. I've move all the imports into the tailwind.css stylesheet so it should be running through postcss-import. While I've been able to resolve usages of `~@kaizen/components` with a resolver in the postcss.config, the styles do not appear in within an @components layer.

I did discover that the all other layers bar the `components` do seem to be coming through as you can see the tokens and reset seems to apply to the styles fine. I've create a quick vid to describe the pains for anyone taking a look.

https://github.com/user-attachments/assets/394d944e-1571-4535-900d-1c3765781658

I'm not sure where to go from here to validate that this will be ok for folks on Tailwind v3

sentience:
OK, I've just watched your video and I want to dig into that deeper when I get done with fixing all the visual diffs in prc-ui.

When you get the styles working at the end, you're changing two things: you're changing the layer name, but you're _also_ changing the way the stylesheet is imported (a relative `@import` vs the import with the `~package-name` format). I want to pin this down by changing only one of those, not both, at a time.

If I were in your place I'd be patching the Kaizen styles using `pnpm patch` to edit the Kaizen package stylesheet, rather than copying Kaizen styles into the local project.

Are you happy to wait for a day or two while I finish fixing these visual diffs, then I'll hit you up to pair on this further?

mcwinter07:
> Are you happy to wait for a day or two while I finish fixing these visual diffs, then I'll hit you up to pair on this further?

Yep, that's not a problem at all :)

sentience:
@mcwinter07 If you haven't already, I'd suggest having a close look at some of the Storybook stories for animations that are built with `@keyframes` in Kaizen's styles. I've just moved these out of the `@layer components` blocks because my reading suggests they are not valid inside of a `@layer`, but I'd like to confirm we haven't broken all the animations in Kaizen before we roll forward with this change.

mcwinter07:
> @mcwinter07 If you haven't already, I'd suggest having a close look at some of the Storybook stories for animations that are built with `@keyframes` in Kaizen's styles. I've just moved these out of the `@layer components` blocks because my reading suggests they are not valid inside of a `@layer`, but I'd like to confirm we haven't broken all the animations in Kaizen before we roll forward with this change.

@sentience Will do! I'll try testing it out again new Well and Card animation :) I'm also going to test the canary in the survey-config repo to validate everything's hunky dory in non-next environment

mcwinter07:
> > @mcwinter07 If you haven't already, I'd suggest having a close look at some of the Storybook stories for animations that are built with `@keyframes` in Kaizen's styles. I've just moved these out of the `@layer components` blocks because my reading suggests they are not valid inside of a `@layer`, but I'd like to confirm we haven't broken all the animations in Kaizen before we roll forward with this change.
> 
> @sentience Will do! I'll try testing it out again new Well and Card animation :) I'm also going to test the canary in the survey-config repo to validate everything's hunky dory in non-next environment

Ok, I've tested pretty extensively in Home today to validate the latest changes. I've posted a comment [here](https://github.com/cultureamp/unified-home/pull/3446#issuecomment-2899893367) detailing the look into the various unified-home pages. Based on that I'm quite confident in the latest update to this PR and its compatibility with with Next.JS repos, including the animation keyframe not being in the same layer. 

Learning from testing in unified-home and things we'll need to call out in a release announcement for this change is:
- There is hard dependancy on next-config@3.4.0 and above for next repos
  - This resolves some CSS source order issues created in Next.js's css minification
- The importance of checking dev builds and farms for normalize or any 3rd party library styles superseding Kaizen component in unexpected ways
  - If something targets an `a` or `button` tag it will likely win out as kaizen's distributed styles will have lower priority
  
I also tried testing in survey-config to validate the impact on React repos. I was able to get the local and deployed storybook running. Again, similar things like target anchor tags seemed to add styles but that was to be expected. These fortunately got [caught by chromatic](https://www.chromatic.com/build?appId=6287140297edd4004a4fed4f&number=2003). As for the dev farm - I don't think I stumbled upon which account was had the right data seeded for Aslan so I have't confirmed that yet.

sentience:
Cool! Let me know if you need me to do anything to get this over the line. Otherwise I'll leave this with you to shepherd through to release.

mcwinter07:
> Cool! Let me know if you need me to do anything to get this over the line. Otherwise I'll leave this with you to shepherd through to release.

All good, I'll spend this arvo getting this ready to release and clearing up some conflict from recent update to main. I'll do a quick pre-release in FE engineering and the Update Design System channel to make sure folks don't have any issues then aim to pull the trigger on the merge and release tomorrow

mcwinter07 (approved):
Final checks in unified home looks good so I'll approve this PR (albeit with some required updates to the normalize that is used).

Just going to check with the team before I merge and release this in case there's any fix / work they want to get out before this version since there will be dependencies in this release @sentience

PR #5702, opened 2025-04-23
feat(components): upgrade react-day-picker from 8.10.1 to 9.6.7

ryanseddon:
Update react-day-picker to version 9.6.7 and update all dependent components to work with the new API. Changes include:

## Why

To support react 19 we need to update react-day-picker to v9

## What

- Refactor component props to use the updated type definitions
- Update component rendering to match the new API requirements
- Modify CSS modules to accommodate new class naming conventions
- Adjust navigation components to use the new Chevron component API
- Add new styles for proper range highlighting and hidden requirements

react-day-picker changed quite a lot between major versions and how it applies class names to the table cell and button based on the type it is, so had to adjust some styles to target button children in some instances.

### UX change

Behaviour between v8 and v9 in respect to how it handles clicking the first day of an active date range has changed.

Instead if clearing the range like it did in v8 the range is now set to the first day.
To better understand what I mean I have two demos:

- v8: <https://codesandbox.io/p/devbox/rdp-v9-docs-example-forked-9d7kl5?file=/src/App.tsx&workspaceId=ws_B237FHzyHgonmzEroF4SKD>
- v9: <https://codesandbox.io/p/sandbox/nkmfsk?file=/src/App.tsx>

If you click the first day in each demo of the active range you'll see the difference.

https://github.com/user-attachments/assets/8eb636d2-1cb8-4cc0-acbd-72c31e437beb

### DIFF
--- a/packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
+++ b/packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
@@ -6,7 +6,7 @@ $cell-border-radius: 3px;
 .month {
   padding: 0 $spacing-sm;
 
-  &:first-child {
+  &:nth-of-type(1) {
     padding-inline-start: 0;
   }
 
@@ -19,7 +19,7 @@ $cell-border-radius: 3px;
   position: relative;
   padding: 0 $spacing-md;
 
-  &:not(:first-child) {
+  &:not(:nth-of-type(1)) {
     &::before {
       position: absolute;
       content: '';
@@ -28,7 +28,7 @@ $cell-border-radius: 3px;
     }
   }
 
-  &:first-child {
+  &:nth-of-type(1) {
     padding-inline-start: 0;
   }
 
--- a/packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
+++ b/packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
@@ -40,18 +40,9 @@ $cell-border-radius: 3px;
 .nav {
   position: absolute;
   display: flex;
-  justify-content: flex-start;
+  justify-content: space-between;
   width: 100%;
   color: $color-purple-800;
-
-  .captionEnd & {
-    flex-direction: row-reverse;
-  }
-
-  .captionStart.captionEnd & {
-    justify-content: space-between;
-    flex-direction: row;
-  }
 }
 
 .dayRangeStart,
@@ -63,4 +54,12 @@ $cell-border-radius: 3px;
 .dayRangeMiddle {
   background-color: $color-blue-100;
   color: $color-blue-500;
+
+  & [class*='_button_'] {
+    color: $color-blue-500;
+  }
+}
+
+.hidden {
+  background-color: transparent;
 }
--- a/packages/components/src/Calendar/CalendarRange/CalendarRange.tsx
+++ b/packages/components/src/Calendar/CalendarRange/CalendarRange.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef } from 'react'
 import { enAU } from 'date-fns/locale'
-import { DayPicker, type ClassNames, type PropsBase, type PropsRange } from 'react-day-picker'
+import { DayPicker, type PropsBase, type PropsRange } from 'react-day-picker'
 import { Icon } from '~components/__next__/Icon'
 import { type OverrideClassName } from '~components/types/OverrideClassName'
 import { baseCalendarClassNames } from '../baseCalendarClassNames'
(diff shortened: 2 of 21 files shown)


packages/components/src/Calendar/LegacyCalendarRange/LegacyCalendarRange.tsx
```diff
@@ -45,7 +45,7 @@ export const LegacyCalendarRange = ({
     day_range_start: styles.dayRangeStart,
     day_range_end: styles.dayRangeEnd,
     day_range_middle: styles.dayRangeMiddle,
-  } satisfies DayPickerRangeProps['classNames']
+  } as ClassNames
```
NJWong, line 48: Same as above - although when I do this:

```
const classNames: Partial<ClassNames> = {
```

But then I get a type error with `nav_button_next` - maybe it's meant to be `button_next`?

packages/components/src/Calendar/CalendarRange/CalendarRange.tsx
```diff
@@ -39,13 +39,12 @@ export const CalendarRange = ({
...
     nav: styles.nav,
-    day_range_start: styles.dayRangeStart,
-    day_range_end: styles.dayRangeEnd,
-    day_range_middle: styles.dayRangeMiddle,
-  } satisfies DayPickerRangeProps['classNames']
+    range_start: styles.dayRangeStart,
+    range_end: styles.dayRangeEnd,
+    range_middle: styles.dayRangeMiddle,
+    hidden: styles.hidden,
+  } as ClassNames
```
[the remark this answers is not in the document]
NJWong: Or potentially `PropsBase['classNames']` would work because that's what `baseCalenderClassNames` is using - might be better to align to that

packages/components/src/Calendar/LegacyCalendarRange/LegacyCalendarRange.tsx
ryanseddon: Went with the PropsBase["classNames"] and I get no complaints but the lsp does mention it is deprecated in favour of button_next. Good catch.

I also dug in to the types and can see it here:
Screenshot 2025-05-07 at 4 19 32 pm

NJWong (approved):

mcwinter07 (commented):
Seems to be some odd thing happening to the styles in the DateRangePicker

**Before**
Screenshot 2025-05-08 at 2 52 00 pm

**After**
Screenshot 2025-05-08 at 2 52 07 pm

They seem ok in the FilterDateRangePicker

Also did chromatic run for this PR or was it skipped it because we're merging into a branch instead of main

packages/components/src/Calendar/baseCalendarClassNames.ts
```diff
@@ -1,22 +1,20 @@
 /* eslint-disable camelcase */
-import { type ClassNames } from 'react-day-picker'
+import { type PropsBase } from 'react-day-picker'
 import styles from './baseCalendarClassNames.module.scss'
 
-export const baseCalendarClassNames: ClassNames = {
-  vhidden: styles.vHidden,
+export const baseCalendarClassNames: PropsBase['classNames'] = {
```
mcwinter07, line 5: I wonder if this is the cause of some of the DateRangerPicker style issues since this is shared in both 🤔 oddly enough it looks like the button next doesn't appear to get any of the classes applied to it, where as previous does?
Screenshot 2025-05-08 at 3 27 48 pm

mcwinter07: ooo dug in a little more and think the issue lies in the `LegacyCalendarRange` classnames  :) We probably just need to update these to the new namespace
Screenshot 2025-05-08 at 3 43 13 pm

ryanseddon:
@mcwinter07 chromatic hasn't ran due to this PR merging into another branch, we should probably have a label for situations like this

packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
```diff
@@ -28,7 +28,7 @@ $cell-border-radius: 3px;
     }
   }
 
-  &:first-child {
+  &:nth-of-type(1) {
```
ryanseddon, line 31: rdpv9 has the nav inside the month container now so first-child didn't match

mcwinter07:
> @mcwinter07 chromatic hasn't ran due to this PR merging into another branch, we should probably have a label for situations like this

@ryanseddon Potentially we can add another check to our run-check for the chromatic job. I know other teams use a `run_chromatic` label to trigger their chromatic builds. They mainly use this as a means of reducing the number of snapshot chromatic takes but add could do something like this on top of our current check in the `chomatic-prop.yaml`:

```
      github.event_name == 'pull_request'
      && github.event.action == 'labeled'
      && contains(github.event.pull_request.labels.*.name, 'run_chromatic')
```

packages/components/src/Calendar/CalendarRange/CalendarRange.module.scss
mcwinter07: Ah that's we're it got us 🤔

mcwinter07 (approved):

PR #5749, opened 2025-05-05
fix: remove root-level React dependencies 

ryanseddon:
## Why

To support react 19 

## What

There seems to hard dependency on react and react-dom in the top level package.json that may be left over from pre KAIO days. From my testing there seems to be no real use for it to be there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,10 +36,6 @@
     "i18n:extract": "pnpm pkg:aio i18n:extract",
     "i18n:extract:remove-obsolete": "pnpm pkg:aio i18n:extract --deleteObsoleteTranslations"
   },
-  "dependencies": {
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0"
-  },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -104,11 +100,25 @@
...
         "react-focus-on>react": "19",
         "react-focus-on>react-dom": "19",
         "react-popper>react": "19",
         "react-popper>react-dom": "19",
+        "react-remove-scroll>react": "19",
+        "react-remove-scroll>react-dom": "19",
+        "react-remove-scroll-bar>react": "19",
+        "react-remove-scroll-bar>react-dom": "19",
+        "react-style-singleton>react": "19",
+        "react-style-singleton>react-dom": "19",
```
ryanseddon, line 121: I had to add a few more transient deps of some of our older packages. I want to be explicit here and not do a catch all so we know when we remove these we can also clean up here

mcwinter07 (approved):

PR #5750, opened 2025-05-08
Update filter multi select using popover api

thanhnguyenp26:

## Why

- Our team is having a customer problem with the dropdown menu filters in the sticky sidebar. When the dropdown box is too long, it gets cut off at the bottom and prevents scrolling.
- Ticket: https://cultureamp.atlassian.net/browse/CABB-4885

## What

- We would want to make the dropdown to float up when there's limited space below it.
- Remove tests where we click the trigger button when the dropdown menu is open. The button becomes inaccessible when the dropdown menu is open due to the use of FocusOn on the dropdown menu.
- Also wanted to try out popover api.


thanhnguyenp26:
Thanks for reminding @ryanseddon. I've updated the PR description.

packages/components/src/Filter/FilterMultiSelect/context/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
```diff
@@ -86,17 +70,6 @@ describe('<MenuTriggerProvider /> - Mouse interaction', () => {
...
-    it('is closed when user clicks on the trigger', async () => {
-      render(<MenuTriggerProviderWrapper defaultOpen />)
-      const trigger = screen.getByRole('button', {
-        name: 'trigger-display-label-mock',
-      })
-      await user.click(trigger)
-      await waitFor(() => {
-        expect(screen.queryByText('menu-content-mock')).not.toBeInTheDocument()
-      })
-    })
```
diffidentDude, line 98: We probably want this test to continue to work right? ie, it should close when you click the trigger

packages/components/src/Filter/FilterMultiSelect/context/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
```diff
@@ -53,21 +52,6 @@ describe('<MenuTriggerProvider /> - Visual content', () => {
...
-      const trigger = screen.getByRole('button', {
-        name: 'trigger-display-label-mock',
-      })
-      await user.click(trigger)
-
-      await waitFor(() => {
-        expect(onOpenChange).toBeCalledTimes(1)
-        expect(onOpenChange).toBeCalledWith(false)
-      })
-    })
```
diffidentDude, line 70: this one should also probably still work right?

packages/components/src/Filter/FilterMultiSelect/context/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
thanhnguyenp26: @diffidentDude with the use of FocusOn on the menu popup element, I think all the non-focused elements get assigned `pointer-events: none` so can't click the trigger. We also close the menu when clicking outside with onClickOutside, and clicking the trigger button also means clicking outside?

diffidentDude: so this test should still work, we just replace the click trigger with click outside?

thanhnguyenp26: @diffidentDude there already exists a test about clicking outside to close the dropdown, which still works. The two tests I removed were clicking the trigger, which is now assigned `pointer-events: none`.

diffidentDude: ah k, 👍🏻

packages/components/src/Filter/FilterMultiSelect/context/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
diffidentDude: this test probably just needs a new way to close it, and it should still work

packages/components/src/Filter/FilterMultiSelect/subcomponents/MenuPopup/MenuPopup.tsx
```diff
@@ -1,58 +1,100 @@
...
+        className={classnames(styles.menuPopup, classNameOverride)}
+        role="dialog"
+        aria-modal="true"
+        // @ts-expect-error: popover is valid in supported browsers
+        popover="manual"
+        {...restProps}
+      >
+        {isLoading && loadingSkeleton ? loadingSkeleton : children}
+      </div>
+    </FocusOn>
```
diffidentDude, line 97: There is a little mislink between these two. In the render, we only render the popover if it's open, but in the useEffect we have logic for handling if it's open or not. We should choose one approach or the other

thanhnguyenp26: @diffidentDude if we remove the conditional rendering, but the tests checking if the menu is open/closed will fail.
Whether the dropdown is open or not, it still exists in the dom, so it'll pass the test `toBeInTheDocument()`. I tried with `toBeVisible()` but I think `display` is manipulated by `:popover-open` pseudo-class (it says popover-over is not supported in the test env), so the visibility check doesn't behave as expected.

diffidentDude: Because the visibility is managed by a useEffect, you may need an `act` to ensure the useEffect has run?

thanhnguyenp26: I wonder if that works. This initial test already fails without any user interaction:
```
it('does not show the menu initially', async () => {
      render(<MenuTriggerProviderWrapper />)
      await waitFor(() => {
        const menu = screen.queryByText('menu-content-mock')
        expect(menu).not.toBeVisible()
      })
    })
```
```
 → expect(element).not.toBeVisible()

Received element is visible:
  <span />
```

diffidentDude: Ok, maybe it would be best to put the conditional back, and replace the useEffect with one that just shows it?

thanhnguyenp26: @diffidentDude yup. it works

packages/components/src/Filter/FilterMultiSelect/context/MenuTriggerProvider/MenuTriggerProvider.spec.tsx
thanhnguyenp26: @diffidentDude I've modified the way to close it and it works.

packages/components/src/Filter/FilterMultiSelect/subcomponents/MenuPopup/MenuPopup.tsx
```diff
@@ -1,58 +1,98 @@
...
+    requestAnimationFrame(() => {
+      buttonRef.current?.focus()
+    })
+  }
+
+  useEffect(() => {
+    if (menuTriggerState.isOpen && floatingElement && referenceElement) {
+      floatingElement.showPopover?.()
+      update()
+    }
```
diffidentDude, line 71:
```suggestion
    if (floatingElement && referenceElement) {
      floatingElement.showPopover?.()
      update()
    }
```

diffidentDude: I would get rid of the isOpen check personally

thanhnguyenp26: @diffidentDude  ah nice, thank you!

diffidentDude (approved):

packages/components/src/Filter/FilterMultiSelect/subcomponents/MenuPopup/MenuPopup.tsx
```diff
@@ -1,58 +1,98 @@
...
+      returnFocus={false}
+      onClickOutside={menuTriggerState.close}
+      onEscapeKey={menuTriggerState.close}
+      onDeactivation={handleReturnFocus}
+    >
+      <div
+        ref={setFloatingElement}
+        style={floatingStyles}
+        className={classnames(styles.menuPopup, classNameOverride)}
+        role="dialog"
```
mcwinter07, line 87: Since we're creating an element with the role dialog, we'll need to give it an accessible name where it's implemented in the `FilterMultiSelect.tsx`, i.e:

```JSX
// ...
    <MenuTriggerProvider {...menuTriggerProps}>
      <div className={className}>
        <MenuTriggerConsumer>{trigger}</MenuTriggerConsumer>
        <MenuPopup {...menuPopupProps} // needs accessible name>
          <SelectionProvider {...selectionProps}>
            <SelectionConsumer>{children}</SelectionConsumer>
          </SelectionProvider>
        </MenuPopup>
      </div>
    </MenuTriggerProvider>
```

I'm thinking that the `label` prop in the FilterMultiSelect should be enough to provide context to what the dialogue belongs to. Would recommend potentially taking a squiz at the [dialog pattern on W3](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/)

mcwinter07 (commented):
Hey @thanhnguyenp26 thanks for your work on this PR. I've left a few comments to look into but overall the update is looking pretty good.

I did have some usability concerns around the double scroll bar that gets introduced by this update in reduced window sizes. 

Screenshot 2025-05-22 at 10 44 31 am

Generally we want to avoid multiple scroll bars within a single element as it can increase cognitive load and visually busy up the UI. Is there any way we can keep to one scroll bar like in the current implementation?

packages/components/src/Filter/FilterMultiSelect/subcomponents/MenuPopup/MenuPopup.module.scss
```diff
@@ -8,7 +8,6 @@ $menu-container-width: 294px;
 $menu-container-max-height: 312px;
 
 .menuPopup {
-  position: absolute;
   z-index: 1000; // from $ca-z-index-dropdown
```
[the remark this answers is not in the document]
thanhnguyenp26: @mcwinter07 I think we should keep it for the browsers that don't support popover.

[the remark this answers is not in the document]
mcwinter07: Ah gotcha 👍 Looking at the [browser coverage](https://caniuse.com/?search=popover) for the popover API it looks pretty good at the moment, but understand the caution since the API still pretty fresh.

Happy to leave it in, but can you add a comment so we don't remove it in the future without checking :)

packages/components/src/Filter/FilterMultiSelect/subcomponents/MenuPopup/MenuPopup.tsx
```diff
@@ -1,58 +1,98 @@
...
+      floating: floatingElement,
     },
-    overlayRef,
-  )
+    strategy: 'fixed',
+    middleware: [
+      size({
+        apply({ availableHeight, availableWidth, elements }) {
+          Object.assign(elements.floating.style, {
+            maxHeight: `${Math.max(availableHeight - 25, 350)}px`,
```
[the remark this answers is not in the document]
mcwinter07: Hmmm that's a good point, we can use Math.min instead and then it would stop at 350px.

I was also revisiting the designs and behavior of the FilterMultiSelect before these update and there a distinct change in where the search input and footer buttons sit. 

**Design**
Screenshot 2025-05-27 at 1 57 29 pm

**Pre-update implementation**
Screenshot 2025-05-27 at 1 59 20 pm

Here we can see that they are both always present and the list content is the only thing that scrolls.

I think we can achieve this by setting specific max heights - it'll still have a double scroll bar when there's limited room available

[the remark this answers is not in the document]
mcwinter07: Hey @thanhnguyenp26 👋 Hiding the menu scrollbar by reducing the width to 0 would likely introduce an accessibility issue if there is overflown content but no indication to users. It may be that we just have to accept that on reduced screen sizes we have two scrolls, since that would be better than a hidden one. 

I'd be keen to pair tomorrow if you have availability so we can exhaust any other avenue and get the last of this PR over the line 😄

NJWong (dismissed):

NJWong (approved):

PR #5791, opened 2025-05-27
[KZN-3169] Fix for flaky storybook Tooltip test

Kitty-Al:
## Why
Bugfix: https://cultureamp.atlassian.net/browse/KZN-3169 
Flaky storybook test for the Tooltip: Tooltip test was failing erratically on `toBeVisible`

## What

Updated the Storybook documentation for the Tooltip to use the Button next and Tabs next components 
Added a fix for the failing test using `checkVisibility` on the Tooltip element - seems to be passing locally on every test


mcwinter07 (approved):
Awesome stuff! Looking forward to not having to re-trigger those the test-storybook actions 🙇

PR #5817, opened 2025-06-06
[KZN-3337] handles async loading of tabs

Kitty-Al:
## Why

When multiple tabs are rendered asynchronously we are not updating where the right/left chevrons should intersect. Therefore if you start with one tab and load more the extra tabs will not cause the chevron to render. 

## What

Updating the use effect so that it re-renders based on the number of tabs.

### DIFF
--- a/packages/components/src/__next__/Tabs/subcomponents/TabList/TabList.tsx
+++ b/packages/components/src/__next__/Tabs/subcomponents/TabList/TabList.tsx
@@ -100,7 +100,7 @@ export const TabList = (props: TabListProps): JSX.Element => {
       firstTabObserver.disconnect()
       lastTabObserver.disconnect()
     }
-  }, [isDocumentReady, containerElement, isRTL])
+  }, [isDocumentReady, containerElement, isRTL, tabListContext?.collection.size])
 
   useEffect(() => {
     if (!isDocumentReady) {


packages/components/src/__next__/Tabs/subcomponents/TabList/TabList.tsx
```diff
@@ -100,7 +100,7 @@ export const TabList = (props: TabListProps): JSX.Element => {
       firstTabObserver.disconnect()
       lastTabObserver.disconnect()
     }
-  }, [isDocumentReady, containerElement, isRTL])
+  }, [isDocumentReady, containerElement, isRTL, tabListContext?.collection.size])
```
mcwinter07, line 103: Had a quick poke to test this locally with the same scenario Viet had and looks like the overlap still happens when the character length of the tab names are just the right (or wrong? 😆 ) size.

Screenshot 2025-06-06 at 11 58 30 am

Story example:

```jsx
export const TestInitial: Story = {
  render: (args) => {
    const [selectedKey, setSelectedKey] = useState<Key>(0)
    return (
      <>
        <Tabs {...args} selectedKey={selectedKey} onSelectionChange={setSelectedKey}>
          <>
            <TabList aria-label="Tabs">
              <Tab id="one">Conversational</Tab>
              <Tab id="two">Personal</Tab>
            </TabList>
            <TabPanel id="one" className="p-24">
              <Text variant="body">Content 1</Text>
            </TabPanel>
            <TabPanel id="two" className="p-24">
              <Text variant="body">Content 2</Text>
            </TabPanel>
          </>
        </Tabs>
        <Button label="Switch to tab 2" onClick={(): void => setSelectedKey('two')} />
      </>
    )
  },
}
```

I gave crack at adding an async state update as well to see if that then calculate the chevron's appearance correctly but it looks like it doesn't consider it overlapping.

mcwinter07: I'd say its also probably worth have a mobile story for this so we can capture the 320px scenario as a snapshot too.

Kitty-Al: In Viet's case the second tab text was `Personal notes` so I think that would be partially fixed with the re-render. But I see the problem you're talking about with the intersection not being fine grained enough and missing the last letter fully if the letter is skinny like i l or 1 and partially cutting other letters off. Updating the threshold to `0.8` seems to handle that situation.

packages/components/src/__next__/Tabs/_docs/Tabs.stories.tsx
```diff
@@ -66,11 +66,57 @@ export const Playground: Story = {
...
+            <Text variant="body">Content 2</Text>
+          </TabPanel>
+        </Tabs>
+        <Button label="Switch to tab 2" onClick={(): void => setSelectedKey('two')} />
+      </>
+    )
+  },
+}
+
+export const AsyncMobileView: Story = {
```
mcwinter07: Since we're already validating that asynchronously loaded tabs will render the arrow right button in the `AsyncLoaded` play Story, I reckon we can either remove this or change this to a `MobileView` story with a few more tabs so we can use it in a mdx in the docs to describe the mobile behavior.

Kitty-Al: Sounds good to me, I've taken out the Story and have left the test

mcwinter07 (approved):

PR #5818, opened 2025-06-06
Modal stacking order

aaronknol:

### DIFF
--- a/.changeset/light-nights-tell.md
+++ b/.changeset/light-nights-tell.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+Add relative position to the top level parent of the Modal components to ensure they are on top of the whole application
--- a/packages/components/src/Modal/GenericModal/GenericModal.module.scss
+++ b/packages/components/src/Modal/GenericModal/GenericModal.module.scss
@@ -68,6 +68,8 @@
   .transitionLayer {
     transition: opacity $ca-duration-fast ease-in-out;
     opacity: 1;
+    position: relative;
+    z-index: $ca-z-index-modal;
 
     .backdropLayer {
       /*


ryanseddon:
Thanks @aaronknol.

So this is happening due to how the browser tries to optimise animations on the gpu. 

When closing the modal it applies a opacity transition which the browser then promotes the transition layer to a GPU layer in the browser rendering context and when that happens the z-index of the backdrop within that layer only applies to within that layer and not the whole document anymore so 10 is higher than 0 and therefore the sticky header appears above.

We can force this GPU layer and see the issue when the modal is open by applying `will-change: opacity` which tells the browser to create the gpu layer straight away rather than JIT it.

Screenshot 2025-06-10 at 9 59 21 am

So long story your fix is the right one! Long term we'll most likely move modals to leverage the built-in dialog which utilises the browser top-layer feature which negates all these esoteric stacking context issues!

ryanseddon:
I think it'd be good to also add some classes to the genericmodal storybook tests so if for whatever reason it'll break the visual tests if this is undone or overridden.

```diff
diff --git a/packages/components/src/Modal/GenericModal/_docs/GenericModal.spec.stories.tsx b/packages/components/src/Modal/GenericModal/_docs/GenericModal.spec.stories.tsx
index fe216ec54..c2a88f0c1 100644
--- a/packages/components/src/Modal/GenericModal/_docs/GenericModal.spec.stories.tsx
+++ b/packages/components/src/Modal/GenericModal/_docs/GenericModal.spec.stories.tsx
@@ -21,7 +21,11 @@ export const TestBase: Story = {
 
     return (
       <>
-        <button type="button" className="border border-gray-500" onClick={() => setIsOpen(true)}>
+        <button
+          type="button"
+          className="border border-gray-500 relative z-10"
+          onClick={() => setIsOpen(true)}
+        >
           Open Modal
         </button>
         <GenericModal
@@ -30,6 +34,7 @@ export const TestBase: Story = {
           onOutsideModalClick={handleDismiss}
           onEscapeKeyup={handleDismiss}
           id="GenericModalTestId"
+          className="will-change-[opacity]"
         >
           <ModalHeader>
             <ModalAccessibleLabel>Test Modal</ModalAccessibleLabel>
```

Add `relative z-10` to the button classes and `will-change-[opacity]` className to the GenericModal component.

ryanseddon (approved):

PR #5850, opened 2025-06-16
fix(Select/next): pass aria-required into Select trigger props 

mcwinter07:
## Why
`select/next` is currently missing the `aria-required` attribute on the Select trigger despite the `isRequired` prop being used. 

While this can impact the ability to submit forms if you pass in `validationBehavior="native"`, not having this attribute means that assistive technologies cannot determine the `required` nature of the select when used in forms.

See [support thread here](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1749615222734579) for more details.

## What

- Updates implementation to pass `isRequired` into the `selectToggleProps` as the value for `aria-required`. 
  - Add interaction tests to validate `isRequired` and native form submission
- Update Stories and API documentation to clarify the behaviour of `isRequired` and how to leverage `validationBehaviour` prop for native form submission
  -  This was not explicitly called out before and I feel like we'll get this question more as people migrate their Selects used in forms that are handling validation in using another library like react-hook-form

### DIFF
--- a/packages/components/src/__next__/Select/_docs/Select.mdx
+++ b/packages/components/src/__next__/Select/_docs/Select.mdx
@@ -98,7 +98,7 @@ When using the `isRequired` property you can also specify the `validationBehavio
 
 <Canvas of={SelectStories.SelectNativeValidationBehavior} />
 
-While both use `aria-required` to announce whether the field has to have value to assistive technologies, the `native` will option will prevent form submissions if the `selectedKey` is `undefined`.
+While both use `aria-required` to announce whether the field has to have a value to assistive technologies, the `native` will option will prevent form submissions if the `selectedKey` is `undefined`.
 
 ### Full width
 


mcwinter07:
I've tested the canary against a local version of home on windows with JAWs - can confirm this no longer announces as role menu would should be a good last step for this component before it is removed

packages/components/src/__next__/Select/_docs/Select.mdx
```diff
@@ -92,6 +92,14 @@ Add validation messages using `status` and `validationMessage`.
 
 <Canvas of={SelectStories.Validation} />
 
+#### isRequired and validationBehavior
+
+When using the `isRequired` property you can also specify the `validationBehavior` to change from `aria` to `native` form validation.
+
+<Canvas of={SelectStories.SelectNativeValidationBehavior} />
+
+While both use `aria-required` to announce whether the field has to have value to assistive technologies, the `native` will option will prevent form submissions if the `selectedKey` is `undefined`.
```
ryanseddon:
```suggestion
While both use `aria-required` to announce whether the field has to have a value to assistive technologies, the `native` will option will prevent form submissions if the `selectedKey` is `undefined`.
```

ryanseddon (approved):

mcwinter07:
Thanks for the review and catching that error in the docs @ryanseddon I've merged in the fix so should be good for a ✅ 🤞😬🤞

ryanseddon (approved):

PR #5851, opened 2025-06-17
fix(LikertScaleLegacy)Pass isRequired to handle required fields

Livia-Santos:
## Why

**LikertScaleLegacy - fix accessibility issue identified by the Intopia audit** 

Required questions are marked as required using asterisks. However, the form elements are not marked as required in the underlying HTML, which means assistive technologies (like screen readers) cannot programmatically determine which questions are mandatory. This creates a gap in accessibility compliance and user experience for those relying on assistive tools.

See audit item: [CUL-743 – Required field cannot be programmatically determined](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/5043421650/A11Y+-+Survey+Capture+Intopia+Audit+Review#8.-Required-field-cannot-be-programmatically-determined-(CUL-743))

JiraCard: https://cultureamp.atlassian.net/browse/ESD-3151

## What

- Updated the LikertScaleLegacy component to handle the isRequired prop by applying `aria-required` to the `radiogroup` element when necessary.

- Added interaction and unit tests to validate the presence of the isRequired attribute.

- Included a changeset to document the update.


packages/components/src/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -109,6 +111,7 @@ export const LikertScaleLegacy = ({
       tabIndex={-1}
       aria-describedby={validationMessageId}
       data-testid={dataTestId}
+      aria-required={isRequired ? true : undefined}
```
mcwinter07: Since `isRequired` is optional it will return undefined if not provided. Could we instead pass in `isRequired` and not set the default value on ln50?

packages/components/src/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -25,6 +25,7 @@ export type LikertScaleProps = {
   'colorSchema'?: ColorSchema | 'classical'
   'validationMessage'?: string
   'status'?: 'default' | 'error'
+  'isRequired'?: boolean
```
mcwinter07, line 31: Since this will only be used to update the aria-required attribute it might be worth calling this out in the type description as it won't prevent form submissions. 

Something like: 
> Sets `aria-required` value on `radiogroup` for assistive technologies. Validation must still be handled.

mcwinter07 (commented):
Heya @Livia-Santos, left a couple of comments on the PR, but was also thinking we could use a Story. I've whipped up a video to talk through some point screen reader issues that can be solved with an accessible label:

https://github.com/user-attachments/assets/a7f3cf0b-a097-481d-9912-4b010dea16eb

Shout out if you want to book in some time to pair on this tomorrow or if you're happy to add a story with some guidance solo :)

mcwinter07 (approved):
Nice work @Livia-Santos Thanks for the contribution 🙇‍♂️

PR #5853, opened 2025-06-20
Update BrandMomentCaptureIntro animation toggle

mcwinter07:
## Why
In the current implementation of `BrandMomentCaptureIntro` the toggle to stop the loop animation is hidden behind a hover or focus event. 

This presents an issue for individuals that have motion sensitivity as they may not know they can pause the animation. This was flagged as a [medium severity issue](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/5043421650/A11Y+-+Survey+Capture+Intopia+Audit+Review) in the recent Intopia Audit of the Capture Survey experience.

While this is technically the case with all of our animate-able `Scene` components, this PR only addresses the `BrandMomentCaptureIntro` component to limit the scope of impact. This is to allow for a potential design solution or rethink of the animation toggle styles and position that has design curators approval, before it is rolled out across the product.

## What

- Adds optional `hasVisibleAnimationToggle` prop to `VideoPlayer`
  - This enabled a style override in the `BrandMomentCaptureIntro` component that removes the initial opacity 0% that hides the animation toggle
  -

### DIFF
--- a/.changeset/sweet-news-search.md
+++ b/.changeset/sweet-news-search.md
@@ -1,0 +1,10 @@
+---
+'@kaizen/components': patch
+---
+
+Add optional prop to `VideoPlayer` and update `BrandMomentCaptureIntro` to always display the animation toggle.
+
+- Add `hasVisibleAnimationToggle` prop to `VideoPlayer`
+- Update VideoPlayer to next `Button` component
+- Remove redundant styles from button migration
+- Add i18n strings for play/pause labels
--- a/packages/components/locales/en.json
+++ b/packages/components/locales/en.json
@@ -188,5 +188,13 @@
   "splitButton.dropdownButton.label": {
     "description": "Label for a dropdown menu holding additional actions",
     "message": "Additional actions"
+  },
+  "videoPlayer.pausePlayBtn.pauseLabel": {
+    "description": "Label for the pausing / stopping an animation",
+    "message": "Play animation"
+  },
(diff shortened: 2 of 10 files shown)


mcwinter07:
I've create a branch we're I've partially implemented the idea Lua had made a [here](https://cultureamp.slack.com/archives/C05D8V4GALX/p1750394068033439) since its less aggressive than what we had before -  Branch [here](https://github.com/cultureamp/kaizen-design-system/pull/5868).

The Button is using the secondary variant and I've also added an onClick to the VideoPlayer to toggle the animation.

**Example - no hover**
Screenshot 2025-06-23 at 11 37 30 am

**Example - hover**
Screenshot 2025-06-23 at 11 37 37 am

packages/components/src/Illustration/subcomponents/Base/Base.module.css
```diff
@@ -0,0 +1,97 @@
...
+  .aspectRatioWrapper {
+    margin: 0;
+    display: flex;
+    align-items: center;
+  }
+
+  .landscape {
+    aspect-ratio: 4/3;
+
+    @supports not (aspect-ratio: auto) {
```
ryanseddon, line 47: I think we can safely use this without a fallback as it's a baseline widely avaiable feature and it landed 4+ years ago in most browsers

ryanseddon (approved):

PR #5869, opened 2025-06-25
[KZN-3418] SingleSelect folder setup

Kitty-Al:
## What

Sets up the folder structure and the basic working example of the React-Aria-Component `Select`. This is the bare bones for this component, styling and complexity will change over time. 

Subcomponents included:

- List
- ListItem
- ListSection

Users can either pass their data in as `items` or render them themselves using the subcomponents`SingleSelect.ListItem` etc. 

**Alpha release**
Setup done for alpha release 
Component released from alpha
Added an inline notification for alpha components to warn consumers against using in production


packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+export type ItemGroup = {
+  label: string
+  options: Item[]
+}
+
+export type Items = {
+  items?: ItemGroup[] | Item[]
+}
+
+export type SingleSelectProps = {
```
ryanseddon, line 7: I would pare this component back to the basics and drop the labelText, inline, validationMessage etc we want to just introduce the basic folder structure and not anything that isn't intentional to this set PR

packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  inline = false,
+  validationMessage,
+  status,
+  description,
+  classNameOverride,
+  children,
+  ...restProps
+}: PropsWithChildren<SingleSelectProps>): JSX.Element => {
+  return (
+    <FieldGroup inline={inline}>
```
ryanseddon: Avoid this as this should be a consumer thing to wrap the select in the field component

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.module.scss
```diff
@@ -0,0 +1,9 @@
+@import '~@kaizen/design-tokens/sass/typography';
```
ryanseddon: We're moving away from sass so we should start with these being css modules

packages/components/src/__alpha__/SingleSelect/subcomponents/PopoverContents/PopoverContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+type Items = {
+  items: ItemGroup[] | Item[]
+}
+
+export const PopoverContent = ({ items }: Items): React.ReactElement => {
+  return (
+    <List>
+      {items.map((item) => {
+        if ('options' in item) {
```
ryanseddon: I think remove this for now too so when we get to tackling recursive menus we can do it then, so the initial stories should just be very simple too

ryanseddon (commented):
Looking good, I think if we just simplify the initial files to bare essentials to get the basics going that would be best

PR #5881, opened 2025-07-01
Deprecation notices

ryanseddon:
## Why
 
This PR adds all the missing deprecation notices for v2 and will warn people via the JSDoc comments in their IDEs and with notification banners in the storybook docs.

## What

This adds two new Notification components to storybook docs to make it clear whether a component is being removed or renamed.

I also added deprecations to the storybook nav for affected items.


packages/components/src/LikertScaleLegacy/LikertScaleLegacy.tsx
```diff
@@ -37,6 +37,9 @@ const SelectedItemIcon = (): JSX.Element => (
 )
 
 /**
+ * @deprecated This component will be renamed to LikertScale in v2.
+ * Start importing as LikertScale instead. Codemod coming soon.
```
ryanseddon: Should I say it's coming soon or just remove that?

Kitty-Al: I think it's best to remove the "Codemod coming soon", that way we don't have to go back in & update when/if the codemod is delivered

Kitty-Al (approved):

Kitty-Al (approved):

docs/components/RenameNotice/RenameNotice.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Unstyled } from '@storybook/blocks'
+import { InlineNotification } from '~components/Notification'
+
+export const RenameNotice = (): JSX.Element => {
+  return (
+    <Unstyled>
+      <InlineNotification variant="cautionary" persistent title="Deprecated">
+        {`This component will be renamed in the upcoming major release.`}
```
mcwinter07: Can we link out to the major releases doc for folks to find the rename value in our docs? It might preemptively nip a couple of questions in the bud for those that won't dive into our types

mcwinter07 (dismissed):
Looks good to me, I left a comment about an optional tweak to the rename notice but not a hard requirement 👍

PR #5885, opened 2025-07-02
[KZN-3415] uses native popover api in SingleSelect

Kitty-Al:
https://cultureamp.atlassian.net/browse/KZN-3415 
## What

- Implements the native popover api which handles moving the popover into the css layer
- ~~uses `useOverlayPosition` to handle the popover positioning instead of using css anchor and having to manually handle the resizing & flipping of the popover~~
- positions popover via css anchor positioning, if anchor positioning not supported we're now positioning manually via javascript

Positioning supports:
- SSR
- RTL layouts
- Window resizing

Notes: Because we now handle the popover visibility manually via the Select component state we have to wire up the selected state, and clone the children passing the appropriate props. 

## Why

Reducing any future risk of z-index and layout issues. These were a regular occurance in the past

Initially positioning fallback was meant to be done with RAC - however RAC was not working in RTL layouts due to quirks of positioning in the css top-layer that RAC doesn't account for.


packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
```diff
@@ -1,23 +1,88 @@
...
+    if (onSelectionChange) {
+      onSelectionChange(key)
+    }
+  }
+
+  const selectedKeys: Iterable<Key> = state.selectedKey
+    ? new Set<Key>([state.selectedKey])
+    : new Set()
+
+  // Clone user children injection selection props
```
ryanseddon: It'd be worth adding you comment here about why we need to clone so it's clear why this is here

packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
```diff
@@ -1,23 +1,88 @@
...
+        setOpen: state.setOpen,
+        selectedKey: state.selectedKey,
+        items: items,
+      }}
+    >
+      <RACSelect
+        onSelectionChange={(key) =>
+          handleOnSelectionChange(key != null ? new Set([key]) : new Set())
+        }
+        className={classNameOverride}
```
mcwinter07: One this I noticed was that we're using `classNameOverride` here instead of `className`. While this was the pattern for our older components, for recent ones like `Button`, `Link` and `LinkButton` we've allowed for `className` to be used so folks can leverage the RACs render props (see [RAC styling section](https://react-spectrum.adobe.com/react-aria/Select.html#styling)). 

Looking at some of the sub-components that use RAC like `ListItem` they have `className` available so was just wondering if we are selectively choosing which ones have access to this?

Kitty-Al: Thanks for calling this out 😄 I never remember whether `classNameOverride` is new or old pattern given that it's still being used in some of our next components. I've removed it now and written myself a note 🧠

packages/components/src/__alpha__/SingleSelect/subcomponents/Trigger/Trigger.tsx
```diff
@@ -1,12 +1,36 @@
...
 
-export const Trigger = (): JSX.Element => {
+function flattenItems(items: (SelectItem | SelectSection)[]): SelectItem[] {
+  return items.flatMap((item) => ('options' in item ? item.options : item))
+}
+
+interface TriggerProps {
+  buttonRef: React.RefObject<HTMLButtonElement>
+}
+
```
mcwinter07: I noticed that were using an `interface` for `TriggerProps` and was wandering if there's a specifc reason for chosing this over `type`?

In the past we made a decision to lean on `type` over `interface` unless we foresaw the need to extend / merge a type. The main reasoning for where we landed was to simplify the decision process of whether something should be an `interface` or `type`, which has lead to most of our codebase using `type`.

packages/components/src/__alpha__/SingleSelect/types.ts
```diff
@@ -0,0 +1,9 @@
+export type SelectItem = {
+  label: string
```
mcwinter07, line 6: Question on the types here, is the item always going to have a string `label` or is will it need to be a React node / element to be able handle Avatars, icons and such?

packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stories.tsx
```diff
@@ -12,7 +13,9 @@ export default meta
 type Story = StoryObj<typeof meta>
 
 export const Playground: Story = {
-  args: {},
+  args: {
+    items: singleMockItems,
```
mcwinter07: I noticed in the playground that when there is no children there appears to be a bullet rendered. 
Screenshot 2025-07-08 at 11 10 09 am

I'm assume we'll tackle what null or an empty children prop looks like later but just wanted to flag it in this story :)

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <div
+        // @ts-expect-error - popover attribute is not included in current ts version, ignore type error
+        popover="manual"
+        ref={popoverRef}
+        style={{
+          ...overlayProps,
+          zIndex: 'none',
+          margin: '0',
+          boxSizing: 'border-box',
+          width: buttonRef.current?.getBoundingClientRect().width ?? '200px',
```
mcwinter07: So this would mean whatever the trigger button size is there popover box will be. Is there plans to give the consumers the ability to control the trigger styles a little more, ie: either a custom trigger button or giving them a className prop?

packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
```diff
@@ -1,23 +1,85 @@
...
+      }}
+    >
+      <RACSelect
+        onSelectionChange={(key) =>
+          handleOnSelectionChange(key != null ? new Set([key]) : new Set())
+        }
+        placeholder=""
+        {...restProps}
+      >
+        <Trigger buttonRef={buttonRef} />
```
mcwinter07, line 77: The trigger at the moment is controlled entirely by our implementation with only the buttonRef exposed. Have we considered any potential needs for our consumers to either have a custom trigger or to be able to pass in data attributes that they may used in testing or for programmatically shifting focus? 

The main thing that gets me thinking about that is [this recent support question](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1751428951027369)

packages/components/src/__alpha__/SingleSelect/SingleSelect.spec.tsx
```diff
@@ -16,11 +17,90 @@ const SingleSelectWrapper = (): JSX.Element => (
...
+  afterEach(cleanup)
   describe('renders', () => {
     it('a basic select component', () => {
       const { getByRole } = render(<SingleSelectWrapper />)
       const select = getByRole('button')
       expect(select).toBeInTheDocument()
     })
   })
+
+  describe('popover interactions', () => {
```
mcwinter07: I keen to get thoughts about potentially moving the interaction tests over to Storybook's play tests 🤔 

I don't think we have a hard and fast rule about when to use one over the other so might also be worth discussing as a team when we're all back :)

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { overlayProps } = useOverlayPosition({
+    targetRef: buttonRef,
+    overlayRef: popoverRef,
+    placement: 'bottom start',
+    offset: 0,
+    isOpen: isOpen,
+    shouldFlip: true,
+  })
+  return (
+    <RACPopover trigger="manual" isOpen={isOpen} onOpenChange={setOpen} ref={racPopoverRef}>
```
mcwinter07: I think we'll need to give this an `aria-labelledby` to ensure the element with the role dialog has a title

packages/components/src/__alpha__/SingleSelect/types.ts
Kitty-Al: This is definitely not set in stone, and will most likely change. My thoughts are that we would support a Key: number or string. Passing in items will be mostly used for handling state and async behaviour. If users want to render a bespoke item they can do so by mapping over their own `ListItem` and passing it as children.

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
Kitty-Al: At the moment these styles are still temporary, as we don't have the designs for the popover yet.

packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
Kitty-Al: Thanks for the callout. I totally appreciate that our users might need more flexibility here. I've written a ticket so that we capture the work on exposing the trigger & popover, trying to give our users as much flexibility as possible  https://cultureamp.atlassian.net/browse/KZN-3462

packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stories.tsx
```diff
@@ -1,21 +1,45 @@
...
+  },
+  decorators: [
+    (Story) => (
+      <div
+        style={{
+          height: '200px',
+          display: 'flex',
+          alignItems: 'center',
+          justifyContent: 'center',
+          position: 'relative',
```
ryanseddon: We can use tailwind classes instead of inline styles

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -28,28 +38,27 @@ export const Popover = ({
     }
   }, [isOpen, popoverRef])
```
ryanseddon: refs are stable objects so this would never trigger the effect is this just here to appease eslint? We should just have an ignore comment here if so

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    setPosition({
+      top,
+      bottom,
+      insetInlineStart: inlineStart,
+      maxHeight,
+    })
+  }, [triggerRef, popoverRef, direction, offset, preferredPlacement])
+
+  const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect
```
ryanseddon: Could we just do a custom hook that checks for window or document and then returns if not found since useEffect doesn't run on the server either.

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const onScrollOrResize = (): void => {
+      frame ??= requestAnimationFrame(() => {
+        updatePosition()
+        frame = null
+      })
+    }
+
+    updatePosition()
+    window.addEventListener('resize', onScrollOrResize)
```
ryanseddon: Can we do this with a ResizeObserver over a global listener and using rAF?

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -0,0 +1,139 @@
...
+    window.addEventListener('scroll', onScrollOrResize, true)
```
ryanseddon: RAC disables scrolling when this is open so do we even need to track this?

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -28,28 +38,27 @@ export const Popover = ({
...
-          zIndex: 'none',
+          position: 'fixed',
+          top,
+          bottom,
+          insetInlineStart,
+          maxHeight,
+          overflowY: 'auto',
+          left: 'auto',
+          right: 'auto',
           margin: '0',
```
ryanseddon: You could move the hardcoded styles to the stylesheet to reduce the non dynamic ones here

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -0,0 +1,139 @@
...
+  preferredPlacement = 'bottom',
+}: UseFixedOverlayPositionProps): Position {
+  const [position, setPosition] = useState<Position>({
+    top: undefined,
+    bottom: undefined,
+    insetInlineStart: 0,
+    maxHeight: undefined,
+  })
+
+  const updatePosition = useCallback(() => {
```
ryanseddon: Was css anchor positioning part of another card? I have it working on my local but couldn't see it in the epic

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <RACPopover trigger="manual" isOpen={isOpen} onOpenChange={setOpen} ref={racPopoverRef}>
+      <div
+        // @ts-expect-error - popover attribute is not included in current ts version, ignore type error
+        popover="manual"
+        ref={popoverRef}
+        // TODO: expect some of these styles to change once we have designs
+        className={styles.popover}
+        style={{
+          visibility: isPositioned ? 'visible' : 'hidden',
```
ryanseddon: We can use the `:popover-open` to fix this so when the popover is visible

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -37,29 +130,23 @@ const usePopoverPositioning = (
...
-        maxHeight,
-      }
+      return getManualPositioningStyles(positionData)
     }
 
     // Use CSS anchor positioning when supported
-    return { '--position-anchor': anchorName } as React.CSSProperties
-  }
+    return getAnchorPositioningStyles(anchorName, positionData)
+  }, [hasAnchorSupport, anchorName, positionData])
```
ryanseddon: I would invert this function and check if `hasAnchorSupport` is true and fallback to manual positioning:

```js
if(!!hasAnchorSupport) {
  return getAnchorPositioningStyles(anchorName, positionData)
}

return getManualPositioningStyles(positionData)
```

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -110,22 +86,25 @@ export function useFixedOverlayPosition({
...
+    setPosition(newPosition)
     setIsPositioned(true)
-  }, [triggerRef, popoverRef, direction, offset, preferredPlacement])
+  }, [triggerRef, popoverRef, direction, offset, preferredPlacement, isSSR])
 
+  // Separate effect for client-side initialization and resize handling
   useEffect(() => {
-    if (typeof window === 'undefined') return // SSR safety
+    // This effect only runs on the client after hydration
+    if (typeof window === 'undefined') return
```
ryanseddon: You'd definitely want to `useLayoutEffect` here as this will run before painting and I would go back to the client hook but make it check if window/document exists and just bail e.g.

```
function useClientLayoutEffect( ...args) {
  if (typeof document === "undefined") return;

  useLayoutEffect(...args); 
} 
```

This way if this gets SSR'd then it'll just not run anything and the name is clear that its client side only

packages/components/src/__alpha__/SingleSelect/utils/useFixedOverlayPosition.ts
```diff
@@ -0,0 +1,130 @@
...
+  popoverRef: RefObject<HTMLElement>
+  direction?: 'ltr' | 'rtl'
+  offset?: number
+  preferredPlacement?: 'top' | 'bottom'
+}
+
+/**
+ * Hook to calculate and update the position of an overlay rendered in the top layer.
+ */
+export function useFixedOverlayPosition({
```
ryanseddon: We can simplify this in the case where the browser supports anchor positioning and we would only need to keep track of maxHeight adjustments otherwise anchor positioning can do everything else.

This is the CSS snippet I have on my local, maybe we pair on this.

 ```css
 @layer kz-components {
   .popover {
     // ...
     @supports (anchor-name: --anchor) {
       position-area: bottom center;
       position-try-fallbacks: flip-block;
       margin-block-start: 4px;
       width: anchor-size(width);
     }
   } 
}

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.module.css
```diff
@@ -0,0 +1,33 @@
...
+
+    /* TODO: update width based on design */
+    width: 200px;
+    visibility: hidden;
+
+    @supports (anchor-name: --anchor) {
+      position-anchor: var(--position-anchor, --trigger);
+      top: var(--popover-top, calc(anchor(bottom) + 4px));
+      bottom: var(--popover-bottom, auto);
+      inset-inline-start: var(--popover-inline-start, anchor(start));
```
ryanseddon: You can drop top, bottom and inset-inline-start as the position-area does that for you. You'll just need to set the margin-inline-start to 4px to get the gap

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/supportsAnchorPositioning.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (cachedAnchorSupport !== null) return cachedAnchorSupport
+
+  if (typeof window === 'undefined' || typeof CSS === 'undefined') {
+    cachedAnchorSupport = false
+    return false
+  }
+
+  try {
+    cachedAnchorSupport =
+      CSS.supports('position-anchor', 'auto') || CSS.supports('position-anchor: auto')
```
ryanseddon: You've got the same thing twice here it'd be better to also check for `position-try-fallbacks` support
```suggestion
      CSS.supports('position-anchor', 'auto') && CSS.supports('position-try-fallbacks: flip-block')
```

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/supportsAnchorPositioning.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Detects if CSS anchor positioning is supported by the browser (cached, SSR-safe)
+ */
+export const supportsAnchorPositioning = (cachedAnchorSupport: boolean | null): boolean => {
+  if (cachedAnchorSupport !== null) return cachedAnchorSupport
```
ryanseddon: I'm a bit confused by this caching as I don't think it's cached at all as the const will get assigned every render of the `usePositioninStyles` hook.

I think the best path forward is to make this use useMemo.

I'd also use the `use` naming convention since this is a hook e.g `useAnchorPositioningSupport` 

```tsx
import { useMemo } from 'react'

export const useAnchorPositioningSupport = (): boolean => {
  return useMemo(() => {
    if (typeof window === 'undefined' || typeof CSS === 'undefined') {
      return false
    }

     return CSS.supports('position-anchor', 'auto') && CSS.supports('position-try-fallbacks: flip-block')
  }, [])
}
```

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/usePositioningStyles.ts
```diff
@@ -0,0 +1,111 @@
...
+  anchorName: string,
+): { popoverStyle: React.CSSProperties; isPositioned: boolean } => {
+  // Cache anchor positioning support detection (SSR-safe)
+  const cachedAnchorSupport: boolean | null = null
+  const { direction } = useLocale()
+  const [hasAnchorSupport, setHasAnchorSupport] = useState<boolean | null>(null)
+
+  useEffect(() => {
+    setHasAnchorSupport(supportsAnchorPositioning(cachedAnchorSupport))
+  }, [])
```
ryanseddon: With the refactor suggested above we can avoid the useState and useEffect altogether and just do:

```tsx
const supportsAnchors = useAnchorPositioningSupport()
```

packages/components/src/__alpha__/SingleSelect/subcomponents/Trigger/Trigger.module.css
```diff
@@ -1,5 +1,6 @@
 @layer kz-components {
   .button {
+    anchor-name: var(--anchor-name, --trigger);
```
ryanseddon: Does `--trigger` ever get set anywhere?

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/supportsAnchorPositioning.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  if (typeof window === 'undefined' || typeof CSS === 'undefined') {
+    cachedAnchorSupport = false
+    return false
+  }
+
+  try {
```
ryanseddon: We don't expect this to throw as we already check it exists above so a try catch block is unnecessary here

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/usePopoverPositioning.ts
```diff
@@ -0,0 +1,130 @@
...
+}: UsePopoverPositioningProps): Position & { isPositioned: boolean } {
+  // Provide SSR-compatible default positioning
+  const [position, setPosition] = useState<Position>({
+    top: preferredPlacement === 'bottom' ? offset : 'auto',
+    bottom: preferredPlacement === 'top' ? offset : 'auto',
+    insetInlineStart: 0,
+    maxHeight: 300, // TODO: update this based on designs
+  })
+  // Start as true for SSR compatibility - we have default positioning
+  const [isPositioned, setIsPositioned] = useState(true)
```
ryanseddon, line 22: This is still running even when we support anchor positioning. We need some logic to only update the min-height in the anchor positioning only fallback to this when not

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.module.css
```diff
@@ -15,14 +15,10 @@
...
-      top: var(--popover-top, calc(anchor(bottom) + 4px));
-      bottom: var(--popover-bottom, auto);
-      inset-inline-start: var(--popover-inline-start, anchor(start));
-      max-height: var(--popover-max-height, 300px);
-      position-area: bottom center;
-      position-try-fallbacks: flip-block;
-      /* stylelint-disable-next-line declaration-property-value-no-unknown */
+      position-anchor: var(--position-anchor);
+      margin-block-start: var(--spacing-4); 
+      margin-block-end: var(--spacing-4);
```
ryanseddon: You can just do the shorthand here `margin-block`

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.module.css
```diff
@@ -15,14 +15,10 @@
     visibility: hidden;
```
ryanseddon: Does this actually do anything? popovers have user agent styles to flip the display when opened or closed

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -0,0 +1,58 @@
...
+      popover.showPopover()
+    } else {
+      popover.hidePopover()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [shouldShowPopover])
+
+  return (
+    <RACPopover trigger="manual" isOpen={isOpen} onOpenChange={setOpen} ref={racPopoverRef}>
+      <div style={{ position: 'relative' }}>
```
ryanseddon: What's the reason for this container?

packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stickersheet.stories.tsx
```diff
@@ -17,22 +17,22 @@ const StickerSheetTemplate: StickerSheetStory = {
     return (
       <StickerSheet isReversed={isReversed} title="SingleSelect" headers={['Items', 'Grouped']}>
         <StickerSheet.Row>
-          <SingleSelect>
+          <SingleSelect items={singleMockItems}>
             <SingleSelect.List>
               {singleMockItems.map((item) => (
-                <SingleSelect.ListItem key={item.value} value={{ value: item.value }}>
+                <SingleSelect.ListItem key={item.value} id={item.value}>
```
ryanseddon, line 23: I'm pretty sure RAC ListItems will automatically add a key prop based on the id prop

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/usePositioningStyles.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Generates anchor positioning styles with CSS custom properties
+ */
+const getAnchorPositioningStyles = (
+  anchorName: string,
+  positionData: {
+    top: number | string | undefined
+    bottom: number | string | undefined
+    insetInlineStart: number | string | undefined
+    maxHeight: number | string | undefined
```
ryanseddon: Pull this out as a shared type and you could probably pull out the individual properties as a shared type too e.g.

```ts
type PositionDataProp = number | string | undefined;
type PositionData = {
  top: PositionDataProp
  //...
}
```

packages/components/src/__alpha__/SingleSelect/subcomponents/Trigger/Trigger.tsx
```diff
@@ -1,13 +1,40 @@
...
-import { Button as RACButton, SelectValue } from 'react-aria-components'
+import React, { useMemo } from 'react'
+import { Button as RACButton } from 'react-aria-components'
 import { Icon } from '~components/__next__/Icon'
+import { useSingleSelectContext } from '../../context'
+import { type SelectItem, type SelectSection } from '../../types'
 import styles from './Trigger.module.css'
 
-export const Trigger = (): JSX.Element => {
+function flattenItems(items: (SelectItem | SelectSection)[]): SelectItem[] {
```
ryanseddon, line 8: Can we leverage RAC here better and perhaps just use [`<SelectValue>`](https://react-spectrum.adobe.com/react-aria/Select.html#customizing-selectvalue)? This seems like it could get expensive on bigger lists, internally RAC should be already tracking this for us without having to flatten lists.

ryanseddon (commented):
There's quite a few redundant comments throughout too, can you delete ones that are obvious as to what they're doing via the code please.

packages/components/src/__alpha__/SingleSelect/subcomponents/Trigger/Trigger.tsx
Kitty-Al: `SelectValue` relies on RAC internally managing the state, from what I've seen we can't use both `SelectValue` and `useSelectState` at the same time. We need to use `useSelectState` to manage the control of the popover opening/closing, as well as handling dynamic and asynchronous lists

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/usePopoverPositioning.ts
```diff
@@ -0,0 +1,130 @@
...
+  maxHeight?: number
+}
+
+type UsePopoverPositioningProps = {
+  triggerRef: RefObject<HTMLElement>
+  popoverRef: RefObject<HTMLElement>
+  direction?: 'ltr' | 'rtl'
+  offset?: number
+  preferredPlacement?: 'top' | 'bottom'
+}
```
ryanseddon: To be consistent should these types be in the types file too?

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -0,0 +1,56 @@
...
+    if (shouldShowPopover) {
+      popover.showPopover()
+    } else {
+      popover.hidePopover()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [shouldShowPopover])
+
+  return (
+    <RACPopover trigger="manual" isOpen={isOpen} onOpenChange={setOpen} ref={racPopoverRef}>
```
ryanseddon: We should set the [`shouldUpdatePosition`](https://react-spectrum.adobe.com/react-aria/Popover.html#popover-1) to false as we're controlling this now and it can avoid RAC from changing inline styles that we don't need updated

packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stickersheet.stories.tsx
Kitty-Al: This is throwing a warning in the console when key is not included

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/utils/usePopoverPositioning.ts
Kitty-Al: With anchor positioning we need the min-height, but also top and bottom calculations due to the `position-try-fallbacks: flip-block` positioning not supporting the behaviour that we're after. Considering that we need all 3 calculations I feel like we should just use the same function instead of breaking it out.

PR #5888, opened 2025-07-04
feat: add Split Button Replacement story and docs to Button & Menu

ryanseddon:
## Why

Split Button is being deprecated so we need a path forward for existing usage

## What

Added stories and API spec docs to both Menu and Button sections

### DIFF
--- a/.changeset/fifty-snails-do.md
+++ b/.changeset/fifty-snails-do.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+add Split Button Replacement story and docs to Button & Menu sections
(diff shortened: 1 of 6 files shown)


.changeset/fifty-snails-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+add Split Button Replacement story and docs to Button & Menu sections
```
mcwinter07: Since this is updating the documentation, do we need to release a patch?

ryanseddon: Good catch, I changed it to patch the docs since it updates content there but the files live in the components dir

mcwinter07 (approved):

PR #5906, opened 2025-07-09
[KZN-3463] fix: add ssr guards in GenericModal

NJWong:
## Why

https://cultureamp.slack.com/archives/C0189KBPM4Y/p1751950032455809

This PR adds SSR guards to GenericModal which has some functionality that uses `window` and `document` - neither of which will be available in static prerendering environments like Next.js server side rendering

Although CultureAmp doesn't have an official mandate to support SSR, we still prerender static versions of the Next.js app as part of the build script

## What

Adds various guards in GenericModal that return early if `window` or `document` is undefined

In the main render block, we return an empty `<></>` fragment if `document` is not available for `createPortal` to use

### DIFF
--- a/packages/components/src/Modal/GenericModal/GenericModal.tsx
+++ b/packages/components/src/Modal/GenericModal/GenericModal.tsx
@@ -3,6 +3,7 @@ import { createPortal } from 'react-dom'
 import { Transition } from '@headlessui/react'
 import classnames from 'classnames'
 import FocusLock from 'react-focus-lock'
+import { useIsClientReady } from '../../utils/useIsClientReady'
 import { warn } from '../util/console'
 import { ModalContext } from './context/ModalContext'
 import styles from './GenericModal.module.scss'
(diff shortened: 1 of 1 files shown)


packages/components/src/Modal/GenericModal/GenericModal.tsx
```diff
@@ -58,6 +58,9 @@ export const GenericModal = ({
   }
 
   const focusOnAccessibleLabel = (): void => {
+    // Add SSR guard
+    if (typeof document === 'undefined') return
```
mcwinter07: Since we're making this check multiple times for document and window what are your thoughts on combining these in a variable or function? Some like:

`const isClientReady = typeof document !== 'undefined' && typeof window !== 'undefined'`

We also have a utility hook `useIsClientReady` we made a little while back that could potentially be used here. It wasn't actually used to check as an SSR guard but looks like it could serve that function.

NJWong: Ah perfect, I think that hook is exactly what we want

PR #5959, opened 2025-07-29
[KZN-3487] styling Link component in notifications

Kitty-Al:
## Why

Passing the `Link` component into any of our notifications was breaking the styling and giving the link 2 underlines. 
toast new link before

## What

Updating the styling to fix the implementation when using the `Link` component without breaking the older implementations that use a link`<a href="/" >` 

`Link` component passed
Screenshot 2025-07-30 at 10 10 12 am

`<a href="/" >` passed
toast old link after

### DIFF
--- a/packages/components/src/Notification/subcomponents/GenericNotification/_mixins.scss
+++ b/packages/components/src/Notification/subcomponents/GenericNotification/_mixins.scss
@@ -266,7 +266,6 @@ $notification-slide-right: transform 300ms ease-out;
       border-bottom: var(--spacing-1) solid $color-blue-500;
       text-decoration: none;
       color: $color-blue-500;
-      text-decoration: none;
 
       &:hover {
         text-decoration: none;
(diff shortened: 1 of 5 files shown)


packages/components/src/Notification/subcomponents/GenericNotification/_mixins.scss
```diff
@@ -263,12 +263,13 @@ $notification-slide-right: transform 300ms ease-out;
     }
 
     a[href] {
+      border-bottom: var(--spacing-1) solid $color-blue-500;
```
ryanseddon, line 266: This is actaully adding two borders to the Link now as `isUnderline` is on by default with link, perhaps we need to add a `:not()` selector to only add the border on non-Link components e.g.

```css
a[href]:not([data-rac]) {
  //...
}
```

Image

ryanseddon (approved):

PR #5977, opened 2025-08-08
[KZN-3505] TitleBlockZen home variant

NJWong:

## Why

Context and background for how the TitleBlockZen component is being used in Home: https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/5332566155/TitleBlockZen+in+Home+Page

### DIFF
--- a/packages/components/src/TitleBlockZen/TitleBlockZen.module.scss
+++ b/packages/components/src/TitleBlockZen/TitleBlockZen.module.scss
@@ -39,6 +39,20 @@
       background-color: $dt-color-background-color-admin;
     }
 
+    @media (min-width: $layout-breakpoints-large) {
+      &.homeVariant {
+        background-color: transparent;
+
+        h1 {
+          color: var(--color-purple-800);
+        }
+
+        .titleRowInner {
+          margin: 0;
+        }
+      }
+    }
+
     @media print {
       display: none;
     }
--- a/packages/components/src/TitleBlockZen/types.ts
+++ b/packages/components/src/TitleBlockZen/types.ts
@@ -157,7 +157,7 @@ export type SectionTitleRenderProps = Pick<
   | 'sectionTitleDescriptionAutomationId'
 >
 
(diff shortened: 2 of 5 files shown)


packages/components/src/TitleBlockZen/TitleBlockZen.module.scss
```diff
@@ -39,6 +39,20 @@
       background-color: $dt-color-background-color-admin;
     }
 
+    @media (min-width: $layout-breakpoints-large) {
+      &.homeVariant {
+        background-color: transparent;
+
+        h1 {
```
mcwinter07, line 46: Generally we avoid use tags directly unless absolutely necessary. This is mainly as classes are a quicker reference tool within our code base and it helps us to avoid cases where a tag might change. In these instances it can be confusing to find the diff as we often don't use tag directly, ie: we use the `heading-1` variant vs an `<h1>` tag.

Would probably be the `titleTextOverride` class instead?

packages/components/src/TitleBlockZen/types.ts
```diff
@@ -157,7 +157,7 @@ export type SectionTitleRenderProps = Pick<
   | 'sectionTitleDescriptionAutomationId'
 >
 
-export type TitleBlockVariant = 'admin' | 'education' // the default is wisteria bg (AKA "reporting")
+export type TitleBlockVariant = 'admin' | 'education' | 'home' // the default is wisteria bg (AKA "reporting")
```
mcwinter07, line 160: Can we also have a type description here to call out that the `home` variant should not be used outside of the home page context. Just so we have some intellisense to dissuade folks from using this.

Also tiny clean up opportunity here, can you please remove the code comment:
`// the default is wisteria bg (AKA "reporting")`

I think wisteria was from the bygone greycliff theme era.

NJWong:
Closing this PR - instead of porting the changes that Home made to TitleBlockZen, we're going to remove it altogether and adopt the Composable Header pattern

PR #5978, opened 2025-08-08
Remove legacy exports from design-tokens package

mcwinter07:
## Why
This is part of the work for V2 readiness to remove all legacy exports from design tokens

## What
- Remove legacy folder with exported legacy variables and utils
- Update internal usages to reference `/lib`
- Update README and CONTRIBUTION docs to current build processes and usage guides
- Add major changeset in line with V2


mcwinter07:
I've tested a canary in Unified-Home and Performance-Review-Cycle-UI and both seem fine with the update 👍

PR #5990, opened 2025-08-13
Remove retired props v2

ryanseddon:
## Why

v2 readiness, removal of retired props desscribed in [upcoming release](https://cultureamp.design/?path=/docs/releases-upcoming-major-releases--docs).

## What

Prop migrations:

- mood to variant
- automationId to data-testid

Prop removals:

- Select/next trigger.ref and disabledValues
- EmptyState layoutContext and illustrationType
- FilterBarContext toggleOpenFilter
- Popover customIcon


NJWong (commented):
Code looks good, just need to fix up the CI errors

mcwinter07:
> Looks like the background and default border color isn't coming in for the popover?



> Looks like the background and default border color isn't coming in for the popover? Screenshot 2025-08-14 at 12 03 06 pm Screenshot 2025-08-14 at 12 06 29 pm

Just re-read the diff. Assuming that's a white bg we should be good there, though this still needs the border color. we can probably also remove those variants from the sticker sheet :)

mcwinter07 (approved):

mcwinter07 (approved):

PR #5991, opened 2025-08-13
Remove old entry points

mcwinter07:
## Why
Part of the V2 major release is consolidation and removing our old entry points. This will need to be followed by an update to the codemod to ensure they are pointing to the right path.

## What
- Remove v1, v2, v3, utilities and future entry points
  - ReversedColors utility has been moved into the base entry point. This will need a codemod
    - I don't think there's a strong case for `/utilities` unless we're planning on producing a lot of utils
- Move re-export of React Aria and React Aria-Components out of `/v3` entry point  in `/libs`
  - ie:  `import { Button as RacButton } from "@kaizen/components/libs/react-aria-components"`
    - Originally I moved this into the base with the idea that we would could resolve this to a base entry point of `/components/react-aria` but that caused local linting issues with package resolution.
    - This also has the added benefit of aligning closer to our other entry points and grouping our external libraries so there's less errant folder lying around
- Update workspace, rollup, package.json to reflect the updated entry points

## Next steps
These updates will need to update the migrateV2NextToCurrent to ensure all entry points are updated correctly (ie: `/libs/react-aria` vs `/react-aria`)


packages/components/src/index.ts
```diff
@@ -45,6 +45,7 @@ export * from './Radio'
 export * from './RichTextEditor'
 export * from './SearchField'
 export * from './Select'
```
ryanseddon, line 47: Select should be deleted right?

packages/components/package.json
```diff
@@ -37,60 +35,20 @@
...
-    "./v3/react-aria": {
-      "types": "./dist/types/__react-aria__/index.d.ts",
-      "import": "./dist/esm/reactAriaV3.mjs",
-      "require": "./dist/cjs/reactAriaV3.cjs"
-    },
-    "./v3/react-aria-components": {
-      "types": "./dist/types/__react-aria-components__/index.d.ts",
-      "import": "./dist/esm/reactAriaComponentsV3.mjs",
-      "require": "./dist/cjs/reactAriaComponentsV3.cjs"
+    "./libs/react-aria": {
```
ryanseddon, line 43: Possibly beyond scope but do we need to add the __alpha__ export here too?

packages/components/src/index.ts
mcwinter07: I thought we were keeping `Select` since `SingleSelect` doesn't have feature parity and the pickers rebuild hasn't finished.

packages/components/package.json
mcwinter07: ~No harm in having it there as long as the components are properly signposted. Might be worth eventually having a doc to describe the entry points in our guides?~ Actually will do in as a seperate task :) Would keen to serve it up with some docs just to make sure folks don't go ham / use it in prod

packages/components/src/index.ts
ryanseddon: Select  /next is becoming SingleSelect and the new one is in alpha

mcwinter07: Yep, but we still have direct replacement for the 'legacy' (react-select) Select right? Without search functionality and MultiSelect bundled together we don't have suggested migration path.

We also haven't [sign posted](https://cultureamp.design/?path=/docs/releases-upcoming-major-releases--docs#other-components-being-removed) the removal, so I don't know if this is meant to be in scope for v2?

PR #6012, opened 2025-08-22
Winter/kzn 3294 create v1 codemod runner

mcwinter07:
## Why
To simplify the process of running the V1 codemods in a single script prior to the v2 release

## What
- Adds a `runV1Codemods` codemod that run all v1 codemod in sequence.
- Adds a snapshot test to validate codemod output
- Adds README to provide guidance
  -  We can probably convert this to a MDX with some reworking based on what we discussed in the release comms plan


packages/components/codemods/runV1Codemods/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  console.log('9. migrateInformationTileMoodToVariant')
+  console.log('10. migrateInlineNotificationTypeToVariant')
+  console.log('11. migrateMultiActionTileMoodToVariant')
+  console.log('12. migrateNotificationTypeToVariant')
+  console.log('13. migrateProgressBarMoodToColor')
+  console.log('14. migrateToastNotificationTypeToVariant')
+  console.log('15. migrateWellVariantToColor')
+  console.log('16. removeInputEditModalMood')
+  console.log('17. removePopoverVariant')
+  console.log('')
```
ryanseddon: I just make this one console.log with back ticks

```suggestion
console.log(`🚀 Running all V1 codemods in sequence

This will run the following codemods in order:
1. upgradeIconV1
2. upgradeV1Buttons
3. migrateGuidanceBlockActionsToActionsSlot
4. migrateBrandMomentMoodToVariant
5. migrateCardVariantToColor
6. migrateConfirmationModalMoodsToVariant
7. migrateEmptyStateIllustrationTypeToVariant
8. migrateGlobalNotificationTypeToVariant
9. migrateInformationTileMoodToVariant
10. migrateInlineNotificationTypeToVariant
11. migrateMultiActionTileMoodToVariant
12. migrateNotificationTypeToVariant
13. migrateProgressBarMoodToColor
14. migrateToastNotificationTypeToVariant
15. migrateWellVariantToColor
16. removeInputEditModalMood
17. removePopoverVariant
`);
```

packages/components/codemods/runV1Codemods/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+- If any individual codemod fails, the process continues with the remaining codemods
+- Detailed logging shows the start and completion status of each codemod
+- A summary is provided at the end showing successful vs failed codemods
+
+## Cautions and caveats
+
+### Reducing the blast radius of the codemods
+
+Thi running will create a lot of diffs and changes to your components as the traverses the entire directory given.
```
ryanseddon:
```suggestion
This runner will create a lot of diffs and changes to your components as the traverses the entire directory given.
```

packages/components/codemods/runV1Codemods/runV1Codemods.ts
```diff
@@ -0,0 +1,187 @@
...
+  let successCount = 0
+  let errorCount = 0
+
+  for (const codemod of codemods) {
+    try {
+      console.log(`🔄 Starting codemod: ${codemod.name}`)
+      await new Promise<void>((resolve) => {
+        codemod.runner(targetDir)
+        resolve()
+      })
```
ryanseddon: AFAICT `codemod.runner` is synchronous so this Promise is unecessary?

packages/components/codemods/runV1Codemods/runV1Codemods.ts
```diff
@@ -0,0 +1,187 @@
...
+  {
+    name: 'removePopoverVariant',
+    runner: (dir) => {
+      transformComponentsInDir(dir, ['Popover'], (tagNames) => [removePopoverVariant(tagNames)])
+    },
+  },
+]
+
+export const runV1Codemods = async (targetDir: string): Promise<void> => {
+  console.log(`📝 Running ${codemods.length} codemods on directory: ${targetDir}`)
```
ryanseddon, line 155: Should we check if the target directory exists?

packages/components/codemods/runV1Codemods/runV1Codemods.ts
```diff
@@ -0,0 +1,187 @@
...
+  for (const codemod of codemods) {
+    try {
+      console.log(`🔄 Starting codemod: ${codemod.name}`)
+      await new Promise<void>((resolve) => {
+        codemod.runner(targetDir)
+        resolve()
+      })
+      console.log(`✅ Completed codemod: ${codemod.name}`)
+      successCount++
+    } catch (error) {
```
NJWong, line 171: I'm not very across how the codemods handle errors, but if there's an error does it rollback the code changes? Wondering if there's a potential scenario where a codemod partially completes then fails, then the next codemod is run on broken/half migrated code

packages/components/codemods/runV1Codemods/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+To reduce the scope, consider either running this on component or page directories or running the mode incrementally, ie:
```
NJWong, line 65: Maybe in the above example `pnpm kaizen-codemod src runV1Codemods` instead of `src` we can use a component or page path?

Might reduce the risk of someone (like myself 😅) just copying and pasting the command then realising I shouldn't have done that

packages/components/codemods/runV1Codemods/runV1Codemods.ts
mcwinter07: Good idea - I've added a check and early exit if it fails to find a file

packages/components/codemods/runV1Codemods/runV1Codemods.ts
mcwinter07: True, has been 🪓'd

packages/components/codemods/runV1Codemods/runV1Codemods.ts
mcwinter07: I think as standard a codemod that fails will only fail on the single node. In that instance you might have some of a single page with some components that were successfully migrated / transformed and some of the same components that were not (they should not be transformed).

The next codemod would still run on this page as there's no condition (at the moment) to early exit on a single failed codemod 🤔 

We could potentially do something like this:

```jsx
    } catch (error) {
      console.log(
        `💥 Error in codemod - Exiting early!`,
      )
      console.error(`${codemod.name} failed with error:\n`)
      console.error(error)
      errorCount++
      process.exit(1)
    }
```

mcwinter07: To elaborate after doing a bit of stress testing with this runner - in this an example, running the transform on a file that has an invalid property (`variant="success"`) / something that will cause an individual codemod to fail:

```jsx
const Page = () => (
<>
  // type "success" will fail in the codemod because this is not a valid type and the codemod will through an error
   <GlobalNotification persistent={true} type="success">
          Global notification content with incorrect type
   </GlobalNotification>

  // This is not be transformed because an error was thrown
   <GlobalNotification persistent={true} type="positive">
          Global notification content that should be transformable
   </GlobalNotification>

  // This will still transformed as its a different codemod
  <Well variant="default">
    <div>Well content</div>
  </Well>
</>
)
```

In this instance, because we're wrapping everything in a `try` `catch` block, that will mean that codemod will cease to run `migrateGlobalNotificationTypeToVariant` but `Well` will still be transformed as its a different codemod.

mcwinter07: On reflecting on this it probably would be better to early exit on the runner 🤔

mcwinter07:
Oky doke -  I've updated this and tested out a canary in goals-ui, which is all looking hunky dory including the condition for passing it an incorrect file path and the early exit for a failed codemod 🎉 

One thing I did notice was a slight vague response from the Icon codemod when there isn't an equivalent icon - we could probably try to return a response so folks know what wasn't updated:

Screenshot 2025-08-27 at 2 52 57 pm

I'll draft up a separate ticket to tackle that one individually.

packages/components/codemods/README.md
```diff
@@ -25,6 +25,12 @@ pnpm kaizen-codemod src migrateWellVariantToColor
 
 ## Available codemods
 
+Released in `1.80.5`
```
ryanseddon: Might want to change this to 1.80.6

packages/components/codemods/runV1Codemods/runV1Codemods.ts
```diff
@@ -0,0 +1,187 @@
...
+      console.log(`🔄 Starting codemod: ${codemod.name}`)
+      codemod.runner(targetDir)
+      console.log(`✅ Completed codemod: ${codemod.name}`)
+      successCount++
+    } catch (error) {
+      console.log(`💥 Error in codemod: ${codemod.name} - exiting early
+
+✅ ${successCount} codemods completed successfully
+😢 ${codemods.length - successCount} codemods did not run
+Address the issues in the follow codemod before proceeding:
```
ryanseddon:
```suggestion
Address the issues in the following codemod before proceeding:
```

packages/components/codemods/runV1Codemods/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+`pnpm kaizen-codemod src/pages/MockPage runV1Codemods`
+
+This can make it easier to review and merge in changes with confidence.
+
+### Prettier diffs
+
+As the codemod will rewrites files, this can cause arbitrary changes that will be resolved by `prettier` formatting. We recommend running the following to reduce the noise in your diffs:
+
+`pnpm prettier --write "**.tsx"`
```
ryanseddon:
```suggestion
`pnpm prettier --write "**/*.tsx"`
```

packages/components/codemods/README.md
mcwinter07: Many thanks :) update this and added those other changes so should be good for a final review 🤞

PR #6013, opened 2025-08-22
fix: remove @deprecated comment from TitleBlock

ryanseddon:
## Why

v2 readiness

## What

TitleBlock is not deprecated this comment was left over from the rename PR, only  saw it due to my editor complaining  in the canary implementation test.


mcwinter07 (approved):
Nice catch! 🌬️ ⚾🧤

PR #6019, opened 2025-08-25
KAIO V1 feature branch (DO NOT MERGE/REMOVE/UPDATE BRANCH)

mcwinter07:
## Why
This feature branch has been create as an interim form of version control for the V1 of our design system.

This PR should remain until we can distribute the v1 documentation to a separate AWS bucket to persist the storybook build for that version.

We should not prune the feature branch until all repo's are off V1 in case we need to backport critical bug fixes to V1. [Refer to guide on how to  backport a fix](https://cultureamp.atlassian.net/wiki/spaces/DesignSystem/pages/5277254482/4.3+Backporting+fixes)

### Additional context
- [Jira ticket](https://cultureamp.atlassian.net/browse/KZN-3517)

## What
- Adds warning to key documentation pages to warn that the user is viewing the old documentation
  - This is also required as we need a diff for chromatic

### DIFF
--- a/.changeset/fine-parrots-return.md
+++ b/.changeset/fine-parrots-return.md
@@ -1,5 +1,0 @@
----
-'@kaizen/components': patch
----
-
-CSS ordering fix - Patch rollup-plugin-postcss to handle multi inputs correctly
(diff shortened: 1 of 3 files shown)


mcwinter07:
Pre-V2 being merged we should set up the branch protection policy on this branch in order to follow best practice

PR #6020, opened 2025-08-27
docs: remove deprected docs and any mentions of next for migrated com…

ryanseddon:
## Why

Support v2 readiness

## What

Added tags to deprecated stories to remove from storybook docs
For mdx files the only solution I found was to delete them, strangely there doesn't seem to be a way to support negation in the stories globs.

Tried ! at the start of a glob as well as a `!(megation)` within a glob and storybook doesn't seem to handle this, they use picomatch under the hood which does support negation but it doesn't work with sotrybook...

I also cleaned up some stray `(next)` and `isNext` strings and props.


mcwinter07:
Code looks good, just added a chromatic flag to check out the storybook and do a once over there :)

mcwinter07 (approved):

NJWong (approved):

PR #6031, opened 2025-09-01
re-add v3 RA library exports to v2

mcwinter07:
## Why
Top reduce the risk of compatibility issues with shared libraries that have not upgraded to KAIO V2 (ie: next-services, goals-shared-ui) we should still export `react-aria` and `react-aria-components` from the V3 endpoint as well as the libs.

This will allow us to remove this as a minor release in the future and will not make the upgrade those packages as a hard dependency for v2.

## What
- re-adds endpoint for v3 with only react-aria and react-aria components exported
- small fix to the types path for the libs export of react-aria

### DIFF
--- a/packages/components/libs/react-aria/package.json
+++ b/packages/components/libs/react-aria/package.json
@@ -1,5 +1,5 @@
 {
   "main": "../../dist/cjs/reactAriaKZ.cjs",
   "module": "../../dist/esm/reactAriaKZ.mjs",
-  "types": "../../dist/types/__react-aria__/index.d.ts"
+  "types": "../../dist/types/__libs__/react-aria/index.d.ts"
 }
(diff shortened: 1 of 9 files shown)


packages/components/libs/react-aria/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "main": "../../dist/cjs/reactAriaKZ.cjs",
   "module": "../../dist/esm/reactAriaKZ.mjs",
-  "types": "../../dist/types/__react-aria__/index.d.ts"
+  "types": "../../dist/types/__libs__/react-aria/index.d.ts"
```
mcwinter07, line 4: Small fix to correct the types for the libs export of RAC since I released this pointing to the incorrect path.

mcwinter07:
Noting the failing test for this should be resolved by @ryanseddon fix so won't handle that in this PR :)

ryanseddon (approved):

PR #6036, opened 2025-09-02
[KZN-3416] refactor SingleSelect to use RA hooks

Kitty-Al:
## Why

Was running into multiple issues with RAC, wherein they would override the state, or pass values to components unnecessarily.
eg. 
- the popover styles
- forcing the popover state.isOpen to be closed/open when not wanted 

Attempted to solely use ComboBox state but also ran into issues with the Trigger component needing Select state to function correctly. 

The outcome has been to simply use a wrapper for the SingleSelect that breaks into using either a Select or ComboBox

## What
 
Refactoring the `SingleSelect` component to use React Aria hooks 
Splitting out Select and ComboBox into separate subcomponents

### DIFF
--- a/packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stickersheet.stories.tsx
+++ b/packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stickersheet.stories.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import { Item, Section } from '@react-stately/collections'
 import { type Meta } from '@storybook/react'
 import { StickerSheet, type StickerSheetStory } from '~storybook/components/StickerSheet'
 import { SingleSelect } from '../index'
@@ -18,25 +17,25 @@ const StickerSheetTemplate: StickerSheetStory = {
       <StickerSheet isReversed={isReversed} title="SingleSelect" headers={['Items', 'Grouped']}>
         <StickerSheet.Row>
           <SingleSelect label="Combobox flat colours" isComboBox>
-            <Item>Red</Item>
-            <Item>Orange</Item>
-            <Item>Yellow</Item>
-            <Item>Green</Item>
-            <Item>Blue</Item>
-            <Item>Purple</Item>
-            <Item>Black</Item>
-            <Item>White</Item>
-            <Item>Lime</Item>
-            <Item>Fushsia</Item>
+            <SingleSelect.Item>Red</SingleSelect.Item>
+            <SingleSelect.Item>Orange</SingleSelect.Item>
+            <SingleSelect.Item>Yellow</SingleSelect.Item>
+            <SingleSelect.Item>Green</SingleSelect.Item>
+            <SingleSelect.Item>Blue</SingleSelect.Item>
+            <SingleSelect.Item>Purple</SingleSelect.Item>
+            <SingleSelect.Item>Black</SingleSelect.Item>
+            <SingleSelect.Item>White</SingleSelect.Item>
+            <SingleSelect.Item>Lime</SingleSelect.Item>
+            <SingleSelect.Item>Fushsia</SingleSelect.Item>
           </SingleSelect>
 
           <SingleSelect label="Select grouped colours">
-            <Section title="Warm Colours">
-              <Item>Red</Item>
-            </Section>
-            <Section title="Cool Colours">
-              <Item>Blue</Item>
-            </Section>
+            <SingleSelect.Section title="Warm Colours">
+              <SingleSelect.Item>Red</SingleSelect.Item>
+            </SingleSelect.Section>
+            <SingleSelect.Section title="Cool Colours">
+              <SingleSelect.Item>Blue</SingleSelect.Item>
+            </SingleSelect.Section>
           </SingleSelect>
         </StickerSheet.Row>
       </StickerSheet>
--- a/packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
+++ b/packages/components/src/__alpha__/SingleSelect/SingleSelect.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { Item, Section } from '@react-stately/collections'
 import { ComboBox, Select } from './subcomponents'
 import { type ComboBoxProps, type SelectItem, type SelectProps } from './types'
 
@@ -13,3 +14,6 @@ export const SingleSelect = <T extends SelectItem>(
 
   return <Select {...(rest as SelectProps<T>)}>{children}</Select>
 }
+
+SingleSelect.Item = Item
+SingleSelect.Section = Section
--- a/packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.spec.stories.tsx
+++ b/packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.spec.stories.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import { Item } from '@react-stately/collections'
 import { type Meta, type StoryObj } from '@storybook/react'
 import { expect, screen, userEvent, waitFor } from '@storybook/test'
 import { SingleSelect } from '../SingleSelect'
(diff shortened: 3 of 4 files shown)


packages/components/package.json
```diff
@@ -74,7 +74,7 @@
     "lint:ts": "tsc --noEmit",
     "build": "ts-patch install -s && pnpm package-bundler build-shared-ui && pnpm build:styles",
     "build:global-styles": "postcss ./styles/global.css --output dist/global.css",
-    "build:combine-styles": "concat-cli -f ./dist/global.css ./dist/styles.css -o ./dist/styles.css && rm ./dist/global.css",
+  "build:combine-styles": "if [ -f ./dist/styles.css ]; then concat-cli -f ./dist/global.css ./dist/styles.css -o ./dist/styles.css; else mv ./dist/global.css ./dist/styles.css; fi && rm -f ./dist/global.css",
```
mcwinter07: Just to get a little more clarity here do you mind walking through this update? Is more a convenience update or is it required from the change to the Select?

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -1,53 +1,81 @@
...
+          <div
+            id="popover-id"
+            {...popoverProps}
+            ref={popoverRef}
+            style={{
+              ...popoverProps.style,
+            }}
+            className={styles.popover}
+          >
+            <DismissButton onDismiss={state.close} />
```
mcwinter07, line 71: I was testing out keyboard behaviour in the "playground"  Story and notice there was 3 tab stops when the popover is active.
First was the trigger, then list, then third tab stop I would have assumed was this -  though was unable to trigger a dismiss action

packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBox.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { SingleSelectContext } from '../context'
+import { type ComboBoxProps, type SelectItem } from '../types'
+import { ComboBoxTrigger } from './ComboBoxTrigger'
+import { List } from './List'
+import { Popover } from './Popover'
+
+export function ComboBox<T extends SelectItem>(props: ComboBoxProps<T>): JSX.Element {
+  const { items, children, label } = props
+
+  const { contains } = useFilter({ sensitivity: 'base' })
```
mcwinter07, line 14: Is the filtering functioning as expected? I tried enabling the isCombobox control and it didn't appear to be reducing items in the list
Screenshot 2025-09-04 at 12 15 36 pm

mcwinter07: Also, could we apply the focus styles to the active-descendant? I know we're pre-styling but makes it a little easier to validate the expected keyboard behaviour from comboboxes

Kitty-Al: Just pushed up some changes to fix the storybook example & make testing this a bit easier

packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
Kitty-Al: Looks like this is an accessibility issue with React Aria wherein if you use a select component with no focusable element following it in the dom, the component tabs to an area from which you cannot escape and can only tab forward/backward - I'll raise it with React Aria & document it in Storybook. But seems a pretty unlikely scenario for users to fall into

packages/components/package.json
```diff
@@ -126,6 +127,7 @@
     "react-focus-on": "^3.10.0",
     "react-popper": "^2.3.0",
     "react-select": "^5.10.2",
+    "react-stately": "^3.41.0",
```
mcwinter07: I do wonder about pulling in the "react-stately" mono package as we have previously used the individual packages like `@react-stately/collections` or `@react-stately/datepicker`, presumably for tree shaking purposes.

Was there any crossover `react-stately` and the packages we already have and the types/components they export?

packages/components/src/__alpha__/SingleSelect/subcomponents/ListSection/ListSection.tsx
```diff
@@ -1,25 +1,47 @@
...
+            role="presentation"
+            aria-hidden
+            className={styles.listSectionHeader}
+          >
+            {section.rendered}
+          </span>
+        )}
+        <ul
+          {...groupProps}
+          style={{
```
mcwinter07: Can we change the inline styles to CSS modules. I know we're not quite at full styling stage but just don't won't to miss this in future reviews 😉

packages/components/package.json
Kitty-Al: Definitely on board with this: I've updated to using either `@react-stately/combobox` or `@react-stately/select` but still need `react-stately` for `Item` and `Section` which are only available there, and used as examples in Storybook. I've moved `react-stately` to be a dev dep

packages/components/src/__alpha__/SingleSelect/subcomponents/List/List.tsx
```diff
@@ -1,18 +1,28 @@
...
+    } else {
+      return <ListItem key={node.key} item={node} state={state} />
+    }
+  }
+
   return (
-    <RACListBox className={classNames(styles.list, className)} {...props}>
-      {children}
-    </RACListBox>
+    <ul {...listBoxProps} ref={listBoxRef} className={styles.list}>
```
mcwinter07, line 24: Was doing another once over in VoiceOver and noticed that the bullet list was being announced - I think this might have been contributing to some of the strange behaviors with VO on Friday.

`list-style: none` is one of the few CSS properties that can impact the accessible tree so would recommend adding that in. Would recommend adding that in and spinning up VO

mcwinter07: Just a quick example of what I mean :) 

https://github.com/user-attachments/assets/d43df31d-6898-4245-a927-1440baa2fbe9

Kitty-Al: Yep - definitely looks like this is what was causing the issues with the number of items being announced incorrectly. Added `list-style: none` and tested in chrome/firefox/safari and the behaviour was correct

packages/components/src/__alpha__/SingleSelect/_docs/SingleSelect.stickersheet.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
+import { Item, Section } from '@react-stately/collections'
```
mcwinter07: So in a consumer implementation would they still be using the exports directly from `@react-stately/collections` or would we re-export them as a component?

I can see that before we have the components exported with dot notation with the and I'm wondering that the safer path 🤔 

Screenshot 2025-09-09 at 2 26 34 pm

My thinking is that our consumer (ideally) shouldn't have to know about `react-stately` and handle that as a dependency. If they can get it all from our SingleSelect, its self documenting in terms of the valid children.

```tsx
import { SingleSelect } from "@kaizen/components/alpha"

const MyLonelySingleSelect = (): JSX.Element => (
  <SingleSelect label="One is the loneliest number" isComboBox>
    <SingleSelect.Item>One</ SingleSelect.Item> // Or even the ListItem if that's at all possible?
  </SingleSelect>
)
```

It will also reduce the risk of mismatched versions of Item and Section being used as children and potentially causing issues.

mcwinter07: Mainly I'm thinking about Ben's [issue](https://cultureamp.slack.com/archives/C0189KBPM4Y/p1753851630932329) from a while back. There's potential that different versions of react-stately being used as Children could lead to unexpected errors

PR #6070, opened 2025-09-10
[KZN-3439] styling states

Kitty-Al:
## What

Applies styling for the `SingleSelect` component's various states:
- disabled
- read only
- primary (default) and secondary variants
- small medium and large sizes 

Handles label variants including:
- label hidden
- horizontal inline label or vertical 

Handles styling of List Items:
- specified disabled items 
- whether to display a selected state: a check or radio
- where to display the selected state: start or end

## Link to the designs 
https://www.figma.com/design/umhYV8q0yC4qwbCOIfJDY5/-Alpha--Select?node-id=474-40251&p=f&t=9gLcpo8O3Z5G18r0-0


packages/components/src/__alpha__/SingleSelect/subcomponents/SelectTrigger/SelectTrigger.tsx
```diff
@@ -23,20 +25,28 @@ export const SelectTrigger = ({
...
   return (
     <button
       {...buttonProps}
       type="button"
       style={{ '--anchor-name': anchorName } as React.CSSProperties}
-      className={styles.trigger}
+      className={classNames(styles.trigger, {
+        [styles.disabled]: isDisabled,
+        [styles.readOnly]: isReadOnly,
+        [styles.secondary]: secondary,
```
ryanseddon, line 36: The secondary styles make the text and arrow jump on hover

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
```diff
@@ -1,23 +1,60 @@
...
   return (
     <li
       key={item.key}
       {...optionProps}
       ref={ref}
       className={classNames(styles.listItem, {
-        [styles.focused]: isFocused,
+        [styles.disabled]: isDisabled,
+        [styles.selected]: isSelected,
+        [styles.selectedStartPosition]: isStart,
```
ryanseddon, line 61: We'd need to consider passing a `className` on the consumer side in the current state this will override it but it should combining it so the spread should be last and the `className` should be plucked off in the destructure and passed to this to combine

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
```diff
@@ -1,23 +1,60 @@
...
 export const ListItem = <T extends SelectItem>({ item, state }: ListItemProps<T>): JSX.Element => {
+  const { selectedIcon, selectedPosition } = useSingleSelectContext()
   const ref = React.useRef(null)
-  const { optionProps, isFocused } = useOption({ key: item.key }, state, ref)
+  const { optionProps, isSelected, isDisabled } = useOption({ key: item.key }, state, ref)
+
+  const isStart = selectedPosition === 'start'
+  const isEnd = selectedPosition === 'end'
+  const isCheck = selectedIcon === 'check'
+  const isRadio = selectedIcon === 'radio'
```
ryanseddon, line 26: I think we should reconsider these props as it should be a child passed to the ListItem with only the tick possibly being defaulted to end

I'd imagine we'd have the following:

Defaults to tick at end (rtl aware):
```tsx
<ListItem>Foo</ListItem>
<ListItem selectedPosition="start">Foo</ListItem>
```

If we go beyond the default it should be a child
```tsx
<ListItem>
  <Radio />
  {Foo}
</ListItem>

<ListItem>
  {Foo}
  <Radio />
</ListItem>
```

This way it aligns with adding an icon/avatar that may coexist with a radio button too
```tsx
<ListItem>
  <Avatar />
  {Foo}
</ListItem>

<ListItem>
  <Icon />
  {Foo}
  <Checkbox />
</ListItem>
```

We can handle this styling complexity with `:has()` to know when the parent ListItem has child or certain type as well as the sibling combinator selectors to know where they sit

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
```diff
@@ -1,23 +1,60 @@
...
+          </div>
+        )
+      }
+
+      return isStart ? <div className={styles.emptySpacer} /> : null
+    }
+
+    if (isRadio) {
+      return (
+        <div className={styles.selectedIconWrapper}>
```
ryanseddon, line 43: If we do the declarative style then we negate challenges here with passing classnames. We want to avoid magic props that render styled children as much as possible.

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
```diff
@@ -1,23 +1,60 @@
...
+          <Radio name="test" id="test" selectedStatus={isSelected} />
```
ryanseddon: Remove the name and id hardcoded strings

packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBoxTrigger/ComboBoxTrigger.tsx
```diff
@@ -31,12 +32,28 @@ export const ComboBoxTrigger = ({
...
+          [styles.large]: size === 'large',
+        })}
+      >
+        <input
+          {...inputProps}
+          ref={inputRef}
+          className={classNames(styles.input, { [styles.smallText]: size === 'small' })}
+        />
+
+        {!isReadOnly && <ChevronButton {...buttonProps} buttonRef={buttonRef} />}
```
ryanseddon, line 117: The clickable area of this button is very small and should be larger

packages/components/src/__alpha__/SingleSelect/subcomponents/Select/Select.module.css
```diff
@@ -0,0 +1,31 @@
...
+    grid-template-columns: auto 1fr;
+    grid-template-rows: auto auto;
+    column-gap: var(--spacing-12);
+    align-items: center;
+  }
+
+  .label {
+    grid-column: 1;
+    grid-row: 1;
+    margin-bottom: var(--spacing-6);
```
NJWong, line 22: Can I get a link to the Figma designs for the side label? Wondering if it's just me or the side label makes the component look unbalanced due to the bottom margin

Screenshot 2025-09-12 at 3 40 52 pm
Screenshot 2025-09-12 at 3 40 59 pm

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
Kitty-Al: I've updated this now so that the `className` can be passed to the wrapper of the item. But currently the user would have no way of updating or changing the disabled, selected, or focused state

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
Kitty-Al: Decided to handle the selected icon/radio internally and have moved the prop onto the `ListItem` itself

packages/components/src/__alpha__/SingleSelect/subcomponents/ListItem/ListItem.tsx
Kitty-Al: Decided to handle the selected radio/icon internally - instead of exposing the selected state and having the user handle the selected styling themselves.

NJWong (approved):
Nice work! Looks good to me

PR #6072, opened 2025-09-11
[KZN-3439] Clear selected item button

Kitty-Al:
## What

Added a `ClearButton` to the combobox - which clears the selected state of the input
The button is only seen once an item has been selected
Clicking the button will clear the selection
The `ClearButton` is accessible via keyboard and can be tabbed to. Once the button is interacted with focus is moved back to the input

Item Selected
Screenshot 2025-09-12 at 8 13 43 am

ClearButton focused
Screenshot 2025-09-12 at 8 13 52 am

### DIFF
--- a/packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBoxTrigger/ComboBoxTrigger.tsx
+++ b/packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBoxTrigger/ComboBoxTrigger.tsx
@@ -59,14 +59,14 @@ const DropdownButton = (props: DropdownButtonProps): JSX.Element => {
   const dropdownButton = formatMessage(
     {
       id: 'singleSelect.dropdownButton',
-      defaultMessage: 'Show {field} suggestions',
+      defaultMessage: 'Show suggestions for {field}',
       description: 'Aria label text for the SingleSelect button to open and close suggestions list',
     },
     { field: fieldLabel },
   )
 
   const { buttonProps } = useButton(
-    { ...props, 'aria-label': String(dropdownButton) },
+    { ...props, 'aria-label': String(dropdownButton), 'aria-labelledby': undefined },
     props.buttonRef,
   )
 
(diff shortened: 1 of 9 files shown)


packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBoxTrigger/ComboBoxTrigger.tsx
```diff
@@ -1,13 +1,69 @@
...
+
+  const { buttonProps } = useButton(
+    {
+      'onPress': () => {
+        if (isComboBox) {
+          state.setSelectedKey(null)
+        }
+        inputRef.current?.focus()
+      },
+      'aria-label': clearButtonAlt,
```
mcwinter07: Instead of `aria-label` can we lean on the `<VisuallyHidden>` as children of the `button` component to hoist up the accessible name?

mcwinter07: We'll also need to make sure that any of the button accessible labels are unique as there could be multiple SingleSelect's on a page, ie:
Screenshot 2025-09-12 at 2 00 55 pm

In this instance we'd be fine but there's a high likelihood in prod that there could be many "Clear Selection"'s listed in the rotor.

packages/components/src/__alpha__/SingleSelect/subcomponents/ComboBoxTrigger/ComboBoxTrigger.tsx
```diff
@@ -1,13 +1,69 @@
...
   const { state } = useSingleSelectContext()
   const { buttonProps } = useButton(props, props.buttonRef)
+  const { formatMessage } = useIntl()
+
+  const chevronButtonClose = formatMessage({
+    id: 'singleSelect.toggleButtonClose',
+    defaultMessage: 'close',
+    description: 'Alt text for button that closes list',
+  })
+
```
mcwinter07, line 67: After looking at the A11Y tree I'm not sure if these icons need alts as the parent buttons will always supersede the annouced content as the use aria-labelled or aria labelledby.

Screenshot 2025-09-12 at 2 15 00 pm

Screenshot 2025-09-12 at 2 15 26 pm

There's also the `aria-expanded` property that's used so that a pretty good indicator to the open and close state :)

mcwinter07: Thinking on this though, we will need to make sure that descriptive text for the "show suggestions" label is unique

Kitty-Al: I've removed the accessible label in favour of relying on the Select or Combobox behaviour stating whether the list is expanded or collapsed

mcwinter07 (approved):

PR #6073, opened 2025-09-11
[KZN-3416] Fix popover regressions

Kitty-Al:
## Why

Recent work done to refactor the alpha `SingleSelect` to manage state instead of using components caused 2 regressions in the popover component. 

1. The popover itself was rendering before all the items loaded - causing the list to render on the bottom even in situations where it would be cutoff.

Screenshot 2025-09-12 at 8 40 38 am

2. Combobox was not rendering correctly: because the inputRef was being passed to the popover instead of the triggerWrapper 

## What

This PR fixes both of these regressions. Calling the updatePosition once the popover has rendered, and passing the correct wrapper ref to the popover for the combobox

### DIFF
--- a/packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
+++ b/packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
@@ -38,24 +38,32 @@ export const Popover = <T extends SelectItem>({
   )
 
   useLayoutEffect(() => {
-    if (!supportsAnchorPositioning) return
-    if (!state.isOpen) return
+    if (!supportsAnchorPositioning || !state.isOpen) return
 
     updatePosition()
-  }, [updatePosition, supportsAnchorPositioning, state.isOpen])
-
-  useLayoutEffect(() => {
-    if (!supportsAnchorPositioning) return
 
     const popover = manualPopoverRef?.current
-    if (!popover?.showPopover || !popover?.hidePopover) return
 
-    if (state.isOpen) {
+    if (popover?.showPopover) {
       popover.showPopover()
-    } else {
+    }
+
+    return () => {
+      if (popover?.hidePopover) {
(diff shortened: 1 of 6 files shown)


packages/components/src/__alpha__/SingleSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -31,12 +31,19 @@ export const Popover = <T extends SelectItem>({
...
 
   const supportsAnchorPositioning = useSupportsAnchorPositioning()
-  const { popoverStyle, isPositioned } = usePositioningStyles(
+  const { popoverStyle, isPositioned, updatePosition } = usePositioningStyles(
     restProps.triggerRef as React.RefObject<HTMLElement>,
     manualPopoverRef,
     anchorName,
   )
 
+  useLayoutEffect(() => {
```
ryanseddon, line 40: I think we can simplify the two useLayoutEffects here so one handles positioning and showing the popover and the other is for hiding. That what we can avoid having logic spread across both of them that handle a single use case:

So this one could change like so:
```js
useLayoutEffect(() => {
  // combine the two into one if
  if (!supportsAnchorPositioning || !state.isOpen) return

  updatePosition()

  const popover = manualPopoverRef?.current

  if (popover?.showPopover) {
    popover.showPopover()
  }

  // We can add a return function to make sure any visible popovers are cleaned up on unmount too
  return () => {
    if (popover?.hidePopover) {
      popover.hidePopover()
    }
  }
}, [state.isOpen, supportsAnchorPositioning, updatePosition])
```

Then the other effect is just for hiding:
```js
useLayoutEffect(() => {
  if (!supportsAnchorPositioning || state.isOpen) return
       
  const popover = manualPopoverRef?.current
    if (popover?.hidePopover) {
      popover.hidePopover()
    }
}, [supportsAnchorPositioning, state.isOpen])
```

PR #6115, opened 2025-09-18
[KZN-3549] Release SingleSelect alpha and update docs

Kitty-Al:
## Why

Releasing the `SingleSelect` alpha - so that it can now be imported
Updating docs in `SingleSelect` 

## What

- exporting alpha from package.json
- adds some more example stories 
- updates the docs with information on incomplete work
- updates args table with types due to Storybook not inferring fully
- fixes a scrolling bug seen on Firefox when anchoring position is not supported


ryanseddon (approved):
Looks good, It'd be worth having a card to follow up on the focus flicker that is happening when you move your mouse around the list items

PR #6116, opened 2025-09-18
feat: patch rollup-plugin-postcss to handle multiple inputs determini…

ryanseddon:
## Why

Our css order seems random and not deterministic usually it's fine but sometimes it's not and it breaks our consumers and css specificity

## What

We need to patch rollup-plugin-postcss to be aware of multi inputs within the rollup config otherwise it doesn't place the css in the right order.

Tried this on a few internal projects and am getting correct results.

### DIFF
--- a/.changeset/fine-parrots-return.md
+++ b/.changeset/fine-parrots-return.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+CSS ordering fix - Patch rollup-plugin-postcss to handle multi inputs correctly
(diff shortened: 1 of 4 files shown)


mcwinter07 (approved):
Nice, patch syntax highlight is a little gnarly to read but I that looks hunky dory in my books 👍 

Non-block wondering 💭: Since this is a patch for V1 I don't forsee any major issue here, but will we have to update this patch version if we run dep updates for the rollup-plugin-postcss-css in V2 (or will we version lock it for now?)

mcwinter07:
> We'll apply the same patch to v2, this plugin is no longer maintained so long term we may need to find an alternative

Oh in that case wew! One of the rare instances where an unmaintained library is a win!

PR #6135, opened 2025-09-23
KAIO v3 upcoming release docs

ryanseddon:
## Why

The upcoming release will be v3 this is is WIP and not fully finalised just yet

## What

Move the old upcoming release to past tense and rename to recent release
Introduce new v3 upcoming release doc

### DIFF
--- a/docs/pages/upcoming-releases.mdx
+++ b/docs/pages/upcoming-releases.mdx
@@ -179,6 +179,8 @@ A codemod will be provided to make the migration off these.
 
 ## `@kaizen/components@v4`
 
+September 1, 2026
+
 ### Components being deprecated
 
 The following components are being deprecated in v3 and will be removed/replaced in a future major version:
--- a/docs/components/ReplacementNotice/ReplacementNotice.tsx
+++ b/docs/components/ReplacementNotice/ReplacementNotice.tsx
@@ -2,11 +2,13 @@ import React from 'react'
 import { Unstyled } from '@storybook/blocks'
 import { InlineNotification } from '~components/Notification'
 
-export const ReplacementNotice = (): JSX.Element => {
+export const ReplacementNotice = ({ isFuture }): JSX.Element => {
+  const wording = isFuture ? 'will be' : 'has been'
+
   return (
     <Unstyled>
       <InlineNotification variant="cautionary" persistent title="Deprecated">
-        {`This component has been replaced by a 'next' version and will be removed in an upcoming major release. `}
+        {`This component ${wording} replaced by a 'next' version and will be removed in an upcoming major release. `}
       </InlineNotification>
     </Unstyled>
   )
--- a/packages/components/src/SingleSelect/_docs/SingleSelect--api-specification.mdx
+++ b/packages/components/src/SingleSelect/_docs/SingleSelect--api-specification.mdx
@@ -6,7 +6,7 @@ import * as SingleSelectStories from './SingleSelect.stories'
 
 # SingleSelect
 
-<ReplacementNotice />
+<ReplacementNotice isFuture={true} />
 
 <ResourceLinks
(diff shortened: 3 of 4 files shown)


docs/pages/upcoming-releases.mdx
```diff
@@ -4,120 +4,44 @@ import { Meta, Markdown } from '@storybook/blocks'
...
       <th>Recommendation</th>
     </tr>
   </thead>
   <tbody>
     <tr>
-      <td>
-        [`Skirt`](https://60a1e4a102f0cb003b5d19d6-xvoxshqkig.chromatic.com/?path=/docs/components-skirt-deprecated--docs)
-      </td>
-      <td>Remove any usages</td>
+      <td>`TileGrid`</td>
```
ryanseddon: Should we add the depreciation notice (`<RemovalNotice />`) to these component docs

docs/pages/recent-releases.mdx
```diff
@@ -0,0 +1,364 @@
...
+      <td>`BrandMoment`</td>
+      <td>`mood`</td>
+      <td>`migrateBrandMomentMoodToVariant`</td>
+      <td>Use `variant` instead</td>
+    </tr>
+    <tr>
+      <td>`Card`</td>
+      <td>`variant`</td>
+      <td>`migrateCardVariantToColor`</td>
+      <td>Use `color` instead</td>
```
StaphSynth, line 150: Everything else seems to use `variant`, how come `Card` uses `color` for this?

docs/pages/recent-releases.mdx
```diff
@@ -0,0 +1,364 @@
...
+      <td>`mood`</td>
+      <td>`migrateMultiActionTileMoodToVariant`</td>
+      <td>Use `variant` instead</td>
+    </tr>
+    <tr>
+      <td>`Well`</td>
+      <td>`variant`</td>
+      <td>`migrateWellVariantToColor`</td>
+      <td>
+        This will not fallback to default variant. default can be used but must be explicitly passed
```
StaphSynth:
```suggestion
        This will not fallback to default variant. Default can be used but must be explicitly passed
```

docs/pages/upcoming-releases.mdx
StaphSynth: I think so

StaphSynth: Are they marked as `@deprecated` in the current v2 export?

docs/pages/upcoming-releases.mdx
```diff
@@ -130,235 +54,160 @@ Imports from the new names will be allowed shortly, and support for the old name
...
-        <ul>
-          <li>
-            Not needed if you are using `KaizenProvider` from `@kaizen/components` or
-            `defaultPreset` from next-services.
-          </li>
-          <li>If necessary, import from `@kaizen/design-tokens/js`</li>
-        </ul>
-      </td>
+      <td>`useResizeObserver`</td>
+      <td>Use native `ResizeObserver` DOM API instead</td>
```
StaphSynth: Perhaps a link to MDN docs here? https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

docs/pages/upcoming-releases.mdx
ryanseddon: No didn't do that yet but I think a long lead time on marking as deprecated makes sense so will add the notice and the jsdoc comments

docs/pages/recent-releases.mdx
ryanseddon: They're are literal colour selections vs the variant being more around the intent of something "primary", "secondary" etc

StaphSynth (approved):

StaphSynth (approved):

PR #6304, opened 2025-12-01
[UUXP-55] Update TitleBlock mobile designs 

AliceMenzie:
## Why

https://cultureamp.atlassian.net/browse/UUXP-55

- Remove the mobile action draw and replace with updated design

## What
- replace mobile drawer with collapsed menu in main titleblock in mobile  view
- removed mobile actions and replace them in main actions
- decrease separator size from 3px to 1px


.changeset/cold-pears-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kaizen/components': patch
```
fantasticmrdavid: Maybe this should be a minor change instead of a patch?

fantasticmrdavid (dismissed):
LGTM great work Alice!

fantasticmrdavid (dismissed):

PR #6580, opened 2026-04-14
fix: Export SelectToggle subcomponent from SingleSelect.

dan-lennox:
Export SelectToggle subcomponent from SingleSelect

The release notes from Kaizen 3.0.0 indicate that these subcomponents should be directly importable "Import SelectToggle and ListBox directly instead of via SingleSelect.TriggerButton/ListBox".

I can't see any cultureamp repos actually using ListBox directly so I've only exported the SelectToggle component which appears to be used by a bunch of repos via the depreciated SingleSelect.TriggerButton alias. Those repos will need to be able to import SelectToggle directly when moving to Kaizen 3.0.0.

See full release notes in e769d8beea2d6a3fce8271fd6ffb2d93e217821a

### DIFF
--- a/.changeset/rude-stars-cover.md
+++ b/.changeset/rude-stars-cover.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+fix: Export SelectToggle subcomponent from SingleSelect.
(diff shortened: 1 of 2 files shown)


ckychris (approved):

Kitty-Al (approved):
Yep all good - this is necessary for users to be able to customise the trigger

PR #6646, opened 2026-05-04
[KZN-3999] Fix SingleSelect cause skipping content in Collapsible

mia-nguyen0411:
## Why
SingleSelect causes VO to skip the remaining content of a Collapsible and jump to the next content block. 
Content between the SingleSelect and the end of the collapsible is unreachable via VO+Right Arrow in Safari on macOS.
Ref ticket [KZN-3999](https://cultureamp.atlassian.net/browse/KZN-3999)

## What
- `Collapsible.tsx`: add `sectionRef` using `useCallBack` to manually add and remove `inert` attributes
- `Collapsible.stories.tsx`: add story to check `SingleSelect` place inside `Collapsible` &rarr; could be removed once issue is solved
- `Collapsible.spec.tsx`: add 2 tests to see if `inert` is detected in the right case

[KZN-3999]: https://cultureamp.atlassian.net/browse/KZN-3999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Collapsible/Collapsible/_docs/Collapsible.stories.tsx
```diff
@@ -100,3 +101,33 @@ export const Controlled: Story = {
   },
   parameters: { docs: { source: { code: controlledSourceCode } } },
 }
+
```
mia-nguyen0411, line 104: will remove this story after ensuring the issue is fixed

packages/components/src/Collapsible/Collapsible/Collapsible.tsx
```diff
@@ -128,8 +141,13 @@ export const Collapsible = ({
...
       </div>
       {(!lazyLoad || isOpen) && (
-        <AnimateHeight height={isOpen ? 'auto' : 0} data-testid={`collapsible-section-${id}`}>
+        <AnimateHeight
+          height={isOpen ? 'auto' : 0}
+          disableDisplayNone
+          data-testid={`collapsible-section-${id}`}
+        >
           <div
+            ref={sectionRef}
```
ryanseddon: We should just add the inert here and make it conditional 

```suggestion
            inert={isOpen ? undefined : true}
```

We're on older react types so it will complain about inert we can ts-ignore or add it to our types

```tsx
import 'react'                                                                                                                                            
                                                                                                                                                         
declare module 'react' {                                                                                                                                  
  interface HTMLAttributes<T> {                                                                                                                           
    inert?: boolean                                                                                                                                       
  }                                                                                                                                                       
}                                                                                                                                                         
```

mia-nguyen0411: Thanks Ryan. I follows your suggestion then working around a little bit to avoid the additional declaration. Happy to know your thoughts!

packages/components/src/Collapsible/Collapsible/_docs/Collapsible.stories.tsx
mia-nguyen0411: Update: I will keep this stories for quick check

packages/components/src/Collapsible/Collapsible/Collapsible.tsx
```diff
@@ -128,12 +128,17 @@ export const Collapsible = ({
...
+          height={isOpen ? 'auto' : 0}
+          disableDisplayNone
+          data-testid={`collapsible-section-${id}`}
+        >
           <div
             id={sectionId}
             className={classnames(styles.section, noSectionPadding && styles.noPadding)}
             role="region"
             aria-labelledby={buttonId}
+            {...(!isOpen && { inert: true as unknown as undefined })}
```
ryanseddon: I think its better to add the `@ts-ignore` with a TODO comment above it as once we move to @types/react@19 this won't complain anymore

mia-nguyen0411: just updated! Thanks Ryan.

ryanseddon (approved):

PR #6648, opened 2026-05-08
feat(BrandMoment, ErrorPage): add optional tag prop

Kitty-Al:
## Summary

- Adds an optional `tag` prop (`ElementType`) to `BrandMoment` that allows consumers to override the default `<main>` HTML element rendered by the component
- Adds the same optional `tag` prop to `ErrorPage`, which passes it through to the underlying `BrandMoment`
- Both props default to `main` when not provided, so existing usage is unaffected

## Motivation

When `BrandMoment` or `ErrorPage` is used in a context where a `<main>` element already exists (or is semantically inappropriate), consumers need a way to swap it for a different element like `<section>` or `<div>`. This is needed for the `AppChrome` rollout, where repos have errorpages wrapped in the old nav, now moving to the `AppChrome` which comes with it's own `<main>` element

## Test plan

- [x] Added `BrandMoment.spec.tsx` with tests for default `main` rendering and custom tag override
- [x] Added tests to `ErrorPage.spec.tsx` verifying the `tag` prop passes through to `BrandMoment`
- [x] All 13 tests pass
- [x] TypeScript compiles with no errors

### DIFF
--- a/.changeset/add-optional-tag-brand-moment.md
+++ b/.changeset/add-optional-tag-brand-moment.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': minor
+---
+
+Add optional `tag` prop to BrandMoment and ErrorPage components to allow overriding the default `main` HTML element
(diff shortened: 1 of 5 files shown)


AliceMenzie (approved):
Have approved, the tests are failing unrelated change though(checkbox cors error), so just trying to rerun them

PR #6752, opened 2026-06-03
feat(components): add ref props to TextField, TextAreaFIeld, DatePicker and TimeField

RoMilton:
## Why

React Hook Form's Controller component provides `field.ref` as a callback ref (a function), but our form field components currently only accept `RefObject` type. Other components are missing the ref prop altogether, forcing consumers to write workarounds. 

## What

- Added props `inputRef` to components DatePicker and Timefield
- Widened the ref prop types from `React.RefObject` to `React.Ref` for TextField. This is backwards compatible — existing RefObject usage continues to work

## Example: Using with react-hook-form

Before this change, this common pattern fails with a type error:
```
import { Controller } from "react-hook-form"
import { TextField } from "@kaizen/components"

<Controller
  name="email"
  control={control}
  render={({ field }) => (
    <TextField
      {...field}
      inputRef={field.ref}  // ❌ Type error: RefCallBack is not assignable to RefObject
    />
  )}
/>

```
After this change, the same code works as expected.

## Demo

Using the prop, we can programatically focus on these fields using RHF `setFocus` method. Here is very simple demo:

https://github.com/user-attachments/assets/5ebd934d-d0e9-4941-8a09-b16db8df37a8

In the real world it will get used to focus when errors occur.

## Out of scope

I didn't add the ref prop to component Rich Text Editor because it uses a 3rd party library. Because it's based on vanilla JS and not React it has no concept of refs. Adding a ref will involve a workaround, or some kind of anti pattern that I'll leave to separate PR.

### DIFF
--- a/packages/components/src/DateInput/DateInput/DateInput.tsx
+++ b/packages/components/src/DateInput/DateInput/DateInput.tsx
@@ -2,7 +2,6 @@ import React from 'react'
 import classnames from 'classnames'
 import { Input, type InputProps } from '~components/Input'
 import { Label } from '~components/Label'
-import { isRefObject } from '~components/utils/isRefObject'
 import styles from './DateInput.module.css'
 
 type OmittedInputProps = 'reversed' | 'type' | 'inputRef'
@@ -23,7 +22,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
         disabled={disabled}
       />
       <Input
-        inputRef={isRefObject(ref) ? ref : undefined}
+        inputRef={ref}
         id={id}
         type="text"
         autoComplete="off"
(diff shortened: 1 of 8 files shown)


packages/components/src/Input/Input/Input.tsx
```diff
@@ -8,7 +8,7 @@ export type InputType = (typeof InputTypes)[number]
 export type InputStatusType = (typeof InputStatus)[number]
 
 export type InputProps = {
-  inputRef?: React.RefObject<HTMLInputElement>
+  inputRef?: React.Ref<HTMLInputElement>
```
RoMilton, line 11: This widening of type makes it accept both callback refs (used by RHF) and useRef style refs, so its backwards compatible.

packages/components/src/DateInput/DateInput/DateInput.tsx
```diff
@@ -23,7 +22,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
         disabled={disabled}
       />
       <Input
-        inputRef={isRefObject(ref) ? ref : undefined}
```
RoMilton, line 26: No need for defensive checks like this - that's TS's job.

mia-nguyen0411 (approved):
Tried some tests from my end and they passed and complied. Thanks for your contribution!

PR #6793, opened 2026-06-15
fix(Tabs): only scroll selected tab into view on selection change

Kitty-Al:
## Summary

`Tabs` scrolled the selected tab into view on **mount**, which jumped the page for consumers rendering `Tabs` below the fold ([KZN-3363](https://cultureamp.atlassian.net/browse/KZN-3363)). The scroll effect now runs only when the selection actually changes via user interaction.

## Change

- [`TabList.tsx`](packages/components/src/Tabs/subcomponents/TabList/TabList.tsx) — guard the `scrollIntoView` effect with the previous `selectedKey`. On mount (and on `containerElement`/`isDocumentReady` settling) `prev === current`, so it no-ops. A genuine selection change scrolls as before.

## Behaviour

- **Land on a page with Tabs below the fold** → page stays put, no jump.
- **Select a tab** → selected tab scrolls into view within the horizontal strip.

## Stories / tests

- `Components/Tabs → Below The Fold` — placeholder page content pushes the Tabs below the fold; demonstrates no scroll on load, scroll on selection.
- `Components/Tabs/Tabs tests → Scrolls into view on selection only` — play test spying on `scrollIntoView`: asserts **0 calls on mount**, **>0 after selecting a tab**.

## Changeset

Patch — `@kaizen/components`.

Closes KZN-3363

[KZN-3363]: https://cultureamp.atlassian.net/browse/KZN-3363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/Tabs/_docs/Tabs.spec.stories.tsx
+++ b/packages/components/src/Tabs/_docs/Tabs.spec.stories.tsx
@@ -117,6 +117,45 @@ export const ArrowsShowingAndHidingRTL: Story = {
   },
 }
 
+let scrollIntoViewCalls = 0
+const originalScrollIntoView = HTMLElement.prototype.scrollIntoView
+
+/**
+ * The selected tab must NOT be scrolled into view on mount (KZN-3363) — only when
+ * the selection changes via user interaction. This protects users who render Tabs
+ * below the fold from having the page jump on load.
+ */
+export const ScrollsIntoViewOnSelectionOnly: Story = {
+  name: 'Scrolls into view on selection only',
+  beforeEach: () => {
+    scrollIntoViewCalls = 0
+    HTMLElement.prototype.scrollIntoView = function (): void {
+      scrollIntoViewCalls += 1
+    }
+    return () => {
+      HTMLElement.prototype.scrollIntoView = originalScrollIntoView
+    }
+  },
+  render: (args) => (
+    <div style={{ maxWidth: '500px' }}>
+      <Tabs defaultSelectedKey="one" {...args} />
+    </div>
+  ),
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement.parentElement!)
+
+    await step('No scroll into view on mount', async () => {
+      expect(scrollIntoViewCalls).toBe(0)
+    })
+
+    await step('Scrolls into view when a tab is selected', async () => {
+      const tabFour = await canvas.findByRole('tab', { name: 'Tab 4' })
+      await userEvent.click(tabFour)
+      await waitFor(() => expect(scrollIntoViewCalls).toBeGreaterThan(0))
+    })
+  },
+}
+
 export const AsyncLoaded: Story = {
   render: () => {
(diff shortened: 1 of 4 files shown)


packages/components/src/Tabs/_docs/Tabs.stories.tsx
```diff
@@ -87,3 +88,117 @@ export const Controlled: Story = {
...
+ * This story mimics a real page: a heading and several sections of content push the
+ * `Tabs` below the fold. On load the page stays at the top. Scroll down and select a
+ * tab — only then does the selected tab scroll into view within the strip.
+ */
+export const BelowTheFold: Story = {
+  parameters: {
+    chromatic: { disable: true },
+  },
+  render: () => (
+    <div style={{ maxWidth: '760px' }}>
```
ryanseddon, line 103: Rather than all this text you could just set a div with height 100dvh so you know the tabs will be out of view for any viewport and just a short sentence

packages/components/src/Tabs/_docs/Tabs.spec.stories.tsx
```diff
@@ -117,6 +117,47 @@ export const ArrowsShowingAndHidingRTL: Story = {
...
+
+/**
+ * The selected tab must NOT be scrolled into view on mount (KZN-3363) — only when
+ * the selection changes via user interaction. This protects users who render Tabs
+ * below the fold from having the page jump on load.
+ */
+export const ScrollsIntoViewOnSelectionOnly: Story = {
+  name: 'Scrolls into view on selection only',
+  beforeEach: () => {
+    scrollIntoViewCalls = 0
```
ryanseddon, line 131:
```suggestion
    let scrollIntoViewCalls = 0
```

packages/components/src/Tabs/_docs/Tabs.spec.stories.tsx
```diff
@@ -117,6 +117,47 @@ export const ArrowsShowingAndHidingRTL: Story = {
...
+    <div style={{ maxWidth: '500px' }}>
+      <Tabs defaultSelectedKey="one" {...args} />
+    </div>
+  ),
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement.parentElement!)
+
+    await step('No scroll into view on mount', async () => {
+      // Let all mount effects flush before asserting
+      await new Promise((r) => setTimeout(r, 500))
```
ryanseddon: Is this needed I thought storybook handles this with a step for you?

packages/components/src/Tabs/_docs/Tabs.spec.stories.tsx
Kitty-Al: `scrollIntoViewCalls` is scoped on line 20 so that it can be accessed via the play test - looks like it breaks if I try and scope it here

ryanseddon (approved):

PR #6796, opened 2026-06-15
[KZN-4066] Add custom icon

AliceMenzie:
## Why

https://cultureamp.atlassian.net/browse/KZN-4066

## What

Falcon - showing custom icon
Screenshot 2026-06-15 at 5 09 53 pm


packages/components/src/index.ts
```diff
@@ -25,6 +25,7 @@ export * from './Focusable'
 export * from './GuidanceBlock'
 export * from './Heading'
 export * from './Icon'
+export * from './Icon/CustomIcons'
```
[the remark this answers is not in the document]
AliceMenzie: this caused build errors as other components consuming `Icon` would also import the custom icon and issues with other packages (design-tokens) not compiling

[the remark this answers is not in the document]
mia-nguyen0411: just me wondering, not really feedback to change anything. Why we need to re-export the `CustomIcons`. If we're already exporting `Icon`, wouldn't `CustomIcons` be exported along with it? If we add more custom icons in the future, will we need to export them separately or we will then update the barrel export to support it?

packages/components/src/Icon/CustomIcons.ts
```diff
@@ -0,0 +1 @@
+export { default as UserSettingsIcon } from '../../assets/svgs/custom-icons/user-settings.svg'
```
ryanseddon: If we just do a named export we can avoid do the `default as`
```suggestion
export { UserSettingsIcon } from '../../assets/svgs/custom-icons/user-settings.svg'
```

packages/components/rollup.config.mjs
```diff
@@ -1,4 +1,21 @@
...
+  export interface Props extends SVGProps<SVGSVGElement> {
+    color?: string;
+    size?: number;
+  }
+
+  const ${variables.componentName}: FC<Props> = ({ size = 24, color = 'currentColor', ...props }) => (
+    ${variables.jsx}
+  );
+
+  export default ${variables.componentName};
```
ryanseddon: I think we should just export the component here rather than default

```suggestion
  export const ${variables.componentName}: FC<Props> = ({ size = 24, color = 'currentColor', ...props }) => (
    ${variables.jsx}
  );
```

packages/components/src/index.ts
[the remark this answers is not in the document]
AliceMenzie: its because of the file structure 
/Icons/index.ts  file exports the icon.tsx file 
but because i export from the customIcons.tsx and don't add it to the Icons/index.tsx it's not included in the "barreling" 

so we are not reexporting it, it's never included and so in the 
components/src/index, when i export 
export * from './Icon'
export * from './Icon/CustomIcons'
I can export them separately

in the future we would just update the customIcons.tx and add the new custom icons there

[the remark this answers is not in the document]
AliceMenzie: i think its confusing because of the `export * from './Icon'`
but it works because of some of the settings bundler settings - moduleResolution

packages/components/src/Icon/CustomIcons.ts
AliceMenzie: updated this, but it is different to solution because of typescript had an issue with the types.d.ts file of how i was defining svg as `*` . 

tried some other options but could only get 2 to work this one and or another option where we would need to have a type file for each svg. 

thinking  my latest change is probably the preferred option let me know

packages/components/src/Icon/CustomIcons.ts
```diff
@@ -1 +1 @@
-export { default as UserSettingsIcon } from '../../assets/svgs/custom-icons/user-settings.svg'
+export { SvgUserSettings } from '../../assets/svgs/custom-icons/user-settings.svg'
```
AliceMenzie: This is the default naming convention from SVGR -- i can update it back to the name+icon or anything else

docs/.storybook/main.ts
```diff
@@ -52,6 +53,45 @@ const config: StorybookConfig = {
...
     config?.plugins?.push(nodePolyfills())
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    const iconTemplate = (variables: any, { tpl }: any) => tpl`
+      import * as React from 'react';
+
+      export const ReactComponent = ({ size = 24, color = 'currentColor', ...props }) => (
+        ${variables.jsx}
+      );
+    `
```
ryanseddon, line 64:
```suggestion
    const iconTemplate = (variables: any, { tpl }: any) => tpl`
      import * as React from 'react';

      const ${variables.componentName} = ({ size = 24, color = 'currentColor', ...props }) => (
        ${variables.jsx}
      );

      export default ${variables.componentName};
    `
```

We should use the componentName and export that as default

packages/components/src/Icon/CustomIcons.ts
```diff
@@ -0,0 +1 @@
+export { ReactComponent as SVGUserSettingsIcon } from '../../assets/svgs/custom-icons/user-settings.svg'
```
ryanseddon:
```suggestion
import UserSettingsIcon from '../../assets/svgs/custom-icons/user-settings.svg'

export { UserSettingsIcon }
```

Using default we need to change the import and reexport here.

packages/components/src/Icon/types.ts
```diff
@@ -9,3 +10,8 @@ type MaterialIconNames = keyof typeof iconsMetadata
 // available in Material Icons, which appears to still work.
 // https://fonts.google.com/icons?icon.query=auto+awesome&icon.set=Material+Icons&icon.size=20&icon.color=%23000000&selected=Material+Icons+Outlined:auto_awesome:
 export type IconNames = MaterialIconNames | 'auto_awesome'
+
+export interface Props extends SVGProps<SVGSVGElement> {
+  color?: string
+  size?: number
+}
```
ryanseddon: With the changes we can then remove these additions

packages/components/rollup.config.mjs
```diff
@@ -1,4 +1,19 @@
...
+
+  export interface Props extends SVGProps<SVGSVGElement> {
+    color?: string;
+    size?: number;
+  }
+
+  export const ReactComponent: FC<Props> = ({ size = 24, color = 'currentColor', ...props }) => (
+    ${variables.jsx}
+  );
+`
```
ryanseddon, line 13:
```suggestion
const iconTemplate = (variables, { tpl }) => tpl`
  import * as React from 'react';
  import type { FC, SVGProps } from 'react';

  const ${variables.componentName}: FC<SVGProps<SVGSVGElement> & { color?: string; size?: number }> = ({ size = 24, color = 'currentColor', ...props }) => (
    ${variables.jsx}
  );

  export default ${variables.componentName};
`
```

We can inline the interface/type

packages/components/types.d.ts
```diff
@@ -17,3 +17,9 @@ declare module '*.png' {
   const content: string
   export default content
 }
+
+declare module '*.svg' {
+  const ReactComponent: import('react').FC<import('./src/Icon/types').Props>
+  export { ReactComponent }
+  export default ReactComponent
+}
```
ryanseddon, line 26:
```suggestion
declare module '*.svg' {
  const Icon: import('react').FC<
    import('react').SVGProps<SVGSVGElement> & { color?: string; size?: number }
  >
  export default Icon
}
```

I think it makes sense to have the type here rather than in Icon/types.ts

packages/components/src/Icon/_docs/IconsCustom.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { type Meta, type StoryObj } from '@storybook/react'
+import { SVGUserSettingsIcon } from '../CustomIcons'
```
ryanseddon: I think we can just drop the SVG prefix and have UserSettingsIcon be the naming convention

docs/.storybook/main.ts
AliceMenzie: updated

packages/components/src/Icon/CustomIcons.ts
AliceMenzie: updated

packages/components/src/Icon/_docs/IconsCustom.stories.tsx
AliceMenzie: updated

packages/components/rollup.config.mjs
AliceMenzie: updated

mia-nguyen0411 (dismissed):
after running `pnpm build`, I could not find the `CustomIcon.mjs` like other icons but has `CustomIcon.d.ts` that reexport the `.svg` instead
image
This cause the type error when import and use in the consumer's app unless they export it by themselves. 
image
Beside from the annoying warning, the icon is still shown after adding a temp check
image

&rarr; should we follow the `IconV1` approach to export the `CustomIcon` so consumers could use it without export the icon themselves? This is a minor user experience things, not really about functionality so I think we could pass it if this is not urgent right now.

AliceMenzie:
> after running `pnpm build`, I could not find the `CustomIcon.mjs` like other icons but has `CustomIcon.d.ts` that reexport the `.svg` instead image This cause the type error when import and use in the consumer's app unless they export it by themselves. image Beside from the annoying warning, the icon is still shown after adding a temp check image
> 
> → should we follow the `IconV1` approach to export the `CustomIcon` so consumers could use it without export the icon themselves? This is a minor user experience things, not really about functionality so I think we could pass it if this is not urgent right now.

Hey @mia-nguyen0411  did you link a canary or a pnpm portal of the kaizen package build in a consuming repo such as unified-home. without that it will not be running on the latest build of kaizen components and so it wont come up as an import. you might need to restart VSC.

 alternatively you can check the build dist for the exported items in kaizen after running a build and see it there 
in either 
```
dist/esm/index.mjs 
/dist/cjs/index.cjs
```

mia-nguyen0411:
> Hey @mia-nguyen0411 did you link a canary or a pnpm portal of the kaizen package build in a consuming repo such as unified-home. without that it will not be running on the latest build of kaizen components and so it wont come up as an import. you might need to restart VSC.
> 
> alternatively you can check the build dist for the exported items in kaizen after running a build and see it there in either
> 
> ```
> dist/esm/index.mjs 
> /dist/cjs/index.cjs
> ```

will try it after lunch. Thanks Alice ദ്ദി(｡˃ ᵕ ˂ )⸝

mia-nguyen0411 (dismissed):
Work as intended with canary release test to consumer's app from my end after the second time testing with VSC restart ♩¨̮(ง ˙˘˙ )ว♩¨̮

packages/components/src/Icon/types.ts
AliceMenzie: unresolving these comments, this latest change has issue with the build types not exporting correctly , they get overridden and the resulting SVG type exported is `string`

will look into another option

packages/components/rollup.config.mjs
AliceMenzie: same as above issue -- 

unresolving these comments, this latest change has issue with the build types not exporting correctly , they get overridden and the resulting SVG type exported is `string`

will look into another option

AliceMenzie:
Screenshot 2026-06-26 at 1 16 53 pm

using pnpm portal of last commits build in  unified-home

docs/.storybook/main.ts
```diff
@@ -52,6 +53,46 @@ const config: StorybookConfig = {
...
+        ${variables.jsx}
+      );
+      
+      export default ${variables.componentName};
+    `
+
+    config?.plugins?.push(
+      svgr({
+        include: '**/custom-icons/*.svg',
+        svgrOptions: {
```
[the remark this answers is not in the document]
AliceMenzie: updated

packages/components/src/Icon/CustomIcons.ts
```diff
@@ -0,0 +1,8 @@
+import type { FC, SVGProps } from 'react'
+import _UserSettingsIcon from '../../assets/svgs/custom-icons/user-settings.svg'
+
+type CustomIconsProps = FC<SVGProps<SVGSVGElement> & { color?: string; size?: number }>
+
+// Cast forces tsc to emit the concrete FC type in .d.ts rather than a .svg path reference,
+// which consumers can't resolve (the declare module '*.svg' in types.d.ts is not included in the package output).
+export const UserSettingsIcon = _UserSettingsIcon as CustomIconsProps
```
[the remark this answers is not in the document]
AliceMenzie: NA: keeping this as opt-in, want to keep the type inline with untitledUI.

packages/components/src/Icon/CustomIcons.ts
```diff
@@ -0,0 +1,8 @@
+import type { FC, SVGProps } from 'react'
+import _UserSettingsIcon from '../../assets/svgs/custom-icons/user-settings.svg'
+
+type CustomIconsProps = FC<SVGProps<SVGSVGElement> & { color?: string; size?: number }>
```
ryanseddon: I think this change is fine only change I would suggest is the name is not for props but a component

```suggestion
type CustomIcon = FC<SVGProps<SVGSVGElement> & { color?: string; size?: number }>
```

packages/components/src/Icon/CustomIcons.ts
ryanseddon:
```suggestion
export const UserSettingsIcon = _UserSettingsIcon as CustomIcon
```

AliceMenzie: thanks, updated

ryanseddon (approved):

PR #6800, opened 2026-06-17
fix: propose span as the default tag for the focusable component wrap…

CrowBe:
…per, with an optional polymorphic as prop

## Why
Fixes a component tree error caused when using Focusable to render tooltips on an inline, non interactable component.

## What
Swaps the default tag for the Focusable element to span, so it's more semantic inline.
Added optional `as` prop to the component in case the consumer prefers a different tag.

### Notes
One thing to call out is that span as the default changes the component tree for all consumers from div > span.
I think that's correct, but open to setting the default to div for now, and I as the downstream consumer can implement span

### DIFF
--- a/packages/components/src/Focusable/Focusable.tsx
+++ b/packages/components/src/Focusable/Focusable.tsx
@@ -1,33 +1,41 @@
-import React, { useRef, type HTMLAttributes, type ReactNode } from 'react'
+import React, { useRef, type ElementType, type HTMLAttributes, type ReactNode } from 'react'
 import classnames from 'classnames'
 import { useFocusable, type FocusableOptions } from 'react-aria'
 import styles from './Focusable.module.css'
 
-export type FocusableProps = {
+export type FocusableProps<T extends ElementType = 'span'> = {
   children: ReactNode
+  /**
+   * The HTML element to render as. Defaults to `span` to avoid block-in-inline
+   * hydration errors when Focusable wraps content inside a `<p>` or other inline context.
+   */
+  as?: T
 } & FocusableOptions &
-  HTMLAttributes<HTMLDivElement>
+  HTMLAttributes<HTMLElement>
 
-export const Focusable = ({ children, className, ...props }: FocusableProps): JSX.Element => {
-  const ref = useRef<HTMLDivElement>(null)
+export const Focusable = <T extends ElementType = 'span'>({
+  children,
+  className,
+  as,
+  ...props
+}: FocusableProps<T>): JSX.Element => {
+  const ref = useRef<HTMLElement>(null)
   const { focusableProps } = useFocusable(props, ref)
+  const Element = as ?? 'span'
 
   return (
-    <div
+    <Element
       ref={ref}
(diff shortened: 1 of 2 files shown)


packages/components/src/Focusable/Focusable.tsx
```diff
@@ -1,33 +1,40 @@
...
 import styles from './Focusable.module.css'
 
-export type FocusableProps = {
+export type FocusableProps<T extends ElementType = 'span'> = {
   children: ReactNode
+  /**
+   * The HTML element to render as. Defaults to `span` to avoid block-in-inline
+   * hydration errors when Focusable wraps content inside a `<p>` or other inline context.
+   */
+  as?: T
```
mia-nguyen0411: Suggesting using `tag` for this since it is used more frequently across Kaizen

packages/components/src/Focusable/Focusable.tsx
```diff
@@ -1,33 +1,40 @@
-import React, { useRef, type HTMLAttributes, type ReactNode } from 'react'
+import React, { useRef, type ElementType, type HTMLAttributes, type ReactNode } from 'react'
 import classnames from 'classnames'
 import { useFocusable, type FocusableOptions } from 'react-aria'
 import styles from './Focusable.module.css'
 
-export type FocusableProps = {
+export type FocusableProps<T extends ElementType = 'span'> = {
```
mia-nguyen0411: It will be better to set the default to `<div>` and allow passing customised `<Focusable tag="span">` tag to solve the SSR hydration issue in your team. Unfortunately, changing the component permanently to `<span>` as default would be a major breaking change for other consumers

packages/components/src/Focusable/Focusable.tsx
mia-nguyen0411, line 1: Thanks for your contribution! Since this will trigger a new package release with changes for consumers, it would be nice if you can add a changeset with `patch` and include a change description for `Focusable` as well.

mia-nguyen0411 (approved):
Looks good to me! All CI and local checks passed. `Focusable` works fine without `tag`, render as expected with customised tag, `Tooltip` and `Tag` with `Focusable` has the same behaviors as before, inline `Focusable` acts fine
Thanks so much for your contribution!

sentience:
FYI just reading the changelog before releasing this, and I would have said this was a minor change (i.e. a new API feature), not a patch. Not a huge deal, though, so I'm rolling ahead with the release anyway. Just thought I'd flag it for future consideration, @CrowBe.

mia-nguyen0411:
> FYI just reading the changelog before releasing this, and I would have said this was a minor change (i.e. a new API feature), not a patch. Not a huge deal, though, so I'm rolling ahead with the release anyway. Just thought I'd flag it for future consideration, @CrowBe.

sorry @sentience it was my fault ૮(˶╥︿╥)ა he clarified with me but then I said `patch` was fine since this does not change the behavior of the `Focusable`. Will check carefully from now on ૮(˶╥︿╥)ა

PR #6897, opened 2026-07-13
feat(TitleBlock): add opt-in sticky top row

Kitty-Al:
## What

Adds an opt-in `sticky` prop to `TitleBlock`. When enabled, the top strip sticks to the top of its scrollable container as content scrolls, positioned below the app-chrome nav via two CSS variables:

- **`--app-chrome-sticky-offset`** — owned by AppChrome, read-only. Positions the strip below the top bar. Dropped to `0` below 1080px to match the collapsed hamburger nav.
- **`--titleblock-sticky-offset`** — opt-in, per consumer. Reserves space for a banner rendered above the TitleBlock. Defaults to `0`, so existing TitleBlocks are unchanged.

Additive-by-default: the shared `--app-chrome-sticky-offset` is never reassigned, so apps that read it for other layout are unaffected.

## Stories

- **Sticky Top Strip In Scrollable Container** — sticky demo across all variants, with a fake app-chrome nav that collapses below 1080px.
- **With Global Notification Above** — a `GlobalNotification` banner pinned above a sticky TitleBlock; banner height measured via `ResizeObserver` and fed into `--titleblock-sticky-offset` so the strip stays flush.

## Notes

- Pattern documented in Confluence: *[Sticky banner above a sticky TitleBlock](https://cultureamp.atlassian.net/wiki/spaces/~712020af04446916ad4662a5df629105aa7b04/pages/6393626806/Sticky+banner+above+a+sticky+TitleBlock)*.
- Supersedes spike branch `ryan/KZN-3981/sticky-header`.
## Testing

`devbox run verify` — build, lint:ts, eslint, stylelint, prettier, vitest all green (TitleBlock: 23 tests ✓). Chromatic will need visual approval.

### DIFF
--- a/.changeset/smooth-doodles-follow.md
+++ b/.changeset/smooth-doodles-follow.md
@@ -1,0 +1,5 @@
+---
+'@kaizen/components': patch
+---
+
+Add `sticky` prop to TitleBlock. When enabled, the top strip sticks to the top of its scrollable container as the content scrolls, offset below the app-chrome nav via the `--app-chrome-sticky-offset` and `--titleblock-sticky-offset` CSS variables.
--- a/packages/components/src/TitleBlock/TitleBlock.module.scss
+++ b/packages/components/src/TitleBlock/TitleBlock.module.scss
@@ -61,6 +61,35 @@
     display: flex;
     width: 100%;
     justify-content: center;
+    position: relative;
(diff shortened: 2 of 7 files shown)


AliceMenzie (approved):
great docs

PR #6960, opened 2026-07-30
add css breakpoint

AliceMenzie:


packages/tailwind/css/breakpoints.css
```diff
@@ -0,0 +1,22 @@
+/**
+ * Kaizen breakpoints for Tailwind CSS v4 (CSS-first config).
```
AliceMenzie, line 2: ai - i don't think this fully explains what is happening correctly fyi , but the fix of having to define these values through css in a theme without a prefix remains

PR #6961, opened 2026-07-31
fix(SingleSelect, MultiSelect): mark select popover as react-aria top layer

Kitty-Al:
Fixes [KZN-4210](https://cultureamp.atlassian.net/browse/KZN-4210). Reported by @nate-mowbray-ca (survey-reporting-ui), who diagnosed the root cause — thank you, this would have been a long hunt otherwise.

## The bug

Opening a `SingleSelect` inside a `MenuPopover` would close the menu automatically, no way of choosing an item from the select.

## Cause

The select popover portals into `document.body` with a plain `createPortal`. react-aria doesn't recognise it as one of its own overlays, so when focus moves into the listbox the menu thinks focus went outside itself and closes.

The check react-aria uses is `isElementInChildOfActiveScope`, and it looks for a `data-react-aria-top-layer` attribute. Our portal doesn't have one. `isNonModal` doesn't help — the close happens through `shouldCloseOnBlur`, which `usePopover` always turns on.

## Fix

Put data-react-aria-top-layer="true" on the popover. react-aria walks up from the focused element, finds it, and leaves the menu open.

## Backwards compatibility

No API change — patch release. No props added or removed, no type-export churn.

**Known trade-off:** a marked element is exempt from all react-aria isolation, so while a select popover is open, clicking inside it won't dismiss an unrelated overlay. That needs two overlays open at once where neither contains the other. It's the same trade react-aria makes for its own overlays and toasts.

[KZN-4210]: https://cultureamp.atlassian.net/browse/KZN-4210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/MultiSelect/subcomponents/Popover/Popover.tsx
```diff
@@ -36,6 +37,8 @@ export const Popover = <RT extends ReferenceType>({
   classNameOverride,
   ...restProps
 }: PopoverProps<RT>): JSX.Element => {
+  useTopLayer(portalContainer)
+
```
ryanseddon, line 41: I removed this as I don't think the whole marking thing is needed just the attribute seems to be fine?

packages/components/src/MultiSelect/subcomponents/Popover/topLayer.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (remaining > 0) {
+        markCounts.set(element, remaining)
+        return
+      }
+
+      markCounts.delete(element)
+      element.removeAttribute(TOP_LAYER_ATTRIBUTE)
+    }
+  }, [element])
+}
```
ryanseddon, line 53: I'm not sure this WeakMap really does anything I reduced the hook to just be an alternate way of adding the `data-react-aria-top-layer` to an element and it seems fine with just that, might be missing some context though. The story for single select still works as expected

```suggestion
export const useTopLayer = (element: Element | DocumentFragment | undefined): void => {
  useLayoutEffect(() => {
    if (!(element instanceof Element) || element.hasAttribute(TOP_LAYER_ATTRIBUTE)) return

    element.setAttribute(TOP_LAYER_ATTRIBUTE, 'true')

    return () => element.removeAttribute(TOP_LAYER_ATTRIBUTE)
  }, [element])
}

```

packages/components/src/MultiSelect/subcomponents/Popover/Popover.tsx
Kitty-Al: Thanks for this Ryan, I've removed this now

AliceMenzie (approved):
Do we need a canary for this/ were we able to validate on a consumer level? 
if NA/we are confident without that 👍

